PR #11, opened 2024-08-19
Add test runner, config, and some basic tests

### LINKED ISSUE #5: Add Jest and implement basic unit tests for core components.

### LINKED ISSUE #7: Add Codecov for test coverage.

aganders3:
### Summary
This adds [vitest](https://vitest.dev/) as a test runner, and some basic tests.

The additional changes to vite and ts config were mostly put in to make the TypeScript compiler happy when checking the `test` directory.

### Related Issue
Closes #5, but we will want a new issue to target increasing coverage.
This is also related to #6 but that is not implemented here.
Closes #7.

### Tests & Checks
I am able to run `npm run test` and `npm run coverage` locally. Here is the coverage report:

```
> viz@0.0.0 coverage
> vitest --coverage --run


 RUN  v2.0.5 /Users/aandersoniii/src/active-learning/imaging-active-learning
      Browser runner started at http://localhost:5173/
      Coverage enabled with istanbul

stdout | test/webgl_renderer.test.ts > Instantiate WebGLRenderer
WebGL version WebGL 2.0 (OpenGL ES 3.0 Chromium)
 ✓ test/layer_manager.test.ts (1)
 ✓ test/webgl_renderer.test.ts (3)

 Test Files  2 passed (2)
      Tests  4 passed (4)
   Start at  16:58:27
   Duration  1.38s (transform 0ms, setup 0ms, collect 1.01s, tests 13ms, environment 0ms, prepare 497ms)

 % Coverage report from istanbul
-----------------------|---------|----------|---------|---------|-------------------
File                   | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s 
-----------------------|---------|----------|---------|---------|-------------------
All files              |   45.09 |    66.66 |   30.43 |   45.09 |                   
 core                  |   41.66 |       50 |   31.25 |   41.66 |                   
  layer.ts             |       0 |      100 |       0 |       0 | 4-11              
  layer_manager.ts     |   33.33 |      100 |       0 |   33.33 | 7-11              
  renderer.ts          |   46.66 |       50 |   41.66 |   46.66 | 23-47,71          
 layers                |       0 |      100 |       0 |       0 |                   
  single_mesh_layer.ts |       0 |      100 |       0 |       0 | 6-9               
 objects/renderable    |       0 |      100 |       0 |       0 |                   
  mesh.ts              |       0 |      100 |       0 |       0 | 5                 
 renderers             |   72.72 |      100 |      40 |   72.72 |                   
  webgl_renderer.ts    |   72.72 |      100 |      40 |   72.72 | 24-29             
-----------------------|---------|----------|---------|---------|-------------------
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,10 @@ dist
 dist-ssr
 *.local
 
+# test output
+coverage
+test/__screenshots__
+
 # Editor directories and files
 .vscode/*
 !.vscode/extensions.json
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "dev": "vite",
     "build": "tsc && vite build",
     "preview": "vite preview",
-    "lint": "eslint './src/**/*.{ts,tsx}'",
+    "lint": "eslint .",
     "format": "prettier --write './**/*.{ts,html,json}'",
     "test": "vitest",
     "coverage": "vitest --coverage"
--- a/package.json
+++ b/package.json
@@ -8,23 +8,30 @@
     "build": "tsc && vite build",
     "preview": "vite preview",
     "lint": "eslint './src/**/*.{ts,tsx}'",
-    "format": "prettier --write './**/*.{ts,html,json}'"
+    "format": "prettier --write './**/*.{ts,html,json}'",
+    "test": "vitest",
+    "coverage": "vitest --coverage"
   },
   "devDependencies": {
     "@eslint/js": "^9.9.0",
     "@types/webgl2": "^0.0.11",
     "@typescript-eslint/eslint-plugin": "^8.1.0",
     "@typescript-eslint/parser": "^8.1.0",
+    "@vitest/browser": "^2.0.5",
+    "@vitest/coverage-istanbul": "^2.0.5",
     "eslint": "^9.9.0",
     "globals": "^15.9.0",
     "husky": "^9.1.4",
+    "jsdom": "^24.1.1",
+    "playwright": "^1.46.1",
     "prettier": "^3.3.2",
     "typescript": "^5.2.2",
     "typescript-eslint": "^8.1.0",
     "vite": "^5.3.1",
     "vite-plugin-eslint": "^1.8.1",
     "vite-plugin-glsl": "^1.3.0",
-    "vite-tsconfig-paths": "^4.3.2"
+    "vite-tsconfig-paths": "^4.3.2",
+    "vitest": "^2.0.5"
   },
   "husky": {
     "hooks": {
--- a/src/core/layer.ts
+++ b/src/core/layer.ts
@@ -1,4 +1,4 @@
-import { RenderableObject } from "@/core/renderable_object";
+import { RenderableObject } from "./renderable_object";
 
 export abstract class Layer {
   private objects_: RenderableObject[] = [];
--- a/src/core/layer.ts
+++ b/src/core/layer.ts
@@ -1,4 +1,4 @@
-import { RenderableObject } from "./renderable_object";
+import { RenderableObject } from "@/core/renderable_object";
 
 export abstract class Layer {
   private objects_: RenderableObject[] = [];
--- a/test/layer_manager.test.ts
+++ b/test/layer_manager.test.ts
@@ -1,0 +1,8 @@
+import { expect, test } from 'vitest'
+
+import { LayerManager } from "@";
+
+test('Instantiate LayerManager', () => {
+  const layerManager = new LayerManager();
+  expect(layerManager).toBeDefined();
+});
--- a/test/layer_manager.test.ts
+++ b/test/layer_manager.test.ts
@@ -1,8 +1,0 @@
-import { expect, test } from "vitest";
-
-import { LayerManager } from "@";
-
-test("Instantiate LayerManager", () => {
-  const layerManager = new LayerManager();
-  expect(layerManager).toBeDefined();
-});
--- a/test/layer_manager.test.ts
+++ b/test/layer_manager.test.ts
@@ -1,8 +1,8 @@
-import { expect, test } from 'vitest'
+import { expect, test } from "vitest";
 
 import { LayerManager } from "@";
 
-test('Instantiate LayerManager', () => {
+test("Instantiate LayerManager", () => {
   const layerManager = new LayerManager();
   expect(layerManager).toBeDefined();
 });
--- a/test/webgl_renderer.test.ts
+++ b/test/webgl_renderer.test.ts
@@ -1,0 +1,24 @@
+import { expect, test } from 'vitest'
+
+import { WebGLRenderer } from '@';
+import { LayerManager } from '@';
+
+
+test('Instantiate WebGLRenderer', () => {
+  document.body.innerHTML = '<canvas id="canvas"></canvas>';
+  const renderer = new WebGLRenderer('#canvas', new LayerManager());
+  expect(renderer).toBeDefined();
+});
+
+test('Instantiate WebGLRenderer with invalid selector', () => {
+  document.body.innerHTML = '';
+  expect(() => new WebGLRenderer('#canvas', new LayerManager())).toThrowError();
+});
+
+test('Instantiate WebGLRenderer with no OpenGL context', () => {
+  document.body.innerHTML = '<canvas id="canvas"></canvas>';
+  const canvas = document.querySelector<HTMLCanvasElement>('#canvas')!;
+  // mock getContext to return null
+  canvas.getContext = () => null;
+  expect(() => new WebGLRenderer('#canvas', new LayerManager())).toThrowError();
+});
--- a/test/webgl_renderer.test.ts
+++ b/test/webgl_renderer.test.ts
@@ -1,24 +1,21 @@
-import { expect, test } from 'vitest'
+import { expect, test } from "vitest";
 
-import { WebGLRenderer } from '@';
-import { LayerManager } from '@';
+import { LayerManager, WebGLRenderer } from "@";
 
-
-test('Instantiate WebGLRenderer', () => {
+test("Instantiate WebGLRenderer", () => {
   document.body.innerHTML = '<canvas id="canvas"></canvas>';
-  const renderer = new WebGLRenderer('#canvas', new LayerManager());
+  const renderer = new WebGLRenderer("#canvas", new LayerManager());
   expect(renderer).toBeDefined();
 });
 
-test('Instantiate WebGLRenderer with invalid selector', () => {
-  document.body.innerHTML = '';
-  expect(() => new WebGLRenderer('#canvas', new LayerManager())).toThrowError();
+test("Instantiate WebGLRenderer with invalid selector", () => {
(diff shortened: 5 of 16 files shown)


test/__screenshots__/webgl_renderer.test.ts/Instantiate-WebGLRenderer-with-no-OpenGL-context-1.png
andy-sweet, line 1: Was this intended to be added as a gold standard screenshot to be compared against? If so, where is the corresponding assertion in the tests? Or this just implied by playwright?

If not, was this automatically generated by the tests, and can it be git-ignored?

andy-sweet (changes requested):
Looks good to me except for the PNG screenshot, which I'm not sure should be added to the repo. I'm also somewhat unsure about the config changes, but I pulled this down and was able to install and run everything.

package.json
```diff
@@ -8,23 +8,30 @@
     "build": "tsc && vite build",
     "preview": "vite preview",
     "lint": "eslint './src/**/*.{ts,tsx}'",
-    "format": "prettier --write './**/*.{ts,html,json}'"
+    "format": "prettier --write './**/*.{ts,html,json}'",
+    "test": "vitest",
+    "coverage": "vitest --coverage"
```
shlomnissan, line 13: Nice! I guess this PR closes this issue too https://github.com/chanzuckerberg/imaging-active-learning/issues/7

src/core/layer.ts
```diff
@@ -1,4 +1,4 @@
-import { RenderableObject } from "./renderable_object";
+import { RenderableObject } from "@/core/renderable_object";
```
shlomnissan: Why is this change necessary? I think relative paths are fine for related objects in the same directory.

Is there a way to add an include path without assigning it to some alias `@`?

test/layer_manager.test.ts
```diff
@@ -0,0 +1,8 @@
+import { expect, test } from 'vitest'
+
+import { LayerManager } from "@";
+
+test('Instantiate LayerManager', () => {
+  const layerManager = new LayerManager();
+  expect(layerManager).toBeDefined();
```
shlomnissan: Is this a temporary test? I’m not sure it’s meaningful, especially since the `LayerManager` class has no constructor logic or meaningful state to validate.

test/webgl_renderer.test.ts
```diff
@@ -0,0 +1,24 @@
+import { expect, test } from 'vitest'
+
+import { WebGLRenderer } from '@';
+import { LayerManager } from '@';
+
```
shlomnissan: - You can combine these import statements.
- There’s an extra empty line here. It doesn’t matter, but I wonder if we can add something to Prettier to delete unnecessary empty lines.

test/webgl_renderer.test.ts
```diff
@@ -0,0 +1,24 @@
...
+
+test('Instantiate WebGLRenderer', () => {
+  document.body.innerHTML = '<canvas id="canvas"></canvas>';
+  const renderer = new WebGLRenderer('#canvas', new LayerManager());
```
shlomnissan: Interesting... So we can actually initialize a WebGL context in tests? I guess that's where jsdom and Playwright come in.

test/webgl_renderer.test.ts
```diff
@@ -0,0 +1,24 @@
...
+  expect(renderer).toBeDefined();
+});
+
+test('Instantiate WebGLRenderer with invalid selector', () => {
+  document.body.innerHTML = '';
+  expect(() => new WebGLRenderer('#canvas', new LayerManager())).toThrowError();
+});
+
+test('Instantiate WebGLRenderer with no OpenGL context', () => {
```
shlomnissan: Nit: Instantiate a WebGLRenderer with no **WebGL** context.

test/webgl_renderer.test.ts
```diff
@@ -0,0 +1,24 @@
...
+  document.body.innerHTML = '<canvas id="canvas"></canvas>';
+  const canvas = document.querySelector<HTMLCanvasElement>('#canvas')!;
+  // mock getContext to return null
```
shlomnissan: Nit: I feel this comment is unnecessary. Comments are a subtle form of coupling. If the code already says something clearly, an additional comment should be avoided.

vite.config.js
```diff
@@ -1,39 +1,52 @@
...
 import { defineConfig } from 'vite';
 import tsconfigPaths from 'vite-tsconfig-paths';
 import eslint from 'vite-plugin-eslint';
 import path from 'path';
 
+// __dirname is not available in ES6 modules
+// https://github.com/vitejs/vite/issues/6946#issuecomment-1041506056
+import { dirname } from 'node:path'
+import { fileURLToPath } from 'node:url'
+const _dirname = dirname(fileURLToPath(import.meta.url));
```
shlomnissan, line 11: This file was using `__dirname` before. How is this different?

shlomnissan (commented):
Overall, I think it looks good. The only thing I want to understand better before I approve it is why we use the alias `@` for the source include directory and the purpose of the empty screenshot (as Andy mentioned), especially since it was generated from the “no OpenGL context” test.

test/__screenshots__/webgl_renderer.test.ts/Instantiate-WebGLRenderer-with-no-OpenGL-context-1.png
aganders3: Oh oops - totally accidentally committed.

test/webgl_renderer.test.ts
aganders3: I will make sure Prettier is even running on the tests dir.

src/core/layer.ts
aganders3: Yeah this is not necessary, but I changed it to be consistent with the others. I tend to prefer keeping everything as an absolute import but it's not a hill I'm going to die on. I can change back the imports from the same dir.

> Is there a way to add an include path without assigning it to some alias @?

I'm not sure what you mean by this. I think the only path you can reference without an alias is the `baseUrl` but I think that was why the `test` directory would not build. I am not an expert here by any means though.

vite.config.js
aganders3: `__dirname` was triggering a lint issue for me, because node treats all `js` files as modules when using `module` in package.json. I guess `__dirname` does not work with ES modules, but maybe whatever context this is actually run in does not care.

test/webgl_renderer.test.ts
aganders3: Yeah, that's my understanding. Without jsdom you can't access `document` and without Playwright you can't create the WebGL context.

package.json
aganders3: Ah I thought that was about connecting to https://about.codecov.io/ or something.

test/__screenshots__/webgl_renderer.test.ts/Instantiate-WebGLRenderer-with-no-OpenGL-context-1.png
aganders3: Ah I think it may have been generated when that test _failed_. I failed it on purpose once to make sure the mock was doing what I wanted.

test/layer_manager.test.ts
aganders3: I would consider all of these temporary tests at this point, this is more meant to just get the infrastructure in place. I can open another PR shooting for increased coverage, but given the current state of implementation that should maybe come with added functionality.

aganders3:
> why we use the alias @ for the source include directory

I haven't done that much web dev, but my impression is that this is just a convention. The preference for absolute over relative imports is my own, however. I am fine with relative imports within the same dir. Maybe there's some lint we can add to standardize? i.e. use absolute imports except within the same dir (or possibly child dirs).

andy-sweet (approved):
Now the screenshot is gone, approved. I tend to side with @shlomnissan on the nits, but they are nits, so feel free to do what you want with those.

aganders3:
I think I figured out how to make tsc happy without the `@` alias. Tests would just have to refer to imports with `../src`. We can then remove that alias from the vite config as well.

Edit: `examples/main.ts` is already using the `@` alias. I don't care much about this either way so don't want to spend too much time here. Maybe we leave the alias in place but only use it in tests and examples?

package.json
shlomnissan: I wanted to use codecov for generating the coverage, but if you think there's value in integrating it with codecov, we can keep the task open.

test/layer_manager.test.ts
shlomnissan: I think that's fine. If these are temporary tests, we should append "example" or "temporary" to the test name so it's clear.

shlomnissan:
> examples/main.ts is already using the @ alias

My intention was to add an alias `@` so examples/tests can access the source directory while keeping the default paths for the library code. This way, examples and tests don't need to use relative paths (`../src`), and library imports reference the source directory as expected.

src/core/layer.ts
shlomnissan: Using absolute paths as the default is fine. The only exception I tend to make is if the file is in the same working directory. This habit comes from C++ code, where private header files should be kept in the same folder as implementation files and referenced using the relative path. However, I'm not sure it makes sense in this context, so I'm happy to default to absolute paths.

My intention for the alias '@' was for external modules (examples, tests) to reference the source directory, but I didn't expect the library code to use it. The library code should have direct access since it all exists in the same root directory.

.gitignore
```diff
@@ -12,6 +12,10 @@ dist
 dist-ssr
 *.local
 
+# test output
+coverage
+test/__screenshots__
```
shlomnissan, line 17: I don’t think screenshots should be ignored if we intend to generate them. Comparing screenshots is how we check for regressions. My comment was specifically about the empty screenshot, which seemed out of place. Then again, I don’t know much about Vitest and didn’t consider testing visual regressions, so I leave it up to you.

vite.config.js
shlomnissan: I don't think the linter should parse this file, but I don't think it matters much anyway.

aganders3:
> My intention was to add an alias @ so examples/tests can access the source directory while keeping the default paths for the library code. This way, examples and tests don't need to use relative paths (../src), and library imports reference the source directory as expected.

This makes sense - I think we can do this but still need the changes in tsconfig. I'll push some changes and re-request review.

.gitignore
aganders3: This actually comes with playwright, but this directory has a different purpose. `__screenshots__` is where screenshots generated during testing - either on failure or from the comparison - get stored. What we probably want is for this dir to be uploaded as an artifact from our workflows.

I believe "golden" screenshots for comparison would live in a directory named for the test file. So in this case:
```
tests/
  webgl_renderer.test.ts
  webgl_renderer.test.ts-snapshots/
    instantiate-webgl-renderer-chromium-linux.png
```

shlomnissan:
No need to request another review. Feel free to merge when you are ready.

PR #12, opened 2024-08-19
Add a shader program to the renderer

shlomnissan:
### Summary

This PR adds a simple shader program to the WebGL renderer. It applies basic
caching by preloading uniform locations when the program is linked and again
on subsequent calls to `gl.getUniformLocation`, which may be necessary if the
uniform is a struct, for example.

### Related Issue

### API
```ts
program = new ShaderProgram(gl);
program.addShader(vertexShaderSrc, gl.VERTEX_SHADER);
program.addShader(fragmentShaderSrc, gl.FRAGMENT_SHADER);
program.link();

// 

program.use();
```

### Tests & Checks

This class offers a more flexible interface than the one in the [webgl-starter repo](https://github.com/shlomnissan/webgl-starter/tree/main). To test it, I integrated it into the webgl-starter repo to ensure everything works. Now that we have a PR for testing, it's worth considering mocking the gl context. I will write tests for this class in a future PR.

### DIFF
--- a/src/renderers/webgl_shaders.ts
+++ b/src/renderers/webgl_shaders.ts
@@ -1,23 +1,37 @@
+type ShaderMap = {
+  [type: number]: {
+    shader: WebGLShader;
+    source: string;
+  };
+};
+
+const regex = /^\s*uniform\s+(?:highp|mediump|lowp|)[\w\s]+\s+(\w+)\s*;\s*$/gm;
+
 export class WebGLShaders {
   private readonly gl_: WebGL2RenderingContext;
-  private readonly program_: WebGLProgram | null;
-  private readonly uniformRegex_ =
-    /^\s*uniform\s+(?:highp|mediump|lowp|)[\w\s]+\s+(\w+)\s*;\s*$/gm;
-
+  private readonly program_: WebGLProgram;
   private uniformLocations_: Map<string, WebGLUniformLocation>;
-  private shaderSources: string[] = [];
-  private shaderObjects: WebGLShader[] = [];
+  private shaders_: ShaderMap = {};
+  private linked_ = false;
 
   constructor(gl: WebGL2RenderingContext) {
     this.gl_ = gl;
-    this.program_ = gl.createProgram();
-    if (!this.program_) {
+    const program = gl.createProgram();
+    if (!program) {
       throw new Error(`Failed to create WebGL shader program`);
     }
+    this.program_ = program;
+
     this.uniformLocations_ = new Map<string, WebGLUniformLocation>();
   }
 
   public addShader(source: string, type: number) {
+    if (this.linked_) {
+      throw new Error(
+        "The program is already linked. Create a new shader program to add shaders."
+      );
+    }
+
     const shader = this.gl_.createShader(type);
     if (!shader) {
       throw new Error(`Failed to create a new shader of type ${type}`);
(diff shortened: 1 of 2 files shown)


src/renderers/webgl_shaders.ts
```diff
@@ -0,0 +1,109 @@
...
+  private uniformLocations_: Map<string, WebGLUniformLocation>;
+  private shaderSources: string[] = [];
+  private shaderObjects: WebGLShader[] = [];
+
+  constructor(gl: WebGL2RenderingContext) {
+    this.gl_ = gl;
+    this.program_ = gl.createProgram();
+    if (!this.program_) {
+      throw new Error(`Failed to create WebGL shader program`);
+    }
```
andy-sweet: Optional: we can add an extra line to make `this.program_` explicitly non-nullable, which I think means we could get rid of the `program` property.

```suggestion
    const program = gl.createProgram();
    if (!program) {
      throw new Error(`Failed to create WebGL shader program`);
    }
    this.program_ = program;
```

I wish there was a more concise way to do this in TS, but I'm not aware of one.

Or, if we want `program_` to be nullable, maybe we should explicitly assign it to `null` in `deleteProgram`?

src/renderers/webgl_shaders.ts
```diff
@@ -0,0 +1,109 @@
...
+      if (location) {
+        this.uniformLocations_.set(name, location);
+      }
+    }
+  }
+
+  private deleteShaderObjects() {
+    if (this.shaderObjects.length) {
+      this.shaderObjects.forEach((s) => this.gl_.deleteShader(s));
+    }
```
andy-sweet: Optional: allowing iteration over the empty array makes this more readable to me.

```suggestion
    this.shaderObjects.forEach((s) => this.gl_.deleteShader(s));
```

src/renderers/webgl_shaders.ts
```diff
@@ -0,0 +1,109 @@
...
+    this.gl_.attachShader(this.program, shader);
+    this.shaderSources.push(source);
+    this.shaderObjects.push(shader);
+  }
+
+  public link() {
+    this.gl_.linkProgram(this.program);
+    if (!this.gl_.getProgramParameter(this.program, this.gl_.LINK_STATUS)) {
+      this.deleteShaderObjects();
+      this.deleteProgram();
```
andy-sweet: If we reach this line (or the similar one below), and the program is deleted, can this instance ever be useful again?

src/renderers/webgl_shaders.ts
```diff
@@ -0,0 +1,109 @@
+export class WebGLShaders {
+  private readonly gl_: WebGL2RenderingContext;
+  private readonly program_: WebGLProgram | null;
+  private readonly uniformRegex_ =
+    /^\s*uniform\s+(?:highp|mediump|lowp|)[\w\s]+\s+(\w+)\s*;\s*$/gm;
+
+  private uniformLocations_: Map<string, WebGLUniformLocation>;
+  private shaderSources: string[] = [];
+  private shaderObjects: WebGLShader[] = [];
```
andy-sweet: As members, should these have trailing underscores?

```suggestion
  private shaderSources_: string[] = [];
  private shaderObjects_: WebGLShader[] = [];
```

andy-sweet (approved):
Generally looks good to me. All my comments are optional to address, though I am a little confused about the useful lifetime of an instance of `WebGLShaders`. That might become more clear once we have at least one example or test that use this.

src/renderers/webgl_shaders.ts
```diff
@@ -0,0 +1,109 @@
...
+    this.gl_.validateProgram(this.program);
+    if (!this.gl_.getProgramParameter(this.program, this.gl_.VALIDATE_STATUS)) {
+      this.deleteShaderObjects();
+      this.deleteProgram();
+      const message = this.gl_.getProgramInfoLog(this.program);
+      throw new Error(`Error validating program: ${message}`);
+    }
+
+    this.deleteShaderObjects();
+    this.shaderSources.forEach(s => this.preprocessUniformLocations(s));
```
aganders3: I think this may fail in the case of adding multiple shaders of the same type. Each is pushed into this array but not removed if another shader of the same type is added. Only the last one will remain attached to the program.

Also perhaps a cache hit here could be wrong if a new shader is added (intending to override a previous one) and it uses the same uniform name. So maybe it needs to be cleared in `addShader`.

aganders3 (commented):
Overall this looks good and straightforward. The API seems to closely match the WebGL API, but I think our interface could be stricter with types. For example each program should have strictly one vertex and one fragment shader, right?

src/renderers/webgl_shaders.ts
shlomnissan: That's a good question. In the current design, it won't be useful again. I'm going to remove `deleteProgram()` so technically, we could add shaders and try to link again. I don't know if it's particularly useful, but it's a small change, and we can figure it out later as we learn more.

src/renderers/webgl_shaders.ts
shlomnissan: Great points! I'll add some extra guardrails:

- Store shaders in a map `[shader_type] ⇒ {object, source}` so that adding another shader of the same type before the program is linked will override the previous one.
- Throw an error if attempting to add a shader after the program is linked. I think it’s reasonable to make the program immutable at this point.

PR #13, opened 2024-08-20
Add lint and test workflow (CI)

### LINKED ISSUE #6: Add a GitHub Actions workflow to ensure that all tests are passing.

aganders3:
### Summary
This adds a workflow to run lint (ESLint + Prettier --check) and tests (with coverage) using GitHub Actions.

### Related Issue
Closes #6
Based on PR #11, so should wait to merge until after that

### Tests & Checks
This PR serves as the test bed to the workflow. I will push commits with lint issues and/or failing tests so we can confirm success/failure.

### DIFF
--- a/.github/workflows/lint-and-test.yml
+++ b/.github/workflows/lint-and-test.yml
@@ -1,0 +1,31 @@
+name: Lint and Test
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+jobs:
+  lint-and-test:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 20
+
+      - name: Install dependencies
+        run: |
+          npm ci
+          npx playwright install chromium
+
+      - name: Lint
+        run: |
+          npm run lint
+          npm run format-check
+
+      - name: Test
+        run: npm run coverage
--- a/.github/workflows/lint-and-test.yml
+++ b/.github/workflows/lint-and-test.yml
@@ -23,9 +23,7 @@ jobs:
           npx playwright install chromium
 
       - name: Lint
-        run: |
-          npm run lint
-          npm run format-check
+        run: npm run lint-all
 
       - name: Test
         run: npm run coverage
--- a/.github/workflows/lint-and-test.yml
+++ b/.github/workflows/lint-and-test.yml
@@ -37,4 +37,4 @@ jobs:
           npm run format-check
 
       - name: Run tests with coverage
-        run: npm run coverage
+        run: npm run test-with-coverage
--- a/.github/workflows/lint-and-test.yml
+++ b/.github/workflows/lint-and-test.yml
@@ -22,8 +22,14 @@ jobs:
           npm ci
           npx playwright install chromium
 
-      - name: Lint
-        run: npm run lint-all
+      - name: Compile (TypeScript)
+        run: npm run compile
+
+      - name: Lint (ESLint)
+        run: npm run lint
+
+      - name: Lint (Prettier)
+        run: npm run format-check
 
       - name: Test
         run: npm run coverage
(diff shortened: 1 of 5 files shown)


aganders3:
I think this is ready to go for now. I think we have some redundant rules enforced by both tsc and eslint, but I don't know of any _conflicting_ rules in our config.

.github/workflows/lint-and-test.yml
```diff
@@ -0,0 +1,35 @@
...
+  push:
+    branches:
+      - main
+  pull_request:
+
+jobs:
+  lint-and-test:
+    runs-on: ubuntu-latest
+
+    steps:
```
shlomnissan, line 13: We probably want to add caching for `node_modules`. I think we can use the `actions/cache` action to cache dependencies between builds.

.github/workflows/lint-and-test.yml
```diff
@@ -0,0 +1,35 @@
...
+        run: npm run compile
+
+      - name: Lint (ESLint)
+        run: npm run lint
+
+      - name: Lint (Prettier)
+        run: npm run format-check
+
+      - name: Test
+        run: npm run coverage
```
shlomnissan: - Should this be `run: npm run test` or at least `run: npm run test && npm run coverage`? Although, I'm not sure what the purpose of running the coverage is -- is there an error state for this check?
- We can rename this to "Test and Coverage" if we decide to keep the coverage.

.github/workflows/lint-and-test.yml
```diff
@@ -0,0 +1,35 @@
...
+          npx playwright install chromium
+
+      - name: Compile (TypeScript)
+        run: npm run compile
+
+      - name: Lint (ESLint)
+        run: npm run lint
+
+      - name: Lint (Prettier)
+        run: npm run format-check
```
shlomnissan: Nit: I think we can simplify the workflow by adding both of these checks into a single step.

.github/workflows/lint-and-test.yml
```diff
@@ -0,0 +1,35 @@
...
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 20
+
+      - name: Install dependencies
+        run: |
+          npm ci
+          npx playwright install chromium
+
+      - name: Compile (TypeScript)
```
shlomnissan: Nit: Compile TypeScript no need for parentheses.

shlomnissan (dismissed):
Have we implemented branch protection rules that require successful status checks prior to merging?

The two primary concerns for me are:

- Ensuring the tests run (this may be part of the `coverage` script, but it's unclear to me.)
- Adding caching to node modules if possible. We don't anticipate many dependencies, but it would be beneficial to implement this before merging this PR.

.github/workflows/lint-and-test.yml
aganders3: Good idea - `setup-node` offers some in-built caching though it doesn't cache `node_modules` directly. I will look into this.

.github/workflows/lint-and-test.yml
aganders3: `npm run coverage` runs the tests under coverage and outputs the coverage report.  `npm run test && npm run coverage` would run the tests twice. We can change the name if it's not clear, this is just how it was configured in the vitest docs.

For now there's no check on the coverage, but this will otherwise fail the same way `npm run test` would.

shlomnissan: If it’s all the same, I have a slight preference towards `npm run test` because it's clearer, but it's up to you.

aganders3:
> Have we implemented branch protection rules that require successful status checks prior to merging?

I just added this to the rules. I will push another failing commit to make sure it's working.

aganders3:
@shlomnissan can you approve so we can test the branch rules? Right now I see "Merging can be performed automatically with 1 approving review."

andy-sweet (approved):
Approving, conditioned on removing the failing test. And also to test the branch protection rules.

aganders3:
Thanks! This looks correct to me:
Screenshot 2024-08-21 at 2 10 47 PM

shlomnissan (approved):
LGTM

PR #16, opened 2024-08-21
Add a temporary mesh source

shlomnissan:
### Summary

This PR adds a plane geometry generator that outputs a temporary data source
(`MeshSource`). The intention is to generate geometry in the mesh layer and make
it available to the renderer. I consider these changes entirely exploratory and
expect most of this code to change once we learn more about efficient data
fetching patterns. The goal of adding this component is to enable rendering of
a single mesh layer.

### Related Issue

### Tests & Checks

I manually tested this code, verifying that the renderer can access the mesh data without errors. However, I didn't add unit tests, as I consider these components temporary and didn't feel it would be beneficial.

### DIFF
--- a/src/data/mesh_source.ts
+++ b/src/data/mesh_source.ts
@@ -29,7 +29,7 @@ type MeshSourceAttributeType = "vertices" | "normals" | "uvs";
 
 export class MeshSource {
   private attributes_: Map<MeshSourceAttributeType, MeshSourceAttribute>;
-  private index_: Float32Array | null = null;
+  private index_: Uint32Array | null = null;
 
   constructor() {
     this.attributes_ = new Map<MeshSourceAttributeType, MeshSourceAttribute>();
@@ -51,10 +51,10 @@ export class MeshSource {
   }
 
   public setIndex(data: number[]) {
-    this.index_ = new Float32Array(data);
+    this.index_ = new Uint32Array(data);
   }
 
-  public getIndex(): Readonly<Float32Array> | null {
+  public getIndex(): Readonly<Uint32Array> | null {
     return this.index_;
   }
 }
--- a/src/data/mesh_source.ts
+++ b/src/data/mesh_source.ts
@@ -1,0 +1,60 @@
+/*
+  This class is temporary. Data source classes will likely be observable proxy
+  objects that hold references to data. A mesh can be loaded from various file
+  types. File loaders (such as zarr or obj) are responsible for loading the
+  necessary data (possibly using a chunk manager) and returning a data source
+  type. This data source is then used to initialize renderable objects (like
+  Mesh or Volume). While renderable objects may request more data through the
+  data source, a loader or chunk manager will be responsible for providing this
+  data. As we're actively exploring this area, I anticipate more clarity in the
+  coming weeks.
+*/
+
+class MeshSourceAttribute {
+  private data_: Float32Array;
+
+  public itemSize: number;
+
+  constructor(data: number[], itemSize: number) {
+    this.data_ = new Float32Array(data);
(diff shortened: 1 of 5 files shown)


src/objects/renderable/mesh.ts
```diff
@@ -1,7 +1,39 @@
+import { MeshSource } from "@/data/mesh_source";
 import { RenderableObject } from "core/renderable_object";
 
 export class Mesh extends RenderableObject {
+  private source_: MeshSource;
+
+  constructor(source: MeshSource) {
```
shlomnissan, line 7: Renderable objects are designed to be initialized with their respective data sources, but they serve as more than mere proxies. While the geometry is stored in the MeshSource (though "GeometrySource" might have been a more apt name), these renderable objects can also contain additional data such as transformations and rendering options or states.

aganders3 (approved):
This generally looks good, and as you note a lot of this is likely to change.

One thing I would suggest is trying to provide errors as early as possible. For example the `MeshSource` should not support attributes that don't match each other (or `indices`) in length. I think this can prevent obtuse low-level errors from the rendering backend.

src/data/mesh_source.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  get data(): Readonly<Float32Array> {
+    return this.data_;
+  }
+}
+
+type MeshSourceAttributeType = "vertices" | "normals" | "uvs";
+
+export class MeshSource {
+  private attributes_: Map<MeshSourceAttributeType, MeshSourceAttribute>;
```
andy-sweet, line 31: What's the motivation to store these in a `Map` rather than just declare each attribute as an optional member of the class? I saw a similar pattern in three.js, but also wasn't sure what motivated that either.

src/data/mesh_source.ts
```diff
@@ -0,0 +1,60 @@
...
+  private index_: Float32Array | null = null;
```
andy-sweet: I would expect indices to be integer valued, so I'd expect the type here to be `UInt32Array` (or similar). The values for these coming from `number`s in `PlaneGeometry`, but I wouldn't expect that always to be the case.

andy-sweet (approved):
Looks good. I have a couple of questions, but no blocking feedback.

src/data/mesh_source.ts
shlomnissan: I noticed this pattern in three.js and thought it was a good idea:

- Reducing boilerplate: If these attributes were all optional properties, I'd expect them to have getters and setters. Encapsulating these accessor methods behind a stable interface allows us to add shared guardrails around them (e.g., validation, tracking state changes, etc) without duplicating logic.
- Flexibility: It's likely we'll add new attributes in the future. These might be known shared attributes (like tangent data) or dynamic attributes for special cases. This structure would make such additions easier.

I don't feel strongly about this pattern. In hindsight, it does seem a bit elaborate for a temporary object. However, I think it's fine for now until we develop a clear strategy for managing data sources.

src/data/mesh_source.ts
shlomnissan: Good catch! This is just an oversight on my part.

shlomnissan:
> One thing I would suggest is trying to provide errors as early as possible. For example the MeshSource should not support attributes that don't match each other (or indices) in length. I think this can prevent obtuse low-level errors from the rendering backend.

@aganders3 I agree that's a great point. However, this is a temporary object, so I didn't take the time to fully think it through. In hindsight, the attributes pattern makes it seem more thought-out than it actually is. I expect we'll completely rewrite this once we start loading data from external files, so I'm tempted to defer on error handling.

PR #17, opened 2024-08-21
Ensure private properties end with an underscore

shlomnissan:
### Summary

This PR introduces an ESLint rule requiring private class members to end with
an underscore.

### Context

I'm adding this for consistency. While I don't feel strongly about it, the way accessor methods work in TypeScript makes this pattern sensible to me. To clarify, I don't believe every class property needs accessor methods—a public property is perfectly fine in some cases.

### Related Issue

I already closed this issue for some reason.

### Tests & Checks

I manually tested this by removing trailing underscores from private class members and confirming that the linter raises an error.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -8,8 +8,17 @@ export default [
   pluginJs.configs.recommended,
   ...tseslint.configs.recommended,
   {rules: {
-    '@typescript-eslint/no-unused-vars': ['error', { argsIgnorePattern: '^_' }],
     'no-duplicate-imports': 'error',
+    '@typescript-eslint/no-unused-vars': ['error', { argsIgnorePattern: '^_' }],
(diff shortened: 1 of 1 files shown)


aganders3 (approved):
LGTM!

I tested locally and see the error both when running `npm run lint` and in my editor via LSP.

aganders3:
Also - something that I didn't know before - there is [a pure JS syntax for private fields](https://www.typescriptlang.org/docs/handbook/classes.html#ecmascript-private-fields) that works a bit differently. Instead of declaring an attribute `private`, you just prefix the name with `#`.

I'm not necessarily advocating using this instead (it doesn't do exactly the same thing), but I found it interesting. Here is some more info on the differences:
https://devblogs.microsoft.com/typescript/announcing-typescript-3-8-beta/#ecmascript-private-fields

PR #18, opened 2024-08-23
Create shader programs and refactor

shlomnissan:
### Summary

This PR refactors the shader program class. It renames `WebGLShaders` to
`WebGLShaderProgram` and simplifies the API -- instead of adding shaders and
linking the program separately, you now supply the vertex and fragment source
codes to the constructor, and the program links automatically.
Additionally, this PR initializes the Vite GLSL plugin and creates a sample
shader program for meshes.

### Related Issue

### Tests & Checks

I manually tested this change, verifying that no errors occurred and confirming the shader program's validity.

### DIFF
--- a/src/renderers/shaders/index.ts
+++ b/src/renderers/shaders/index.ts
@@ -1,0 +1,11 @@
+import meshVertexShader from "./mesh_vert.glsl";
+import meshFragmentShader from "./mesh_frag.glsl";
+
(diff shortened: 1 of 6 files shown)


aganders3 (approved):
LGTM!

My only nit here is that we might also want the shaders to be named specifically for WebGL or GLSL. This is also not a problem unless/until we have another renderer though.

PR #19, opened 2024-08-26
Add a perspective camera and refactor

shlomnissan:
### Summary

This PR implements the following changes:
- Extracts the main loop from the renderer
- Adds the `glMatrix` dependency and associated TypeScript types
- Creates a base camera class and a derived perspective camera class
- Sets the projection uniform in the WebGL renderer

### Related Issue

### Tests & Checks

I manually tested this change, verifying that no errors occurred.

### DIFF
--- a/src/core/renderer.ts
+++ b/src/core/renderer.ts
@@ -1,19 +1,18 @@
 import { LayerManager } from "./layer_manager";
 import { Mesh } from "objects/renderable/mesh";
+import { Camera } from "objects/cameras/camera";
 
 export abstract class Renderer {
   private readonly canvas_: HTMLCanvasElement | null;
-  private lastTimestamp_ = 0;
   private width_ = 0;
   private height_ = 0;
-  private layerManager_: LayerManager;
+  private activeCamera_: Camera | null = null;
 
   protected abstract resize(width: number, height: number): void;
   protected abstract renderMesh(mesh: Mesh): void;
   protected abstract clear(): void;
 
-  constructor(selector: string, layerManager: LayerManager) {
-    this.layerManager_ = layerManager;
+  constructor(selector: string) {
     this.canvas_ = document.querySelector<HTMLCanvasElement>(selector);
     if (!this.canvas_) {
       throw new Error(`Canvas element not found for selector "${selector}"`);
@@ -25,19 +24,10 @@ export abstract class Renderer {
     });
   }
 
-  public render() {
-    const nextFrame = (timestamp: number) => {
-      const dt = (timestamp - this.lastTimestamp_) / 1000;
-      this.lastTimestamp_ = timestamp;
-      this.renderFrame(dt);
-      requestAnimationFrame(nextFrame);
-    };
-    requestAnimationFrame(nextFrame);
-  }
-
-  private renderFrame(_: number) {
+  public render(layerManager: LayerManager, camera: Camera) {
     this.clear();
-    this.layerManager_.layers.forEach((layer) => {
+    this.activeCamera_ = camera;
+    layerManager.layers.forEach((layer) => {
       layer.objects.forEach((obj) => {
         // Before sending the object to the renderer backend, we must verify
         // its visibility by checking its render state and location relative
@@ -70,7 +60,12 @@ export abstract class Renderer {
     return this.height_;
   }
 
-  protected get layers() {
-    return this.layerManager_;
+  protected get camera() {
+    if (this.activeCamera_ === null) {
+      throw new Error(
+        "Attempted to access the active camera before it was set."
+      );
+    }
+    return this.activeCamera_;
   }
 }
--- a/src/core/renderer.ts
+++ b/src/core/renderer.ts
@@ -60,7 +60,7 @@ export abstract class Renderer {
     return this.height_;
   }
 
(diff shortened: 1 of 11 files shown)


src/core/renderer.ts
```diff
@@ -1,19 +1,18 @@
...
+  private activeCamera_: Camera | null = null;
 
   protected abstract resize(width: number, height: number): void;
   protected abstract renderMesh(mesh: Mesh): void;
   protected abstract clear(): void;
 
-  constructor(selector: string, layerManager: LayerManager) {
-    this.layerManager_ = layerManager;
+  constructor(selector: string) {
     this.canvas_ = document.querySelector<HTMLCanvasElement>(selector);
```
aganders3, line 16: As an aside - should the renderer take an `HTMLCanvasElement` directly instead of querying for one? We may also want to make it optional, and create it here instead. I wonder if this would make it easier to use for things like offscreen rendering. I'm not sure what is possible though.

shlomnissan: Yes, I was thinking about this as well. I think your suggestion is similar to Three API, where you can pass a canvas as an optional parameter. If you don't, the renderer creates one, and you can attach it to the DOM yourself: `document.body.appendChild( renderer.domElement );`.

I think it's out of scope for this PR, but I will add an issue (feel free to own it if you are itching to write code.)

shlomnissan:
@aganders3 I noticed you didn’t approve the PR. Is there anything blocking or are you planning to take another pass?

aganders3:
Nothing blocking, I just didn't really take a close look at it. I should be able to review it today though.

src/renderers/webgl_shader_program.ts
```diff
@@ -45,6 +47,10 @@ export class WebGLShaderProgram {
     return this.uniformLocations_.get(name)!;
   }
 
+  public setUniform(name: string, value: mat4) {
+    this.gl_.uniformMatrix4fv(this.getUniformLoc(name), false, value);
+  }
```
aganders3, line 52: Not all uniforms will be `mat4`, right? Is this just a placeholder for now?

src/objects/cameras/perspective_camera.ts
```diff
@@ -0,0 +1,35 @@
...
+const DEFAULT_ASPECT_RATIO = 1.77; // 16:9
+
+export class PerspectiveCamera extends Camera {
+  private fov_: number;
+  private aspectRatio_: number;
+  private near_: number;
+  private far_: number;
+
+  constructor(
+    fov: number = DEFAULT_FOV,
```
aganders3, line 14: It feels more natural to me to specify FOV in degrees and convert to radians.

src/objects/cameras/perspective_camera.ts
```diff
@@ -0,0 +1,35 @@
+import { Camera } from "./camera";
+import { mat4 } from "gl-matrix";
+
+const DEFAULT_FOV = 1.047197551; // 60 degrees in radians
+const DEFAULT_ASPECT_RATIO = 1.77; // 16:9
+
+export class PerspectiveCamera extends Camera {
```
aganders3, line 7: Should this override the parent `type`? Not necessary until it's useful, but we should be consistent with different inheritance hierarchies.

src/core/renderer.ts
```diff
@@ -70,7 +60,12 @@ export abstract class Renderer {
     return this.height_;
   }
 
-  protected get layers() {
-    return this.layerManager_;
+  protected get camera() {
```
aganders3: I think this is unused right now - no worries if that's intentional. I think I do like that `render` rakes a `camera` for each call, so I'm just curious what you have in mind for accessing the active camera like this.

aganders3 (approved):
I left a couple comments/questions, but nothing blocking.

src/renderers/webgl_shader_program.ts
shlomnissan: Yes, there are several ways to organize uniform assignments, and it depends on how we want to structure shader programs. In the short term, I thought we could add setter functions for uniforms as needed. This is an area open to exploration. Here's an example of how three is organizing uniforms.

https://github.com/mrdoob/three.js/blob/dev/src/renderers/webgl/WebGLUniforms.js

src/objects/cameras/perspective_camera.ts
shlomnissan: That's fair. I will modify this.

src/core/renderer.ts
shlomnissan: It's used. The render function assigns it to the local `activeCamera_` member. The `renderMesh` in the `WebGLRenderer` uses it to retrieve the active camera. I think renaming this to `get activeCamera()` might be a good idea.

src/objects/cameras/perspective_camera.ts
shlomnissan: Good call! I'll update it now.

src/core/renderer.ts
aganders3: Ahh, I see now - thanks. This way `renderer.render` takes a camera, but it doesn't need to be explicitly passed into each `renderX` function for the renderable objects.

PR #20, opened 2024-09-03
Render mesh

### LINKED ISSUE #4: Add all the necessary components for rendering a single mesh layer.

shlomnissan:
### Summary
This PR adds all the necessary plumbing to render a quad on the screen. Specifically:

- It adds management of vertex buffers and vertex array objects.
- It generates UUIDs for renderable elements.
- It updates the projection when the window resizes.

### Notes

1. All renderable objects should have some type of geometry that includes vertex attributes and additional metadata. Even a 3D volume will have a plane we use to project the volume onto. Right now, this is represented by the `MeshSource` class, but it shouldn't be tied to a source. `MeshSource` should probably become something like `BufferGeometry` in Three.js, which makes sense because it is modeled after it.
2. However, modeling after `BufferGeometry` (using individual attributes) might not have been ideal. Three.js stores vertex attributes in separate buffers by default (as do this PR), which is beneficial for dynamic data like animations, where the position buffer may change while the color buffer remains static. The downside is that configuring each attribute in its own buffer may lead to cache misses. Interleaved data might serve us better, though I'm not certain. Ultimately, we may need a combination of both approaches.

### Related Issue
Closes #4

### Tests & Checks

I manually tested this change, verifying that no errors occurred. See visual verification video below.

https://github.com/user-attachments/assets/f60354d2-cd97-4dd5-a1b7-73df862e2082


src/data/mesh_source.ts
```diff
@@ -20,16 +20,24 @@ class MeshSourceAttribute {
...
+    return this.data_.length;
+  }
 }
 
 type MeshSourceAttributeType = "vertices" | "normals" | "uvs";
 
 export class MeshSource {
   private attributes_: Map<MeshSourceAttributeType, MeshSourceAttribute>;
-  private index_: Uint32Array | null = null;
+  private index_: Uint16Array | null = null;
```
andy-sweet: Is the drop from uint32 to 16 because gl.UNSIGNED_INT (through [an extension](https://developer.mozilla.org/en-US/docs/Web/API/OES_element_index_uint)) may not be available in webgl 1?

If it's guaranteed to be available by default in webgl 2, I'd consider switching this back, but not a big issue.

andy-sweet (approved):
Looks good to me. I have a question about the drop in index width, but I'm uncertain about it and assume it is not blocking.

src/core/renderer.ts
```diff
@@ -46,17 +45,27 @@ export abstract class Renderer {
...
+
+    if (this.canvas.width !== this.width_) this.canvas.width = this.width_;
+    if (this.canvas.height !== this.height_) this.canvas.height = this.height_;
+
+    if (this.activeCamera_) {
+      if (this.activeCamera_ instanceof PerspectiveCamera) {
+        this.activeCamera_.setAspectRatio(this.width_ / this.height_);
+      }
+      this.activeCamera_.updateTransforms();
+    }
```
aganders3, line 57: I understand the intent here, but I'm not sure I like this coupling. It's not blocking though and we can discuss later.

src/core/utils.ts
aganders3, line 1: It's probably inevitable, but I'm not a fan of generic "utils" modules. I think it's just a pet peeve so this is non-blocking and we can make more descriptive modules later.

src/objects/cameras/perspective_camera.ts
```diff
@@ -21,6 +21,12 @@ export class PerspectiveCamera extends Camera {
     this.aspectRatio_ = aspectRatio;
     this.near_ = near;
     this.far_ = far;
+
+    this.updateProjectionTransform();
+  }
+
+  public setAspectRatio(aspectRatio: number) {
+    this.aspectRatio_ = aspectRatio;
   }
```
aganders3, line 30: Should setting a new aspect ratio also update the transform? I guess there may be performance benefits to keep setting/updating separate, but in that case maybe we could set the transform as "stale" when changing values.

src/objects/renderable/mesh.ts
```diff
@@ -13,27 +13,11 @@ export class Mesh extends RenderableObject {
...
-  }
-
-  public get normals() {
-    return this.source_.getAttribute("normals");
-  }
-
-  public get uvs() {
-    return this.source_.getAttribute("uvs");
+  public get source(): Readonly<MeshSource> {
+    return this.source_;
```
aganders3, line 17: I kind of liked the decoupling that was here before, but maybe we just add it back if/when needed.

src/core/renderer.ts
```diff
@@ -1,6 +1,7 @@
 import { LayerManager } from "./layer_manager";
 import { Mesh } from "objects/renderable/mesh";
 import { Camera } from "objects/cameras/camera";
+import { PerspectiveCamera } from "@/objects/cameras/perspective_camera";
```
aganders3:
```suggestion
import { PerspectiveCamera } from "objects/cameras/perspective_camera";
```

src/renderers/webgl_bindings.ts
```diff
@@ -0,0 +1,62 @@
+import { RenderableObject } from "core/renderable_object";
+import { Mesh } from "@/objects/renderable/mesh";
```
aganders3:
```suggestion
import { Mesh } from "objects/renderable/mesh";
```

aganders3 (approved):
Looks good! It's great to see a scene.

src/renderers/webgl_bindings.ts
```diff
@@ -0,0 +1,62 @@
...
+    let objectVAO = this.VAOs_.get(object.uuid) || null;
+
+    if (objectVAO && objectVAO === this.currentVAO_) {
+      // this object's VAO is already active
+      return;
+    }
+
+    if (!objectVAO) {
+      objectVAO = this.createVAO();
+      this.createBuffers(object as Mesh);
```
andy-sweet, line 27: How do you anticipate generalizing this to other `RenderableObject` types?

src/renderers/webgl_bindings.ts
```diff
@@ -0,0 +1,62 @@
...
+    this.gl_.bindVertexArray(objectVAO);
+    this.currentVAO_ = objectVAO!;
+  }
+
+  private createVAO() {
+    const vao = this.gl_.createVertexArray();
+    if (!vao) {
+      throw new Error(`Unable to generate a vertex array object name`);
+    }
+    this.gl_.bindVertexArray(vao);
```
andy-sweet: Maybe it doesn't matter, but do you need to call `bindVertexArray` both here and in `bind`?

src/data/mesh_source.ts
shlomnissan: Great catch! I encountered some rendering issues, which were resolved by switching to uint16. The problem stemmed from how I called the drawing function. I'll update the PR accordingly.

src/core/renderer.ts
shlomnissan: What's your specific concern here? I agree that having the renderer maintain an active camera instance might not be ideal. However, I suspect your comment is about checking the underlying camera type. I believe we can generalize this by adding mutators for width and height to the base class, which an orthographic camera would also need, but I want to clarify with you first

src/core/utils.ts
shlomnissan: What specific concerns do you have with a single module exporting utility functions? The only issues I can think of are unoptimized tree shaking—which isn't a concern here—and the potential for this module to become overloaded with unrelated functions. However, I'm not worried about the latter at this point.

src/objects/cameras/perspective_camera.ts
shlomnissan: The camera node will eventually have two transforms: one that represents the projection, and another that represents position and direction (which will be inverted.) The latter will be part of the renderable object base class.

1. The projection transform updates whenever we initialize the camera or change the window size (the aspect ratio, at least, should not change otherwise). At this stage, I don't believe tracking its state adds any value.
2. The transform belonging to the renderable object base class (comparable to Object3D in Three.js) will need to track state changes, but this hasn't been implemented yet.

I'm trying to keep things as simple as possible so we can stay flexible as we introduce new components.

src/objects/renderable/mesh.ts
shlomnissan: I felt the previous accessor methods introduced unnecessary indirection. In the PR summary, I expressed my doubts about the necessity of the “attributes pattern” and highlighted the need to better define the boundary between mesh and geometry. This change helped remove some of the boilerplate so we can punt on this decision.

src/renderers/webgl_bindings.ts
shlomnissan: My PR summary touch on this, but didn't address it directly. I think every renderable object should have geometry assigned to it. This approach would ensure consistent bindings across all objects, while allowing drawing configurations to vary based on the object type at the renderer level. What are your thoughts on this approach?

src/renderers/webgl_bindings.ts
shlomnissan: We should bind the VAO before specifying the vertex attributes. This approach would result in a double bind, which I initially didn't consider problematic. However, I've reorganized the code to avoid it. I hope it improves clarity as well.

src/renderers/webgl_bindings.ts
andy-sweet: I think I get it.

By geometry you mean the vertices, indices, normals, as well as optional attributes like colors or texture uvs. And that geometry interface can effectively be shared across all renderable objects, so we only need one definition of `createBuffers` that might have a few `if` statements to handle those optional attributes depending on the object.

So we don't need a separate `createMeshBuffers`, `createPointCloudBuffers`, etc. It just looks a little awkward right now because `RenderableObject` has a very limited interface and `Mesh` is our only example of one.

That sounds good to me. Thanks for clarifying.

src/renderers/webgl_bindings.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  private createBuffers(mesh: Mesh) {
+    let idx = 0;
+    for (const [, value] of mesh.source.attributes) {
+      const buffer = this.gl_.createBuffer();
+      const bufferType = this.gl_.ARRAY_BUFFER;
+      const size = value.itemSize;
+      this.gl_.bindBuffer(bufferType, buffer);
+      this.gl_.bufferData(bufferType, value.data, this.gl_.STATIC_DRAW);
+      this.gl_.vertexAttribPointer(idx, size, this.gl_.FLOAT, false, 0, 0);
```
andy-sweet, line 50: Doe `idx` need to unique across all objects? Or just within this one?

If it does need to be unique, no need to fix in this PR, but a TODO or similar would be useful.

src/core/renderer.ts
aganders3: Right - I was thinking about the branching based on camera type. I think you're right about how to generalize, but no need to do it here.

I think it's reasonable to have the renderer keep an active camera. I can't think of an issue with that, at least in a single-threaded context.

src/core/utils.ts
aganders3: It's more the latter - that "utils" ends up being a junk drawer. As I said though I think it's just a pet peeve of mine and I'm not going to fight you on it!

src/objects/cameras/perspective_camera.ts
aganders3: I might be just missing your point. My comment was about the implicit API introduced here. The consumer (right now just the `renderer`) needs to know to call `updateTransforms` after `setAspectRatio` is called. This is related to the comment above about the `renderer` branching on the camera type.

shlomnissan: Upon reflection, I understand/agree with your take. We'll need to check and potentially update the world transform for every component in the scene every frame. These updates should occur implicitly, without requiring the consumer to make explicit calls.

I initially thought the current approach was acceptable because it's unlikely we'd change the aspect ratio for other reasons. However, we'll probably want to update other parameters. I'm assuming you don't consider this a blocking issue? I think implementing this change would be more appropriate once I've added transforms to renderable objects.

src/renderers/webgl_bindings.ts
shlomnissan: Each "attribute" is assigned to a separate buffer. The index specifies the location of GLSL input variables. Depending on our shader management approach, the index might need to be specific as well (i.e. 0 = position, 1 = normals, etc). While we can technically load all data into a single buffer (even if it's not interleaved), we still need to map the attributes to different variables.

To answer your question, the index needs to be unique per buffer, but it doesn't need to be unique across all objects. It's related to the shader code, and we may want to keep it consistent for every object.

src/objects/cameras/perspective_camera.ts
aganders3: Definitely not blocking!

Issue #21, opened 2024-09-05
Support loading a 2D region from an OME-Zarr image


andy-sweet:
The main goal of this issue is to add basic support for loading a 2D region from an OME-Zarr image.

The implementation should have the following the features.

- Support [v0.4 of the OME-NGFF specification](https://ngff.openmicroscopy.org/0.4/).
- Exercise the [layer/data-type/data-source technical design proposal of the viewer components](https://docs.google.com/document/d/1aw0TVdweqjKjqZv9XHW-lC5K3RRSMezw_Ue2V23uKO0/edit#heading=h.f3a0okcg78n1).
- Data should be loaded asynchronously (i.e. it should not block the main event loop).
- At least one example application that uses this implementation to display the loaded 2D image data as a texture (depends on support for textures).

It does not need any of the following features, though may include them and should use them to inform technical design.

- Tiled/progressive loading of sub-regions within the requested region.
- Support non-axis aligned regions. 
- Use units of the OME-Zarr image.

Another goal of this issue is to inform the technical design of data loading. There is [a work-in-progress design document](https://docs.google.com/document/d/1b8NhTejl7LIKUO0g4tCuIgo0_Ygk2oeQcJjKs_vDDLM/edit#heading=h.wx60qweyuyn9) on that topic, but it is missing many important details. This work should help fill in some of those.

andy-sweet:
In some early progress, I've been using [zarrita](https://zarrita.dev/) instead of zarrjs. I haven't got something functional yet, but my impression is that the API is pretty interchangeable. And the purported support for zarr v3 is nice, though not immediately useful for this issue until there is a stable OME-Zarr spec for zarr v3.

PR #25, opened 2024-09-10
Render texture

### LINKED ISSUE #22: Add a Texture2D class
- Implement a Texture2D class to handle image loading
- Develop a GLTexture2D class that generates a 2D texture in WebGL and transfers image data to the GPU
- Incorporate an optional texture property into the Mesh type
- Enhance the WebGL renderer and modify shader code to display the texture on a plane

shlomnissan:
### Summary

This PR adds all the necessary components to render a texture. Specifically:
- It introduces rudimentary layer states.
- It adds a virtual `update` function for layers that will be called every frame.
- It implements a Texture base class and a derived Texture2D class.
- It includes a basic WebGL texture manager.

### Notes
- This PR doesn't account for multiple textures using texture units.
- Currently, Mesh combines geometry (currently MeshSource) and textures. We may eventually need a simpler "material" abstraction, similar to three.js.

### Related Issue
Closes #22 

### Tests & Checks

I manually tested this change, verifying that no errors occurred. See the visual verification screenshot below.

Screenshot 2024-09-10 at 9 20 54 AM


src/objects/textures/texture.ts
```diff
@@ -0,0 +1,7 @@
+import { RenderableObject } from "core/renderable_object";
```
shlomnissan, line 1: Inheriting from RenderableObject is temporary.

src/objects/textures/texture_2d.ts
```diff
@@ -0,0 +1,35 @@
+import { Texture } from "./texture";
+
+export class Texture2D extends Texture {
```
shlomnissan, line 3: At this point, this class is simply a wrapper around an image. However, we can start adding texture properties that are shared and supported across multiple rendering backends, such as format, wrapping, filtering, etc. when needed.

src/renderers/webgl_textures.ts
```diff
@@ -0,0 +1,75 @@
...
+    }
+    return texture;
+  }
+
+  private configuredTexture2D(texture: Texture2D) {
+    const gl = this.gl_;
+    const format = gl.RGBA;
+    const type = gl.UNSIGNED_BYTE;
+    const image = texture.image;
+
```
shlomnissan, line 69: The following configuration will be based on Texture2D properties.

public/texture_test.png
andy-sweet, line 1: Nit: would be good to add a README or similar in this directory to indicate the sources of these images.

src/objects/textures/texture_2d.ts
```diff
@@ -0,0 +1,35 @@
...
+  private readonly image_ = new Image();
+  private loaded_ = false;
+
+  constructor(imagePath: string) {
+    super();
+    this.image_.src = imagePath;
+
+    // using bind to ensure that the callbacks have the correct context
+    this.image_.onload = this.onLoad.bind(this);
+    this.image_.onerror = this.onError.bind(this);
```
andy-sweet, line 13: This `Texture` constructor is very coupled to an `HTMLImageElement` and its loading pattern. Given that's the only thing we have right now, that's fine, but it's also the first thing I would look to decouple when adding an OME-Zarr loader.

Do you have any thoughts about what to pass through the constructor in general? I'm thinking that width, height, and pixel format (e.g. RGBA) could be the bare minimum, but there might be others. I'm thinking [canvas' `ImageData`](https://developer.mozilla.org/en-US/docs/Web/API/ImageData/ImageData) is not a bad place to start.

src/layers/single_mesh_layer.ts
```diff
@@ -1,13 +1,21 @@
...
     super();
+    this.state_ = "loading";
+    this.addObject(new Mesh(this.plane_.meshSource, this.texture_));
+  }
 
-    const plane = new PlaneGeometry(5, 5, 1, 1);
-    const mesh = new Mesh(plane.meshSource);
-    this.addObject(mesh);
+  public update(): void {
+    if (this.texture_.loaded) {
```
andy-sweet, line 17: If we wanted to support custom loaders (i.e. outside of `HTMLImageElement`'s), then I assume this is where the driving the logic should be.

E.g. If I wanted to do something really specific for a first pass to support #21, I could define an `OmeZarrImageLayer` that asynchronously fetches the image data for the texture, then updates a similar `loaded` state once it's done.

andy-sweet (approved):
This looks good to me and should be good enough as a base for me to implement some other image data loading (i.e. OME-Zarr). I had some general questions/advice about how to proceed with some of those next steps, so would appreciate some thoughts there.

public/texture_test.png
shlomnissan: It's just a random screenshot that I assumed would be deleted quickly.

src/objects/textures/texture_2d.ts
shlomnissan: > the first thing I would look to decouple when adding an OME-Zarr loader.

100%. This aligns with my expectations. The goal of this PR (which I should have stated more clearly in the summary) is to display a texture on the screen. I tried to create stable abstractions, but this coupling is certainly temporary.

> Do you have any thoughts about what to pass through the constructor in general?

My initial intuition was that a texture will have an instance of an `Image` which will include the width, height, and pixel format, and maybe even the data itself, and that this constructor will accept an image, and texture-specific properties like wrapping, filtering, etc. I’m not sure if this maps well to the data-loading patterns that you’re exploring so I will leave it to you to decide.

Take a look at [DataTexture.js](https://github.com/mrdoob/three.js/blob/dev/src/textures/DataTexture.js). It includes the parameters you mentioned, as well as some we might want to incorporate in the future (if and when they become necessary).

src/layers/single_mesh_layer.ts
shlomnissan: That's what I had in mind. I'm unsure about what will drive subsequent fetches. For example, a layer showing an image at a certain time-point might listen to an onTimeSlider change event, triggering changes outside the update function. However, the update function itself might be responsible for resolving the layer’s state—if that makes sense.

Issue #27, opened 2024-09-11
Add React and SDS to the existing example


aganders3:
The [existing example](https://github.com/chanzuckerberg/imaging-active-learning/blob/main/examples/index.html) uses the viewer library to display a mesh. We'd like to use React and SDS to develop the additional UI for the prototype application. The first step is changing the example to use these libraries. 

* Add React and SDS dependencies
* Connect the renderer to a canvas on a React-rendered page
* Use SDS to add a header component

andy-sweet:
As we now have two examples, I'll leave those alone as plain typescript examples. Instead, I'll target creating a skeleton application for the prototype.

PR #28, opened 2024-09-11
Render a 2D image from an OME-Zarr 5D volumetric time-series

### LINKED ISSUE #21: Support loading a 2D region from an OME-Zarr image

andy-sweet:
### Summary
This adds a rudimentary implementation of asynchronously loading then rendering a 2D slice from an OME-Zarr multi-scale volume. This implements enough to satisfy #21, but there are plenty of future design choices needed before this is more generally useful (e.g. how to support OME-Zarr images with different dtypes).

This adds a dependency on zarrita as a way to expedite the implementation of the prototype (and to also exercise zarrita as a solid dependency that we might want to upstream to), but we may consider implementing our own zarr chunk fetcher in the future.

Screenshot 2024-09-16 at 4 14 04 PM

### Related Issue
Closes #21

### Tests & Checks
Manually tested this with the newly added example.


examples/ome_zarr.ts
```diff
@@ -0,0 +1,31 @@
...
+  OmeZarrMultiscaleVolumeSource,
+} from "@";
+
+const url =
+  "https://public.czbiohub.org/royerlab/zebrahub/imaging/single-objective/ZSNS001.ome.zarr/";
+const layersManager = new LayerManager();
+const renderer = new WebGLRenderer("#canvas");
+const camera = new PerspectiveCamera(60, renderer.width / renderer.height);
+
+// Source is 5D, so provide indices at 3 dimensions to project to 2D.
```
andy-sweet: I wanted to at least pull out the region definition into the example script, so it can be easily modified. Eventually, I'd expect the region to be automatically defined by canvas and camera properties and passed through `Layer.update` (or similar).

src/layers/ome_zarr_2d_slice_layer.ts
```diff
@@ -0,0 +1,35 @@
+import { Layer } from "core/layer";
+import { Mesh } from "objects/renderable/mesh";
+import { PlaneGeometry } from "objects/geometry/plane_geometry";
+import { DataTexture2D } from "objects/textures/data_texture_2d";
+import { OmeZarrMultiscaleVolumeSource } from "data/ome_zarr_source";
+import { Region } from "data/region";
+
+export class OmeZarr2DSliceLayer extends Layer {
```
andy-sweet: I interpret a layer as being responsible for maintaining the geometry/textures/attributes associated with some view/slice/subregion of some data source (here `OmeZarrMultiscaleVolumeSource`).

Eventually, I'd expect `Layer.update` to pass through that view's associated region. And the layer is responsible for mapping from that region to its data source, then executing the associated async load, and updating the geometry/textures/attributes so the renderer can render the updates the next time its polling detects a change.

shlomnissan: I'd like your thoughts on separating the file type from the data source. In this case, the data type would be OME Zarr, and the data source would be a multi-scale volume. While this might be over-abstraction at this point, ideally:

- A 2D slice layer could use multiple data sources (initially, only `MultiscaleVolumeSource` will be supported).
- A `MultiscaleVolumeSource` could be exported from multiple file types (initially, only OME Zarr will be supported).

This approach mirrors Neuroglancer's implementation. I considered it a potential template for us to follow, though I haven't spent too much time considering the trade-offs.

examples/index.html
```diff
@@ -21,6 +21,6 @@
   </head>
   <body>
     <canvas id="canvas"></canvas>
-    <script type="module" src="./main.ts"></script>
+    <script type="module" src="./ome_zarr.ts"></script>
```
shlomnissan: It would be nice to add a sidebar to the HTML file that allows toggling between different examples. It's not a priority, of course, and certainly not related to this PR, but I think it will streamline development.

examples/ome_zarr.ts
```diff
@@ -0,0 +1,31 @@
...
+  OmeZarr2DSliceLayer,
+  WebGLRenderer,
+  OmeZarrMultiscaleVolumeSource,
+} from "@";
+
+const url =
+  "https://public.czbiohub.org/royerlab/zebrahub/imaging/single-objective/ZSNS001.ome.zarr/";
+const layersManager = new LayerManager();
+const renderer = new WebGLRenderer("#canvas");
+const camera = new PerspectiveCamera(60, renderer.width / renderer.height);
```
shlomnissan: Note: We need to reconsider the camera API. The renderer automatically adjusts the aspect ratio on window resize, so we must choose a consistent approach. We can automatically calculate the aspect ratio when attaching the camera to the renderer if a value was not provided to the constructor, or always require the user to update the aspect ratio, even during window resizes. I believe the latter approach might be preferable.

examples/ome_zarr.ts
```diff
@@ -0,0 +1,31 @@
...
+// Source is 5D, so provide indices at 3 dimensions to project to 2D.
+const region = [
+  { dimension: "t", start: 0 },
+  { dimension: "c", start: 0 },
+  { dimension: "z", start: 277.76 },
+];
+OmeZarrMultiscaleVolumeSource.open(url).then((source) => {
+  const layer = new OmeZarr2DSliceLayer(source, region);
+  layersManager.add(layer);
+});
```
shlomnissan: I believe a more robust public API would involve creating an instance of `OmeZarrMultiscaleVolumeSource` with the URL as a constructor parameter. This object would then be passed to the layer, which would internally handle file loading and subsequently load the chunks. This approach would simplify the API for users and encapsulate all loading-related functionality within the layer itself.

src/core/layer.ts
```diff
@@ -1,6 +1,6 @@
 import { RenderableObject } from "./renderable_object";
 
-type LayerState = "loading" | "ready";
+type LayerState = "initialized" | "loading" | "ready";
```
shlomnissan, line 3: I believe `initialized` or perhaps `initializing` would be a more appropriate default. The `state_` variable shouldn't be set to `ready` by default.

src/data/ome_zarr_source.ts
```diff
@@ -0,0 +1,95 @@
+import * as zarr from "zarrita";
+import { Slice as ZarrSlice } from "@zarrita/indexing";
+
+import { DataLoadInput, VolumeChunk } from "data/region";
+
+interface ScaleTransform {
```
shlomnissan: I understand that `type` and `interface` can be used interchangeably for simple types in TypeScript. The `interface` keyword offers slightly more ergonomic features, such as declaration merging and support for virtual classes with `class implements`. While I've never considered this flexibility "a good thing", I don't have strong feelings about it either way. I'm curious if the use of `interface` here is deliberate, and I'd like to know your thoughts on what our default approach should be.

src/layers/ome_zarr_2d_slice_layer.ts
andy-sweet: I'm open to it, but don't see the motivation yet either here (because our library is simple) or in neuroglancer (because i don't understand everything there).

In order to initially support multiple formats, I think we could just have an interface similar to the following

```typescript
interface ImageLayer<ArrayType> {
  async loadChunks(input: DataLoadInput): Promise<VolumeChunk<ArrayType>[]>;
};
```

that is implemented by different data sources which are coupled with their respective formats.

Right now, most of the code in `OmeZarrMultiscaleVolumeSource` is pretty tightly coupled to the OME-Zarr format. I can see some logic that could be factored out in anticipation of other formats, but it feels mostly functional right now (e.g. mapping the input region to the source's dimensions).

src/data/ome_zarr_source.ts
```diff
@@ -0,0 +1,95 @@
...
+}
+
+interface Dataset {
+  path: string;
+  coordinateTransformations: Array<ScaleTransform>;
+}
+
+interface Axis {
+  name: string;
+  type?: "space" | "time" | "channel";
```
shlomnissan: What does it mean for an axis not to have a type? I tend to prefer named default values for union types (e.g., “none”) because it’s more explicit.

src/data/ome_zarr_source.ts
```diff
@@ -0,0 +1,95 @@
...
+}
+
+export class OmeZarrMultiscaleVolumeSource {
+  root: zarr.Group<zarr.FetchStore>;
+  axes: Array<Axis>;
+  datasets: Array<Dataset>;
+
+  constructor(root: zarr.Group<zarr.FetchStore>) {
+    this.root = root;
+    const multiscales = this.root.attrs.multiscales as Array<Multiscale>;
```
shlomnissan: Is this cast safe? Can we guarantee that "multiscales" is a valid attribute? It doesn't appear to be when I examine the underlying type. I'm uncertain about the outcome of casting an undefined value to an array. However, if it simply triggers the error case, that should be acceptable.

src/data/ome_zarr_source.ts
```diff
@@ -0,0 +1,95 @@
...
+      throw new Error(
+        `Can only handle one multiscale image. Found ${multiscales.length}`
+      );
+    }
+    const image = multiscales[0];
+    this.axes = image.axes;
+    this.datasets = image.datasets;
+  }
+
+  static async open(url: string): Promise<OmeZarrMultiscaleVolumeSource> {
```
shlomnissan: - What is the benefit of having a factory member function, as opposed to creating an uninitialized instance of the object and having a public method that loads the data?
- If there is merit to sticking with a factory function, I feel it should be called by the layer (as mentioned earlier). We could also choose a more suitable name for the factory, such as `create`, since it doesn't just open the file but creates an instance of the class.

src/data/ome_zarr_multiscale_image_source.ts
```diff
@@ -0,0 +1,134 @@
...
+    this.root_ = root;
+    const multiscales = this.root_.attrs.multiscales as Array<Multiscale>;
+    if (multiscales.length !== 1) {
+      throw new Error(
+        `Can only handle one multiscale image. Found ${multiscales.length}`
+      );
+    }
+    const image = multiscales[0];
+    this.axes_ = image.axes;
+    this.datasets_ = image.datasets;
```
shlomnissan:
```suggestion
    const { axes, datasets } = multiscales[0];
    this.axes = axes;
    this.datasets = datasets;
```

src/data/ome_zarr_multiscale_image_source.ts
```diff
@@ -0,0 +1,134 @@
...
+  type: "translation";
+  translation: Array<number>;
+}
+
+interface ScaleTransform {
+  type: "scale";
+  scale: Array<number>;
+}
+
+type Transform = IdentityTransform | TranslationTransform | ScaleTransform;
```
shlomnissan: The name of this type might conflict with the Transform object I'm considering creating. Can we come up with a more specific name for it?

src/data/ome_zarr_multiscale_image_source.ts
```diff
@@ -0,0 +1,134 @@
...
+    const image = multiscales[0];
+    this.axes_ = image.axes;
+    this.datasets_ = image.datasets;
+  }
+
+  // Opens an OME-Zarr multiscale volume from the URL of the top-level Zarr group.
+  static async open(url: string): Promise<OmeZarrMultiscaleImageSource> {
+    const store = new zarr.FetchStore(url);
+    const root = await zarr.open.v2(store, { kind: "group" });
+    console.debug("opened root ", root, root.attrs);
```
shlomnissan: Note: I wonder if it's worth setting up a simple logging mechanism to assign categories and priorities to individual logs. I suspect that filtering logs will be beneficial even in early stages of development.

src/data/ome_zarr_multiscale_image_source.ts
```diff
@@ -0,0 +1,131 @@
...
+
+  // Opens an OME-Zarr multiscale image from the URL of its Zarr group.
+  static async open(url: string): Promise<OmeZarrMultiscaleImageSource> {
+    const store = new zarr.FetchStore(url);
+    const root = await zarr.open.v2(store, { kind: "group" });
+    console.debug("opened root ", root, root.attrs);
+    return new OmeZarrMultiscaleImageSource(root);
+  }
+
+  async loadChunks(input: DataLoadInput): Promise<ImageChunk<Uint16Array>[]> {
```
shlomnissan: Do we consistently return a `Uint16Array` for now? If so, why do we need a typed `VolumeChunk`?

src/data/ome_zarr_multiscale_image_source.ts
```diff
@@ -0,0 +1,131 @@
...
+    // TODO: use the input to determine what level to load.
+    // For now, just use the lowest.
+    const datasetIndex = this.datasets_.length - 1;
```
shlomnissan: Consider renaming this variable to `lowestLevelIdx` if the last index represents the lowest level and the first represents the highest (assuming multiple levels are available). This would clarify the variable's purpose and eliminate the need for an explanatory comment.

src/data/ome_zarr_multiscale_image_source.ts
```diff
@@ -0,0 +1,131 @@
...
+    const dataset = this.datasets_[datasetIndex];
+    const indices = regionToIndices(input.region, dataset, this.axes_);
+    console.debug("loading dataset with indices", dataset, indices);
+
+    const array = await zarr.open.v2(this.root_.resolve(dataset.path), {
```
shlomnissan: What happens if any of these requests fail? Should we implement a retry mechanism? This might be premature, but at the very least, all `awaits`—especially those resulting in network calls—should be wrapped in try/catch blocks to report errors.

src/data/ome_zarr_multiscale_image_source.ts
```diff
@@ -0,0 +1,131 @@
...
+function regionToIndices(
+  region: Region,
+  dataset: Dataset,
+  axes: Array<Axis>
+): Array<Slice | number | null> {
+  const indices: Array<Slice | number | null> = [];
+  for (const [i, axis] of axes.entries()) {
+    let index = null;
+    const regionIndex = region.find((s) => s.dimension == axis.name);
+    if (regionIndex) {
```
shlomnissan: If this is false, does it mean that we might inadvertently push null or a duplicate value to the indices array. Is this behavior intentional?

src/layers/image_layer.ts
```diff
@@ -0,0 +1,48 @@
...
+        const chunk = chunks[0];
+        if (chunk.shape.length !== 2) {
+          throw new Error(
+            `Expected a 2D chunk. Instead chunk has shape ${chunk.shape}`
+          );
+        }
+        const [height, width] = chunk.shape;
+        const texture = new DataTexture2D(chunk.data, width, height);
+        this.addObject(new Mesh(this.plane_.meshSource, texture));
+        this.state_ = "ready";
```
shlomnissan: Note: It would be beneficial to have a mutator handle this assignment. This approach would enable us to incorporate logging and additional code, such as hooks for state transitions.

src/objects/textures/data_texture_2d.ts
```diff
@@ -0,0 +1,30 @@
+import { Texture } from "./texture";
+
+export class DataTexture2D extends Texture {
```
shlomnissan: I'm fine with renaming this to `2DTexture` and removing the object I created. As you mentioned in my PR, we don't need to tie the texture to an image element.

src/renderers/webgl_textures.ts
```diff
@@ -72,4 +77,24 @@ export class WebGLTextures {
     gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_T, gl.CLAMP_TO_EDGE);
     gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MIN_FILTER, gl.LINEAR);
   }
+
+  private configuredDataTexture2D(texture: DataTexture2D) {
+    const gl = this.gl_;
+    gl.texImage2D(
+      gl.TEXTURE_2D, // target
+      0, // level
```
shlomnissan: Nit: While it's not a major issue here due to the stable interface, I view comments as a subtle form of coupling. We can improve clarity by creating named variables for unclear parameters, even if they're used only once (e.g., const level = 1).

src/renderers/webgl_textures.ts
```diff
@@ -72,4 +77,24 @@ export class WebGLTextures {
...
+      texture.width,
+      texture.height,
+      0, // border
+      gl.RED_INTEGER, // format
+      gl.UNSIGNED_SHORT, // type
+      texture.data // source
+    );
+    gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_S, gl.CLAMP_TO_EDGE);
+    gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_T, gl.CLAMP_TO_EDGE);
+    // I think we need NEAREST for R16UI.
```
shlomnissan: Reminder: Remove this comment or change it to explain why we need NEAREST for R16UI.

shlomnissan (commented):
I'm submitting this as an initial review. The airplane Wi-Fi is unreliable, and the PR has already been updated. I'll do another pass later today or tomorrow.

src/data/region.ts
```diff
@@ -0,0 +1,42 @@
...
+export interface DataLoadInput {
+  region: Region;
+  renderSize?: [number, number];
+}
+
+// The output of loading a region of a data source.
+export interface ImageChunk<ArrayType> {
+  region: Region;
+  data: ArrayType;
+  shape: Array<number>;
```
shlomnissan: We can define a Shape type and map the array onto it, allowing clients of the data source to avoid dealing with implementation-specific details.

andy-sweet:
This is ready for review now.

But it is not mergeable because I still need to resolve how to support multiple shaders for the RGB HTML `Image` and the uint16 OME-Zarr image chunks.

src/data/ome_zarr_multiscale_image_source.ts
andy-sweet: Maybe, but this type is not exported so it might be OK to let the local namespace take precedence here. It is possible that we might need to import the `Transform` type you mention here though, in which case we could either alias that or we could use the more verbose OME-NGFF term of `CoordinateTransform` here instead. I'll probably wait for that to be a problem before doing anything though.

src/data/ome_zarr_multiscale_image_source.ts
andy-sweet: Good idea. I used `lowestResolutionIndex` and removed the comment, but kept the TODO.

src/renderers/webgl_textures.ts
andy-sweet: Interesting, I hadn't considered this particular case of comments to name positional arguments as something to avoid. However, I'm not sure that the coupling here is any worse than the coupling of explicit variable names. Both have the potential for the same mistake (i.e. mapping a name to the wrong position), while this has the benefit of being more concise. Maybe a better argument for variables here is that the mistake is more clear during interactive debugging?

src/data/ome_zarr_multiscale_image_source.ts
andy-sweet: Is the goal of the try/catch to throw an error with a more application specific message? If so, that's reasonable.

src/renderers/webgl_textures.ts
andy-sweet: Good point. I believe that generally integer valued textures not texture filterable, so we must use `NEAREST` here for both these filters (instead of the default `LINEAR`), so I added a comment to explain that. I also added the best URL source I could find for this, which is https://webgl2fundamentals.org/webgl/lessons/webgl-data-textures.html but let me know if you can find a more original source (e.g. from the OpenGL or WebGL2 specs).

src/data/ome_zarr_source.ts
andy-sweet: From [the OME-NGFF spec](https://ngff.openmicroscopy.org/0.4/#axes-md)

> SHOULD contain the field "type". It SHOULD be one of "space", "time" or "channel", but MAY take other values for custom axis types that are not part of this specification yet.

src/data/ome_zarr_multiscale_image_source.ts
andy-sweet: Yes. In zarrita a `null` index effectively means to take the whole extent of an array dimension (e.g equivalent to `:` in numpy).

src/data/region.ts
andy-sweet: I'm confused. By `shape`, I mean the `shape` of the array equivalent to numpy's `ndarray.shape`, which is effectively just an array of numbers and should be easy for clients to consume.

src/data/ome_zarr_multiscale_image_source.ts
andy-sweet: The example OME-Zarr dataset has a dtype of uint16 and the implementation fails if it finds another dtype, so I could just make `ImageChunk` non generic for now. But I'd hope to solve the more general design problem for loading and rendering in the not too distant future.

I'll remove the generic for now.

src/data/ome_zarr_multiscale_image_source.ts
andy-sweet: This seems mostly stylistic and not a big improvement, so I'll probably ignore (especially since the indentation is broken in the suggestion).

src/data/ome_zarr_source.ts
andy-sweet: It is not safe. For that to be the case, we'd have to validate the generally unstructured attributes mapping to ensure it reflects at least this part of the OME-Zarr spec.

src/data/ome_zarr_source.ts
andy-sweet: As a relative typescript beginner, not deliberate. I'm just looking for a way to define structured duck typing. Happy to switch to `type ... = {` instead.

examples/ome_zarr.ts
andy-sweet: I like the simplification for the client, so I'm down to do some version of this. Probably the hardest part to work out here are meaningful type names.

src/layers/ome_zarr_2d_slice_layer.ts
shlomnissan: I think that’s reasonable :+1:

src/data/ome_zarr_multiscale_image_source.ts
shlomnissan: It's a nit, so feel free to disregard. I believe it slightly improves readability, but I don't have a compelling argument for it in this particular case.

src/data/ome_zarr_multiscale_image_source.ts
shlomnissan: > this type is not exported so it might be OK

Even within a local scope, we should be thoughtful about the names we use. If `Transform` means one thing in 90% of our code but something slightly different here, it could lead to confusion when reasoning about the code.

src/data/ome_zarr_source.ts
shlomnissan: I've worked with TypeScript before, but I'm by no means an expert, so I don't have a definitive answer on "what's the right thing" here. Apart from a few minor differences, it seems like they can be used interchangeably. However, from what I've seen, `type` appears to be the preferred approach for "plain-old-data" types, and `interface` appears more common for defining virtual (abstract) classes, so we can stick to that for now.

To be clear, it’s not a big deal either way.

src/data/ome_zarr_multiscale_image_source.ts
shlomnissan: > The goal of the try/catch to throw an error with a more application specific message

Yep, that's the goal.

src/data/ome_zarr_multiscale_image_source.ts
shlomnissan: Without this additional context, this code may appear to have a bug. I suggest we add a comment or, better yet, modify the code slightly to clarify that pushing null to the array is intentional. This is a minor suggestion; feel free to disregard if you disagree.

src/renderers/webgl_textures.ts
shlomnissan: Generally, I believe most comments are unnecessary at best. My rationale is twofold: computers can't understand them (which relates to your point about debugging), and developers often struggle to keep them up-to-date (creating a subtle coupling problem). Of course, there are exceptions—some code is inherently complex, automatic documentation generators can be beneficial, etc. You're right that the argument for preferring variable names over positional argument comments in the context of "coupling" isn't very convincing for a stable public API. If this were our own implementation, we could have created a data structure to represent the parameters. I agree that a stronger argument lies in the debugging benefits.

src/data/region.ts
shlomnissan: I'm not very familiar with `ndarray.shape`, so I'm happy to defer. I'm concerned about lines like 
```
const [height, width] = chunk.shape;
```
If this is how we're using the shape property, I'd prefer returning a type with explicit width and height members, rather than making assumptions about the array structure.

shlomnissan (approved):
This PR is excellent! I have some questions regarding the abstraction level for data types and layers. However, I'm uncertain whether addressing these concerns is worthwhile at this stage, given that it’s unclear (to me) how things might evolve.

shlomnissan:
A few open threads remain, which you've mentioned will be addressed—particularly moving file loading from the example code to the layer. I assume these will be resolved before merging, though I don't consider them blocking issues.

src/layers/ome_zarr_2d_slice_layer.ts
andy-sweet: I do now have a separation between an image source, which a user creates and passes to an image layer, and an image loader which is an implementation detail that a user doesn't have to think about. I'm not sure if that's quite the same as neuroglancer's terminology, but it makes the intended API more concise (since the user does not have to worry about loading and related state changes).

src/renderers/webgl_textures.ts
andy-sweet: I pulled out some less clear arguments into local variables (e.g. `level`, `internalFormat`) and left others as bare values or property references that are easier to interpret (e.g. `gl.TEXTURE_2D` and `texture.width`).

src/data/region.ts
andy-sweet: I do now throw if the length of shape is not 2 to at least fail when necessary with a meaningful message. But now I realize this might be a naming and scope issue. I intended `ImageChunk` to be n-dimensional (e.g. handling 2D image tiles, and 3D image subvolumes), so I could think about the required design/implementation for that.

But that is probably too premature - especially for something we want to merge and when we probably only initially want 2D presentation of ND data. I think I'm going to change all the image stuff (including chunks) to be explicitly 2D only.

examples/index.html
aganders3: I think this could fit with the work in #27

andy-sweet:
The last thing I want to address before making this mergeable is to make dedicated shaders for the image, which

- avoids modifying the existing mesh shaders that are used with an HTML Image texture data source;
- allows us to define image-specific shader parameters (e.g. a range or function that could be used to normalize the pixel values in the fragment shader), though I'm not sure if I'll get to that in this PR

src/layers/ome_zarr_2d_slice_layer.ts
aganders3: > I do now have a separation between an image source, which a user creates and passes to an image layer, and an image loader which is an implementation detail that a user doesn't have to think about.

I'll look more closely but my gut reaction to this is positive. It sounds like a good separation such that two layers could share a loader if data is coming from the same place. It also seems like a practical separation for eventually moving the loader to a WebWorker.

src/core/layer.ts
```diff
@@ -1,6 +1,6 @@
 import { RenderableObject } from "./renderable_object";
 
-type LayerState = "loading" | "ready";
```
aganders3, line 3: We should carefully consider the names and meanings for these states, and how they relate to the whole system state.

src/data/region.ts
```diff
@@ -0,0 +1,15 @@
+// The semi-closed interval [start, stop).
+export interface Interval {
+  start: number;
+  stop: number;
+}
+
+// An index for a specific dimension or axis in a region.
+// TODO: add a unit for the index value(s).
+export interface DimensionalIndex {
+  dimension: string;
```
aganders3, line 11: I'm curious why the difference in naming here between "dimension" and "axis" (or perhaps more accurately `axis.name`/`axis_name`).

src/layers/image_layer.ts
```diff
@@ -0,0 +1,71 @@
...
+        this.open();
+        break;
+      case "opened":
+        this.load(this.region_);
+        break;
+      case "loading":
+      case "ready":
+        break;
+    }
+  }
```
aganders3: Not necessary for this PR but this is a neat trick to get compile-time checking that your switch statement is exhaustive.
```suggestion
  public update(): void {
    switch (this.state) {
      case "initialized":
        this.open();
        break;
      case "opened":
        this.load(this.region_);
        break;
      case "loading":
      case "ready":
        break;
      default: {
        const exhaustiveCheck: never = this.state;
        throw new Error(`Unhandled LayerState case: ${exhaustiveCheck}`);
      }
    }
  }
```

src/layers/image_layer.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  private async open() {
+    this.loader_ = await this.source_.open();
+    this.state_ = "opened";
+  }
+
+  private async load(region: Region) {
+    if (this.loader_ === null) {
+      throw new Error(`Trying to load chunks from unopened source.`);
+    }
```
aganders3: nit: I would use `state_` in this check to be consistent. If the state does not reflect the presence of the `loader_` then it reveals a bug.

aganders3: I guess on that note - do we need these distinct states? If we try to `load` before `open` could we just do the open at that point?

src/data/ome_zarr_image_loader.ts
```diff
@@ -0,0 +1,140 @@
...
+      throw new Error(
+        `Expected to receive a 2D subarray. Instead chunk has shape ${subarray.shape}`
+      );
+    }
+
+    if (subarray.stride[1] !== 1) {
+      throw new Error(
+        `Expected to receive packed rows. Instead found a column stride of ${subarray.stride[1]}`
+      );
+    }
```
aganders3: Confirming my own understanding here - will this mean `subarray.stride[0]` (aka `rowLength`) will always be the same as `subarray.shape[1]`?

I'm not sure I like the `rowLength` name if this ever may change, but that's not blocking.

src/renderers/shaders/mesh_frag.glsl
```diff
@@ -1,13 +1,14 @@
...
 layout (location = 0) out vec4 fragColor;
 
-uniform sampler2D texture0;
-
+uniform usampler2D texture0;
 in vec2 TexCoords;
 
 void main() {
-    fragColor = vec4(texture(texture0, TexCoords).rgb, 1.0);
+    float value = float(texture(texture0, TexCoords).r) / 256.0;
```
aganders3: More confirming my own understanding: is 256.0 here just a hardcode for contrast limits?

src/renderers/webgl_textures.ts
```diff
@@ -37,9 +38,10 @@ export class WebGLTextures {
   private textureType(texture: Texture) {
     switch (texture.type) {
       case "Texture2D":
+      case "Uint16Texture2D":
         return this.gl_.TEXTURE_2D;
       default:
-        throw new Error(`Unknown texture type`);
+        throw new Error(`Unknown texture type ${texture.type}`);
```
aganders3, line 44: See previous comment about getting compile-time exhaustiveness checking using `never`

src/core/layer.ts
andy-sweet: Changed the default of `Layer.state_` to be `"initialized"`.

andy-sweet:
I think this is mergeable now. My solution to support both existing examples is to define a new fragment shader specifically for uint16 images, which I chose as the shortest path to functionality. Another option is to create an entirely new renderable object (e.g. image2d) that will only ever a plane geometry, but I think that can be done in a follow-up PR.

More comments and questions welcome.

src/layers/image_layer.ts
```diff
@@ -0,0 +1,71 @@
...
+  loadChunks(input: Region): Promise<ImageChunk[]>;
+}
+
+// Loads data from an image source into renderable objects.
+export class ImageLayer extends Layer {
+  private source_: ImageLayerSource;
+  // TODO: remove this when region is passed through to update.
+  private region_: Region;
+  // TODO: plane geometry should be defined by data source extents and region.
+  private plane_ = new PlaneGeometry(3, 3, 1, 1);
```
aganders3: I'm not sure this is the case, but happy to keep it as an open issue if you have ideas.

aganders3 (approved):
Great work! This is really exciting to see. None of my comments are blocking. Most are actually questions to just make sure I understand what you've done.

src/data/region.ts
andy-sweet: This is just a personal preference of dimension over axis, partially due to dimension's associated with dimensional analysis and units.

[OME-NGFF uses `axis` and describes it as a dimension](https://ngff.openmicroscopy.org/latest/#axes-md), and I didn't want to couple the libraries names and definitions too tightly to NGFF.

But I guess `axis` is more concise and I don't have strong opinions here, so happy to change if desired.

src/layers/image_layer.ts
andy-sweet: I was actually wanting for something like this, so added. Thanks!

src/layers/image_layer.ts
andy-sweet: I switched to using `state_` instead.

I think your comment and question revealed a bug though.

I was trying to use these distinct states to control the number of times `open` and `load` are called. I want them to be called exactly once to avoid unnecessary async tasks and to avoid buggy interleaving (e.g. changing the state to `"loading"` prevents a second call to `load`). The bug is that I have no `"opening"` state, so it is possible to get a second call to `open` that might interleave with a call to `load`, in which case we will add a second object.

Therefore, instead of removing a state, I'm planning to add an `"opening"` state. Does that make sense?
Actually, re-reading your comment, I think I prefer removing the `"opened"` state and moving the body of that method into `load`. I was anticipating progressive rendering where we may want to create the objects after opening, and update the textures with `load`, but we're not there in this PR. Sorry for the stream of consciousness!

All this reiterates your point that we should think carefully about the loading states. I think polling with a set of well defined states seems like a good place to start, but let me know if you have other ideas.

src/data/ome_zarr_image_loader.ts
andy-sweet: No, I meant that each row is packed tightly (i.e. there are no byte strides/steps between columns). Which probably means this message is confusing and I should update it!

Happy to consider alternatives to `rowLength`. When `ImageChunk` was n-dimensional (in an earlier instance of this PR), I just had the stride, but for 2D that seems excessive.

src/renderers/shaders/mesh_frag.glsl
andy-sweet: Yes, I added a TODO to explain that in the new `uint16_image_frag.glsl` and also created an issue to support something like user defined contrast limits.

src/layers/image_layer.ts
andy-sweet: See the issue I created for more detailed thoughts: https://github.com/chanzuckerberg/imaging-active-learning/issues/35

Ideally, I'd like the geometry of the plane to define the physical coordinates of the extent of the image, so we can work in some shared space across objects. If we have a transform associated with the geometry, then maybe that can be captured there instead? But in that case, I think we'd still need to know the shape of the lowest resolution array, right?

src/renderers/webgl_textures.ts
andy-sweet: Actually, the type of `texture.type` is just a plain `string` right now, not a union of literal strings. We could change that in the future, but won't make the update in this PR.

src/layers/image_layer.ts
andy-sweet: I removed `"opened"` for now, which simplifies the current code. Though I suspect we'll reintroduce it later likely with other states too.

src/data/ome_zarr_image_loader.ts
andy-sweet: I simplified (and hopefully clarified) the error message here to

`Expected to find a column stride of 1. Instead found ${subarray.stride[1]}`

I'm planning to leave `rowLength` for now since that is consistent with OpenGL terminology. We should feel to change it later.

src/layers/image_layer.ts
shlomnissan: Nice! I had assumed ESLint provided exhaustiveness checks by default, but it doesn't. Let's enable it with `'@typescript-eslint/switch-exhaustiveness-check': 'error'`.

src/layers/image_layer.ts
aganders3: Thanks - makes sense. I suppose something along these lines may make it easier to load tiles or represent other visuals in the same space.

src/data/region.ts
aganders3: I think some people get picky about "channel" in particular not being a "dimension", but in the end it's all discrete. My mind is happy with dimension as it's the "d" in ndarray or ND-Image.

I think the only reason to change it here would be to reserve the word for some specific other meaning, but I don't have one in mind.

src/core/layer.ts
shlomnissan: I agree. We should take the time to think/specify life-cycle states and events for layers. I've been considering life-cycle events in the application space as a way for layers to communicate with the application. For instance:
```ts
const singleMeshLayer = new SingleMeshLayer();

singleMeshLayer.onLoading = () => {
    // ... data is loading, show a spinner?
};

singleMeshLayer.onLoaded = () => {
    // ... data is loaded, hide spinner
};

```
These events could map directly to layer state changes. While this is a simple example, it's possible to envision different layers offering specific events that send data back to the application.

PR #45, opened 2024-09-18
Add geometry buffers to renderable objects

### LINKED ISSUE #24: Refactor "Renderable Objects"
- Create a Geometry type to manage vertex and index data, with vertex data assumed to be interleaved.
- The Geometry type should have configurable attributes corresponding to vertex attributes (item type, item size, offset).
- Each "renderable object" should have a Geometry instance.
- After this refactor, we should be able to remove the MeshSource class, which shouldn't have been considered a "data source" on its own. This will simplify the code-base, making easier to maintain or modify if new requirements come up.

shlomnissan:
### Summary

This PR implements several refactoring tasks discussed in previous
conversations:

- It replaces the "mesh source" type with a geometry type that stores all
vertex data in a single buffer.
- It attaches geometry to every renderable object.
- It introduces a node type for generating unique IDs for non-renderable
objects such as geometry and textures.

### Notes

Before moving on from refactoring, I have a couple of quick follow-ups to address:
1. Disposing of geometry and texture data from the GPU
2. Adding a simple logger class

### Related Issue
Closes #24 

### Tests & Checks

- Manual verification. Everything works as expected.


src/objects/geometry/plane_geometry.ts
```diff
@@ -1,18 +1,16 @@
...
+export class PlaneGeometry extends Geometry {
   constructor(
     width: number,
     height: number,
     widthSegments: number,
     heightSegments: number
   ) {
+    super();
+    const vertex: number[] = [];
+    const index: number[] = [];
```
shlomnissan, line 12: I could likely pre-calculate the vertex and index sizes and write them directly to the buffers, rather than storing the data in intermediate number arrays. However, I don't consider pressing at the moment.

src/renderers/webgl_bindings.ts
```diff
@@ -42,25 +41,37 @@ export class WebGLBindings {
...
+
+    const bufferType = this.gl_.ARRAY_BUFFER;
+    this.gl_.bindBuffer(bufferType, buffer);
+    this.gl_.bufferData(bufferType, vertexData, this.gl_.STATIC_DRAW);
+
+    attributes.forEach((attr) => {
+      let idx = -1;
+      if (attr.type === "position") idx = 0;
+      if (attr.type === "normal") idx = 1;
+      if (attr.type === "uv") idx = 2;
```
shlomnissan, line 58: We may want this mapping to be defined elsewhere, or to be more flexible, perhaps by adding an index to the geometry attribute data structure.

src/core/geometry.ts
```diff
@@ -0,0 +1,52 @@
...
+
+type GeometryAttribute = {
+  type: GeometryAttributeType;
+  itemSize: number;
+  offset: number;
+};
+
+export class Geometry extends Node {
+  protected vertexData_: Float32Array = new Float32Array();
+  protected indexData_: Uint32Array = new Uint32Array();
```
andy-sweet: Nit: the default values of these members are always clobbered in the constructor, so remove the assignments?

```suggestion
  protected vertexData_: Float32Array;
  protected indexData_: Uint32Array;
```

src/renderers/webgl_bindings.ts
andy-sweet: Agreed it would be good to improve this in the future, especially because the order of these interleaved values is also needed when populating `Geometry.vertexData_`.

andy-sweet: Also, should we have some kind of explicit exhaustive check here for `attr.type`, or is that being handled well enough by standard TypeScript checks?

andy-sweet (approved):
Definitely an improvement over `MeshSource`, so looks good to me. Would be nice to have more explicit structure for the interleaved attributes, but that's not blocking and I don't have immediate suggestions.

src/renderers/webgl_renderer.ts
```diff
@@ -20,29 +22,33 @@ export class WebGLRenderer extends Renderer {
...
-    const program = this.getShaderProgram(programName).use();
+  protected renderObject(object: RenderableObject) {
+    const program = this.getShaderProgram(this.getProgramName(object)).use();
     program.setUniformMat4("Projection", this.activeCamera.projectionTransform);
     program.setUniformMat4("ModelView", this.activeCamera.viewTransform);
 
-    this.bindings_.bind(mesh);
-    if (mesh.texture !== null) {
-      this.textures_.bind(mesh.texture);
+    this.bindings_.bind(object);
```
aganders3, line 35: naming nit: seeing this next to `this.textures_.bind()` makes me think "bindings_" and "WebGLBindings" may not be the right name. Maybe `WebGLBuffers` or something?

src/renderers/webgl_renderer.ts
```diff
@@ -20,29 +22,33 @@ export class WebGLRenderer extends Renderer {
...
+
+    if (object.type === "Mesh") {
+      const mesh = object as Mesh;
+      if (mesh.texture !== null) {
+        this.textures_.bind(mesh.texture);
+      }
     }
```
aganders3: Can we avoid checking the object type here? Other renderable objects might have textures as well so it seems like it should be okay.

Also as a future thought - some renderable objects may need multiple textures so `WebGLTextures.bind()` could take a list of them (which may be empty). That would get rid of the conditional here entirely.

src/renderers/webgl_renderer.ts
```diff
@@ -55,6 +61,14 @@ export class WebGLRenderer extends Renderer {
     this.gl.clear(this.gl.COLOR_BUFFER_BIT);
   }
 
+  private getProgramName(object: RenderableObject) {
+    if (object.type === "Mesh") {
+      const mesh = object as Mesh;
+      return mesh.texture?.type === "Uint16Texture2D" ? "uint16Image" : "mesh";
+    }
+    return DEFAULT_PROGRAM;
```
aganders3: Does it make sense to have a default program if the object type is not recognized? Maybe this should throw instead.

src/objects/geometry/plane_geometry.ts
```diff
@@ -31,9 +29,9 @@ export class PlaneGeometry {
 
         // 'z = -5' is temporary until we add a transform to renderable objects
         // which will allow us to control the camera position
-        this.vertices_.push(x, -y, -5);
-        this.normals_.push(0, 1, 0);
-        this.uvs_.push(u, v);
```
aganders3, line 36: nit: I think you can remove these comments and replace `vertex.push(0, 0, 1)` with

```
const normal = [0, 0, 1];
vertex.push(...normal);
```

You could do the same for position but that's pretty clear here.

src/core/geometry.ts
aganders3: This is more than a nit I think. An empty typed array is truthy, so I think having this affects the checks on `if (indexData)` (in `WebGLBindings.bind` and `WebGLRenderer.renderObject`).

src/renderers/webgl_bindings.ts
```diff
@@ -42,25 +41,37 @@ export class WebGLBindings {
...
+
+      this.gl_.vertexAttribPointer(
+        idx,
+        attr.itemSize,
+        this.gl_.FLOAT,
+        false,
+        stride * Float32Array.BYTES_PER_ELEMENT,
+        attr.offset * Float32Array.BYTES_PER_ELEMENT
```
aganders3: This can go into future work, but right now this prevents mixing data types in the interleaved buffer. Could `stride` and `offset` return bytes instead of counts?

src/objects/geometry/plane_geometry.ts
aganders3, line 1: From prior discussion I thought the intent was to _not_ use `Geometry` as a base class. Could this instead be a function that returns a `Geometry` object?

aganders3 (changes requested):
I agree this is definitely an improvement!

I have some concerns that need to be ironed out to marry this with my PR #46 that introduces a new renderable object, but I'm happy to make the needed changes on that PR (or an intermediate one). So I'm happy to approve this in spirit, but I think there are a few changes that should be made or captured as issues.

My main thought here is that renderable objects are strongly (and I think necessarily?) coupled to their shaders. They likely all need different attributes (and orders) and uniforms. This means the object could be aware of its own program (obviating `WebGLRenderer.getProgramName`).

To me this also means `Geometry` should not restrict the available attributes. I think that's basically what you're getting at with this thread: https://github.com/chanzuckerberg/imaging-active-learning/pull/45#discussion_r1765759351.

I don't have a great idea of _where_ this info should live though. My first thought is that it belongs on the renderable object itself, which could provide `getAttributes` and `getUniforms` for example. I guess this turns the `Geometry` class into more of a utility class to make it easier to pack data into an interleaved buffer and bookkeep the stride + offsets.

aganders3 (approved):
I'm preemptively approving after our discussion today so my time zone does not become a blocker. Feel free to merge when ready!

src/objects/geometry/plane_geometry.ts
shlomnissan: We discussed and agreed that it's acceptable for the plane geometry object to inherit directly from the Geometry object. However, we remain open to changing this structure if needed.

PR #46, opened 2024-09-18
Initial Line renderable object + sample layer

### LINKED ISSUE #29: Add a "Line" renderable object
We want to be able to draw 3D lines to visualize cell tracking data (e.g. as part of a "Tracks" layer).

Drawing lines in 3D can be complicated. [This blog post ](https://mattdesl.svbtle.com/drawing-lines-is-hard) provides a good overview of different methods for drawing lines.

WebGL line primitives (GL_LINES) are generally not very flexible. It may be suitable for drawing bounding boxes or other simple lines, but I think we should avoid it for styled lines.

I like the look of the "screen-space projected" lines, and plan to investigate this first.

Another option is to use triangulated lines. This is the most flexible, but also has a lot of edge cases (no pun intended) to handle.

On the [zebrahub viewer](https://github.com/royerlab/points-web-viewer) we used the three.js ["Line2" visual](https://github.com/mrdoob/three.js/blob/dev/examples/jsm/lines/Line2.js). This uses instanced mesh rendering to render triangulated line _segments_. This is nice because we can kind of ignore beveling, etc. Especially with transparency this produces some artifacts not necessarily ideal for all applications, but for scientific visualization I think it can actually be a benefit as it's a bit easier to see what the actual data is showing.

Here is what the zebrahub instanced mesh lines look like up close:
Screenshot 2024-09-12 at 10 10 49 AM

aganders3:
### Summary
This adds a new "Line" renderable object, and a sample layer that renders a single pre-defined line.

This uses [screen-space projected](https://mattdesl.svbtle.com/drawing-lines-is-hard#screenspace-projected-lines_2) line rendering method. This PR does not implement any additional end caps or joins.

Here is what the example looks like now:
Screenshot 2024-09-27 at 12 15 01 PM

Here's a debug screenshots to give an idea of how this works, from rendering the same geometry with and LINE_STRIP:
Screenshot 2024-09-27 at 12 17 46 PM

This also moves examples into subdirs, and creates a very simple index that links to each. The `examples` folder is now also watched by the dev server for reloading.

### Related Issue
Closes #29

### Tests & Checks
So far this has only been tested/debugged manually and visually.


examples/main.ts
```diff
@@ -2,12 +2,15 @@ import {
...
   SingleMeshLayer,
+  SingleLineLayer,
   WebGLRenderer,
 } from "@";
 
 const singleMeshLayer = new SingleMeshLayer();
+const singleLineLayer = new SingleLineLayer();
 const layersManager = new LayerManager();
 layersManager.add(singleMeshLayer);
+layersManager.add(singleLineLayer);
```
andy-sweet: Is the mesh layer visible in the example anymore? If not, create a new example instead. Then we should just have an index to navigate to any of them.

src/objects/renderable/line.ts
```diff
@@ -0,0 +1,119 @@
+import { vec3 } from "gl-matrix";
+
+import { RenderableObject } from "core/renderable_object";
+import { Geometry } from "objects/geometry";
+
+export class Line extends RenderableObject {
+  // TODO: this geometry should be replaced after refactoring
+  private geometry_: Geometry;
+
+  // TODO: support varaible color and width along the path
```
andy-sweet:
```suggestion
  // TODO: support variable color and width along the path
```

src/objects/renderable/line.ts
```diff
@@ -0,0 +1,119 @@
...
+
+  public get width() {
+    return this.width_;
+  }
+
+  public set width(value: number) {
+    this.width_ = value;
+  }
+
+  private createVertices(path: number[]): Float32Array {
```
andy-sweet: Unsure how permanent the structure of this array is, but it might be good to document that this flat array contains the path vertices in the form `[x0, y0, z0, x1, y1, z1, ...]` somewhere. Maybe it's obvious to most readers, but I had to follow the calls up to work that out.

src/objects/renderable/line.ts
```diff
@@ -0,0 +1,119 @@
...
+    // "current", "previous", and "next" points can be indexed in one buffer
+    const length = path.length / 3;
+    const vertices = new Float32Array((length + 2) * 4 * 2);
+    const positiveOffset = 1.0;
+    const negativeOffset = -1.0;
+
+    let c = 0;
+    vertices[c++] = path[0];
+    vertices[c++] = path[1];
+    vertices[c++] = path[2];
```
andy-sweet: Nit: consider setting the values at once.

```suggestion
    vertices.set(path.slice(0, 3), c++);
```

After writing it, I'm not actually sure this is more readable, so take it or leave it!

andy-sweet (commented):
I'll wait to review this more thoroughly after this resolves conflicts on `main` (which should be relatively minor?) and with #45 (which may be more substantial). I also have not reviewed the math in the vertex shader yet either.

src/renderers/webgl_renderer.ts
```diff
@@ -39,6 +42,28 @@ export class WebGLRenderer extends Renderer {
...
+    const angle = (performance.now() / 1000) * Math.PI;
+    mat4.rotateY(modelView, this.activeCamera.viewTransform, angle);
+    program.setUniformMat4("Projection", this.activeCamera.projectionTransform);
+    program.setUniformMat4("ModelView", modelView);
+    program.setUniformVec2("Resolution", [this.width, this.height]);
+    program.setUniformFloat("LineWidth", line.width);
+    program.setUniformVec3("LineColor", line.color);
+
+    this.bindings_.bind(line);
+    const type = this.gl.TRIANGLES;
```
andy-sweet: Rendering points and/or line strips with the triangles is nice for debugging purposes, as in the PR description. It was easy enough to render points or line strips instead of triangles by changing this accordingly, but I guess the PR description requires another call to draw?

How hard would it be to make this easy to enable for debugging geometry in general? Would be super useful for (especially when we start to think about tiling textures and other geometries). Definitely doesn't need to be done in this PR though.

src/objects/renderable/line.ts
aganders3: I'm not sure this works either because it only increments `c` once, right? I basically stole this pattern from some of the example code in the lines repo I was working off, but I like having the separate index for the typed array that is continually incremented.

src/objects/renderable/line.ts
aganders3: Good point - it should at minimum be documented. In any case we have to iterate it so maybe a nested list is better? It may also make sense to support omitting the z-coord and defaulting it to zero.

examples/main.ts
aganders3: It is visible, but I agree this should be a separate example. I think you added an index to the examples on `main` so I will separate them when I update.

src/renderers/webgl_renderer.ts
aganders3: Good question - yes for the images in the PR description I had to update the color uniform and add another draw call.

Each renderable object may need some different configuration for one or more debug modes. For example on a mesh it can be really useful to render with colors representing normals or texture coords, or with a built-in debugging texture such as a checkerboard.

I'm not sure where this would best be set - maybe at the layer level? The renderer level?

aganders3: I created a new issue for this in #47

src/objects/renderable/line.ts
andy-sweet: You're right that this doesn't work, so definitely leave this alone!

aganders3:
There's still a problem here. I think all vertices may need to be doubled to do proper joins at the corners. Here's a short segment (three points) to illustrate the issue.
Screenshot 2024-09-19 at 1 15 58 PM

aganders3:
The [previous bug](https://github.com/chanzuckerberg/imaging-active-learning/pull/46#issuecomment-2361759555) is fixed with some minor updates to how the line normal is calculated in the shader. This basically gives a cheap miter join - not very accurate with tight corners, but also doesn't shoot off to infinity.
Screenshot 2024-09-20 at 2 29 09 PM Screenshot 2024-09-20 at 2 39 35 PM

There is a still a problem with lines running directly perpendicular to the camera plane. Basically at these points the xy diff of the line (how I'm estimating the normal) goes to zero, so the width goes to zero. If you have a straight line directly into the plane it just doesn't show up. It's particularly distracting in an animated context, though adding more points along the line is decent at hiding it. I'm inclined to leave it as-is for now, as these lines are still useful for mostly-in-plane tracks. You can see it in the demo and the video attached.

https://github.com/user-attachments/assets/55c8eb42-f3da-4b20-aaa1-9830e13e915b

examples/main.ts
aganders3: I updated the examples to be links from a main page. I'm sure there's something fancier we could to to avoid having to manually enter the links to each new example, but it's a start.

src/objects/renderable/line.ts
aganders3: I updated this to take a nested list (`[number, number, number][]`) in [571515c](https://github.com/chanzuckerberg/imaging-active-learning/pull/46/commits/571515cc59ef028cef86e6049ae0a4b26fddd66c).

src/layers/single_line_layer.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export class SingleLineLayer extends Layer {
+  constructor(
+    path: [number, number, number][],
+    color: [number, number, number] = [1.0, 0.7, 0.0],
+    width: number = 0.2
+  ) {
+    super();
+
+    // TODO: add color and with to the Line constructor?
```
andy-sweet:
```suggestion
    // TODO: add color and width to the Line constructor?
```

src/renderers/webgl_renderer.ts
```diff
@@ -31,6 +33,28 @@ export class WebGLRenderer extends Renderer {
     const program = this.getShaderProgram(this.getProgramName(object)).use();
     program.setUniformMat4("Projection", this.activeCamera.projectionTransform);
     program.setUniformMat4("ModelView", this.activeCamera.viewTransform);
+    try {
+      program.setUniformVec2("Resolution", [this.width, this.height]);
+    } catch {
+      // Ignore the error if the shader does not have a 'Resolution' uniform.
```
andy-sweet: Optional: unsure how optimized exception handling is, but given that we will hit this for every call with a mesh (and other objects that don't use a shader with a Resolution uniform), I'd expect and prefer an explicit check instead (e.g. `if (program.hasUniform("Resolution")) ...`).

That way you can also get rid of the catch block and exception.

src/core/renderable_object.ts
```diff
@@ -21,4 +27,12 @@ export abstract class RenderableObject extends Node {
   public set geometry(geometry: Geometry) {
     this.geometry_ = geometry;
   }
+
+  protected addUniform(name: string, type: UniformType, value: UniformValue) {
```
andy-sweet: The name `RenderableObject.addUniform` in combination with this setter is unfortunate because it makes me think a new uniform is being added for each call, whereas it is actually clobbering the key in a map. `RenderableObject::setUniform` would make this more clear to me.

src/layers/single_line_layer.ts
```diff
@@ -0,0 +1,22 @@
...
+    const line = new Line(path);
+    line.color = color;
+    line.width = width;
+    this.addObject(line);
+  }
+
+  public update(): void {
+    this.state_ = "ready";
```
andy-sweet: Nit: I think this object is ready on construction (because it doesn't need to load anything), so you could set this once in the constructor instead of every time `update` is called.

andy-sweet (commented):
I'm still digesting the bigger changes here, so I'm going to leave another indecisive review with some minor comments. Then I'm going to take a look at #53 to understand some alternative approaches to handling shaders more generally. Should be able to take a more decisive pass here after that.

aganders3:
Thanks - good comments and I appreciate you taking a look!

I'm about to push the "decorator" change I mentioned in Slack to #53, but feel free to review with and/or without.

src/layers/single_line_layer.ts
aganders3: Thanks - I just copied this from somewhere and still don't really grok the layer states, but that makes sense.

examples/image2d_from_omezarr5d/index.html
```diff
@@ -0,0 +1,26 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Web Viewer</title>
+    <style>
```
shlomnissan, line 7: An improvement would be to implement a static side navigation that loads the appropriate script based on the current selection, rather than navigating to a new HTML page. This approach would allow for easy toggling between examples and eliminate the need to repeatedly add HTML boilerplate for each example.

This is a nice-to-have that's probably out of scope for this PR.

examples/single_line_layer/single_line_layer.ts
```diff
@@ -0,0 +1,41 @@
+import {
+  LayerManager,
+  PerspectiveCamera,
+  SingleLineLayer,
+  WebGLRenderer,
+} from "@";
```
shlomnissan: Note: It would be beneficial to import the entire library namespace, similar to three:

```jsx
import * as LIBRARY from 'library';

const renderer = new LIBRARY.WebGLRenderer("#canvas");
```
I don't know what the impact is on tree shaking, but we could offer both options.

Also, we still don’t have a name for the library 😅

src/layers/single_line_layer.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export class SingleLineLayer extends Layer {
+  constructor(
+    path: [number, number, number][],
+    color: [number, number, number] = [1.0, 0.7, 0.0],
+    width: number = 0.2
+  ) {
+    super();
+
+    // TODO: add color and width to the Line constructor?
```
andy-sweet: Any reason why you haven't done this yet?

src/objects/renderable/line.ts
```diff
@@ -0,0 +1,134 @@
+import { vec3 } from "gl-matrix";
+
+import { RenderableObject } from "core/renderable_object";
+
+export class Line extends RenderableObject {
+  // TODO: support variable color and width along the path
```
andy-sweet: Do you think some version of this will be required for some version of a prototype we plan to deliver in Q4? If so, it may be worth translating this to an issue.

src/objects/renderable/line.ts
```diff
@@ -0,0 +1,134 @@
...
+    });
+    this.geometry.addAttribute({
+      type: "direction",
+      itemSize: 1,
+      offset: 3 * Float32Array.BYTES_PER_ELEMENT,
+    });
+
+    // TODO: this is a hack
+    this.color = this.color_;
+    this.width = this.width_;
```
andy-sweet: Given the property setters are just doing assignment now, this can just be removed?

```suggestion
```

examples/single_line_layer/single_line_layer.ts
```diff
@@ -0,0 +1,41 @@
...
+const heightIncrement = 1.5;
+
+for (let i = 0; i < totalPoints; i++) {
+  const angle = (i / pointsPerTurn) * 2 * Math.PI;
+  const z = radius * Math.cos(angle) - 5.0;
+  const y =
+    (i / pointsPerTurn) * heightIncrement - (heightIncrement * turns) / 2;
+  const x = radius * Math.sin(angle);
+  path.push([x, y, z]);
+}
```
shlomnissan: I suggest we either extract this code into a well-named function or write the result to a file that exports an array and import it as hard-coded data.

src/renderers/webgl_renderer.ts
```diff
@@ -33,7 +34,18 @@ export class WebGLRenderer extends Renderer {
     program.setUniform("Projection", this.activeCamera.projectionTransform);
     program.setUniform("ModelView", this.activeCamera.viewTransform);
 
-    // TODO: set uniforms for other types of renderable objects here
+    switch (object.type) {
+      case "Line": {
+        program.setUniform("Resolution", [
```
andy-sweet, line 47: To confirm, this must be done inside this case (as opposed to outside the switch statement) because the mesh vertex shader does not use this uniform, right?

I'm fine with this for now, but if there are certain uniforms that are used by multiple object types, then I wonder if we should have something like

```
if (program.hasUniform("Resolution")) {
   program.setUniform("Resolution", [this.canvas.width, this.canvas.height]);
}
```

I think that's effectively the maybe-behavior you had before with setUniformIfNeeded, so not asking to bring this back now.

examples/single_line_layer/single_line_layer.ts
```diff
@@ -0,0 +1,41 @@
...
+const helix = new SingleLineLayer(path);
+
+path.length = 0;
+path.push([0.0, -2.5, -5.0], [0.0, 2.5, -5.0]);
+const line = new SingleLineLayer(path, [0.0, 0.7, 0.0], 0.5);
```
shlomnissan: Note: I think we could start leveraging destructured parameters. This would make it easier to understand what these values represent without having to navigate to the layer's signature.
```ts
new SingleLineLayer({path, color: [0.0, 0.7, 0.0], width: 0.5});
```
Additionally, introducing a simple Color datatype would be beneficial.

examples/single_line_layer/single_line_layer.ts
```diff
@@ -0,0 +1,41 @@
+import {
```
shlomnissan: Nit: This script is already scoped by the directory. I'd suggest naming this file `index.ts`. This applies to other example scripts.

src/core/geometry.ts
```diff
@@ -1,6 +1,12 @@
 import { Node } from "core/node";
 
-type GeometryAttributeType = "position" | "normal" | "uv";
+type GeometryAttributeType =
```
shlomnissan, line 3: This change prompts me to reconsider the default approach of "interleaving data" as opposed to representing arbitrary attributes in separate buffers, which is what we had before.

src/core/geometry.ts
```diff
@@ -1,6 +1,12 @@
...
+  | "position"
+  | "normal"
+  | "uv"
+  | "nextpos"
+  | "prevpos"
```
shlomnissan: Nit: Spelling out full names and adding underscores can slightly improve readability.

```suggestion
  | "next_position",
  | "prev_position
```

src/core/geometry.ts
```diff
@@ -12,6 +18,7 @@ export class Geometry extends Node {
   protected vertexData_: Float32Array;
   protected indexData_: Uint32Array;
   private attributes_: GeometryAttribute[];
+  private stride_: number | null = null;
```
shlomnissan: Why is manual stride setting necessary? I intentionally aimed to avoid this when we moved over to interleaved buffers.

src/core/geometry.ts
```diff
@@ -29,21 +36,36 @@ export class Geometry extends Node {
   }
 
   public get stride() {
+    if (this.stride_ !== null) {
```
shlomnissan: Functions that return different values based on internal state can make it harder to reason about what the function does and introduce ambiguity. Ideally, functions should have a clear and consistent purpose, and in this case, it could lead to confusion as the return value depends on whether `this.stride_` is set or not.

I'm hesitant to recommend an approach because I don't fully grasp the necessity of manually setting the stride. This could be due to my own lack of understanding. But if it is necessary, I would probably rename the stride function to `computedStride` and have the stride accessor return the private member.

src/layers/single_line_layer.ts
```diff
@@ -0,0 +1,22 @@
+import { Layer } from "core/layer";
+import { Line } from "objects/renderable/line";
+
+export class SingleLineLayer extends Layer {
```
shlomnissan: I'm not convinced that layers should be named based on the renderables they display. I assumed layers would have domain-specific names, such as "tracks layer". Additionally, I'm not convinced this needs to be limited to a "single line." Couldn't we pass multiple paths to this layer? This consideration also applies to the "single mesh" layer, which I think we can remove. "Image layer" is a good example of how I would define the scope of a layer.

What are your thoughts on the scope/naming of the layer in this scenario?

Also, I previously mentioned that using the term "line" for a path doesn't seem appropriate to me. It's particularly confusing as it can be easily mistaken for a line primitive.

src/layers/single_line_layer.ts
```diff
@@ -0,0 +1,22 @@
...
+  constructor(
+    path: [number, number, number][],
+    color: [number, number, number] = [1.0, 0.7, 0.0],
+    width: number = 0.2
```
shlomnissan: I generally avoid default parameter values for non-obvious cases. I prefer requiring users of this layer to specify the color explicitly.

src/layers/single_line_layer.ts
shlomnissan: This change is minor, so could implement it before merging this PR.

src/renderers/webgl_buffers.ts
```diff
@@ -56,6 +56,9 @@ export class WebGLBuffers {
       if (attr.type === "position") idx = 0;
```
andy-sweet, line 57: Optional

I wonder if we should consider an `Enum` with values auto-incrementing from 0 to handle the attribute type? Not necessary in this PR unless you feel strongly.

src/objects/renderable/line.ts
```diff
@@ -0,0 +1,134 @@
+import { vec3 } from "gl-matrix";
+
+import { RenderableObject } from "core/renderable_object";
+
+export class Line extends RenderableObject {
```
shlomnissan: My previous comment about using the term "line" to describe the layer applies here too. It's perhaps even more applicable in this context, as I would have associated this with a `GL_LINE` primitive.

src/objects/renderable/line.ts
```diff
@@ -0,0 +1,130 @@
...
+    vertices[c++] = path[0][1];
+    vertices[c++] = path[0][2];
+    vertices[c++] = positiveOffset;
+
+    vertices[c++] = path[0][0];
+    vertices[c++] = path[0][1];
+    vertices[c++] = path[0][2];
+    vertices[c++] = negativeOffset;
+
+    for (let i = 0; i < length; i++) {
```
andy-sweet: Optional: what do you think about removing the explicit repeat lines and absorbing into a fancier iteration?

```suggestion
    for (const x of [0, ...Array(length).keys(), length - 1]) {
```

src/objects/renderable/line.ts
```diff
@@ -0,0 +1,130 @@
+import { vec3 } from "gl-matrix";
+
+import { RenderableObject } from "core/renderable_object";
+
+export class Line extends RenderableObject {
```
andy-sweet: I'd appreciate a reference to explain the approach this type of `Line` takes - ideally both for the geometry it stores and its corresponding shader (since they are very coupled). I think you did that in the PR description, but I think a comment in the code would be helpful  - especially if we want to rename this in the future to reflect that approach.

src/objects/renderable/line.ts
shlomnissan: I suggest introducing a "path geometry" factory object that extends Geometry. It's essentially performing the same function. In fact, I don't consider this a "renderable object" at all. You're essentially generating a mesh, and we already have a "mesh" renderable (which is probably where we should add the color property).

I assume you added a new renderable type due to the different shader we're using. This relates to my earlier question about precomputing all mesh data on the CPU and using it directly as a mesh. Even if a new shader is necessary, I believe we should still separate the geometry generator from the "renderable type".

src/objects/renderable/line.ts
```diff
@@ -0,0 +1,134 @@
...
+  // TODO: support variable color and width along the path
+  private color_: vec3 = [1.0, 0.7, 0.0];
+  private width_: number = 0.2;
+
+  constructor(path: [number, number, number][]) {
+    // path is a list of 3D points [x, y, z]
```
shlomnissan: I'm not convinced that this comment adds any value.

src/objects/renderable/line.ts
```diff
@@ -0,0 +1,134 @@
...
+  public set width(value: number) {
+    this.width_ = value;
+  }
+
+  private createVertices(path: [number, number, number][]): Float32Array {
+    // each point on the path is split into two vertices
+    // these are pushed in opposite directions in screen-space in the vertex shader
+
+    // first and last points are duplicated so
+    // "current", "previous", and "next" points can be indexed in one buffer
```
shlomnissan: If this algorithm is derived from an external source or article, consider adding a comment with a link for proper attribution and reference.

src/core/geometry.ts
```diff
@@ -1,6 +1,12 @@
 import { Node } from "core/node";
 
-type GeometryAttributeType = "position" | "normal" | "uv";
+type GeometryAttributeType =
+  | "position"
+  | "normal"
+  | "uv"
+  | "nextpos"
```
shlomnissan: If we're converting points to a mesh, why do we need additional buffers? Couldn't we precompute the entire mesh on the CPU?

src/objects/renderable/line.ts
```diff
@@ -0,0 +1,130 @@
...
+  // TODO: support variable color and width along the path
+  private color_: vec3 = [1.0, 0.7, 0.0];
+  private width_: number = 0.2;
+
+  constructor(path: [number, number, number][]) {
+    // path is a list of 3D points [x, y, z]
+    super();
+    this.geometry.vertexData = this.createVertices(path);
+    this.geometry.indexData = this.createIndex(path.length);
+    const stride = 4 * Float32Array.BYTES_PER_ELEMENT;
```
andy-sweet: I keep re-reading this, but I can't keep the byte layout of these attributes clear in my head. That might be fixed after lunch, so don't do anything yet.

andy-sweet (commented):
Almost ready to approve. I just need to better understand the use of stride here.

src/renderers/shaders/line_vert.glsl
```diff
@@ -0,0 +1,47 @@
+#version 300 es
```
shlomnissan: Note: I haven’t reviewed the geometry generator code or this shader code closely because it will take me time and I don’t think it should block this PR.

src/renderers/webgl_renderer.ts
shlomnissan, line 47: Ah, I see why you implemented the "if needed" function. I hadn't considered GLSL's optimization of unused uniforms. While I still prefer this approach over adding a method at the program level, I think that if we want to check for shared attributes, it should be placed in the renderer. Something like:

```ts
if (program.hasUniform("Resolution")) {
  program.setUniform("Resolution", ...);
}

File: src/renderers/webgl_renderer.ts [hunk 34]
```diff
@@ -59,13 +71,17 @@ export class WebGLRenderer extends Renderer {
 
   protected clear() {
     this.gl.clearColor(0, 0, 0, 1.0);
-    this.gl.clear(this.gl.COLOR_BUFFER_BIT);
+    this.gl.clearColor(0.12, 0.13, 0.25, 1.0);
+    this.gl.clear(this.gl.COLOR_BUFFER_BIT | this.gl.DEPTH_BUFFER_BIT);
+    this.gl.enable(this.gl.DEPTH_TEST);
+    this.gl.depthFunc(this.gl.LEQUAL);
```
shlomnissan: Interesting... I didn't know you could select different depth functions. What difference does this selection make?

shlomnissan (changes requested):
I believe I have enough questions to request changes before merging but most of them are not blocking. My main concerns are:
- Could we generate the path geometry on the CPU and treat it as a mesh?
- If not, we should at least separate the "geometry generator" code from the renderable object, similar to how mesh and plane geometry are distinct.
- I don't fully understand why we need to set the stride manually. I'm probably just missing something.

src/objects/renderable/line.ts
aganders3: Yes - I thought I had already removed it last night, but it's removed now!

src/renderers/webgl_renderer.ts
aganders3: Yep, you have it correct.

I think it makes sense to leave as-is and reorganize this as it grows.

I am fine using the look-before-you-leap `if (program.hasUniform("Resolution"))` pattern, but I feel like it's mixing a bit of logic responsibilities. How do we decide:
* sometimes we're saying "set this uniform if the shader needs it" 
* other times we're saying "we expect the shader needs this uniform"

When we go to draw there will be an error if we _don't_ set a needed uniform, and likewise an error if we _do_ set an unnecessary uniform.

This was kind of my intent with the other PR - the shader always already knows which uniforms it needs.

examples/single_line_layer/single_line_layer.ts
aganders3: Agreed - it makes me sad this is the only way to do named arguments in JS but so it goes.

examples/single_line_layer/single_line_layer.ts
aganders3: I don't think it's generally useful code, but I'm happy to put it into a function with a name in the example.

examples/image2d_from_omezarr5d/index.html
aganders3: Yeah I started doing this but gave up. It would also be nice to have it automatically add links for new examples but I think this would need to run some kind of script at build time.

aganders3:
Thanks for the reviews! I'll start working on the immediately actionable items and try to answer questions here. Though I guess if we're going to go with a totally different method for drawing lines I'd rather change course quickly.

> Could we generate the path geometry on the CPU and treat it as a mesh?

I don't think this generalizes to 3D lines, because the a line meshed on the CPU will end up as a ribbon. You would need to re-mesh every time the camera moved to make sure it's facing the right way. Maybe this is fine for our application? Certainly for the prototype we can do this as we're only handling 2D data right now. Moving the vertices in the vertex shader here is effectively billboarding, and allows the line to be viewed from any angle (except of course straight into the plane).

I guess for 3D we could compute a closed tube shape and use that as a 3D line - maybe this is what you mean? I don't know much about any pitfalls of that approach but can start looking into it.

I do think an instanced-mesh based approach like Three.js's `Line2` (in addons) is worth investigating as well. I think long-term there's probably value in having more than one line implementation.

> If not, we should at least separate the "geometry generator" code from the renderable object, similar to how mesh and plane geometry are distinct.

Yes, we can do this. I was trying to avoid having a pairing of `XRenderable` and `XGeometry` for every renderable object, but I agree they're separate responsibilities so that's fine with me.

> I don't fully understand why we need to set the stride manually. I'm probably just missing something.

The other option is to repeat the vertices more times in the buffer. This will work, it's just less efficient. Hopefully the descriptions below can help, but if not I can probably draw with more fidelity in a vector drawing tool.

Here is the current layout in this PR. Note that after the first point `prev` will reuse memory previously pointed by the `curr` buffer. Total buffer size is 8N + 16 floats (where N is the length of the path).
```
|--- first and last point repeated twice -----| |--- each point repeated once ----|
x0 y0 z0 -1 x0 y0 z0 +1 x0 y0 z0 -1 x0 y0 z0 +1 x1 y1 z1 -1 x1 y1 z1 +1 x2 y2 z2 -1 x2 y2 z2 +1 ...
|-prev-|                |-curr-||d|             |-next-|
|--stride-|
```

The computed stride here will be `(3 + 3 + 3 + 1) * sizeof(float)`, so we'd need to lay it out like this - none of the memory gets reused. Total buffer size is 20N (or 20N + 20?) floats. This is probably not a huge deal for the size of data we're considering right now. I can imagine using something (splines) to interpolate the lines though and they would get much bigger.

```
x0 y0 z0 x0 y0 z0 x1 y1 z1 -1 x0 y0 z0 x0 y0 z0 x1 y1 z1 +1 x0 y0 z0 x1 y1 z1 x2 y2 z2 -1 ...
|-prev-| |-curr-| |-next-||d|
|--------stride-------------|
```

src/objects/renderable/line.ts
aganders3: I tried to answer in https://github.com/chanzuckerberg/imaging-active-learning/pull/46#issuecomment-2374643051 with some ASCII diagramming, if that helps.

src/renderers/webgl_renderer.ts
aganders3: Good catch - actually we don't need to change the function we just need to set the other parts. I added this during debugging but the default (`LESS`) should be okay.

I suppose `LEQUAL` maybe is more intuitive since fragments with the same z-value will show up if drawn _later_, but I don't care about this and sticking with defaults is usually a good start.

src/layers/single_line_layer.ts
aganders3: It was more a question of - does this belong in the constructor? I think this gets to your points about what a layer should represent, and whether a "single line layer" is a good idea. The problem now is creating a new layer is really the only way to test a new renderable object.

src/core/geometry.ts
aganders3: I agree in principle - this is not the correct fix if we determine we need manual setting stride, but I also understand wanting to make it unnecessary.

src/layers/single_line_layer.ts
aganders3: I agree with both points here. I'm not convinced a "tracks layer" is general enough to belong in the library either - it feels pretty specific to cell tracking data. I think napari is adding a more general "graph" layer that might be a better abstraction?

Regarding "line" I totally support renaming - it's just kind of tedious with renaming files and code. I'm hoping to only have to do it once and I don't yet have a good name for this "kind" of line.

src/renderers/webgl_buffers.ts
aganders3: I agree this needs improvement and an Enum is probably a good start. I don't feel like this is the right PR for it so we should probably just create an issue.

src/renderers/webgl_renderer.ts
aganders3: Andy had the same comment and I responded with a few more thoughts https://github.com/chanzuckerberg/imaging-active-learning/pull/46#discussion_r1775528675

I think this may deserve more consideration in the future but not sure this is the right spot for it.

src/objects/renderable/line.ts
aganders3: Yeah, this is probably a little nicer. Thanks!

examples/single_line_layer/single_line_layer.ts
shlomnissan: If it's not useful, consider copying the output data into an array and importing it separately. This will enhance readability for this file and separate the data from the logic needed to demonstrate this example.

src/layers/single_line_layer.ts
shlomnissan: I think it’s fine. Until we implement a system for conditional shader inclusion based on renderable object properties, we may need to create temporary layers. We should mark these layers with comments to indicate that.

shlomnissan:
> Moving the vertices in the vertex shader here is effectively billboarding, and allows the line to be viewed from any angle (except of course straight into the plane).

Ah, that makes sense! I'm accustomed to these operations being handled in the geometry shader, but that's not an option here. Instancing might simplify some of this complexity, though it's not a critical issue at this stage.

> The other option is to repeat the vertices more times in the buffer. This will work, it's just less efficient.

I think repeating vertices is fine for now - it's simpler and easier to understand. I'm not experienced with generating lines like this, but if this becomes a bottleneck, we could look into other options - see how three.js handles it, or we tweak the Geometry class to be more flexible if needed. Manually specifying the stride feels like a hack for a premature optimization, but I’m still missing some context so I will leave this decision to you.

src/objects/renderable/line.ts
aganders3: I'd probably put this at P1, but I'm going to remove the TODO and make it an issue. Color along the path may be useful for indicating direction of motion, for example.

src/core/geometry.ts
aganders3: I added a comment about this in #62 to capture this. I think there's probably space for both in different applications. It's not strictly either/or - we can also do arbitrary attributes still in an interleaved buffer.

aganders3:
Thanks again - I made a bunch of changes based on your suggestions. The main change is updating the layout of interleaved attributes to be less confusing and not require a manually-set stride. I also created issues where I felt comments were important to capture but out of scope for this PR.

I think the main thing remaining is to rename a bunch of stuff, and I plan to do that in the morning. I am leaning towards renaming to `ProjectedLine` and `ProjectedLineGeometry` for the renderable object. I think the layer could be `ProjectedLineLayer` (or just `LineLayer` if it supported adding different types of lines at some point). I'll add one more comment when I'm done with this and re-request review.

andy-sweet (approved):
Looks good to me now. Could consider renaming the example source files to `main.ts` given they are in named sub-directories now, but definitely optional.

examples/projected_lines/main.ts
```diff
@@ -0,0 +1,60 @@
...
+layersManager.add(layer);
+
+const helixA = generateHelix({
+  radius: 1.0,
+  turns: 3.0,
+  pointsPerTurn: 100,
+  pitch: 1.5,
+  phase: 0.0,
+});
+layer.addLine({ path: helixA, color: [1.0, 0.7, 0.0], width: 0.1 });
```
shlomnissan, line 19: I think that what we eventually want is passing "data sources" to the layer instead of having inline functions to add geometry, but we can better understand how this looks when we start rendering lines using the mock data.

src/layers/projected_line_layer.ts
```diff
@@ -0,0 +1,25 @@
+import { Layer } from "core/layer";
+import { ProjectedLine } from "objects/renderable/projected_line";
+import { ProjectedLineGoemetry } from "objects/geometry/projected_line_geometry";
+
+interface LineParameters {
+  path: [number, number, number][];
+  color: [number, number, number];
+  width: number;
+}
```
shlomnissan, line 9: Nits:

- For consistency, let's stick with `type` for plain data structures.
- We could use `vec3[]` for paths and potentially for colors too (though we might want to create a dedicated `Color` type at some point).

shlomnissan (approved):
Looks good to me :+1:

src/layers/projected_line_layer.ts
aganders3: Thanks - I'm fine with type-by-default for a convention. I'll also read up on the difference (again ).

I actually had the color as `vec3` but reverted it. My thought was that `vec3` is not providing any useful information or functionality over the builtin and I could avoid the import.

I agree a `Color` type would be better. We probably generally want colors to be vec4 with a default 1.0 alpha as well. I started on this but it's out of scope here.

I'm a little confused about what the `vec3` type is, because it seems like any function that accepts it will also take `[number, number, number]`. Under the hood I think it's a Float32Array though. For path it probably makes sense.

aganders3: I'm going to leave this here, but added a reference to your comments in #58

PR #48, opened 2024-09-19
[WIP] Add a separate React + SDS application that uses the visualization package

### LINKED ISSUE #27: Add React and SDS to the existing example

andy-sweet:
### Summary

This change current conflates a few things related to initializing the prototype application.

Screenshot 2024-09-19 at 11 08 53 AM

#### Adds a React + SDS application that uses the visualization package

To keep things simple and smaller, I just added enough to exercise the visualization library in addition to some playback controls (borrowed from https://github.com/royerlab/points-web-viewer), so that it forms a very early skeleton for the prototype app.

My combined use of React's state management and render loop with the visualization library already has some issues. That might be my misuse of those things, but it could also provide a way to drive improvements too (e.g. allowing the playback controls to mutate a region that can be passed through to the animating update).

**I think it's also worth confirming that we want to use React for the prototype.** It will force us to think about usage of the library in certain ways that are surely important if we want this to be adopted widely within CZI. But it may also slow us down too much and may pollute our design of the library without certain guardrails.

#### Puts the application in a separate `apps` sub-directory

I created a new sub-directory `apps/ultrack-learns` to contain the prototype application we intended to build, and gave its own `package.json`. It also got its own config files because I used `npm create vite@latest ...` and then updated them.

I think having the separate app package is a desirable goal, but I'm currently taking some undesirable shortcuts to get there (e.g. forcing relative imports in the viz package, redefining some common configs). This seems fixable once I better understand import/export path resolution, especially when dealing with two local packages.

#### Adds non-functional playback controls to the application.

For this to be functional, we need to at least allow the `ImageLayer` to load new texture data based on the current time point. This requires a few changes (e.g. https://github.com/chanzuckerberg/imaging-active-learning/issues/33), but is generally a good driver of the library's API design and functionality.

### Related Issue
Closes #27

### Tests & Checks
I manually tested installing and running the new app and existing examples.


andy-sweet:
I'm going to pause further work on this until we can discuss the issues in the PR description as a group.

andy-sweet:
Closing in favor of the simpler https://github.com/chanzuckerberg/imaging-active-learning/pull/49

Issue #33, opened 2024-09-16
Pass the region to be rendered to `Layer.update`


andy-sweet:
In #28, we added support for loading uint16 image data from OME-Zarr sources. The associated example defines a region and passes it to the constructor of `ImageLayer`. Instead this region should be passed through to a call of `Layer.update`.

The renderer can use its camera to partially define this region, but we may need to also define some state related to non-visualized dimensions (e.g. time). This can temporarily be stored in `Renderer`, but we may want to consider `LayerManager` or something new instead.

shlomnissan:
While I may be missing some context here, I believe the state should be stored in the structure representing the data type. This relates to the separation between file loaders (such as OME-Zarr) and the data type itself (e.g., 2D images with a time component, or whatever the appropriate shared term might be).

I understand why the camera needs to be accessible at a top level (such as in the renderer or layer manager) since it affects multiple layers. However, as I understand it, the time point is specific to the data itself.

andy-sweet:
At a high level, I think of some high level component (not necessarily the renderer) being responsible for defining some region of data it wants to present in some space shared across all layers (e.g. the union of all their dimensions). For a 2D presentation, the camera defines some sub-region in the two dimensions of interest (e.g. X + Y), but other things may define other parts of the sub-region (e.g. in an app, the value of an input range slider may define the Z depth).

It's up to the loader to map that region to the dimensions of a specific data source. If they don't map in a way that is compatible with the layer to be presented, then I would expect an error.

Note that I think it would be reasonable (and maybe quite powerful!) to pass along a sub-region to the layer that indicates what is of interest (e.g. for the ultrack project, a sub-region in time that we want to loop over). But that's different to what I'm including in this issue and should also be taken into account when performing the mapping described above.

I'm still trying to understand what data type means in the context of the library, so I can't comment on how that fits here.

andy-sweet:
I'm going to close this because it was created to capture an idea early on Idetik's development. There has been some recent related development (e.g. viewports) and discussion, but I don't think this issue represents an active work item.

PR #49, opened 2024-09-19
Add a React + SDS ultrack prototype

### LINKED ISSUE #27: Add React and SDS to the existing example

andy-sweet:
### Summary

This adds a React + SDS prototype application to `ultrack-prototype/frontend` that uses the visualization package to render some OME-Zarr image data.

Screenshot 2024-09-19 at 11 08 53 AM

To keep things simple and small, I just added enough to exercise the visualization library in addition to some playback controls (borrowed from https://github.com/royerlab/points-web-viewer), so that it forms a very early skeleton for the prototype app.

The playback controls are interactive, but have no meaninful side-effects on visualization yet. For this to be functional, we need to at least allow the `ImageLayer` to load new texture data based on the current time point. This requires a few changes (e.g. https://github.com/chanzuckerberg/imaging-active-learning/issues/33), and is generally a good driver of the library's early API design and functionality.

As this is currently a prototype for developers, React and SDS are dev dependencies. You can serve the prototype by running

```shell
npm run dev -- --mode prototype
```

### Related Issue
Closes #27

### Tests & Checks
I manually tested the application by running the command above.


src/renderers/webgl_renderer.ts
```diff
@@ -51,7 +51,7 @@ export class WebGLRenderer extends Renderer {
   }
 
   protected clear() {
-    this.gl.clearColor(0.12, 0.13, 0.25, 1.0);
+    this.gl.clearColor(0, 0, 0, 1.0);
```
andy-sweet, line 58: I made the background color of the canvas black as a better presentation for the prototype. This should probably be configurable at some point, but black seems like a reasonable default for scientific images.

ultrack-prototype/frontend/components/App.tsx
```diff
@@ -0,0 +1,70 @@
...
+const region = [
+    // TODO: when the region is state associated with the renderer or
+    // layer manager, and we have a reference to that, then sync it
+    // with React state that captures the time-point.
+    { dimension: "t", index: 400 },
+    { dimension: "c", index: 0 },
+    { dimension: "z", index: 300 },
+];
+const layer = new ImageLayer(source, region);
+const layerManager = new LayerManager();
```
andy-sweet: I put the source, layer, and layer manager in this file because I think some version of them must sit at this level of the component hierarchy so that the playback controls can be updated with the layer's relevant metadata after opening (e.g. number of time points), and so that the playback controls can modify the region to load.

These are not functional yet, so I could move these into the renderer component for now to remove the need for renderer props.

aganders3 (approved):
LGTM, thanks!

I was able to pull and run this branch with just:
```
npm install
npm run dev -- --mode prototype
```

src/renderers/webgl_renderer.ts
aganders3: I agree it should be configurable at some point. I think having it non-black can help while debugging because you can see the edges of the image, but I am okay with this change.

shlomnissan: Yep, I think making it a property of the generalized Renderer class is a sensible approach.

tsconfig.json
```diff
@@ -27,5 +28,5 @@
 
     "types": ["vite-plugin-glsl/ext"]
   },
-  "include": ["examples", "src", "test"]
+  "include": ["examples", "src", "test", "ultrack-prototype/frontend"]
```
shlomnissan, line 31: Nit: I think a more common directory name combinations would be `app` and `server`, or `client` and `server`.

ultrack-prototype/frontend/components/App.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { useEffect, useState } from "react";
+
+// Source is 5D, so provide indices at 3 dimensions to project to 2D.
+const url =
+    "https://public.czbiohub.org/royerlab/zebrahub/imaging/single-objective/ZSNS001.ome.zarr/";
+const source = new OmeZarrImageSource(url);
+const region = [
+    // TODO: when the region is state associated with the renderer or
+    // layer manager, and we have a reference to that, then sync it
+    // with React state that captures the time-point.
```
shlomnissan: I raised this question in Issue #33, but why do we want the region to be associated with the renderer or the layer manager instead of the data source (which might be managed through a specialized layer)?

ultrack-prototype/frontend/components/App.tsx
shlomnissan: Why not encapsulate all viewer-related code in a single component and manage state at this level to coordinate updates between UI components and the viewer?

ultrack-prototype/frontend/components/App.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+const playbackFPS = 16;
+const playbackIntervalMs = 1000 / playbackFPS;
+
+export default function App() {
+  // TODO: this state should be kept in sync with the corresponding
+  // state in the visualization library. For now, use known values
+  // associated with the data we are using here.
+  const [curTime, setCurTime] = useState<number>(0);
+  const [numTimes, _] = useState<number>(791);
```
shlomnissan:
```suggestion
  const [numTimes] = useState<number>(791);
```

ultrack-prototype/frontend/components/App.tsx
```diff
@@ -0,0 +1,70 @@
...
+    if (playing) {
+      const interval = setInterval(
+        () => {setCurTime((curTime + 1) % numTimes)},
+        playbackIntervalMs,
+      );
+      return () => {
+        clearInterval(interval);
+      };
+    }
+  }, [numTimes, curTime, playing]);
```
shlomnissan: I think all this logic can be encapsulated within the `PlaybackControls` component. If you need to pass data back to the viewer through this component, you can provide callbacks to the playback component. Something along the lines of:
```
const onTimeChange = (currTime: number) => {}
const onPlayToggle = (isPlaying: boolean) = {}
```

ultrack-prototype/frontend/components/Main.tsx
```diff
@@ -0,0 +1,21 @@
+import React from "react";
+import { createRoot } from "react-dom/client";
+import { ThemeProvider as EmotionThemeProvider } from "@emotion/react";
+import { StyledEngineProvider, ThemeProvider } from "@mui/material/styles";
```
shlomnissan, line 4: Is this part of SDS's recommended modules for theme management? I'd expect SDS to provide some kind of internal theme provider. I have the same question about `<Box>` — I would have assumed that SDS would offer a grid system, or at least fixed measurements to ensure consistency in gaps and margins.stent.

ultrack-prototype/frontend/components/PlaybackControls.tsx
```diff
@@ -0,0 +1,39 @@
+import { Box } from "@mui/material";
+import { Button, InputSlider } from "@czi-sds/components";
+
+interface PlaybackControlsProps {
+    playing: boolean;
+    curTime: number;
+    numTimes: number;
+    setPlaying: (isPlaying: boolean) => void;
+    setCurTime: (curTime: number) => void;
```
shlomnissan: This comment builds on my earlier point. The callbacks here seem reversed. Passing mutators down to a child component can break encapsulation. Typically, it's better for a child component to notify the parent of events via callbacks, allowing the parent to handle state updates:
- The child modifies the parent's state, leading to tight coupling and making reuse harder.
- The parent should manage state; passing mutators down mixes responsibilities.
- Direct state manipulation by the child complicates testing, debugging, and reasoning.

ultrack-prototype/frontend/components/PlaybackControls.tsx
```diff
@@ -0,0 +1,39 @@
...
+            <Button
+                icon="Play"
+                sdsSize="large"
+                sdsType="primary"
+                sdsStyle="icon"
+                onClick={() => props.setPlaying(!props.playing)}
+            />
+
+            <InputSlider
+                id="time-frame-slider"
```
shlomnissan: Why is it necessary to specify an ID in this case?

vite.config.js
```diff
@@ -13,11 +13,19 @@ const _dirname = dirname(fileURLToPath(import.meta.url));
 
 const plugins = [tsconfigPaths(), eslint(), glsl()];
 
+function modeToRoot(mode) {
+  if (mode === 'development') {
+    return 'examples';
+  } else if (mode === 'prototype') {
+    return 'ultrack-prototype/frontend';
+  }
+  return undefined;
```
shlomnissan: Nit: I don't see much value in returning undefined and would have defaulted to examples.

shlomnissan (changes requested):
Overall, I think this PR looks good.

I’m going to request changes because I believe there are opportunities to improve the composition of the components—specifically, moving objects and some state out of `App.tsx` and into the relevant components.

I also have questions about how we update regions and why we need to store this information in the renderer or layer manager. However, this isn't directly related; we can discuss it in our next engineering sync.

src/renderers/webgl_renderer.ts
andy-sweet: I created an issue here: https://github.com/chanzuckerberg/imaging-active-learning/issues/52

ultrack-prototype/frontend/components/PlaybackControls.tsx
andy-sweet: Not necessary. Removed.

vite.config.js
andy-sweet: Don't feel too strongly, but I was assuming that `undefined` (as was returned before this PR in a similar case) would not serve successfully, which seemed like desirable behavior if an unrecognized mode was passed (e.g. due to a typo). It is not served successfully, but there is no indication (server or client side) of that failure (other than a 404 for the client). What about using `console.error` here and continuing to return `undefined`?

ultrack-prototype/frontend/components/App.tsx
andy-sweet: I'm not sure I can give a better explanation that the first comment.

To clarify, for the functionality of this PR, I can push these and related state/effects further down into only the components they affect. I will do that to keep this PR as simple as possible. But I suspect they will be lifted up soon in some form.

ultrack-prototype/frontend/components/App.tsx
andy-sweet: I left a more detailed comment on #33. The short summary is that I think the region here defines the sub-region that we want to present across all layers. Each layer and its loader should use its metadata to map that sub-region to the dimensions/regions associated with its underlying data, then fetch and populate vertices/textures accordingly

ultrack-prototype/frontend/components/App.tsx
andy-sweet: I think that `curTime` is an example of state that will need to be shared between the slider and whatever is responsible for triggering a load of the corresponding data at that time.

But that is not the case right now because the app doesn't do anything interesting, so your point holds - we can just push this all into `PlaybackControls` so I'll do that for now.

ultrack-prototype/frontend/components/PlaybackControls.tsx
andy-sweet: Sorry, I think the points here mostly stem from anticipating functionality that will shortly be in the application, but is not currently (i.e. `curTime` is currently more boring that it will be soon).

I am assuming that `curTime` will be state that is shared across multiple components, in which case I believe the recommended design pattern in React is to lift that state up into the parent component and pass the dispatchers (or functions that call them) down in to the children: https://react.dev/learn/sharing-state-between-components

Not sure I understand the point about callbacks though - that's how I would describe the dispatchers like `setCurTime`. They are callbacks for the React state machine to work out what to do with the state change (e.g. re-render). From what I understand of React, I would not describe them as mutators

ultrack-prototype/frontend/components/Main.tsx
andy-sweet: I copied this straight from Ashley's work on the zebrahub viz tool, but using the material provider is consistent with the SDS examples (e.g. see `index.tsx` in the first example): https://sds.czi.design/009eaf17b/p/169911-theming/b/36245e

The next line imports an SDS theme (the default one), which is used below.

I can't find any SDS-specific layouts/grids in the [SDS docs](https://sds.czi.design/009eaf17b/p/55ad7a-introduction), but maybe I just didn't look hard enough yet.

tsconfig.json
andy-sweet: I'm going to leave this now since this matches the prototype architecture diagram and it's a nit. Feel free to change later.

ultrack-prototype/frontend/components/App.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+export default function App() {
+  return (
+    <Box sx={{
+      display: "flex",
+      flexDirection: "column",
+      gap: "1em",
+    }}>
+      <Renderer></Renderer>
+      <PlaybackControls></PlaybackControls>
```
shlomnissan, line 15: As we discussed during our Zoom call, we'll need to manage shared state in this component. However, this doesn't necessarily mean the top-level component should handle state directly related to its children. When defining the child components, we should consider their internal scope and their public interface carefully. For the public interface, it makes sense, for example, to pass the URL to the renderer component or the playback FPS to the playback component. I do consider this configuration “speculative” at this point, but they are sensible enough that it’s probably okay.

ultrack-prototype/frontend/components/PlaybackControls.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { useEffect, useState } from "react";
+
+const playbackFPS = 16;
+const playbackIntervalMs = 1000 / playbackFPS;
+// Use an arbitrary number of times just to exercise playback.
+const numTimes = 100;
+
+export default function PlaybackControls() {
+  const [curTime, setCurTime] = useState<number>(0);
+  const [playing, setPlaying] = useState<boolean>(false);
```
shlomnissan: TypeScript should be able to deduce the underlying type based on the value passed to `useState`.

```suggestion
  const [curTime, setCurTime] = useState(0);
  const [playing, setPlaying] = useState(false);
```

ultrack-prototype/frontend/components/PlaybackControls.tsx
```diff
@@ -0,0 +1,52 @@
...
+    if (playing) {
+      const interval = setInterval(
+        () => {setCurTime((curTime + 1) % numTimes)},
+        playbackIntervalMs,
+      );
+      return () => {
+        clearInterval(interval);
+      };
+    }
+  }, [numTimes, curTime, playing]);
```
shlomnissan: - I haven't tested this, but I think using `curTime + 1` directly can lead to stale closures because `curTime` may not be the latest state when the interval runs. Instead, you can use a functional update for `setCurTime` to always get the latest value of `curTime`.
- Nit: The return statement can be simplified.
```suggestion
  useEffect(() => {
    console.debug("effect-playback: ", curTime, playing);
    if (playing) {
      const interval = setInterval(() => {
        setCurTime((prevCurTime) => (prevCurTime + 1) % numTimes);
      }, playbackIntervalMs);
      return () => clearInterval(interval);
    }
  }, [numTimes, curTime, playing]);
```

shlomnissan (approved):
I left a couple of suggestions, but otherwise, LGTM

ultrack-prototype/frontend/components/PlaybackControls.tsx
andy-sweet: I was unaware of this form - thanks!

The other improvement (in addition to potential stale-ness) this has is that the effect is no longer explicitly dependent on `curTime` (other than the debug log, which can be removed), so it will only re-run when playing changes which seems like the right approach with set/clearInterval. Before, the effect would re-run on every change to `curTime` so the interval was never really being used in the happy path of the event loop keeping up (i.e. set/clearTimeout would be more suitable).

I made this change myself to remove `curTime` from the dependency list and to clean up the arrow function formatting a little more

PR #50, opened 2024-09-20
Add support for uint8 image chunks and textures

### LINKED ISSUE #31: Support loading and rendering image textures of data types other than uint16
In #28, we added support for loading uint16 image data from OME-Zarr sources. This hard-coded many things coupled to uint16 (e.g. texture, fragment shader, loader).

The goal of this task is to implement support for other data types. In my experience, the most common in scientific imaging tend to be uint8, uint16, and float32, but ideally it should be straightforward to add support for new data types as needed.

Before starting this issue, we may want to wait until we know the data type of the driving example data we plan to use for the prototype.

andy-sweet:
### Summary

This adds support for loading uint8 image chunks from OMEZarr and using them as WebGL textures.

Screenshot 2024-09-24 at 4 25 58 PM

### Related Issue
Closes #31

While this only adds support for uint8 in addition to uint16, it allows new similar scalar data types (e.g. float32, uint32) to be added somewhat easily. A new fragment shader will be needed for some types to handle texture sampling.

Even though they are currently the same, I decided to define the supported array types separately for `DataTexture2D`, `ImageChunk`, and `OmeZarrImageLoader`, because I think they should be allowed to diverge in the future.

### Tests & Checks
I manually tested my changes using the new and existing examples.


src/data/ome_zarr_image_loader.ts
```diff
@@ -35,6 +35,14 @@ interface Multiscale {
   datasets: Array<Dataset>;
 }
 
+const dataTypes = [Uint8Array, Uint16Array] as const;
```
andy-sweet, line 38: I went all-in on some fancy typescript here and I might regret it.

The advantage of this construct is that I can just add to the array of `dataTypes` and automatically get type inference and a nice error message.

The disadvantage is that I need to sometimes re-read this and understand what it's doing.

aganders3: I don't have a problem with this. I think I did almost the same thing in #53 with a very small comment. I'm encouraged that we ended up on a similar solution!

https://github.com/chanzuckerberg/imaging-active-learning/blob/c54037711570cfdfc6e1b99583550f5fd72a0785/src/renderers/webgl_shader_program.ts#L167-L177

shlomnissan:
> A new fragment shader will be needed for some types to handle texture sampling.

Prioritizing conditional inclusion for fragment shaders might be a good idea. This would allow us to select the sampler or sampling method based on the internal texture format. My current thinking is that a shader program should be bound to the renderable object type and be configurable according to its internal properties.

src/layers/image_layer.ts
```diff
@@ -58,11 +58,15 @@ export class ImageLayer extends Layer {
       throw new Error(`Expected one chunk. Instead found ${chunks.length}`);
     }
     const chunk = chunks[0];
-    const texture = new Uint16Texture2D(
+
+    // TODO: DataTexture2D and ImageChunk share an almost identical interface.
+    // Should we simplify this?
```
aganders3: They're also quite close to the format Zarrita is already returning. I think it's okay to convert to our own type though.

I think the difference here is that `DataTexture2D` will be tightly packed in order to facilitate transfer to GPU. That's true now for `ImageChunk` as well (column stride == 1) but that seems like a more arbitrary choice. I could be missing something though.

The interface could be unified though - maybe just provide an attribute that would "pack" data if necessary.

I'm also not totally sold on the `ImageChunk` name - seems like it conflicts with the "chunks" in a Zarr.

None of that should be changed here though.

src/renderers/webgl_textures.ts
```diff
@@ -78,17 +78,15 @@ export class WebGLTextures {
     gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MIN_FILTER, gl.LINEAR);
   }
 
-  private configuredUint16Texture2D(texture: Uint16Texture2D) {
+  private configuredDataTexture2D(texture: DataTexture2D) {
```
aganders3: I missed this in previous PRs, but I feel like I'm reading this incorrectly. Is the method meant to be `configureDataTexture2D`? I don't understand the use of past-tense `configured`.

src/data/ome_zarr_image_loader.ts
aganders3: One thought on this: could you share the `TextureDataType` from `data_texture_2d.ts` here somehow? I think it probably makes sense to keep them separate but it's just a little suspicious they match 1:1 right now.

edit: I think we do want them separate, as we may have loaders that don't support all the types the texture could.

aganders3 (approved):
LGTM! Nothing blocking, just comments.

src/data/ome_zarr_image_loader.ts
andy-sweet: I was sharing that definition at some point, but actually prefer the separation like you mention in your update. I think having a general boundary between rendering and loading is desirable for the reason you mentioned. It's also possible that a loader supports more types than a texture. In either case, it may be the responsible of something else in the middle to handle conversion of loaded types to renderable types.

src/layers/image_layer.ts
andy-sweet: I'll probably just remove this TODO. Again, I think it's OK to have a little redundancy to allow for loading and rendering to diverge as needed to support a variety of types/layouts and the necessary conversions.

Open to other names than chunks, though neuroglancer also uses that terminology for more than just image/array chunks: https://github.com/google/neuroglancer/tree/master/src/chunk_manager

src/renderers/webgl_textures.ts
andy-sweet: Huh, I completely missed that as I was copy/pasting the method name from `configuredTexture2D`. 

@shlomnissan : this is just a typo, right? If so, I'll go ahead and make the corrections in this PR.

examples/image2d_from_omezarr5d_u8.ts
```diff
@@ -0,0 +1,31 @@
+import {
```
shlomnissan: If PR #46 is merged first, we may need to make some adjustments — adding an `index.html` file and moving this file into a separate directory.

src/data/ome_zarr_image_loader.ts
```diff
@@ -35,6 +35,14 @@ interface Multiscale {
...
+const dataTypeNames = dataTypes.map((DataType) => DataType.name);
+type DataType = InstanceType<(typeof dataTypes)[number]>;
+
+function isDataType(value: unknown): value is DataType {
+  return dataTypes.some((DataType) => value instanceof DataType);
+}
```
shlomnissan, line 44: > The disadvantage is that I need to sometimes re-read this

I prefer a simpler approach, but I don't feel strongly about it.
```suggestion
const dataTypeNames = ["Uint8Array", "Uint16Array"];
function isDataType(value: unknown) {
  return value instanceof Uint8Array ||
         value instanceof Uint16Array;
}
```

src/layers/image_layer.ts
shlomnissan: > I think it's OK to have a little redundancy

I completely agree. `DataTexture2D` is likely to diverge rapidly. We should consider adding default properties soon, such as wrapping settings, filtering options, mipmap generation, and encoding parameters.

src/objects/textures/data_texture_2d.ts
```diff
@@ -0,0 +1,54 @@
...
+  private readonly height_: number;
+  private readonly rowStride_: number;
+  private readonly rowAlignmentBytes_: number;
+
+  constructor(
+    data: TextureDataType,
+    width: number,
+    height: number,
+    rowStride: number,
+    alignmentBytes: number
```
shlomnissan, line 17: In the previous PR, I suggested we could potentially replace the existing `Texture2D` class that loads an image with this more general implementation. To achieve this, we should make the interface more flexible by assigning default values to `rowStride` (0) and `alignmentBytes` (4), allowing users to update these using mutators. Additionally, it's worth noting that valid values for `UNPACK_ALIGNMENT` are limited to 1, 2, 4, or 8. We might want to create a specific type for `alignmentBytes` rather than allowing any number.

src/objects/textures/data_texture_2d.ts
```diff
@@ -0,0 +1,54 @@
...
+  public get rowStride() {
+    return this.rowStride_;
+  }
+
+  public get rowAlignmentBytes() {
+    return this.rowAlignmentBytes_;
+  }
+
+  public get type() {
+    return "DataTexture2D";
```
shlomnissan, line 52: I didn't realize you could reflect on the class name using `this.constructor.name`. I wonder if, instead of an abstract method, and manually defining `type()` for each object, we could have a default implementation in the base class that handles this automatically?

src/renderers/webgl_textures.ts
shlomnissan: Yep, typo.

src/renderers/webgl_textures.ts
```diff
@@ -78,17 +78,15 @@ export class WebGLTextures {
...
     const gl = this.gl_;
-    // Use an unpack alignment of 2 to support any row length of 2-byte
-    // uint16 pixel values.
-    gl.pixelStorei(gl.UNPACK_ALIGNMENT, 2);
-    gl.pixelStorei(gl.UNPACK_ROW_LENGTH, texture.rowLength);
+    gl.pixelStorei(gl.UNPACK_ALIGNMENT, texture.rowAlignmentBytes);
+    gl.pixelStorei(gl.UNPACK_ROW_LENGTH, texture.rowStride);
```
shlomnissan, line 84: This should be conditional IMO, even for the `DataTexture2D` type. We should only call this when the default value has actually changed. This relates to my earlier comment about increasing flexibility.

src/renderers/webgl_textures.ts
```diff
@@ -78,17 +78,15 @@ export class WebGLTextures {
...
     const level = 0;
-    const internalFormat = gl.R16UI;
+    const internalFormat = this.dataTexture2DGlInternalFormat(texture);
```
shlomnissan: I believe the internal format should be determined based on the texture's input format and type, which in turn should be derived from texture properties. There's a useful table in the "OpenGL ES 3 Programming Guide" (attached) that lists and explains different valid combinations for integer textures. _R_ denotes renderable (can be used as a render target) and _F_ stands for filterable. I'm curious about the best approach to determine these values. I haven't yet investigated how other libraries handle this issue.
Screenshot 2024-09-27 at 1 12 25 PM

src/renderers/webgl_textures.ts
```diff
@@ -78,17 +78,15 @@ export class WebGLTextures {
...
     const border = 0;
     const format = gl.RED_INTEGER;
```
shlomnissan: I think we also want the format to be inferred from the texture object, but we can deal with this later.

src/renderers/webgl_textures.ts
```diff
@@ -102,10 +100,26 @@ export class WebGLTextures {
     );
     gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_S, gl.CLAMP_TO_EDGE);
     gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_T, gl.CLAMP_TO_EDGE);
-    // Use NEAREST with the R16UI internal format because integer valued textures
-    // are not generally texture filterable.
+    // Use NEAREST because unsigned integer valued textures are not generally
+    // texture filterable.
     // https://webgl2fundamentals.org/webgl/lessons/webgl-data-textures.html
     gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MIN_FILTER, gl.NEAREST);
     gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MAG_FILTER, gl.NEAREST);
```
shlomnissan, line 107: This may not be necessary. `GL_NEAREST` is typically the default value for the magnification filter, though not for the minification filter (which assumes mipmap usage by default).

shlomnissan (approved):
Looks good to me. I've left some comments, but none of them are blocking.

src/objects/textures/data_texture_2d.ts
andy-sweet: I'll probably leave that for this PR. One reason I left the existing `Texture2D` alone is that I think there could be some value in having a specialized RGB or RGBATexture. Sometimes biomedical images comes like this (e.g. HPA), or we might want to design in such a way to support other scientific imaging data (e.g. geospatial).

src/objects/textures/data_texture_2d.ts
andy-sweet: I'm not sure it is not safe in general because of things like minification (i.e. there's no guarantee the name of the class' constructor is what you defined it as in the original source code). But I think it should be safe for built in classes (e.g. `Uint8Array`).

src/renderers/webgl_textures.ts
andy-sweet: Do you mean the WebGL default value? Is that only accessible in docs, or would we need to check it at runtime once we have access to the GL context?

Another approach Nathan mentioned is to store all the GL state you're going to modify before doing so, and resetting it once you've updated/drawn what you need to.

src/renderers/webgl_textures.ts
andy-sweet: I also looked at similar tables when defining the values I was using here, but I'm happy to make this more explicit and exhaustive. Will think about this and might try something for this PR.

src/renderers/shaders/uint_image_frag.glsl
```diff
@@ -1,11 +1,10 @@
 #version 300 es
 
 precision mediump float;
-precision mediump usampler2D;
 
 layout (location = 0) out vec4 fragColor;
 
-uniform usampler2D texture0;
+uniform mediump usampler2D texture0;
```
shlomnissan, line 7: Interesting. I didn't know you needed to specify precision for sampler uniforms.

src/renderers/webgl_textures.ts
andy-sweet: When I do not set this explicitly (e.g. by commenting out this line), I don't see an error, but the texture does not seem to be drawn either.

According to gl.docs (for ES3), the default is GL_LINEAR: https://docs.gl/es3/glTexParameter

andy-sweet:
Thanks for the reviews both. I'll have a think about better defining the mapping between some of the data types and GL texture types before merging this.

src/renderers/shaders/uint_image_frag.glsl
andy-sweet: If I omit it, I get

```
compiling shader: ERROR: 0:7: 'usampler2D' : No precision specified
```

I folded the precision into this line instead of having a separate line above as before.

I wasn't sure how to pick the precision for the sampler though, so advice there would be appreciated!

src/renderers/webgl_textures.ts
shlomnissan: I was thinking runtime check like:
```ts
if (texture.rowAlignmentBytes != DataTexture2D.DEFAULT_UNPACK_ALIGNMENT) {
    gl.pixelStorei(gl.UNPACK_ALIGNMENT, texture.rowAlignmentBytes);
}
```
But maybe it's not a big deal.

src/renderers/webgl_textures.ts
shlomnissan: Definitely not a blocker. We can create a task to look into it later.

src/renderers/webgl_textures.ts
shlomnissan: Ah, I was under the impression it was NEAREST, given that OpenGL typically defaults to the most efficient option.

src/renderers/shaders/uint_image_frag.glsl
shlomnissan: Yeah, I looked at the documentation, and as I understand it, it appears this is necessary. GL ES lacks default precision, requiring us to define it for floats. However, we can't define precision for integers globally as we do with floats and associate them with integer samplers. Instead, we must explicitly state the precision each time we use an integer sampler.

shlomnissan: This doesn't apply to normalized texture samplers.

src/renderers/webgl_textures.ts
andy-sweet: A possible simple improvement to this PR is to couple the internalFormat, format, and type by returning from a single lookup function like

```typescript
  private dataTexture2DGlTextureParams(texture: DataTexture2D) {
    const gl = this.gl_;
    if (texture.data instanceof Uint8Array) {
      return {
        internalFormat: gl.R8UI,
        format: gl.RED_INTEGER,
        type: gl.UNSIGNED_BYTE,
      }
    }
    if (texture.data instanceof Uint16Array) {
      return {
        internalFormat: gl.R16UI,
        format: gl.RED_INTEGER,
        type: gl.UNSIGNED_SHORT,
      }
    }
    const exhaustiveCheck: never = texture.data;
    throw Error(`Unsupported data type: ${exhaustiveCheck}`);
  }
```

which I prefer over the current form because the implementation explicitly reflects that these properties are not independent and (for `DataTexture2D` as it's defined now) can be derived just from the type of the underlying data buffer.

This also means that adding support for a new data buffer type (e.g. `Float32Array`) only needs modification in one place (at least for textures - the shaders are another issue).

What do you think?

src/renderers/webgl_textures.ts
andy-sweet: See my recent comment for a simple way to communicate this: https://github.com/chanzuckerberg/imaging-active-learning/pull/50/files#r1781316530

src/renderers/webgl_textures.ts
andy-sweet: To take that example specifically, I think that condition will always be true (because `DEFAULT_UNPACK_ALIGNMENT` should be 4), so right now this just introduces a condition and more state for no immediate benefit.

Therefore, I'm not going to include any changes in this PR. I think you're planning to work on generalizing our own texture definition, so I think this could become more useful then.

One problem about this call to `pixelStorei` (which you mentioned on Slack), is that it's modifying global state. So either every call to `texImage2D` or similar should set this or the setter here (and elsewhere) should reset the value to what is was before. This is a problem before and after this PR. As I wrote that response, I realized that maybe you meant the combination of the code above and a reset to the default after calling is a way to fix that issue. That's also pretty simple and reasonable with the main downside being that we'll need to define the defaults (or perhaps look them up on once at runtime on initialization of the context).

src/renderers/shaders/uint_image_frag.glsl
aganders3: > This doesn't apply to normalized texture samplers.

I didn't understand this prior, and did some quick reading on this.

Would it make sense to use a normalized texture sampler here? For example for the uint8 data we could use this, which would allow using a `sampler2D` instead of a `usampler2D`. This means we'd also be able to do linear filtering (if desired).

InternalFormat | Format | Type
-- | -- | --
R8 | RED | UNSIGNED_BYTE

andy-sweet: That will work for uint8 support, but not for uint16 support (without converting the data on the CPU) based on my reading of Table 2 (Sized Internal Formats) on docs.gl: https://docs.gl/es3/glTexImage2D

In that case, if we want to support both uint8 and uint16 without CPU conversion, we'll need separate fragment shaders (treating uint8 as a special case) or will need more sophisticated shader composition/generation.

I guess one advantage of normalizing all ints to floats (on CPU or via GL) is that the shader programs may become much simpler - I have some vague memory that napari/vispy does something similar?

aganders3: > That will work for uint8 support, but not for uint16 support

Ahh, okay that makes sense. Thanks!

> I guess one advantage of normalizing all ints to floats (on CPU or via GL) is that the shader programs may become much simpler - I have some vague memory that napari/vispy does something similar?

My memory of this was also vague, so I looked into it in vispy just now. By default it does [normalization on CPU and stores data on the GPU as an unsigned integer](https://vispy.org/api/vispy.visuals.image.html#:~:text=provided%20through%20%E2%80%98custom_kernel%E2%80%99.-,texture_format,-%3A%20numpy.dtype%20%7C%20str). This only supports 8 and 16, but this is confusing (and annoying) because I don't see R16 in the tables. This may be a difference with OpenGL/WebGL. From what I can tell Vispy _does not_ support non-normalized formats (e.g. `8i`, `8ui`, `32i`, `32ui`) at all.

PR #51, opened 2024-09-23
Adding script to create tracks mock data

JoOkuma:
### Summary

This PR adds a script to create mock tracks data using the 2D multi-color dataset, https://public.czbiohub.org/royerlab/ultrack/multi-color/

This is the updated version, where `positions` were replaced with individual coordinate keys (e.g., `x`, `y`, `z`).

The online data currently does not reflect this change; once we reach a consensus, I'll update it.

### Tests & Checks

I manually checked if the script ran.


andy-sweet (changes requested):
Thanks Jordão! I was able to read the Python module easily enough and run the script successfully, so this is good enough for me in terms of functionality.

My requested changes are to move the script into a more suitable place (we added `ultrack-prototype` to the top level) and to add a few early development utilities (i.e. `ruff`) to keep Python code consistent and runnable.

I created a PR on top of this branch that catches this branch up with main and makes those suggested changes: https://github.com/chanzuckerberg/imaging-active-learning/pull/54

tools/create_tracks_mock_data.py
```diff
@@ -0,0 +1,156 @@
...
+
+RNG = np.random.default_rng(42)
+
+
+class TrackData(BaseModel):
+    track_id: int
+    t: list[int]
+    z: list[float]
+    y: list[float]
+    x: list[float]
```
aganders3: This doesn't quite match the data in https://public.czbiohub.org/royerlab/ultrack/multi-color/mock_data.json, right?

aganders3:
Sorry, I'm finally getting around to looking at this. It looks great! I just have one question about the models because to me it doesn't quite match what is in the sample data JSON you shared.

In #55 I added a server that just spits out static data but I think it's probably better if we copy this functionality into the mock server to make it more dynamic. I think that can happen in a new PR once both are merged, though.

tools/create_tracks_mock_data.py
JoOkuma: Indeed, @andy-sweet requested to use this structure because `time` and `position` differed.
Once we are on the same page, I'll update the public data.

JoOkuma:
@aganders3, I agree with you.
I reviewed #54 and I think we can be merged into this PR.
Then, once this PR and #55 are merged, we can have a new PR to remove the redundant functionalities and have a dynamic server.

andy-sweet (approved):
This looks good to me now. Assume we can handle the mock data schema elsewhere.

PR #53, opened 2024-09-23
Generalize handling of uniforms in shaders/renderer/renderable objects

aganders3:
### Summary
This improves the handling of uniforms in the `WebGLShaderProgram` object. It uses WebGL methods to look up uniform location and type, and uses this information to choose the correct function when setting each uniform.

### Related Issue
This was motivated by work on #46, where I needed to add some new uniforms not used in the existing shader.

### Tests & Checks
Tested manually with `main` and with #46


tsconfig.json
```diff
@@ -10,6 +10,8 @@
       "@": ["."],
       "@/*": ["./*"]
     },
+    "experimentalDecorators": true,
```
andy-sweet: I'm confused about typescript's decorators. I think this option corresponds to what's called [stage 2 decorators](https://www.typescriptlang.org/docs/handbook/decorators.html), but there are also [non-experimental stage 3 decorators available in TS 5+](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/#decorators)

aganders3: Yeah, I'm confused too. From what I understand the stage 3 decorators don't support properties, so can't replace all usage of the stage 2. They can also at least partially coexist? I think the other option here (`exportMetadata`) maybe is incompatible with stage 3 decorators or something.

In any case I've removed the decorators!

src/renderers/webgl_renderer.ts
```diff
@@ -29,8 +29,19 @@ export class WebGLRenderer extends Renderer {
...
+
+    program.setUniformIfNeeded(
+      "Projection",
+      this.activeCamera.projectionTransform
+    );
+    program.setUniformIfNeeded("ModelView", this.activeCamera.viewTransform);
+    program.setUniformIfNeeded("Resolution", [this.width, this.height]);
+
+    switch (object.type) {
+      // TODO: set uniforms for other types of renderable objects here
```
shlomnissan: Yep, this approach seems straightforward for now :+1:

src/renderers/webgl_shader_program.ts
```diff
@@ -32,39 +32,65 @@ export class WebGLShaderProgram {
...
-    // Preprocessing uniform locations doesn’t work for all use cases, so we
-    // can’t assume all uniform locations are cached on initialization.
-    if (!this.uniformLocations_.has(name)) {
-      const location = this.gl_.getUniformLocation(this.program_, name);
-      if (location) {
-        this.uniformLocations_.set(name, location);
-      } else {
-        throw new Error(`Uniform ${name} not found in shader program`);
-      }
+  public setUniformIfNeeded(name: string, value: unknown) {
```
shlomnissan: What's the rationale behind including an "if needed" function? I'm curious about scenarios where not raising an error would be preferable.

src/renderers/webgl_shader_program.ts
```diff
@@ -32,39 +32,65 @@ export class WebGLShaderProgram {
...
+    this.setUniformUnsafe(location, value, info.type);
   }
 
-  public setUniformMat4(name: string, value: mat4) {
-    this.gl_.uniformMatrix4fv(this.getUniformLoc(name), false, value);
-  }
-
-  public setUniformVec2(name: string, value: vec2) {
-    this.gl_.uniform2fv(this.getUniformLoc(name), value);
+  public setUniform(name: string, value: unknown) {
```
shlomnissan, line 35: I may be overlooking something, but I believe having a single method called `setUniform` that combines both error handling logic and the type-unsafe assignment would provide a more robust interface.

src/renderers/webgl_renderer.ts
```diff
@@ -29,8 +29,19 @@ export class WebGLRenderer extends Renderer {
...
+      case "Mesh":
+        break;
+    }
```
andy-sweet: Assuming this PR is getting merged to `main`, then I'd prefer the TODO alone.

```suggestion
    // TODO: set uniforms for other types of renderable objects here
```

shlomnissan (approved):
Overall, these changes look good to me. However, I'm not convinced that the `setUniformIfNeeded` function is a good idea. I'll approve the PR since we can modify this later, but if there's no compelling reason for this function, I'd prefer to address it before merging.

andy-sweet (approved):
Other than removing the switch/case, this looks good to me. And using the explicit uniform location check is a worthwhile improvement. Ideally, the TODO about type checking and earlier error handling would be done here, but not required from me.

src/renderers/webgl_shader_program.ts
aganders3: I was trying to separate the logic for uniforms that come from the framework (renderer) and those that come from a renderable object. The problem there is that of course not every shader will need all the possible uniforms from the renderer. So for example trying to set the `Resolution` uniform for a `Mesh` will error.

The other option I suppose is to put this into the switch below - which is fine with me.

aganders3: I removed this.

aganders3:
> Ideally, the TODO about type checking and earlier error handling would be done here, but not required from me. (@andy-sweet)

I went ahead and updated that here, too.

src/renderers/webgl_shader_program.ts
aganders3: I combined these methods - I think that's what you were suggesting but let me know if I didn't get what you meant!

src/renderers/webgl_shader_program.ts
```diff
@@ -128,3 +143,35 @@ export class WebGLShaderProgram {
...
+  WebGL2RenderingContext.SAMPLER_BINDING,
+]);
+
+// using an array and converting to a set allows us to also create a type here
+const SUPPORTED_UNIFORM_TYPES_ = [
+  WebGL2RenderingContext.FLOAT,
+  WebGL2RenderingContext.FLOAT_VEC2,
+  WebGL2RenderingContext.FLOAT_VEC3,
+  WebGL2RenderingContext.FLOAT_MAT4,
+] as const;
```
aganders3, line 173: It's a little tedious but we should probably just go ahead and make this work for all the uniforms we expect.

PR #54, opened 2024-09-25
Move script to new directory with some development utilities

andy-sweet:
### Summary
This moves the script defined in #51 into a new more suitable sub-directory that is now present after merging with `main`.

It also adds a requirements file for very basic tracking of Python dependencies and ruff + pre-commit for consistent linting and formatting of Python source files. Then fixes a couple of minor issues found by ruff.

PR #55, opened 2024-09-25
Add a backend server skeleton (FastAPI) for ultrack-prototype

### LINKED ISSUE #26: Skeleton FastAPI server and models
We plan to use FastAPI to generate the Active Learning Server component. The server will work with a PostgreSQL database that will store tracking data and annotations. Image data will be stored separately on static file hosting, as OME-Zarr.

The tentative plan is to host the application on Bruno, which will give it local access to data and compute.

* Initially two routes: GET `/question` and POST `/answer` (actually names may be TBD)
* Possibly one more route to trigger a model update
* Two models corresponding to "payloads" for these routes
* "Question" payload will include
  * Question ID
  * Bounds of a _region of interest_ in the corresponding image
  * Track annotations: points (nuclei center), tracks (edges between points)
  * Question type: validate cell death, validate cell division (enum)
  * Question-specific annotation: labeled point of cell death or division
* "Answer" payload will include
  * Question ID
  * Session ID and/or Answer ID (up to the backend to handle questions answered multiple times)
  * Answer: Positive | Negative | Undetermined (enum)

Jordão will implement logic in the routes, but for now we can create a skeleton project to include a script to launch the server and database.

aganders3:
### Summary
This creates a new FastAPI application with basic models matching the provided initial [mock_data](https://public.czbiohub.org/royerlab/ultrack/multi-color/mock_data.json), as well as endpoints for serving the data.

If you visit http://127.0.0.1:8000/mock_data/task you will get the full list of tasks from our initial [mock_data](https://public.czbiohub.org/royerlab/ultrack/multi-color/mock_data.json), which has just been copied into the server for simplicity at the moment. Individual task payloads are then available at `http://127.0.0.1:8000/mock_data/task/<task_id>`.

This does not implement any communication between backend/frontend right now. It also does not set up any database integration.

### Related Issue
Closes #26

### Tests & Checks
If you run this you can visit the URL it gives you (probably http://127.0.0.1:8000/) in your browser and you should see "Hello world".


aganders3:
@JoOkuma can you make sure you can install/run this using the instructions in `ultrack-prototype/backend/README.md`? It's my first time using `uv` for project management - just let me know if that's a bad idea or you're not comfortable with it.

ultrack-prototype/backend/.python-version
```diff
@@ -0,0 +1 @@
+3.11
```
andy-sweet, line 1: Is this needed for `uv`? If so, what does it mean?

ultrack-prototype/backend/README.md
```diff
@@ -0,0 +1,28 @@
+# Ultrack Prototype Backend
+
+This is the backend for the Ultrack prototype. It is a simple API for the frontend to GET questions
+and POST answers.
+
+## Installation
+
+This project is meant to be managed with `uv`. To install `uv`, run the following command (or see
```
andy-sweet, line 8: I tried out `uv` for a personal project and loved it. Very similar to the `npm` experience of project package management - i.e. it mostly works with no effort and it's fast.

andy-sweet (approved):
Looks good to me and I love the use of `uv`!

I suggested some more adhoc ways of tracking Python package dependencies and linting/formatting in #54 for the `tools` sub-directory. Those are not incompatible with this PR, but it might be nice to add a dependency group called `dev` or something that includes the dependencies to run those tools. That work can be done after both PRs are merged.

ultrack-prototype/backend/.python-version
aganders3: I don't think it's _needed_ for `uv` but `uv` [does create it and use it](https://docs.astral.sh/uv/guides/projects/#python-version). My understanding is that it tells `uv` which version of python to use by default when creating a new env.

We can delete and ignore this if it's problematic. I _think_ without this file it will use the minimum Python version from `pyproject.toml:project.requires-python` (and then create the file).

ultrack-prototype/backend/src/ultrack_learns/mock_data.py
```diff
@@ -0,0 +1,429 @@
+from ultrack_learns.models import Task
+mock_data = [
```
andy-sweet: Optional

There's also `ultrack-prototype/data/mock_data.json`, so ideally we'd have some way to merge this and that so we have one source for the mock data. It would also be ideal to generate it from Jordão's script in `tools`. Not required for a merge though.

aganders3: I looked into this a bit, but once again confused myself about "package data" etc. and how to include non-code into the repo. In 42cf354 this PR now enables the server to dynamically generate sample data.

aganders3:
This PR now moves the functionality from `tools` into the `mock_data` part of this repo. I'm not positive this is a net benefit but it does allow the server to dynamically generate sample data. This also means currently we _could_ remove the `data` and `tools` directories, but I didn't do that yet. I figure there's still a chance we'll want to store something there.

The first run to generate mock data is slightly slower because it's using a SQLite DB instead of Pandas directly. If this DB is already seeded it's about the same.

There's a lot to do here but I'm going to move on to other stuff. I'd be happy to merge this and fix later, but I am also fine to keep the PR open and improve it in-place since it should not conflict with anything else we add.

ultrack-prototype/backend/pyproject.toml
```diff
@@ -0,0 +1,22 @@
...
+    "sqlalchemy>=2.0.35",
+]
+
+[project.scripts]
+write-mock-data-json = "ultrack_learns.mock_data:write_mock_data_json"
+
+[tool.uv]
+dev-dependencies = [
+    "pandas>=2.2.3",
+]
```
aganders3, line 22: I added this with `uv add --dev pandas` and was surprised to see it under `[tool.uv]` and not `[project.optional-dependencies]`. I'm happy to change this but need to understand then how `uv` would deal with specifying the extras.

Actually, this seems kind of nice to treat "dev" as a special case?
https://docs.astral.sh/uv/concepts/dependencies/#development-dependencies
> Unlike optional dependencies, development dependencies are local-only and will not be included in the project requirements when published to PyPI or other indexes. As such, development dependencies are included under [tool.uv] instead of [project].

andy-sweet: Looks like you just need to scroll up a bit more to find [optional dependencies](https://docs.astral.sh/uv/concepts/dependencies/#optional-dependencies):

The equivalent here could be something like

```
uv add pandas --optional tools
```

But I agree that `pandas` is intended to be a dev/local only, so looks like the right choice here.

PR #57, opened 2024-09-25
Add orthographic camera and a simple transform

### LINKED ISSUE #23: Implement a Transform data structure for renderable objects
- Include vectors representing position and scale
- Include a quaternion representing the current orientation
- Public member functions to translate, scale, and rotate the transform
- Maintain an instance of a matrix representing the transformation
- Implement an accessor function that updates the matrix if needed (when the transform is stale) and returns the transformation matrix
- By the end of this task, we should be able to modify the mesh transformation within the SingleMeshLayer.

shlomnissan:
### Summary

This PR introduces an orthographic camera and a basic transform for all
renderable objects. While the transform is currently incomplete,
I've put this together this to test the orthographic camera and
illustrate my intended approach.

### Related Issue
Closes #23 

### Tests & Checks

Visual confirmation that the transform and camera function as intended. See screenshot below.

Screenshot 2024-09-25 at 2 27 36 PM

### DIFF
--- a/src/core/renderer.ts
+++ b/src/core/renderer.ts
@@ -2,6 +2,7 @@ import { LayerManager } from "./layer_manager";
 import { Camera } from "objects/cameras/camera";
 import { RenderableObject } from "core/renderable_object";
 import { PerspectiveCamera } from "objects/cameras/perspective_camera";
+import { OrthographicCamera } from "@/objects/cameras/orthographic_camera";
 
 export abstract class Renderer {
   private readonly canvas_: HTMLCanvasElement | null;
@@ -46,6 +47,9 @@ export abstract class Renderer {
     if (this.activeCamera_) {
       if (this.activeCamera_ instanceof PerspectiveCamera) {
         this.activeCamera_.setAspectRatio(this.width_ / this.height_);
+      }
+      if (this.activeCamera_ instanceof OrthographicCamera) {
+        this.activeCamera_.setFrame(0, this.width, 0, this.height);
       }
       this.activeCamera_.updateTransforms();
     }
--- a/src/core/renderer.ts
+++ b/src/core/renderer.ts
@@ -49,9 +49,14 @@ export abstract class Renderer {
         this.activeCamera_.setAspectRatio(this.width_ / this.height_);
       }
       if (this.activeCamera_ instanceof OrthographicCamera) {
-        this.activeCamera_.setFrame(0, this.width, 0, this.height);
+        this.activeCamera_.setFrame(
+          -this.width / 2,
+          this.width / 2,
+          -this.height / 2,
+          this.height / 2
+        );
       }
-      this.activeCamera_.updateTransforms();
+      this.activeCamera_.update();
     }
   }
 
--- a/src/core/transform.ts
+++ b/src/core/transform.ts
@@ -1,0 +1,27 @@
+import { vec3, mat4 } from "gl-matrix";
+
+export class Transform {
+  private dirty_ = true;
+  private matrix_ = mat4.create();
+  private position_ = vec3.create();
+
+  public translate(axis: vec3, distance: number) {
+    const v = vec3.scale(vec3.create(), axis, distance);
+    vec3.add(this.position_, this.position_, v);
+    this.dirty_ = true;
+  }
+
+  // TODO: add scale
+
+  // TODO: add rotate
+
+  public get matrix() {
+    if (this.dirty_) {
+      this.matrix_ = mat4.create();
+      mat4.translate(this.matrix_, this.matrix_, this.position_);
+
+      this.dirty_ = false;
+    }
+    return this.matrix_;
+  }
+}
--- a/src/objects/cameras/camera.ts
+++ b/src/objects/cameras/camera.ts
@@ -3,7 +3,8 @@ import { mat4 } from "gl-matrix";
 
 export abstract class Camera extends RenderableObject {
   protected projectionTransform_ = mat4.create();
-  protected viewTransform_ = mat4.create();
+  protected near_ = 0;
+  protected far_ = 0;
 
   protected abstract updateProjectionTransform(): void;
 
@@ -25,6 +26,6 @@ export abstract class Camera extends RenderableObject {
   }
 
   get viewTransform() {
-    return this.viewTransform_;
+    return this.transform.matrix;
   }
 }
--- a/src/objects/cameras/camera.ts
+++ b/src/objects/cameras/camera.ts
@@ -2,30 +2,21 @@ import { RenderableObject } from "core/renderable_object";
 import { mat4 } from "gl-matrix";
 
 export abstract class Camera extends RenderableObject {
-  protected projectionTransform_ = mat4.create();
+  protected projectionMatrix_ = mat4.create();
   protected near_ = 0;
   protected far_ = 0;
 
-  protected abstract updateProjectionTransform(): void;
-
-  constructor() {
-    super();
-    this.updateProjectionTransform();
-  }
+  protected abstract updateProjectionMatrix(): void;
 
   public get type() {
     return "Camera";
   }
 
-  public updateTransforms() {
-    this.updateProjectionTransform();
+  public update() {
+    this.updateProjectionMatrix();
   }
 
-  get projectionTransform() {
-    return this.projectionTransform_;
-  }
-
-  get viewTransform() {
-    return this.transform.matrix;
+  get projectionMatrix() {
+    return this.projectionMatrix_;
   }
 }
--- a/src/objects/cameras/orthographic_camera.ts
+++ b/src/objects/cameras/orthographic_camera.ts
@@ -24,7 +24,7 @@ export class OrthographicCamera extends Camera {
     this.near_ = near;
     this.far_ = far;
 
-    this.updateProjectionTransform();
+    this.updateProjectionMatrix();
   }
 
   public setFrame(left: number, right: number, bottom: number, top: number) {
@@ -38,9 +38,9 @@ export class OrthographicCamera extends Camera {
     return "OrthographicCamera";
   }
 
-  protected updateProjectionTransform() {
+  protected updateProjectionMatrix() {
     mat4.ortho(
-      this.projectionTransform_,
+      this.projectionMatrix_,
       this.left_,
       this.right_,
       this.bottom_,
--- a/src/objects/cameras/orthographic_camera.ts
+++ b/src/objects/cameras/orthographic_camera.ts
@@ -1,0 +1,52 @@
+import { Camera } from "./camera";
+import { mat4 } from "gl-matrix";
+
+export class OrthographicCamera extends Camera {
+  private left_: number;
+  private right_: number;
+  private bottom_: number;
+  private top_: number;
+
+  constructor(
+    left: number,
+    right: number,
+    bottom: number,
+    top: number,
+    near = 0.1,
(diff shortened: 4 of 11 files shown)


src/core/renderer.ts
```diff
@@ -47,6 +48,9 @@ export abstract class Renderer {
       if (this.activeCamera_ instanceof PerspectiveCamera) {
         this.activeCamera_.setAspectRatio(this.width_ / this.height_);
       }
+      if (this.activeCamera_ instanceof OrthographicCamera) {
```
shlomnissan, line 51: I know we don’t like this, but I'll maintain this pattern for now until we devise a better solution for managing camera responses to viewport dimension changes.

src/core/transform.ts
```diff
@@ -0,0 +1,27 @@
+import { vec3, mat4 } from "gl-matrix";
+
+export class Transform {
```
andy-sweet: The exported name `Transform` is vague and as a client of a class like this, I'd hope for the name to be more specific and descriptive (e.g. `AffineTransform`, `MatrixTransform`, `ProjectiveTransform`). Based on the follow-up work described in the TODOs, I think we can avoid a rename for now though.

src/core/renderer.ts
andy-sweet: Seems fine for now.

src/core/transform.ts
```diff
@@ -0,0 +1,27 @@
...
+  private dirty_ = true;
+  private matrix_ = mat4.create();
+  private position_ = vec3.create();
+
+  public translate(axis: vec3, distance: number) {
```
andy-sweet: Optional

Not too important, but given my expectations of what `Transform.translate` would do and our current usage of it, I would just use a single argument here which is the translation vector (instead of the unit vector and the scale). That also avoids the need for the client to understand the position of arguments and an extra function call.

src/core/transform.ts
```diff
@@ -0,0 +1,27 @@
+import { vec3, mat4 } from "gl-matrix";
+
+export class Transform {
+  private dirty_ = true;
+  private matrix_ = mat4.create();
+  private position_ = vec3.create();
```
andy-sweet: Nit: I don't think of transforms having positions, so I'd prefer `translation_` or similar here.

```suggestion
  private translation_ = vec3.create();
```

src/core/transform.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  public translate(axis: vec3, distance: number) {
+    const v = vec3.scale(vec3.create(), axis, distance);
+    vec3.add(this.position_, this.position_, v);
+    this.dirty_ = true;
```
andy-sweet: With only translation, the dirty pattern is unnecessarily complex - you could just call `mat4.translate` with the existing `mat4` instance assigned to `this.matrix_`. But I assume it's here to support composing other parameters in some order (e.g. scale, rotation), right?

src/objects/cameras/camera.ts
```diff
@@ -25,6 +26,6 @@ export abstract class Camera extends RenderableObject {
   }
 
   get viewTransform() {
-    return this.viewTransform_;
+    return this.transform.matrix;
```
andy-sweet: It's confusing to have both `Camera.transform` (that returns a `Transform`) and `Camera.viewTransform` (that returns a `mat4`) when they both contain the same information.

Why not just remove `viewTransform` and have the client deal with accessing the `mat4` through the `matrix` property getter when they need it?

For consistency, I might also expect `projectionTransform` to return a `Transform` now that we have it.

src/objects/cameras/orthographic_camera.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  constructor(
+    left: number,
+    right: number,
+    bottom: number,
+    top: number,
+    near = 0.1,
+    far = 100.0
+  ) {
+    super();
```
andy-sweet, line 18: This call to `Camera`'s constructor will call `updateProjectionTransform` when `this.left_` (and similar values) are undefined. Is that safe? Is it desirable? If not, I'd be fine with just removing the call in the `Camera` constructor and force the non-abstract classes to populate `projectionTransform_` appropriately. Or maybe there's a way to use similar dirty flag logic as in the `Transform` implementation.

andy-sweet (changes requested):
Mostly lots of comments, questions, and nits.

The only requested change is to explain or fix the comment about the call to `super()` in `OrthographicCamera`, which isn't entirely new as `PerspectiveCamera` does something similar.

I think it might also be worth implementing support for scale and rotation to better understand the implementation of `Transform`, but we can also do that when needed.

src/objects/cameras/orthographic_camera.ts
shlomnissan: Good catch! I hadn't realized I was calling it in the `Camera` class. I'll remove the call from the `Camera` class for now. Having a base class invoke an abstract function defined in a derived class seems like a bad design choice anyway.

src/core/transform.ts
shlomnissan: I'll rename it to `AffineTransformas`. I agree it's more descriptive.

src/core/transform.ts
shlomnissan: I'm fine incorporating your suggestion. In previous implementations, I maintained generic transformation functions while introducing axis-aligned transformations (such as rotateX, rotateY, etc.) at a higher level when necessary, but I don’t think it’s necessary at this point.

src/core/transform.ts
shlomnissan: This naming convention aligns with engines and renderers I'm familiar with (like Three.js and Godot, Ogre3D). However, I don't have a strong preference here. Your suggestion seems reasonable, so I'll implement the change.

src/core/transform.ts
shlomnissan: The current approach is particularly valuable in the context of a scene graph, where multiple translations occur as transformations propagate through the graph. In such cases, performing matrix multiplication once (if at all) is beneficial. It's still applicable to our scenario, as you might want to apply multiple translations in a single rendering iteration.

My proposed idea involves introducing a vector to represent scale and a quaternion to represent the current orientation. If the dirty flag is set, we can assemble the transformation matrix inline, rather than using utility functions like `mat4.translate`, `mat4.scale`, etc., however, this using helper functions is probably fine as an intermediate state.

I'm inclined to leave it as is, unless you have strong feelings about it. I don't find it overly complex, and I don't think it's a crucial issue at this stage. However, I'm open to changing it.

src/objects/cameras/camera.ts
shlomnissan: - Another option is to change the visibility of the transform accessor to protected and let the underlying rendering object define how the object can be transformed. Either way, I agree it's confusing. I'll remove `viewTransform` for now.
- I'm unsure about converting the projection matrix to an affine transform object. This conversion might not be necessary, as I don't anticipate modifying it without considering the camera's parameters. However, I'm willing to make this change if it addresses an active use case. In the meantime, I'll rename `projectionTransform` to `projectionMatrix` to clarify that it doesn't return an affine transform object. Does this approach sound okay to you?

shlomnissan:
> I think it might also be worth implementing support for scale and rotation to better understand the implementation of Transform, but we can also do that when needed.

@andy-sweet I agree, this will be a quick follow-up. My initial goal was to implement an orthographic camera in this PR, but the transform was necessary for easy testing. I'll continue working on it in a subsequent PR.

src/objects/cameras/camera.ts
andy-sweet: Removing `viewTransform` and renaming to `projectionMatrix` sounds good to me!

src/objects/cameras/camera.ts
```diff
@@ -2,14 +2,14 @@ import { RenderableObject } from "core/renderable_object";
...
 export abstract class Camera extends RenderableObject {
-  protected projectionTransform_ = mat4.create();
-  protected viewTransform_ = mat4.create();
+  protected projectionMatrix_ = mat4.create();
+  protected near_ = 0;
+  protected far_ = 0;
 
   protected abstract updateProjectionTransform(): void;
 
   constructor() {
```
andy-sweet: Nit: after the removal of `this.updateProjectionTransform()`, I'm assuming javascript gives us this constructor as the default constructor, so this can also be removed?

andy-sweet (approved):
Looks good to me now.

src/renderers/webgl_renderer.ts
```diff
@@ -30,8 +32,14 @@ export class WebGLRenderer extends Renderer {
   protected renderObject(object: RenderableObject) {
     const program = this.getShaderProgram(this.getProgramName(object)).use();
 
-    program.setUniform("Projection", this.activeCamera.projectionTransform);
-    program.setUniform("ModelView", this.activeCamera.viewTransform);
+    const modelView = mat4.multiply(
+      mat4.create(),
+      object.transform.matrix,
+      this.activeCamera.transform.matrix
+    );
```
aganders3, line 39: Since this is not GL-specific code I feel like it _might_ belong in the parent class `render` method. Not blocking, just something to think about later.

src/objects/cameras/camera.ts
```diff
@@ -2,14 +2,14 @@ import { RenderableObject } from "core/renderable_object";
 import { mat4 } from "gl-matrix";
 
 export abstract class Camera extends RenderableObject {
-  protected projectionTransform_ = mat4.create();
-  protected viewTransform_ = mat4.create();
+  protected projectionMatrix_ = mat4.create();
+  protected near_ = 0;
+  protected far_ = 0;
 
   protected abstract updateProjectionTransform(): void;
```
aganders3: nit: with `projectionTransform` renamed `projectionMatrix` this could be renamed `updateProjectionMatrix` as well.

src/core/renderer.ts
aganders3: Yeah I don't think the general pattern needs to change here, but I'm still in favor of changing it!

If I were going to change anything here I'd center the orthographic frame on (0, 0), as I think that's how the perspective camera works.

aganders3 (approved):
Looks good, thanks! This will be really useful for playing with the sample data.

I think Cameras and Transforms could probably be more closely related later. It seems like you started down that road here and backed it out, so I'm fine with keeping them separate for now.

shlomnissan:
> I think Cameras and Transforms could probably be more closely related later. It seems like you started down that road here and backed it out, so I'm fine with keeping them separate for now.

@aganders3  Could you clarify what you mean by "closely related"? Is there a specific aspect or feature you think is missing from the current relationship between Cameras and Transforms?

src/renderers/webgl_renderer.ts
shlomnissan: That's a good point. I will leave it as is for now. I like the fact that it's an abstract method and that all the code is in one place, but we could reconsider when we introduce a new experimental renderer.

aganders3:
Sorry, it was just poorly worded. What I meant was right now we have objects that can have an `AffineTransform` (our code) and cameras that use `gl-matrix` operations to store transforms as `mat4`. If we have our own library of transforms (even if they're mostly just wrapping `gl-matrix` operations) it makes sense to me to use them in the Cameras as well.

src/core/renderer.ts
shlomnissan: That's a good point! I centered the orthographic plane on `(0, 0)`.

PR #65, opened 2024-10-02
Load video from prototype OME-Zarr volume and render frames based on slider value

### LINKED ISSUE #64: Load a range of time points from 2D+t sample image data
The goal of this task is to load a range of time points from the 2D+t sample image (video) data and allow each video frame to be rendered in the canvas. The output of the task is a new example that adds a input range slider that controls the time point that is currently visualized.

The sub-range of time points should be specifiable. I.e. we don't always want to allow loading of all the frames.

Loading frames does not need to be lazy. I.e. we can wait to load all the frames of the interest, and doing so is probably desirable for the prototype because we don't expect to load many frames (i.e. < 100).

andy-sweet:
### Summary
This adds a new `ImageSeriesLayer` type that loads multiple images from the prototype OME-Zarr  and allows one to be rendered at a time. It adds a new example to demonstrate such a usage and also modifies the prototype React app to do something similar.

I have some more complicated versions of this PR that start to think about different generalizations that we'll likely need to think about at some point (e.g. loading m-dimensional image chunks from n-dimensional image sources), but I kept things as simple as I could in this change.

In the same spirit, there is a lot of opportunity for optimization here. In particular, support for updating texture data in place or support for 2D texture arrays is an obvious improvement, but likely can be done in near-term future work.

### Related Issue
Closes #64

### Tests & Checks
I manually tested the new example, the prototype React app, and the existing examples.


ultrack-prototype/frontend/components/Renderer.tsx
```diff
@@ -1,37 +1,45 @@
...
+    layer.setTimeIndex(curTime);
+  }, [curTime]);
+
   // Use the mount-effect so that the renderer can find the corresponding
   // element by its ID.
   useEffect(() => {
+    console.debug("Renderer::mount");
     let lastRequestId = 0;
     const renderer = new WebGLRenderer(`#${canvasId}`);
+    // TODO: use an Orthographic camera.
```
andy-sweet: I started doing this, but need to think through how values should be set here on subsequent window resizes.

ultrack-prototype/frontend/components/App.tsx
```diff
@@ -1,8 +1,10 @@
 import { Box } from "@mui/material";
 import Renderer from "./Renderer";
 import PlaybackControls from "./PlaybackControls";
+import { useState } from "react";
 
 export default function App() {
+  const [curTime, setCurTime] = useState(0);
```
andy-sweet: @shlomnissan : this is how I imagined lifting React state up and then passing through the value and dispatcher appropriately. I think this simple application is relatively clean because only the slider component gets the dispatcher and the render only gets the value (so the corresponding data can be fetched), but I'd be interested in your thoughts and alternative approaches here.

ultrack-prototype/frontend/video_layer_props.ts
```diff
@@ -0,0 +1,12 @@
+// Source is 5D, so provide an interval in T and scalar indices in C (first of
```
andy-sweet: This is a hopefully short lived way of keeping the video layer props in one place, while allowing both components access to some of the state (e.g. the time interval).

At some point, I imagine this information will be in the payload from the server, so will change every time a new task in displayed.

src/data/ome_zarr_image_loader.ts
```diff
@@ -66,7 +66,7 @@ export class OmeZarrImageLoader {
     this.datasets_ = image.datasets;
   }
 
-  async loadChunks(region: Region): Promise<ImageChunk[]> {
+  async loadChunk(region: Region): Promise<ImageChunk> {
```
andy-sweet, line 69: I simplified the signature here since we only supported one chunk anyway and this makes the video data loading logic simpler. In the future, we'll want better generalization here, but I think this PR is a premature place for that.

ultrack-prototype/frontend/components/Renderer.tsx
andy-sweet: @aganders3 : if you're using this PR as a base for combined viz of points/tracks with the imaging data, you may need to fix/workaround this and the plane geometry to get proper alignment.

aganders3: Thanks - I am thinking to just base that work on `main` and keep it in an example for now. Either way I'll need to think about this so I think it's fine to ignore for this PR.

src/layers/video_layer.ts
```diff
@@ -0,0 +1,108 @@
...
+    }
+    this.state_ = "loading";
+    const loader = await this.source_.open();
+    // Wait to load the whole region over all time points.
+    this.dataChunks_ = [];
+    const region = structuredClone(this.region_);
+    for (let t = this.timeInterval_.start; t < this.timeInterval_.stop; ++t) {
+      region[this.timeDimensionIndex_].index = t;
+      const chunk = await loader.loadChunk(region);
+      this.dataChunks_.push(chunk);
```
aganders3: I think this will wait for each chunk to be loaded before requesting the next one. Maybe this could instead be done by pre-allocating the `dataChunks_` array and using `loader.loadChunk(region).then` to make promises that put each chunk into the array at the correct index. Then await `Promise.all` or `allSettled` before setting the state to `"ready"`.

src/layers/video_layer.ts
```diff
@@ -0,0 +1,108 @@
...
+    this.region_ = region;
+    this.timeDimensionIndex_ = region.findIndex(
+      (x) => x.dimension == timeDimension
+    );
+    if (this.timeDimensionIndex_ === -1) {
+      throw new Error(
+        `Could not find dimension ${timeDimension} in ${JSON.stringify(region)}`
+      );
+    }
+    const timeIndex = this.region_[this.timeDimensionIndex_].index;
```
aganders3: Scope-wise it's not a problem but this name clashes with `setTimeIndex`. Why not `timeInterval`?

src/layers/video_layer.ts
```diff
@@ -0,0 +1,108 @@
...
+      case "ready":
+        break;
+      default: {
+        const exhaustiveCheck: never = this.state;
+        throw new Error(`Unhandled LayerState case: ${exhaustiveCheck}`);
+      }
+    }
+  }
+
+  public async setTimeIndex(index: number) {
```
aganders3: Right now this is all synchronous - I guess you want the API to be async for the future?

Should `index` be stored then as state? I can imagine two `setTimeIndex` calls racing if they need to await data, and they could finish out of order.

aganders3 (commented):
I have a couple questions but generally this looks good and is cool to see!

I understand the intent is mostly to put _some_ structure in place and move the prototype forward. The description acknowledges intentionally leaving out some optimizations, and I think that's okay as long as ideas captured as issues and/or TODOs.

src/layers/video_layer.ts
andy-sweet: Oops, `async` is a leftover from a previous effort when this was doing async fetching. I'll just remove it for now and we can think about ordering of multiple async calls later.

src/layers/video_layer.ts
andy-sweet: One reason I pull this out here is to check if this really an interval instead of a scalar, so `timeInterval` seems presumptive.

The naming and typing in the region type could probably be improved since `index` is ambiguous and may better indicate a single value. But I don't have strong thoughts about this.

I think the clash with `setTimeIndex` is OK because the arg type there is `number`.

src/layers/video_layer.ts
andy-sweet: I just realized how slow this is when these requests are not cached... so will definitely look into the optimization described above and maybe some other ones for the loader (e.g. caching opened groups and arrays).

andy-sweet: I implemented your suggestion, which helps speed the download times up by about a factor of 2 when the responses are not cached.

I also tried a few other things with performance tracing enabled and didn't find anything too surprising. There is limited performance benefit in caching the opened array and doing so adds to the complexity of the code (so I did not add it here).

Passing the whole 2D+t region to zarrita does not speed things up much given that its still fetching the same chunks (again in the parallel). That's the approach I would like to take in the future - though not because of performance, but because it makes the loader more general purpose and forces us to think about how to remap sub-regions/chunks correctly back to the objects/textures.

examples/video2d_from_omezarr5d_u8/main.ts
```diff
@@ -0,0 +1,43 @@
...
+// Source is technically 5D (even though Z is unitary),
+// so provide indices at 3 dimensions to project to 2D.
+const source = new OmeZarrImageSource(url);
+const timeInterval = { start: 100, stop: 120 };
+const region = [
+  { dimension: "T", index: timeInterval },
+  { dimension: "C", index: 0 },
+  { dimension: "Z", index: 0 },
+];
+const layer = new VideoLayer(source, region, "T");
```
shlomnissan: I'm not sure "VideoLayer" is the best abstraction here. I associate "video" strongly with file formats and underlying implementations, which might not be the focus in this context. Perhaps an image array would be more appropriate? It's also unclear whether a separate layer is entirely necessary. I had envisioned it as an image layer where the application selects the current frame.

examples/video2d_from_omezarr5d_u8/main.ts
```diff
@@ -0,0 +1,43 @@
...
+const url =
+  "https://public.czbiohub.org/royerlab/ultrack/multi-color/image.zarr/";
+const layerManager = new LayerManager();
+const renderer = new WebGLRenderer("#canvas");
+const camera = new PerspectiveCamera(60, renderer.width / renderer.height);
+
+// Source is technically 5D (even though Z is unitary),
+// so provide indices at 3 dimensions to project to 2D.
+const source = new OmeZarrImageSource(url);
+const timeInterval = { start: 100, stop: 120 };
```
shlomnissan: Is this interval representing a unit of time or frame indices? If it's the latter, we could rename it to make that explicit.

examples/video2d_from_omezarr5d_u8/main.ts
```diff
@@ -0,0 +1,43 @@
...
+  { dimension: "C", index: 0 },
+  { dimension: "Z", index: 0 },
+];
+const layer = new VideoLayer(source, region, "T");
+layerManager.add(layer);
+
+const slider = document.querySelector<HTMLInputElement>("#slider");
+if (slider === null) throw new Error("Time slider not found.");
+slider.min = timeInterval.start.toString();
+slider.max = (timeInterval.stop - 1).toString();
```
shlomnissan: What's the reason for using a half-open interval?

examples/video2d_from_omezarr5d_u8/main.ts
```diff
@@ -0,0 +1,43 @@
...
+slider.addEventListener("input", (event) => {
+  if (!(event.target instanceof HTMLInputElement)) return;
+  const value = event.target.valueAsNumber;
+  console.log("slider changed: ", value);
+  layer.setTimeIndex(value);
+});
```
shlomnissan: The early return seems unnecessary here, particularly as you've already verified the HTML element exists.
```suggestion
slider.addEventListener("input", (event) => {
  const value = (event.target as HTMLInputElement).valueAsNumber;
  layer.setTimeIndex(value);
});
```

src/layers/video_layer.ts
```diff
@@ -0,0 +1,119 @@
...
+import { ImageChunk } from "data/image_chunk";
+import { DataTexture2D } from "objects/textures/data_texture_2d";
+
+interface ImageLayerSource {
+  open(): Promise<ImageChunkLoader>;
+}
+
+interface ImageChunkLoader {
+  loadChunk(input: Region): Promise<ImageChunk>;
+}
```
shlomnissan: I believe these interfaces are already defined in the image layer module. Would it be appropriate to simply export them from there?

src/layers/video_layer.ts
```diff
@@ -0,0 +1,119 @@
...
+    }
+  }
+
+  public setTimeIndex(index: number) {
+    if (this.state_ !== "ready") {
+      console.warn(`Trying to set time index before ready: ${this.state_}`);
+      return;
+    }
+    const { start, stop } = this.timeInterval_;
+    const chunkIndex = Math.round(index - start);
```
shlomnissan: Why do we need to round the difference?

src/layers/video_layer.ts
```diff
@@ -0,0 +1,119 @@
...
+    if (chunkIndex < 0) {
+      throw new Error(`Time index ${index} is before the start time: ${start}`);
+    } else if (chunkIndex >= this.dataChunks_.length) {
+      throw new Error(`Time index ${index} is after the stop time: ${stop}.`);
+    }
+    const chunk = this.dataChunks_[chunkIndex];
+    // TODO: create one object and update the texture in-place.
+    // Or use a texture array and update the index for the shader.
```
shlomnissan: Is there a GitHub issue for this particular task? I believe either approach would be beneficial to implement sooner rather than later, and I'm happy to take on the task.

src/layers/video_layer.ts
```diff
@@ -0,0 +1,119 @@
...
+    );
+    this.clearObjects();
+    this.addObject(new Mesh(this.plane_, texture));
+  }
+
+  private async load() {
+    if (this.state_ !== "initialized") {
+      throw new Error(`Trying to open chunk loader more than once.`);
+    }
+    this.state_ = "loading";
```
shlomnissan: I think creating a state mutator function for the layer type would be beneficial so we could add implicit state-change logs.

src/layers/video_layer.ts
```diff
@@ -0,0 +1,119 @@
...
+    // have a scale associated with them. We should load the whole
+    // region in and map back the chunks appropriately.
+    for (let t = start; t < stop; ++t) {
+      const region = structuredClone(this.region_);
+      region[this.timeDimensionIndex_].index = t;
+      loadPromises.push(
+        loader
+          .loadChunk(region)
+          .then((chunk) => (this.dataChunks_[t - start] = chunk))
+      );
```
shlomnissan: I imagine this will be part of a chunk loader in the future, loading all the data upfront. While I'm not sure if there's a clear advantage, is it possible to load the entire range in a single operation rather than loading chunks in a loop?

ultrack-prototype/frontend/components/App.tsx
shlomnissan: I think this approach is okay, but I would structure it slightly differently. The current time state and mutator could reside within the `PlaybackControls` component. Although this might result in some duplication, it would encapsulate all the logic necessary for the controls to operate independently, without relying on the parent component.

```ts
// App.tsx

const [curTime, setCurTime] = useState(0);

<PlaybackControls
  startTime={curTime}
  onTimeChange={(t: number) => setCurTime(t)}
></PlaybackControls>

// PlaybackControls.tsx

type PlaybackControlsProps = {
  startTime: number;
  onTimeChange: (curTime: number) => void;
}

export default function PlaybackControls(props: PlaybackControlsProps) {
  const [curTime, setCurTime] = useState(0);
  const [playing, setPlaying] = useState(false);

  useEffect(() => { setCurTime(props.startTime); }, []);

  useEffect(() => { props.onTimeChange(curTime); }, [curTime]);

  useEffect(() => {
    console.debug("PlaybackControls::useEffect::playing: ", playing);
    if (playing) {
      const interval = setInterval(
        () =>
          setCurTime((prevCurTime) => minTime + ((prevCurTime + 1) % numTimes)),
        playbackIntervalMs
      );
      return () => clearInterval(interval);
    }
  }, [playing, setCurTime]);

  // return ...
}
```

I haven't tested it, so I might be overlooking something. However, this approach allows the `PlaybackControls` component to function independently of external state. I recognize that it adds complexity, and I'm of the opinion that it's worth it, but I don't feel strongly about it.

ultrack-prototype/frontend/components/PlaybackControls.tsx
```diff
@@ -1,26 +1,35 @@
...
-// Use an arbitrary number of times just to exercise playback.
-const numTimes = 100;
+const minTime = videoLayerTimeInterval.start;
+const maxTime = videoLayerTimeInterval.stop - 1;
+const numTimes = maxTime - minTime + 1;
+
+interface PlaybackControlsProps {
+  curTime: number;
+  setCurTime: Dispatch<SetStateAction<number>>;
+}
```
shlomnissan: As I mentioned in an earlier PR, it would be beneficial to maintain consistency between `interface` and `type` declarations.

src/layers/video_layer.ts
```diff
@@ -0,0 +1,119 @@
...
+        throw new Error(`Unhandled LayerState case: ${exhaustiveCheck}`);
+      }
+    }
+  }
+
+  public setTimeIndex(index: number) {
+    if (this.state_ !== "ready") {
+      console.warn(`Trying to set time index before ready: ${this.state_}`);
+      return;
+    }
```
shlomnissan: > The only fixes I can think of require more foundational changes (e.g. observing layer state changes), which should be separate from this PR.

Referring back to your PR summary, I believe we should prioritize implementing a callback mechanism. This would allow clients who instantiate a layer to listen for and respond to layer state changes. Lets add a GitHub issue for it. I might be able to pick it up this week.

shlomnissan (approved):
Overall, this looks good! I'd like us to discuss the choice of using a video layer instead of sticking to an image layer and selecting different frames within the data source. While I don't think this should block progress, I'm curious about the rationale behind introducing this new abstraction.

examples/video2d_from_omezarr5d_u8/main.ts
andy-sweet: They are effectively frame indices since we don't have units yet and the time-axes in the target OME-Zarr have unit scale, no offset, and no units. Open to ideas for better names.

src/layers/video_layer.ts
andy-sweet: I tried that too. From some basic tracing there was no performance benefit (since the majority of time is spent waiting for chunks to download).

There is a complexity cost, but one that is likely worth it in the long term. In general, I view one of the layer's main responsibility's as requesting some region of data, getting some chunks that cover that region, then mapping those back to renderable objects.

I went for the simple approach initially to get us closer to a functional prototype.

ultrack-prototype/frontend/components/PlaybackControls.tsx
andy-sweet: Do you mean pick one, and stick to it? If so, I'm happy to use `type` whenever possible. I can make the relevant changes in this PR, then follow up with another one to make them everywhere.

src/layers/video_layer.ts
andy-sweet: https://github.com/chanzuckerberg/imaging-active-learning/issues/68

src/layers/video_layer.ts
andy-sweet: https://github.com/chanzuckerberg/imaging-active-learning/issues/69

src/layers/video_layer.ts
andy-sweet: Currently, we don't. In the future, the time interval could have non-integer index values, so we would need this then, but we would also need more. So I can just remove this for now.

src/layers/video_layer.ts
andy-sweet: Maybe. Given how small they are right now, I prefer the lack of a dependency between the two layers instead right now.

examples/video2d_from_omezarr5d_u8/main.ts
andy-sweet: No reason in particular. For integer values, that's my default, as it tends to simplify most related operations.

But as these values are likely to be non-integer in the future, I think we may want to switch using closed intervals to specify regions.

examples/video2d_from_omezarr5d_u8/main.ts
andy-sweet: I originally implemented it in the existing image layer, but that led to one of two downsides.

1. Lots of time/frame specific logic, that may not apply to some image layers.

or

2. The need for a more general purpose mapping between requested regions, chunk regions, and texture regions (i.e. beyond time specifically).

I think we should aim for (2) in the not too distant future (i.e. some version of it in the next month), but I elected to define a `VideoLayer` here to allow for (1) without the downside.

I was also interested in exercising the idea that we might want to create lots of different layer types, to avoid any one layer absorbing lots of complex logic to handle many cases. This one is maybe a little forced, but allows for a simple public API (i.e. `setTimeIndex`) and allows for potential optimizations (e.g. using a 2D texture array).

> I associate "video" strongly with file formats and underlying implementations, which might not be the focus in this context.

Interesting. I don't associate "video" with formats/implementations anymore than I associate "image" with formats/implementations, which is not much at all. And I would expect this layer to support stand video containers/codecs given a suitable implementation of a loader.

I don't feel too strongly about the name though. `ImageArray` is fine, though we would need a new name for `setTimeIndex` (maybe `setArrayIndex`). Or maybe we could use `ImageSeries` which would set us up for `VolumeSeries` if it is needed.

andy-sweet: Since this topic/discussion has the potential to get pretty long, let's discuss it in the sync today and summarize back here.

ultrack-prototype/frontend/components/App.tsx
andy-sweet: Thanks for clarifying. I think I see the trade-off now.

Your approach makes `PlaybackControls` something that can be used and tested without any explicit dependencies (other than the general dependency on React). But then if you want something else to be affected by a change to `curTime`, you need to define duplicate state that is synced via an observer callback.

The approach in this PR follows the approach of [lifting React shared state up](https://react.dev/learn/sharing-state-between-components), which means we can avoid the duplicate state and syncing. But that means we need to inject the `curTime` state and its mutator/dispatcher as props when using it in the application or for unit testing.

It's hard to evaluate the trade-off now because the app is so simple. I think the approach in this PR is preferred when we want to allow external control of `curTime` (e.g. this is one way that sharing a URL to share application state could work without needing to reload the entire application), so I'll probably stick with it for now.

Also a good topic for general discussion in a sync meeting, though I don't think we need to make a decision now.

ultrack-prototype/frontend/components/PlaybackControls.tsx
andy-sweet: I changed all current use of `interface` to `type` since the latter are slightly simpler.

examples/video2d_from_omezarr5d_u8/main.ts
aganders3: I don't think we came to a conclusion yesterday so I'm going to think out loud a bit here.

I think I'm in favor of separate layers for these uses (currently `Image` and `Video`)

I'm weakly in favor of a different name - I like `ImageSeries` currently.

Regarding separate layers. I am thinking a little about what kind of "explosion" we might see if we have sort of different axes of functionality. For example:
* Image
* ImageSeries
* Volume
* VolumeSeries
* MultiChannelImage
* MultiChannelVolume
* MultiChannelVolumeSeries

I think this might actually be fine? The concern I suppose is how they get constructed to keep the code somewhat DRY.

examples/video2d_from_omezarr5d_u8/main.ts
aganders3: Specifically for setting the slider though - it's half-open because `region.ts` defines `Interval` to be half-open, right?

andy-sweet: Yes. Which is consistent with [`TypedArray.slice`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/slice) and other libraries (e.g. zarrita) when thinking about integer intervals.

aganders3 (approved):
I only commented before, but in general this LGTM and I was able to (easily) build off it for #71 so I am happy to approve as-is.

I think we have some clear open questions but I don't think they should block this from going in, in particular as it moves the prototype significantly forward.

examples/video2d_from_omezarr5d_u8/main.ts
andy-sweet: `ImageSeries` is good with me. It also evokes (to me) that the duration between each image may be long and  irregular, while a video may imply short and mostly regular durations.

andy-sweet: I created https://github.com/chanzuckerberg/imaging-active-learning/issues/75 as an issue to capture one way of implementing the functionality of loading/rendering individual frames using the existing `ImageLayer`. I don't think it's an immediate priority, but is something I think we should revisit while thinking about the longer term technical design for the library.

andy-sweet: I also renamed `VideoLayer` to `ImageSeriesLayer`. I left `ImageSeriesLayer.setTimeIndex` alone since that still fits.

ultrack-prototype/frontend/components/App.tsx
andy-sweet: I'm going to merge this as is, but I'm still open to the alternative approach described here.

andy-sweet:
Thanks for the feedback both. I hope I addressed some of it and otherwise opened useful issues to address the rest in the future.

src/layers/video_layer.ts
andy-sweet: Another possible place for this is in `data/image_chunk` since the definitions are all related and pertain to loading image chunk data. Let me move them there for now and also rename `ImageLayerSource` to `ImageChunkSource`.

PR #67, opened 2024-10-07
Allow contrast limits to be set for images

### LINKED ISSUE #32: Control the normalized value of scalar pixel types
In #28, we added support for loading uint16 image data from OME-Zarr sources. The fragment shader performs normalization of the uint16 texture values by converting to float, then dividing by 256.0 which was chosen to present a specific example dataset well.

Instead, we should parametrize this normalization and allow these values to be passed through as part of the definition of the `ImageLayer`. This gives clients of the library (e.g. the prototype application) a way to control the presentation of scalar values.

andy-sweet:
### Summary
This allows the contrast limits of an image to be configured on layer creation. These limits are defined by a low and high value, which the current fragment shader uses to rescale the values.

The contrast limits are passed down to the `Mesh` renderable object and used by the renderer to set them as a uniform in the related fragment shaders. Alternatively, I considered storing them in the related texture types to which they are more closely related, but I avoided this because it would make the renderer uniform setting logic more complicated.

I also changed the image and image-series layer constructors to use de-structuring, which is not required, but feels more ergonomic in the long term, especially for arguments like `contrastLimits` which should be optional.

https://github.com/user-attachments/assets/0e74dcaa-0564-4c3a-9a7d-80c771e8ae16

### Related Issue
Closes #32

### Tests & Checks
I manually tested the existing image examples and the Ultrack prototype.


src/renderers/shaders/mesh_frag.glsl
```diff
@@ -5,9 +5,16 @@ precision mediump float;
 layout (location = 0) out vec4 fragColor;
 
 uniform sampler2D texture0;
+// This will likely be [0, 1], in which case it has no effect.
+// TODO: consider removing this to simplify and optimize this
+// shader, and instead handle conditionally setting the uniform
+// elsewhere.
```
andy-sweet, line 11: Elsewhere for the library means the logic currently in `webgl_renderer.ts` that sets the uniforms based on the type of renderable object. We could look at the texture type to determine if the associated shader has this uniform. More generally, this falls under the renderable-object/texture/program clean-up.

src/layers/image_layer.ts
```diff
@@ -2,19 +2,29 @@ import { Layer } from "core/layer";
 import { Region } from "data/region";
 import { ImageChunkSource } from "data/image_chunk";
 import { makeImageMesh, makeImageTexture } from "layers/image_utils";
+import { Mesh } from "objects/renderable/mesh";
+
+type ImageLayerProps = {
+  source: ImageChunkSource;
+  region: Region;
+  contrastLimits?: [number, number];
```
andy-sweet, line 10: I expect this to become something like `ChannelProps[]` where each element specifies per-channel properties like contrast limits, visibility, opacity, etc.

andy-sweet:
Closing this in favor of #152 which also supports multi-channel images, which will be needed for the organelle box component.

PR #70, opened 2024-10-08
Add layer state mutator and callback mechanism 

### LINKED ISSUE #68: Implement a callback mechanism to observer layer state changes
> The only fixes I can think of require more foundational changes (e.g. observing layer state changes), which should be separate from this PR.

Referring back to your PR summary, I believe we should prioritize implementing a callback mechanism. This would allow clients who instantiate a layer to listen for and respond to layer state changes. Lets add a GitHub issue for it. I might be able to pick it up this week.

_Originally posted by @shlomnissan in https://github.com/chanzuckerberg/imaging-active-learning/pull/65#discussion_r1790886211_

shlomnissan:
### Summary

This PR adds a layer state mutator function to layer objects that includes
state change logs and a callback mechanism allowing clients to listen to
state change events. 

### Related Issue
Closes #68

### Tests & Checks

- Automated tests.

### DIFF
--- a/src/core/layer.ts
+++ b/src/core/layer.ts
@@ -2,9 +2,15 @@ import { RenderableObject } from "./renderable_object";
 
 type LayerState = "initialized" | "loading" | "ready";
 
+type StateChangeCallback = (
+  newState: LayerState,
+  prevState?: LayerState
+) => void;
+
 export abstract class Layer {
   private objects_: RenderableObject[] = [];
-  protected state_: LayerState = "initialized";
+  private state_: LayerState = "initialized";
+  private callbacks_: StateChangeCallback[] = [];
 
   public abstract update(): void;
 
@@ -14,6 +20,17 @@ export abstract class Layer {
 
   public get state() {
     return this.state_;
+  }
+
+  public onStateChange(callback: StateChangeCallback) {
+    this.callbacks_.push(callback);
+  }
+
+  protected setState(newState: LayerState) {
+    const prevState = this.state_;
+    this.state_ = newState;
+    console.log(`${this.constructor.name} state change: ${newState}`);
+    this.callbacks_.forEach((callback) => callback(newState, prevState));
   }
 
   protected addObject(object: RenderableObject) {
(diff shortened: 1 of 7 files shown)


src/core/layer.ts
```diff
@@ -16,6 +22,17 @@ export abstract class Layer {
     return this.state_;
   }
 
+  public onStateChange(callback: StateChangeCallback) {
```
andy-sweet, line 25: Optional

I'm used to these methods being called something like `addStateChangeCallback` or `pushStateChangeListener`, where as `onStateChange` being something more like the name of a callback that is passed to such a method or for assignment to a single callback.

I'm open to this as a more concise form though.

andy-sweet (approved):
Looks great!

src/core/layer.ts
shlomnissan: I like `onStateChange` because it allows for intuitive code structures, such as:
```ts
layer.onStateChange((state) => {
    // handle state change logic
});
```
This approach is similar to other JavaScript events like `window.onresize`, though that's a global member that doesn't support multiple callbacks. For now, I'll keep it as is, but I'm open to changing it if there's a consensus that this isn't the right naming convention.

aganders3: If we want to mimic JS events, should we just derive from ("extend") [`EventTarget`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget)? This gives the same interface so it would be:
```
layer.addEventListener("state", (event) => {
   // handle state change logic
});
```

This gives a very familiar pattern and has the benefit that we don't have to implement some of the underlying logic of managing callbacks.

shlomnissan: I haven't encountered an example of extending event targets before, and I believe this implementation is straightforward and easy to understand. However, I'm open to reconsidering if you have strong opinions about it. In the meantime, I'm inclined to merge it.

andy-sweet:
Will try this out in #65 where it should improve some of the initialization.

PR #71, opened 2024-10-08
Display tracks on image series layer (single task)

### LINKED ISSUE #66: Co-visualize tracks and image
For the ultrack prototype, we need to co-visualize tracks (lines, maybe points) and images. With #28, #46, #50, and #57 - I think we have all the pieces necessary to do this.

My plan is to create a new example that hard-codes a single task from the [initial sample data](https://public.czbiohub.org/royerlab/ultrack/multi-color/mock_data.json). I will use these data to define two layers: an ImageLayer and a ProjectedLineLayer. The image layer will display a region of [sample image](https://public.czbiohub.org/royerlab/ultrack/multi-color/image.zarr/). The line layer will display the tracklets from the selected task overlaid on the image.

The goal here is to think about how to manage coordinates, units, and transforms to align objects from multiple layers and/or data sources. For this version, I plan to simply use pixel-space 2D coordinates. Still, this will require us to think about an origin at the very least.

Additional consideration of time (i.e. #65) out of scope for this task - showing a single time frame will be sufficient to close this issue (noting that a tracklet is actually a representation of multiple frames).

aganders3:
### Summary
This takes a single Task (hard-coded from [mock_data.json](https://public.czbiohub.org/royerlab/ultrack/multi-color/mock_data.json)) and displays it as an example.

This combines two layers:
* a image series layer to display the image (thus this branch is based on #65)
* a projected line layer to display the tracks

The specific example here is a possible "division" event. The identified cell moves for several frames (red track) before dividing into two (green and blue tracks).

The region around the tracks is determined by the tracks themselves, with some additional padding.

first-tracks

### Related Issue
Closes #66

### Tests & Checks
This is only manually tested


aganders3:
I removed the spline interpolation (#81) here so the tracks now look a bit worse, but it's less to review in one go.

The changes to the other examples, etc. here are just to keep the other examples working with changes to default camera FOV, etc. I'm happy to change that here or in a future PR if we come to agreement on what these coordinates should be. Normalizing to something like NDC was just the easiest way for me to reason about bringing the data into the same space for now.

examples/tracks/main.ts
```diff
@@ -0,0 +1,161 @@
...
+  [1842.0, 1356.0, 0.0],
+  [1842.0, 1356.0, 0.0],
+  [1840.0, 1367.0, 0.0],
+  [1844.0, 1378.0, 0.0],
+];
+const [center, xIndex, yIndex] = regionFromTrack([
+  ...trackA,
+  ...trackB,
+  ...trackC,
+]);
```
aganders3: One more thing _not_ handled here is the aspect ratio of the image patch itself. This looks okay because this example is close-ish to square (54x76) but it's being squished/stretched to fit a square.

examples/single_mesh_layer/main.ts
```diff
@@ -11,11 +11,17 @@ layersManager.add(singleMeshLayer);
 
 const renderer = new WebGLRenderer("#canvas");
 
+const aspect = renderer.width / renderer.height;
+const FOV = 2.5;
```
andy-sweet: Is this is effectively the inverse of a zoom factor? I.e. 2.5 is chosen to cover the [-1, 1] normalized coordinate range, plus a little extra for padding.

src/objects/cameras/orthographic_camera.ts
```diff
@@ -6,6 +6,8 @@ export class OrthographicCamera extends Camera {
   private right_: number;
   private bottom_: number;
   private top_: number;
+  // TODO: may want to configure which FOV direction is fixed
```
andy-sweet: Could this be based on whether the aspect ratio is > 1? I think that would give us some simple letter-box-like behavior instead of the clipping behavior we get currently when change the width of the window/canvas.

examples/single_mesh_layer/main.ts
aganders3: Yes, exactly. Zoom factor is probably a better way to think about it, as FOV may imply units where zoom is more of a ratio.

src/objects/cameras/orthographic_camera.ts
aganders3: Yeah, I think so.

examples/image_series_from_omezarr5d_u8/index.html
```diff
@@ -19,6 +19,13 @@
         width: 100%;
         height: 100%;
       }
+      #canvas {
+        flex: 1;
+        min-height: 0;
```
shlomnissan, line 24: I don't think it is necessary to specify a minimum height.

examples/single_mesh_layer/main.ts
shlomnissan: - If that is the case, should we rename the `FOV` variable?
- My experience with orthographic projection has always involved mapping the projection to screen space and treating units as pixels. I’m not entirely clear on why this wouldn’t work here, especially if we are dealing with images?

examples/tracks/main.ts
```diff
@@ -0,0 +1,164 @@
...
+  OmeZarrImageSource,
+  ProjectedLineLayer,
+  WebGLRenderer,
+} from "@";
+import { Interval } from "@/data/region";
+
+// payload roughly equivalent to task 0 from
+// https://public.czbiohub.org/royerlab/ultrack/multi-color/mock_data.json
+const trackA: [number, number, number][] = [
+  [1822.0, 1350.0, 0.0],
```
shlomnissan: - Are these values representing a position in 3D space? If so, it might be better to use an array of `vec3` objects so it’s slightly clearer.
- Will we draw the entire line at once, as shown in this example, or draw it progressively across frames?

examples/tracks/main.ts
```diff
@@ -0,0 +1,164 @@
...
+const timeInterval = { start: 28, stop: 39 };
+const region = [
+  { dimension: "T", index: timeInterval },
+  { dimension: "C", index: 0 },
+  { dimension: "Z", index: 0 },
+  { dimension: "Y", index: yIndex },
+  { dimension: "X", index: xIndex },
+];
+const regionWidth = xIndex.stop - xIndex.start;
+const regionHeight = yIndex.stop - yIndex.start;
```
shlomnissan: Are we planning to render individual regions (in this case, a 54x76 image), or will we render a large slice and zoom into the specific region?

examples/tracks/main.ts
```diff
@@ -0,0 +1,164 @@
...
+
+const imageSeriesLayer = new ImageSeriesLayer(source, region, "T");
+
+// scale to normalized device coordinates [-1, 1]
```
shlomnissan: I think this transformation should be handled within the line layer.

examples/tracks/main.ts
```diff
@@ -0,0 +1,164 @@
...
+  path: trackC,
+  color: [0.0, 0.0, 1.0],
+  width: 0.01,
+});
+
+const renderer = new WebGLRenderer("#canvas");
+
+const layerManager = new LayerManager();
+layerManager.add(lineLayer);
+layerManager.add(imageSeriesLayer);
```
shlomnissan, line 60: I appreciate how overlaying layers just works 👍

examples/tracks/main.ts
```diff
@@ -0,0 +1,164 @@
...
+  path: trackB,
+  color: [0.0, 1.0, 0.0],
+  width: 0.01,
+});
+
+lineLayer.addLine({
+  path: trackC,
+  color: [0.0, 0.0, 1.0],
+  width: 0.01,
+});
```
shlomnissan: Is this code redundant? It appears you're already adding the lines in the constructor.

```ts
const lineLayer = new ProjectedLineLayer([
  { path: trackA, color: [1.0, 0.0, 0.0], width: 0.01 },
  { path: trackB, color: [0.0, 1.0, 0.0], width: 0.01 },
  { path: trackC, color: [0.0, 0.0, 1.0], width: 0.01 },
]);
```

examples/image2d_from_omezarr5d_u16/main.ts
```diff
@@ -11,6 +11,7 @@ const url =
 const layerManager = new LayerManager();
 const renderer = new WebGLRenderer("#canvas");
 const camera = new PerspectiveCamera(60, renderer.width / renderer.height);
+camera.transform.translate([0, 0, 2]);
```
andy-sweet: I tried to think of a few ways to avoid these, but I think we can just allow them here then fix things when using an orthographic camera for all these 2D image presentation examples.

examples/image_series_from_omezarr5d_u8/index.html
```diff
@@ -19,6 +19,13 @@
         width: 100%;
         height: 100%;
       }
+      #canvas {
```
andy-sweet, line 22: We will also need the equivalent in the prototype, but that does not have to happen in this PR.

examples/tracks/main.ts
```diff
@@ -0,0 +1,164 @@
+import { vec3 } from "gl-matrix";
```
andy-sweet, line 1: This is about as long as I'd want an individual example to get. I think this one is fine, but we should probably slow down the rate at which we add examples, and think more about adding functionality to the prototype directly (especially for combinations of features we know we need there).

src/objects/cameras/orthographic_camera.ts
```diff
@@ -32,6 +37,17 @@ export class OrthographicCamera extends Camera {
     this.right_ = right;
     this.bottom_ = bottom;
     this.top_ = top;
+    const aspect = (right - left) / (top - bottom);
+    this.fov_ = (right - left) / aspect;
+  }
+
+  public updateFrame(aspect: number) {
```
andy-sweet: I don't love the long-term consequence of this function (as it is now) in the public API because it throws out the coordinates the a user provided on construction or set with setFrame. I think it's OK as a temporary workaround to get things working though, though so not blocking.

I'm still re-acclimating to OpenGL's coordinate systems, but I think we're absorbing the view, projection, and viewport transforms into a single transform in a way that couples the size of the window/canvas with the field of view.

examples/image2d_from_omezarr5d_u16/main.ts
andy-sweet: If we're making these changes in this PR, we should ideally make it in `Renderer.tsx` too.

andy-sweet (approved):
It's good to see everything so far come together in this PR!

I have a few issues with only allowing OpenGL clip-space coordinates for geometry coordinates, but I understand why you did that here, so I'm OK merging as long as we have issues to think about improvements there (i.e. there are GitHub issues to fix that).

I think this is one of those things in the visualization library I would prioritize getting right early, so I feel a bit dirty approving this, but also don't have clear suggestions that would improve this PR yet.

src/objects/cameras/orthographic_camera.ts
aganders3: I don't think it totally throws out the coordinates, because at least one direction of FOV is preserved, but I agree with this. At very least I need a TODO here indicating there is work to be done. I think it's probably better to do this in a specific PR that improves resizing for both cameras.

> I think we're absorbing the view, projection, and viewport transforms into a single transform in a way that couples the size of the window/canvas with the field of view

I'm not quite sure what you mean here - probably because I'm similarly acclimating to the coordinate systems.

examples/tracks/main.ts
aganders3: I agree with this. I'd be happy for this to to be removed as an example once integrated with the prototype. I could do that here or in a separate PR.

aganders3:
> I think this is one of those things in the visualization library I would prioritize getting right early, so I feel a bit dirty approving this, but also don't have clear suggestions that would improve this PR yet.

I totally agree with this. I think an option here is to pare this down further and keep as much of the undecided bits in the prototype application code.

examples/tracks/main.ts
```diff
@@ -0,0 +1,164 @@
...
+  const paddedZ = { start: zIndex.start - zPad, stop: zIndex.stop + zPad };
+
+  const center = vec3.fromValues(
+    (xIndex.start + xIndex.stop) / 2.0 - 0.5,
+    (yIndex.start + yIndex.stop) / 2.0 - 0.5,
+    (zIndex.start + zIndex.stop) / 2.0 - 0.5
+  );
+
+  return [center, paddedX, paddedY, paddedZ];
+}
```
shlomnissan: - I expect this logic to be hidden from the client and implemented by the layer?
- Another alternative is creating a `Track` type with a computed property that returns the bounding box.
- I understand this is just an example, but it took me a moment to grasp what you were trying to do here. I believe this function could be clearer. Below is a quick potential rewrite. I'm not adding this as a suggestion since I haven't tested it yet.

```ts
function getTrackBoundingBox(track: vec3[], padFactor: number = 0.5) {
  function getAxisBounds(index: number) : Interval {
    const values = track.map((point) => point[index]);
    return { start: Math.min(...values), stop: Math.max(...values) + 1 };
  }

  function applyPadding(index: Interval) : Interval {
    const padding = padFactor * (index.stop - index.start);
    return { start: index.start - padding, stop: index.stop + padding };
  }

  // This could probably be implemented on the Interval type
  function calculateCenter(interval: Interval, offset = 0.5) {
    return (interval.start + interval.stop) / 2.0 - offset;
  }

  const xInterval = applyPadding(getAxisBounds(0));
  const yInterval = applyPadding(getAxisBounds(1));
  const zInterval = applyPadding(getAxisBounds(2));

  const center = vec3.fromValues(
    calculateCenter(xInterval),
    calculateCenter(yInterval),
    calculateCenter(zInterval)
  );

  return [center, xInterval, yInterval, zInterval];
}
```

examples/image2d_from_omezarr5d_u16/main.ts
aganders3: Directly manipulating the camera transform is definitely not ergonomic. I am more used to setting a position, target, and up-vector; or using a function like `lookAt`.

shlomnissan (approved):
It is great seeing line overlays in the viewport 🎉

I'm not comfortable with some of these changes, especially around the FOV and working in NDC space. I still don't fully understand this decision. Unless I'm missing something, it seems like we don't have a clear plan on how to deal with "units." I'm approving because I don't have any useful suggestions at this point and I don't want to block it, but I would love to hear more about it in tomorrow's engineering sync, which will hopefully result in clear tasks for follow-up improvements.

examples/image2d_from_omezarr5d_u16/main.ts
shlomnissan: > I am more used to setting a position, target, and up-vector; or using a function like lookAt.

- The vectors you mentioned are typically used as the input to a `lookAt` function.
- I think it's reasonable to add a `lookAt` function to the transform object.

shlomnissan: I also think it's reasonable to move away from the idea that a camera is a special case of a renderable object. I was influenced by Three.js when I implemented it this way, but it may not be suitable (or at least not beneficial at the moment) for our needs.

examples/tracks/main.ts
shlomnissan: In my view, this could be seen as an indication that we can do a better job with how we define abstractions. I’m not sure that is clearly the case here, but as I mentioned in my comments, I would expect that some operations in this example will be implemented internally.

examples/tracks/main.ts
aganders3: `vec3` is correct - it just needs to be changed in the line layer/renderable object as well.

For now with a relatively short window (10 frames) I think we will draw the entire line at once. Eventually I think tracks would be rendered as a point + a line defining a short window of time. Time can also be encoded as line width and/or color.

examples/image_series_from_omezarr5d_u8/index.html
aganders3: You're probably right - I don't know CSS well at all - but this was the only way I figured out to make the canvas shrink before the slider was covered when shrinking the window.

examples/tracks/main.ts
andy-sweet: I have been wanting for something like napari's layer extent in the library (and the related camera reset operation) and this feels like a step towards, so I'm in favor.

examples/tracks/main.ts
aganders3: Good question! I think ideally we'd be rendering a larger slice and zooming, but obviously that's not what's going on here.

Thanks for bringing this up - I think it's clarifying some parts for me about responsibilities of data loading, etc.

examples/tracks/main.ts
aganders3: I want to be a little careful there because layer extent can also be pretty slow in some circumstances. This can be mitigated by caching but then you also have to deal with caching.

aganders3:
Thanks for the reviews - I'm also not happy with the transforms all into [-1, 1] space but it helped me to write the code to see it. Some of these comments have been clarifying and I will take another shot at it here. This also has two approvals so feel free to merge if it's blocking, but I don't mind it lingering open either.

I'm still struggling a bit with how the layers may coordinate with each other and the camera to define the FOV and transform their own data. I will give this some thought/effort in the morning and we can discuss in the sync tomorrow if needed. This feels a bit like what https://github.com/chanzuckerberg/imaging-active-learning/issues/33 is getting at as well, but that seems a bigger change than what I was intending to take on here.

examples/tracks/main.ts
shlomnissan: > napari's layer extent

Can you share more about this?

andy-sweet:
> I'm still struggling a bit with how the layers may coordinate with each other and the camera to define the FOV and transform their own data. I will give this some thought/effort in the morning and we can discuss in the sync tomorrow if needed.

One thing that I've been working through in #78 is that the `OrthographicCamera` allows us to fully specify the frame/field of view, but the renderer also resizes the viewport canvas to complete fill the size of the window component.

If we want to maintain the aspect ratio of an image (either specified by its geometry or a transform), then we either need to letter-box the GL canvas in the component, or we need to only allow specification of either the horizontal or vertical field of view of the camera (not both as we currently do).

In my experience, I'm used to canvases rendering as much content as we can, so I'd lean towards limiting specification of the orthographic camera's field of view (e.g. by specifying the center of the field of view and a zoom level). Though we might want to support disabling maintain the aspect ratio (seems very unusual in a scientific application though).

Can discuss more in the sync tomorrow, where I'll hopefully have a proposed solution.

examples/single_mesh_layer/main.ts
aganders3: > My experience with orthographic projection has always involved mapping the projection to screen space and treating units as pixels. I’m not entirely clear on why this wouldn’t work here, especially if we are dealing with images?

Maybe it's unfounded (or I'm not understanding) but I have two thoughts on using pixels as the "world" units.

First is mixing pixel sizes in the same renderer. I guess this is where layer transforms would come into play. So an image layer sets up geometry matching its "pixel" size (highest resolution if it's multi-resolution). Then it would use metadata about physical pixel spacing (and position and orientation) to set a layer transform to bring that to some common world space. This just moves our need to define to that common world space.

My second thought is about float precision when supporting very large images (40k+ pixels per-side). Maybe that's not an issue until we're into millions of pixels per-side.

@andy-sweet - I think you mentioned using some physical units (like millimeters) to define the world space. This is starting to make the most sense to me. It becomes a bit odd when working with more abstract data like visualizing a latent-space projection, or like a slot-camera type image, but I think that can be worked out by treating it as arbitrary units or having some time-to-space scale factor.

examples/tracks/main.ts
andy-sweet: > > napari's layer extent
> 
> Can you share more about this?

Effectively, the axis-aligned bounding box around all the data associated with a layer.

When you have a 2D presentation of one layer, resetting the view/region to that extent allows you to reset the frame of the camera so that you can see all the data in some form.

When you have more than one layer, you can take the union of their regions to define a frame to which to reset.

examples/tracks/main.ts
aganders3: This now loads the full slice but zooms into the ROI.

examples/single_mesh_layer/main.ts
aganders3: This is updated to be based on #85 now.

An issue remains which is that the tracks are specified with respect to the image, with (0, 0) in the top left. For now I've just manually adjusted the track coordinates, but I consider this a temporary workaround.

examples/tracks/main.ts
aganders3: > Will we draw the entire line at once, as shown in this example, or draw it progressively across frames?

We discussed changing the `ProjectedLineLayer` into a `TracksLayer`. I think this is a good idea, and would make it more natural to include a "time" component for each track, but I think this should come in a follow-up PR where the time is used for something.

examples/image2d_from_omezarr5d_u16/main.ts
aganders3: These are not relevant now with all 2D examples using the orthographic camera and the lines example relatively untouched.

examples/image_series_from_omezarr5d_u8/index.html
andy-sweet: This can be done in #74

examples/tracks/main.ts
```diff
@@ -0,0 +1,93 @@
...
+  LayerState,
+  OrthographicCamera,
+  OmeZarrImageSource,
+  ProjectedLineLayer,
+  WebGLRenderer,
+} from "@";
+
+// payload roughly equivalent to task 0 from
+// https://public.czbiohub.org/royerlab/ultrack/multi-color/mock_data.json
+// TODO: we don't want to transform the coordinates manually like this
```
andy-sweet, line 14: Add a link to the related issue:

```suggestion
// TODO: we don't want to transform the coordinates manually like this
// https://github.com/chanzuckerberg/imaging-active-learning/issues/88
```

examples/tracks/main.ts
```diff
@@ -0,0 +1,93 @@
...
+const { xMin: left, xMax: right, yMin: bottom, yMax: top } = lineLayer.extent;
+// TODO: instead of padding, we should add zoom to the camera
+const padding = 0.25 * Math.max(right - left, top - bottom);
+const camera = new OrthographicCamera(
+  left - padding,
+  right + padding,
+  bottom - padding,
+  top + padding,
+  0.0,
+  100
```
andy-sweet, line 71: These are the defaults now, so these could be removed.

```suggestion
  0.0,
  100
```

andy-sweet (approved):
Looks good! I only have minor comments and suggestions.

Issue #60, opened 2024-09-27
Support varying color along the path of a line


aganders3:
> Color along the path may be useful for indicating direction of motion, for example.

_Originally posted by @aganders3 in https://github.com/chanzuckerberg/imaging-active-learning/pull/46#discussion_r1777569286_

PR #72, opened 2024-10-09
ProjectedLine fixes

aganders3:
### Summary
In developing #71 I encountered a problem with the `ProjectedLine` renderable object. The width of a line from `[-1, 0, 0]` to `[1, 0, 0]` did not match that of a unit-square (plane).

The problem here was the `offset` in the shader having a homogeneous component of 1.0. This would cause the line to shrink in screen space. Removing this also made `currZ` unnecessary - this makes sense to me as we only want to offset in `x` and/or `y` in screen space.

The other fix here is that the attribute locations for previous/next positions were reversed. This didn't make much of a difference except perhaps at the very endpoints of a line.

Finally - I think my pattern of `const previous = path[i - 1] || path[i];` should instead be `const previous = path[i - 1] ?? path[i];` because I was intending this to check for nullish values, non falsey ones.

### Related Issue
N/A

### Tests & Checks
Tested manually by drawing horizontal and vertical lines on a unit-square plane with an orthographic camera

### DIFF
--- a/src/objects/geometry/projected_line_geometry.ts
+++ b/src/objects/geometry/projected_line_geometry.ts
@@ -45,12 +45,12 @@ export class ProjectedLineGoemetry extends Geometry {
         vertices[c++] = current[1];
         vertices[c++] = current[2];
 
-        const previous = path[i - 1] || path[i];
+        const previous = path[i - 1] ?? path[i];
         vertices[c++] = previous[0];
         vertices[c++] = previous[1];
         vertices[c++] = previous[2];
 
-        const next = path[i + 1] || path[i];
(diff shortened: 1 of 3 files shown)


andy-sweet (approved):
Code changes (i.e. indexing) look good to me.

I haven't taken long enough to understand the initial math/geometry related to these lines, so I can't vouch for the math changes here though.

aganders3:
> I haven't taken long enough to understand the initial math/geometry related to these lines, so I can't vouch for the math changes here though.

Thanks - I think that's good enough for now. This is probably worth revisiting when adding more features to the line (i.e. #60).

Issue #74, opened 2024-10-09
Visualize Task payload(s) in prototype app


aganders3:
#66 describes visualizing a single task in an example

This issue is about moving this example into the prototype app, and structuring the data as from the server. I think it's still okay here to hard-code the data in the app so this doesn't need the server component to run.

Bonus points here if it hard-codes several tasks and allows for cycling through them.

PR #79, opened 2024-10-09
Add Layer.transform property and additional AffineTransform features

### LINKED ISSUE #73: Add a 'transform' property on Layer (base)
Since #57 we have transforms on cameras and (other) renderable objects.

Without going full scenegraph it makes sense to have some transform chain starting with the Layer itself. This allows separate Layers to be transformed to match each other, without needing to iterate and modify their individual RenderableObjects.

The main open question here is how to get the transform into the rendering context, as currently `renderObject` does not have a reference to the layer. We discussed this a little in Slack:

@aganders3 
> You also noted that you don't think we should include a reference to the layer in `renderObject`, but I feel like it may be a reasonable change. This would kind of create three "providers" for shader parameters (uniforms, textures, transforms, etc.):
> * RenderableObject
> * Layer
> * Renderer (including active camera)
>
> This would also make it possible to use a renderable object in multiple layers (though I'm not sure that's a good thing to enable).

@andy-sweet 
> I had imagined that there the number of renderable object types might be smaller than the number of layer types, so implementing a new renderer (e.g. for WebGPU) may be an easier task if we keep layer-only properties out of the renderer's implementation.
> But maybe that doesn't matter if we have some semi-automatic way of pulling shader parameters? I.e. maybe the renderer implementation does not need to explicitly know about specific layer-only properties.

aganders3:
### Summary
This adds a `transform` (`AffineTransform`) property to `Layer`.

In the current implementation this transform is composed with the `RenderableObject.transform` and the camera transform in the `Renderer` base class, and passed into `renderObject` as `modelView`. I don't think this is necessarily the right pattern/implementation but hopefully we can sort that out here.

Other options:
* Pass `layer.transform` directly to `renderObject` (perhaps default to identity)
* Introduce an `activeLayer` property similar to `activeCamera` on the renderer, and access the layer transform that way
* Pass the whole layer into `renderObject` along with its object
* Make sure any layer parameters are copied into/onto or passed through its renderable objects
* Give renderable objects a reference to their "parent" layer

This also adds `scale` and `rotate` method to `AffineTransform`. One thought after using this a little is that we may also want `setScale`, `setRotation`, and `setTranslation` methods that don't apply the new transform incrementally.

Finally, this adds an `inverse` property to `AffineTransform`. I use this to apply the inverse of the camera transform to the object to get the `modelView` matrix. This makes more sense to me than using `matrix` here but I could be thinking about it wrong.

### Related Issue
Closes #73 

### Tests & Checks
Tested manually and created a new example to show how it works.

### DIFF
--- a/examples/index.html
+++ b/examples/index.html
@@ -34,6 +34,7 @@
       </li>
       <li><a href="/single_mesh_layer/index.html">Single Mesh Layer</a></li>
       <li><a href="/projected_lines/index.html">Projected Lines</a></li>
+      <li><a href="/layer_transform/index.html">Layer Transform</a></li>
     </ul>
   </body>
 </html>
--- a/examples/layer_transform/main.ts
+++ b/examples/layer_transform/main.ts
@@ -8,30 +8,52 @@ import {
 
 const layersManager = new LayerManager();
 
-const helixA = generateHelix({
-  radius: 1.0,
-  turns: 3.0,
-  pointsPerTurn: 100,
-  pitch: 1.5,
-  phase: 0.0,
-});
-
-const helixB = generateHelix({
-  radius: 1.0,
-  turns: 3.0,
-  pointsPerTurn: 100,
-  pitch: 1.5,
-  phase: 90.0,
-});
-
+const movingLayer = new ProjectedLineLayer([
+  {
+    path: [
+      [-1, 0, -5],
+      [0, -1, -5],
+      [1, 0, -5],
+      [0, 1, -5],
+      [-1, 0, -5],
+    ],
+    color: [0.7, 0.7, 0.0],
+    width: 0.1,
+  },
+]);
 const stationaryLayer = new ProjectedLineLayer([
-  { path: helixA, color: [1.0, 0.7, 0.0], width: 0.1 },
+  {
+    path: [
+      [-1, 0, -5],
+      [1, 0, -5],
+    ],
+    color: [0.7, 0.0, 0.0],
+    width: 0.1,
+  },
+  {
+    path: [
+      [0, -1, -5],
+      [0, 1, -5],
+    ],
+    color: [0.0, 0.7, 0.0],
+    width: 0.1,
+  },
+  {
+    path: [
+      [0, 0, -6],
+      [0, 0, -4],
+    ],
+    color: [0.0, 0.0, 0.7],
+    width: 0.1,
+  },
 ]);
-const transformedLayer = new ProjectedLineLayer([
-  { path: helixB, color: [0.0, 0.7, 0.0], width: 0.2 },
-]);
+stationaryLayer.transform.translate([0.0, 0.0, 5]);
+stationaryLayer.transform.rotate(
+  quat.fromEuler(quat.create(), 45.0, -45.0, 0.0)
+);
+stationaryLayer.transform.translate([0.0, 0.0, -5]);
 layersManager.add(stationaryLayer);
-layersManager.add(transformedLayer);
+layersManager.add(movingLayer);
 
 const renderer = new WebGLRenderer("#canvas");
 const camera = new PerspectiveCamera(60, renderer.width / renderer.height);
@@ -39,35 +61,13 @@ const camera = new PerspectiveCamera(60, renderer.width / renderer.height);
 animate();
 
 const rotate = () => {
-  transformedLayer.transform.rotate(
-    quat.fromEuler(quat.create(), 0.0, 0.0, 0.2)
-  );
+  movingLayer.transform.translate([0.0, 0.0, 5]);
+  movingLayer.transform.rotate(quat.fromEuler(quat.create(), 0.1, 0.2, 0.5));
+  movingLayer.transform.translate([0.0, 0.0, -5]);
 };
 setInterval(rotate, 5);
 
 function animate() {
   renderer.render(layersManager, camera);
   requestAnimationFrame(animate);
-}
-
-function generateHelix(params: {
-  radius: number;
-  turns: number;
-  pointsPerTurn: number;
-  pitch: number;
-  phase: number;
-}) {
-  const { radius, turns, pointsPerTurn, pitch, phase } = params;
-  const phaseRad = (phase * Math.PI) / 180;
-  const path: [number, number, number][] = [];
-  const totalPoints = turns * pointsPerTurn;
-
-  for (let i = 0; i < totalPoints; i++) {
-    const angle = (i / pointsPerTurn) * 2 * Math.PI;
-    const z = radius * Math.cos(angle + phaseRad) - 5.0;
-    const y = (i / pointsPerTurn) * pitch - (pitch * turns) / 2;
-    const x = radius * Math.sin(angle + phaseRad);
-    path.push([x, y, z]);
-  }
-  return path;
 }
--- a/examples/layer_transform/main.ts
+++ b/examples/layer_transform/main.ts
@@ -1,0 +1,73 @@
+import { quat } from "gl-matrix";
+import {
+  LayerManager,
+  PerspectiveCamera,
+  ProjectedLineLayer,
+  WebGLRenderer,
+} from "@";
+
+const layersManager = new LayerManager();
+
+const helixA = generateHelix({
+  radius: 1.0,
+  turns: 3.0,
+  pointsPerTurn: 100,
+  pitch: 1.5,
+  phase: 0.0,
+});
+
+const helixB = generateHelix({
+  radius: 1.0,
+  turns: 3.0,
+  pointsPerTurn: 100,
+  pitch: 1.5,
+  phase: 90.0,
+});
+
+const stationaryLayer = new ProjectedLineLayer([
+  { path: helixA, color: [1.0, 0.7, 0.0], width: 0.1 },
+]);
+const transformedLayer = new ProjectedLineLayer([
+  { path: helixB, color: [0.0, 0.7, 0.0], width: 0.2 },
+]);
+layersManager.add(stationaryLayer);
+layersManager.add(transformedLayer);
+
+const renderer = new WebGLRenderer("#canvas");
+const camera = new PerspectiveCamera(60, renderer.width / renderer.height);
+
+animate();
+
+const rotate = () => {
+  transformedLayer.transform.rotate(
+    quat.fromEuler(quat.create(), 0.0, 0.0, 0.2)
+  );
+};
+setInterval(rotate, 5);
+
+function animate() {
+  renderer.render(layersManager, camera);
+  requestAnimationFrame(animate);
+}
+
+function generateHelix(params: {
+  radius: number;
+  turns: number;
+  pointsPerTurn: number;
+  pitch: number;
+  phase: number;
+}) {
+  const { radius, turns, pointsPerTurn, pitch, phase } = params;
+  const phaseRad = (phase * Math.PI) / 180;
+  const path: [number, number, number][] = [];
+  const totalPoints = turns * pointsPerTurn;
+
+  for (let i = 0; i < totalPoints; i++) {
+    const angle = (i / pointsPerTurn) * 2 * Math.PI;
+    const z = radius * Math.cos(angle + phaseRad) - 5.0;
+    const y = (i / pointsPerTurn) * pitch - (pitch * turns) / 2;
+    const x = radius * Math.sin(angle + phaseRad);
+    path.push([x, y, z]);
+  }
+  return path;
+}
--- a/examples/layer_transform/main.ts
+++ b/examples/layer_transform/main.ts
@@ -1,5 +1,6 @@
 import { quat } from "gl-matrix";
 import {
+  Layer,
   LayerManager,
   PerspectiveCamera,
   ProjectedLineLayer,
@@ -47,11 +48,6 @@ const stationaryLayer = new ProjectedLineLayer([
     width: 0.1,
   },
 ]);
-stationaryLayer.transform.translate([0.0, 0.0, 5]);
-stationaryLayer.transform.rotate(
-  quat.fromEuler(quat.create(), 45.0, -45.0, 0.0)
-);
-stationaryLayer.transform.translate([0.0, 0.0, -5]);
 layersManager.add(stationaryLayer);
 layersManager.add(movingLayer);
 
@@ -60,12 +56,14 @@ const camera = new PerspectiveCamera(60, renderer.width / renderer.height);
 
 animate();
 
-const rotate = () => {
-  movingLayer.transform.translate([0.0, 0.0, 5]);
-  movingLayer.transform.rotate(quat.fromEuler(quat.create(), 0.1, 0.2, 0.5));
-  movingLayer.transform.translate([0.0, 0.0, -5]);
+const rotateAboutCenter = (layer: Layer, x: number, y: number, z: number) => {
+  layer.transform.translate([0.0, 0.0, 5]);
+  layer.transform.rotate(quat.fromEuler(quat.create(), x, y, z));
+  layer.transform.translate([0.0, 0.0, -5]);
 };
-setInterval(rotate, 5);
+
+rotateAboutCenter(stationaryLayer, 45.0, -45.0, 0.0);
+setInterval(rotateAboutCenter, 5, movingLayer, 0.1, 0.2, 0.5);
 
 function animate() {
   renderer.render(layersManager, camera);
--- a/src/core/renderer.ts
+++ b/src/core/renderer.ts
@@ -15,7 +15,7 @@ export abstract class Renderer {
   protected abstract resize(width: number, height: number): void;
   protected abstract renderObject(
     object: RenderableObject,
-    modelMatrix: mat4
+    modelView: mat4
   ): void;
   protected abstract clear(): void;
 
@@ -38,12 +38,18 @@ export abstract class Renderer {
       layer.update();
       if (layer.state === "ready") {
         layer.objects.forEach((obj) => {
-          const modelMatrix = mat4.multiply(
+          const modelView = mat4.multiply(
             mat4.create(),
             obj.transform.matrix,
             layer.transform.matrix
           );
-          this.renderObject(obj, modelMatrix);
+          mat4.multiply(
+            modelView,
+            modelView,
+            this.activeCamera.transform.inverse
+          );
+
+          this.renderObject(obj, modelView);
         });
       }
     });
--- a/src/core/renderer.ts
+++ b/src/core/renderer.ts
@@ -13,7 +13,10 @@ export abstract class Renderer {
   private activeCamera_: Camera | null = null;
 
   protected abstract resize(width: number, height: number): void;
-  protected abstract renderObject(object: RenderableObject, modelMatrix: mat4): void;
+  protected abstract renderObject(
+    object: RenderableObject,
+    modelMatrix: mat4
+  ): void;
   protected abstract clear(): void;
 
   constructor(selector: string) {
@@ -40,7 +43,7 @@ export abstract class Renderer {
             obj.transform.matrix,
             layer.transform.matrix
           );
-          this.renderObject(obj, modelMatrix)
+          this.renderObject(obj, modelMatrix);
         });
       }
     });
--- a/src/core/renderer.ts
+++ b/src/core/renderer.ts
@@ -1,3 +1,5 @@
+import { mat4 } from "gl-matrix";
+
 import { LayerManager } from "./layer_manager";
 import { Camera } from "objects/cameras/camera";
 import { RenderableObject } from "core/renderable_object";
@@ -11,7 +13,7 @@ export abstract class Renderer {
   private activeCamera_: Camera | null = null;
 
   protected abstract resize(width: number, height: number): void;
-  protected abstract renderObject(object: RenderableObject): void;
+  protected abstract renderObject(object: RenderableObject, modelMatrix: mat4): void;
   protected abstract clear(): void;
 
   constructor(selector: string) {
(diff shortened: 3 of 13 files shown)


src/core/transforms.ts
```diff
@@ -1,26 +1,50 @@
...
+      );
+      this.inverse_ = null;
       this.dirty_ = false;
     }
+    this.inverse_ = mat4.invert(mat4.create(), this.matrix_);
     return this.matrix_;
   }
+
+  public get inverse() {
+    if (this.inverse_ === null) {
```
andy-sweet: Does the inverse also need to be recomputed when the dirty flag has been set? I'm thinking about the case when the transform is dirty, but `inverse` is accessed before `matrix`.

src/layers/projected_line_layer.ts
```diff
@@ -11,7 +11,7 @@ interface LineParameters {
 export class ProjectedLineLayer extends Layer {
   constructor(lines: LineParameters[] = []) {
     super();
-    lines.forEach(this.addLine);
+    lines.forEach((line) => this.addLine(line));
```
andy-sweet, line 14: I assume this is related to binding `this`, but why did this need to be added in this PR and not prior?

Unrelated, there is a typo a few lines below from a previous PR `ProjectedLineGoemetry` should be `ProjectedLineGeometry`. I'd just fix that in this PR, but also feel free to create a new one instead.

src/core/transforms.ts
```diff
@@ -1,26 +1,50 @@
...
+      mat4.fromRotationTranslationScale(
+        this.matrix_,
+        this.rotation_,
+        this.translation_,
+        this.scale_
+      );
+      this.inverse_ = null;
       this.dirty_ = false;
     }
+    this.inverse_ = mat4.invert(mat4.create(), this.matrix_);
```
andy-sweet: Why compute the inverse every time the `matrix` property is accessed instead of doing it lazily in the property getter?

src/core/transforms.ts
```diff
@@ -1,26 +1,50 @@
-import { vec3, mat4 } from "gl-matrix";
+import { mat4, vec3, quat } from "gl-matrix";
 
 export class AffineTransform {
   private dirty_ = true;
   private matrix_ = mat4.create();
+  private inverse_: mat4 | null = null;
```
andy-sweet: Does this need to be nullable? Could you not just create it here as for `matrix_` and rely on the dirty mechanism to update it?

andy-sweet (commented):
Still need to think a bit on the transform composition and camera transform inverse here.

src/core/transforms.ts
aganders3: Yes, I realized this while falling asleep last night. Good catch!

src/core/transforms.ts
aganders3: I think it would need a separate `dirty_` flag unless we compute it automatically when re-computing the `matrix_`. Now that I write that down though, that's all the `null` variant is so maybe it's better to just use the same pattern.

andy-sweet:
> In the current implementation this transform is composed with the `RenderableObject.transform` in the `Renderer` base class, and passed into `renderObject`. I don't think this is necessarily the right pattern/implementation but hopefully we can sort that out here.
> 
> Other options:
> 
> * Pass `layer.transform` directly to `renderObject` (perhaps default to identity)
> * Introduce an `activeLayer` property similar to `activeCamera` on the renderer, and access the layer transform that way
> * Pass the whole layer into `renderObject` along with its object
> * Make sure any layer parameters are copied into/onto or passed through its renderable objects
> * Give renderable objects a reference to their "parent" layer

At first glance, these options all seem worse than the one in this PR unless they have secondary benefits.

src/core/transforms.ts
aganders3: Ugh sorry - I left this in after considering two options:
* calculate the inverse eagerly whenever the matrix is updated
* calculate the inverse lazily, and set to `null` when the matrix is updated

src/layers/projected_line_layer.ts
aganders3: I only caught it in this PR because previously I was using the empty constructor and `addLine` after so this was skipped (untested). I can take the opportunity to add some tests in this PR as well.

Good spot on the typo as well. I'm happy to fix here if you're okay with it. Oof I'm so bad with typos.

src/core/renderer.ts
```diff
@@ -32,7 +37,14 @@ export abstract class Renderer {
     layerManager.layers.forEach((layer) => {
       layer.update();
       if (layer.state === "ready") {
-        layer.objects.forEach((obj) => this.renderObject(obj));
+        layer.objects.forEach((obj) => {
+          const modelMatrix = mat4.multiply(
```
andy-sweet: I'd expect the result of this composition to be unchanged in most use cases, so I think we should think about caching the result ASAP (maybe even in this PR) given that `render` is called frequently.

src/renderers/webgl_renderer.ts
```diff
@@ -30,13 +30,13 @@ export class WebGLRenderer extends Renderer {
...
-  protected renderObject(object: RenderableObject) {
+  protected renderObject(object: RenderableObject, modelMatrix: mat4) {
     const program = this.getShaderProgram(this.getProgramName(object)).use();
 
     const modelView = mat4.multiply(
       mat4.create(),
-      object.transform.matrix,
-      this.activeCamera.transform.matrix
+      modelMatrix,
+      this.activeCamera.transform.inverse
```
andy-sweet: I'd also think about caching the result of this composition because in many use cases, I'd expect the camera to be static.

aganders3:
> At first glance, these options all seem worse than the one in this PR unless they have secondary benefits.

The secondary benefit of some would be if we eventually need _other_ data from the layer in the rendering context. - as an example in your contrast limits PR it could remove the need to pass through contrast limits to the renderable object.

One other option that I kind of like is to move the whole `modelView` construction out of `renderObject` and calculate (+cache as you mention - good idea) the `modelView` matrix in the base `Renderer` class. I kind of like this as well because as I mentioned on the initial PR for this it's not WebGL-specific code in any way.

aganders3:
Thanks for the feedback. I'm going to take this opportunity to add some tests and look at possibly caching the `modelView` computation, then will re-request review.

src/core/transforms.ts
aganders3: I switched to use the same `dirty_` pattern for both `matrix` and `inverse` properties.

src/core/renderer.ts
aganders3: I'm punting on caching for this PR, but opened #80 and will consider taking it on.

src/core/transforms.ts
```diff
@@ -1,26 +1,49 @@
...
+      );
+      this.dirtyMatrix_ = false;
+      this.dirtyInverse_ = true;
     }
     return this.matrix_;
   }
+
+  public get inverse() {
+    if (this.dirtyMatrix_ || this.dirtyInverse_) {
+      mat4.invert(this.inverse_, this.matrix);
```
andy-sweet: `dirtyInverse_` is never set to `false`, but I think it should be set in the body here

```suggestion
    if (this.dirtyMatrix_ || this.dirtyInverse_) {
      mat4.invert(this.inverse_, this.matrix);
      this.dirtyInverse_ = false;
```

examples/layer_transform/main.ts
```diff
@@ -0,0 +1,73 @@
...
+stationaryLayer.transform.translate([0.0, 0.0, -5]);
+layersManager.add(stationaryLayer);
+layersManager.add(movingLayer);
+
+const renderer = new WebGLRenderer("#canvas");
+const camera = new PerspectiveCamera(60, renderer.width / renderer.height);
+
+animate();
+
+const rotate = () => {
```
andy-sweet: I think this is intended to be a rotation around the center of the path, so `centeredRotate` may be a better local name. Also if you pass the Euler angles as arguments you could reuse this above.

At some point, we might actually want some general utilities like this in the library (passing a center too), but we can hold off on that for now.

examples/layer_transform/main.ts
```diff
@@ -0,0 +1,73 @@
+import { quat } from "gl-matrix";
```
andy-sweet, line 1: Definitely the most fun example so far and also helpful for understand what the shader is doing.

test/transform.test.ts
```diff
@@ -0,0 +1,134 @@
...
+import { expect, test } from "vitest";
+
+import { AffineTransform } from "@/core/transforms";
+
+// NOTES:
+// * mat4.fromValues is column-major
+// * use mat4.equals instead of expect(t.matrix).toEqual because it's better for comparing floats,
+// even though it provides worse output on failure
+
+test("rotate", () => {
```
andy-sweet, line 16: I was tempted to suggest avoiding too many tests, while we still in "prototype mode" (whatever that is...). But I think we can be confident about the permanence of defining some library specific transforms, so I think these tests are valuable and are unlikely to change.

I've been holding off on writing tests because I feel like the things I've been defining to get closer to the prototype are more transient. Let me know if I should change that.

src/core/transforms.ts
aganders3: 🤦 
Of course it works if the inverse is always dirty. So much for my tests.

test/transform.test.ts
aganders3: Ehh, I'm similarly ambivalent about it right now and as discussed I'm not sure these tests are super valuable. I think it's worth adding some just to get coverage but I wouldn't put too much effort into it.

test/transform.test.ts
```diff
@@ -0,0 +1,134 @@
...
+  const t = new AffineTransform();
+  const q = quat.rotateZ(quat.create(), quat.create(), Math.PI / 2);
+  t.rotate(q);
+  expect(
+    mat4.equals(
+      t.matrix,
+      mat4.fromValues(0, 1, 0, 0, -1, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1)
+    )
+  ).toBe(true);
+  t.rotate(quat.invert(quat.create(), q));
```
andy-sweet, line 28: Nit

I find the main action under test gets a little lost with the formatting here. It's also hard (for me) to read the matrix. Is it possible/easy to refactor a utility assert/expect function to handle the matrix equality here (e.g. where we could pass a nested array for better readability).

andy-sweet (approved):
Looks good to me now!

test/transform.test.ts
aganders3: Yes and Prettier is doing no favors here. I'll see if I can make something quickly that looks better.

aganders3:
Thanks! I'm going to merge to make progress on #71 and hopefully unblock #74. I'm happy to update later with additional feedback.

examples/index.html
```diff
@@ -39,6 +39,7 @@ <h1>Examples</h1>
       </li>
       <li><a href="/single_mesh_layer/index.html">Single Mesh Layer</a></li>
       <li><a href="/projected_lines/index.html">Projected Lines</a></li>
+      <li><a href="/layer_transform/index.html">Layer Transform</a></li>
```
shlomnissan, line 42: I have reservations about adding this to the examples directory. Applying a transform to a layer doesn't strike me as a distinct "feature." Moreover, viewing this example without additional context beyond its name doesn't clearly convey what's being demonstrated.

examples/layer_transform/main.ts
```diff
@@ -0,0 +1,71 @@
...
+
+const renderer = new WebGLRenderer("#canvas");
+const camera = new PerspectiveCamera(60, renderer.width / renderer.height);
+
+animate();
+
+const rotateAboutCenter = (layer: Layer, x: number, y: number, z: number) => {
+  layer.transform.translate([0.0, 0.0, 5]);
+  layer.transform.rotate(quat.fromEuler(quat.create(), x, y, z));
+  layer.transform.translate([0.0, 0.0, -5]);
```
shlomnissan, line 62: This seems to contradict the architecture as I understand it. The client shouldn't directly transform a layer — all layer transformations should be handled internally. Is there a specific use case for making the transform public? If not, I suggest making this transform private and, as previously mentioned, removing this example entirely.

src/core/renderer.ts
```diff
@@ -11,7 +13,10 @@ export abstract class Renderer {
   private activeCamera_: Camera | null = null;
 
   protected abstract resize(width: number, height: number): void;
-  protected abstract renderObject(object: RenderableObject): void;
+  protected abstract renderObject(
+    object: RenderableObject,
+    modelView: mat4
```
shlomnissan, line 18: I think this approach is reasonable. It makes sense to me to add another transform instance on the renderable object that represents the "final" transformation (similar to three.js's local transform and world transform) and update it (if needed) at this point. However, I haven't spent much time considering alternatives, so we can address this later when we have more context. For example, we may need to pass a reference to the layer into the "render object" anyway.

src/core/renderer.ts
```diff
@@ -32,7 +37,22 @@ export abstract class Renderer {
     layerManager.layers.forEach((layer) => {
       layer.update();
       if (layer.state === "ready") {
-        layer.objects.forEach((obj) => this.renderObject(obj));
+        layer.objects.forEach((obj) => {
+          // TODO: cache the modelView matrix
+          // https://github.com/chanzuckerberg/imaging-active-learning/issues/80
```
shlomnissan, line 42: This is where storing another transform on the renderable object (previous comment) makes sense to me. You can update the renderable object “world position” only if the layer transform is dirty.

src/core/transforms.ts
```diff
@@ -1,26 +1,58 @@
-import { vec3, mat4 } from "gl-matrix";
+import { mat4, vec3, quat } from "gl-matrix";
 
 export class AffineTransform {
-  private dirty_ = true;
+  private dirtyMatrix_ = true;
+  private dirtyInverse_ = true;
```
shlomnissan, line 5: I believe having an inverse function for the transform is reasonable, but I'm unsure about implementing it this way or tracking changes separately. Since this is only applicable to cameras, I suggest moving the logic there. Specifically, add an accessor method for the camera's transform that returns its inverse. If you want to implement caching, you can handle it there as well. Tracking two dirty flags seems unnecessarily complicated to me.

src/core/transforms.ts
```diff
@@ -1,26 +1,58 @@
...
-    this.dirty_ = true;
+    this.dirtyMatrix_ = true;
   }
 
-  // TODO: add scale
-
-  // TODO: add rotate
+  public scale(vec: vec3) {
+    vec3.multiply(this.scale_, this.scale_, vec);
+    vec3.multiply(this.translation_, this.translation_, vec);
```
shlomnissan, line 25: Why multiply the translation vector by the scale vector? I think it's reasonable to assume a specific order of operations.

test/projected_line_layer.test.ts
```diff
@@ -0,0 +1,38 @@
+import { expect, test } from "vitest";
+
+import { ProjectedLineLayer } from "@";
+
+test("lines in constructor", () => {
```
shlomnissan, line 5: While I don't find these tests particularly valuable, I'm not opposed to keeping them. However, I suggest reconsidering their names to better describe what you're actually testing. Perhaps:

- construct layer with multiple lines
- add lines to layer

Does `AddLine` support multiple lines? If so, maybe change it to "AddLines"?

vite.config.js
```diff
@@ -19,8 +19,13 @@ function modeToRoot(mode) {
     return 'examples';
   } else if (mode === 'prototype') {
     return 'ultrack-prototype/frontend';
+  } else if (mode === 'production') {
+    // no root for production mode (library build)
+  } else if (mode == 'test') {
+    // no root for test mode
```
shlomnissan, line 25: I suggest we eliminate unused code branches or, at the very least, add logging statements to track their execution.

shlomnissan (commented):
I have a few concerns about this PR:

- The addition of an example for the layer transform seems unnecessary.
- Unless there's a compelling reason to make the layer's transform public, I suggest removing it. I've never envisioned clients interacting with layers so directly.
- Tracking `dirtyInverse_` appears confusing and potentially unnecessary.

src/core/transforms.ts
shlomnissan: https://github.com/mrdoob/three.js/blob/dev/src/cameras/Camera.js#L49

I think this is a good example of "inversion" being the responsibility of the camera.

src/core/transforms.ts
aganders3: This to me would be the difference between `scale` and (something like) `setScale`.

Calling `scale([1, 2, 3])` implies to me multiplying the existing transform by a scaling matrix, which would include scaling the translation component.

I can think of uses for both. I'm not sure how I feel about assuming order of operations otherwise unless we can make it explicit.

examples/index.html
aganders3: I'm happy to remove the example, it was just something to develop/debug against. This feels true of most of our examples at this point.

examples/layer_transform/main.ts
aganders3: I think this is probably true about the API in a longer term, but the transform needs to be public until we solve coordinates/units in a more general way that can be applied across layers. I think there needs to be some public API on layers to allow two layers with different data to coordinate into the same space. Directly exposing the transform may not be it.

src/core/renderer.ts
aganders3: This would probably make it easier - you're right. Another option is to introduce the concept of a "transform chain".

There's still no mechanism for determining if the layer transform is dirty because we need to know if it's dirty since its last use in each context. This can be solved by another callback but I'm not eager to add more callbacks/events to the layer class, and I'm even less sure about adding that to AffineTransform.

I'm not sure I like the idea of storing this on the renderable object though. It feels like leaking information about the relationship that the renderable object should not need to understand.

src/core/transforms.ts
aganders3: It feels pretty natural to me that "matrix (transform) inversion" is the responsibility of a transform, not a camera. Why do you think the inverse would only be applicable to cameras? I'd also rather not split the responsibility of caching calculations between multiple classes.

test/projected_line_layer.test.ts
aganders3: The tests here are simple, but would have caught an issue identified in this PR just by adding coverage. Similarly I think having coverage can help (in the long run) identify breaking changes in the API. We can't rely on examples for coverage unless we're actively checking them all.

I agree the names are not great and can update them.

> Does AddLine support multiple lines? If so, maybe change it to "AddLines"?

`addLine` does not support multiple lines. My plan is to remove this function anyway. You brought up before the idea that layers should be provided with a data source, rather than having a public API to modify them. I think I'm generally on board with that but might propose layers be constructed _either_ with a data source, _or_ with all the data they need at construction. In this scenario `addLine` would become private.

examples/index.html
shlomnissan: > This feels true of most of our examples at this point.

I agree that there isn't a clear definition for the scope of an example, but I think there's a slight difference -- the other examples demonstrate unique visualizations or data types, whereas this one deals with an internal property that should not even be public (the layer's transform).

examples/layer_transform/main.ts
shlomnissan: > the transform needs to be public until we solve coordinates/units in a more general way

If that's the case, this API definitely should not be part of an example. We should also add a TODO comment to ensure we don't forget to make it private.

src/core/renderer.ts
shlomnissan: > There's still no mechanism for determining if the layer transform is dirty

I'm not sure I understand what you mean by "no mechanism." The transform object already has a "dirty" flag, doesn't it?

> I'm not sure I like the idea of storing this on the renderable object though.

I think I disagree. Storing the final position in the renderable object seems logical to me. This approach aligns with scene graph implementations I'm familiar with, such as Three.js and Godot. However, I don't have a strong preference on this matter.

src/core/transforms.ts
shlomnissan: I agree that it's the transform’s responsibility to implement `invert()`. However, I disagree that it needs to track a dirty state for it. If an object needs the inverse of its transform, it can simply call the inverse function to get a copy before returning it.

src/core/transforms.ts
shlomnissan: > I can think of uses for both

I initially assumed that translation wouldn't be affected by scale, but I may lack the experience to fully appreciate the value of supporting both approaches, particularly the current implementation.

> I'm not sure how I feel about assuming order of operations

"Assuming" may have been a bad word choice. It's more accurate to say that thinking about geometric transformations in this way is a logical approach IMO.

shlomnissan:
I feel strongly about removing `dirtyInverse_` from the transform object and adding a `TODO` to make the layer's transform private once we figure out units. I’m happy to have this conversation over Zoom if you disagree about the former as I would like to better understand your view.

src/core/renderer.ts
aganders3: > I'm not sure I understand what you mean by "no mechanism." The transform object already has a "dirty" flag, doesn't it?

This is an implementation detail of the transform, though. Say we compose `layer.transform` with `object.transform` and store it on the renderable object in `object.finalTransform` - how does the object know when one of the input transforms has changed and the composite needs to be recomputed? Even if `dirty` was a public component of the transform it would be set false at first access (possibly by some other entity).

There are ways to do this, but it doesn't exist now.

> I think I disagree. Storing the final position in the renderable object seems logical to me.

I also don't feel too strongly about this - it probably comes down to the implementation for me. If we don't need to decide it now I'm happy to revisit as-needed.

aganders3:
That's good enough for me - I can open a PR with these changes and continue conversation there if needed.

shlomnissan:
> That's good enough for me

Just to be clear - I might feel strongly about something, but that doesn't mean I expect it to change. I just want to keep the conversation going. I trust you to make the final call on the code you're writing if you think it's the right way to go.

aganders3:
Sorry I'm not trying to be passive aggressive.

I feel more strongly about some things than others, and I have a few open questions. Either way it's easier to have a conversation where we can look at possible changes in the context of the code. I'm happy to open another PR to continue conversation there.

shlomnissan:
I didn't think you were passive-aggressive at all 😅 I was mostly worried that my responses came across a little aggressive.

PR #81, opened 2024-10-11
Add cubic spline interpolation

aganders3:
### Summary
This adds some pretty basic spline interpolation. Given the simplicity of the joins in `ProjectedLine` this offers a significant improvement in appearance.

The improvement is especially apparent for lines with few points and tight curves, where the line width can deviate significantly from what is specified due to the lack of miter joins (red line).

Simply adding more points along the line (linear interpolation) helps a lot with this, but the interpolation here offers some additional smoothing of the path.

Screenshot 2024-10-11 at 10 20 10 AM

### Related Issue
N/A

### Tests & Checks
I tested manually and made a new example to demonstrate usage.

### DIFF
--- a/src/core/splines.ts
+++ b/src/core/splines.ts
@@ -1,0 +1,84 @@
+import { vec3 } from "gl-matrix";
+
+type BezierParams = {
+  path: vec3[];
+  n: number;
+  x?: number[];
+  fac?: number;
+};
+
+export function cubicHermiteInterpolation(params: BezierParams): vec3[] {
+  const { path, n } = params;
+  let { x, fac } = params;
+  if (!x) {
+    x = new Array(path.length).fill(0).map((_, i) => i);
+  }
+  if (!fac) {
+    fac = 1.0 / 3.0;
+  }
+  const tangents = pathTangents(path, x);
+
+  const out = Array((path.length - 1) * n);
+
+  // a cubic bezier curve is defined by 4 control points: a, b, c, d
+  // for interpolation of a segment:
+  // * a and d are the endpoints of the curve segment
+  // * b and c are control points that define curvature
+  // here we use 1/3 of the tangent at each endpoint as the control points
+  // this is equivalent to a cubic Hermite spline
+  for (let i = 0; i < path.length - 1; i++) {
+    const a = path[i];
+    const d = path[i + 1];
+    const b = vec3.clone(tangents[i]);
+    vec3.scaleAndAdd(b, a, b, fac);
+    const c = vec3.clone(tangents[i + 1]);
+    vec3.scaleAndAdd(c, d, c, -fac);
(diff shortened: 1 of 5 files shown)


src/core/splines.ts
```diff
@@ -0,0 +1,85 @@
...
+    const prevX = x[i - 1] ?? x[i];
+    const currX = x[i];
+    const nextX = x[i + 1] ?? x[i];
+
+    const m0 = vec3.create();
+    const m1 = vec3.create();
+    tangents[i] = vec3.create();
+
+    if (i !== 0) {
+      vec3.sub(m0, curr, prev);
```
aganders3, line 65: I think this can be optimized to reuse the previous `m1` calculation as `m0`.

src/core/splines.ts
```diff
@@ -0,0 +1,85 @@
...
+  path: vec3[];
+  pointsPerSegment: number;
+  x?: number[];
+  tangentFactor?: number;
+};
+
+export function cubicBezierInterpolation({
+  path,
+  pointsPerSegment,
+  x,
```
andy-sweet, line 13: I understood `x` as something like a time-point of a point following the overall path, so that if two points are close in time, then their differentiation (and resulting integration) will be handled appropriately when forming the curve. Is that correct?

Regardless, it would be good to rename this variable to better indicate that. Unless this is standard notation for Bezier curves or similar - in which I'd appreciate a link out to a reference.

Also, we're not currently using this functionality, so I also wonder if we should include it as it complicates the implementation too.

src/core/splines.ts
```diff
@@ -0,0 +1,85 @@
...
+  tangentFactor = 1.0 / 3.0,
+}: BezierParams): vec3[] {
+  if (!x) {
+    x = new Array(path.length).fill(0).map((_, i) => i);
```
andy-sweet, line 17: Nit: I think they are equivalent, but it would be good to stick to either `new Array(` and `Array` - at least locally (you use `Array` without new just below).

andy-sweet (approved):
I assume this could also be done in the shader, but I don't know the trade-offs and I'm fine doing this on the CPU and passing responsibility to the client for now.

src/core/splines.ts
aganders3: Yeah, it's the "independent variable" in case your path is not evenly spaced. I think you're right it could just be removed for now.

aganders3:
I'm not totally sure this can be done in the shader, but would need to think about it some more. It might require a different line-drawing method.

Another option would be to pull this into the application (prototype) code as a utility function. I think interpolations are generally useful in graphics applications, but if it's handled by the client on CPU maybe this is too much scope creep at the moment for the library.

aganders3:
> Another option would be to pull this into the application (prototype) code as a utility function.

Closing in favor of this approach for now.

PR #82, opened 2024-10-11
Add immutable textures

### LINKED ISSUE #69: Allow sub-regions of textures to be updated
Currently, any time texture data needs to be updated, we need to create an entirely new texture and maybe even an entirely new renderable object.

Since presenting 2D image views of n-dimensional image arrays will be a common pattern for the visualization library, we should allow sub-regions of texture data to be updated instead.

The goal of this task is to provide some mechanism, so that layer's can do that when loading new data that needs to be pushed into an existing texture.

shlomnissan:
### Summary

This PR introduces immutable textures, refactors the texture class to support
internal properties that can be configured by layers, and moves texture
property mapping to the GLTextures module.

### Related Issue
This PR doesn't close #69 but it's related to it. The following PR will address it directly.

### Tests & Checks

Manual verification

https://github.com/user-attachments/assets/89626ebd-f8f3-4ab0-9f13-8e5f7e58e86c

### DIFF
--- a/src/data/image_chunk.ts
+++ b/src/data/image_chunk.ts
@@ -1,4 +1,5 @@
 import { Region } from "data/region";
+import { TextureUnpackRowAlignment } from "objects/textures/texture";
 
 // One 2D chunk of n-dimensional image data.
 // TODO: include the region of this chunk.
@@ -10,7 +11,7 @@ export type ImageChunk = {
     height: number;
   };
   rowStride: number;
-  rowAlignmentBytes: number;
+  rowAlignmentBytes: TextureUnpackRowAlignment;
 };
 
 export type ImageChunkSource = {
--- a/src/layers/image_layer.ts
+++ b/src/layers/image_layer.ts
@@ -4,7 +4,6 @@ import { PlaneGeometry } from "objects/geometry/plane_geometry";
 import { Region } from "data/region";
 import { ImageChunkSource } from "data/image_chunk";
 import { DataTexture2D } from "objects/textures/data_texture_2d";
-import { TextureUnpackRowAlignment } from "objects/textures/texture";
 
 // Loads data from an image source into renderable objects.
 export class ImageLayer extends Layer {
@@ -53,12 +52,11 @@ export class ImageLayer extends Layer {
 
     texture.dataFormat = "red_integer";
     if (chunk.data instanceof Uint16Array) {
-      texture.datatType = "unsigned_short";
+      texture.dataType = "unsigned_short";
     }
 
     texture.unpackRowLength = chunk.rowStride;
-    texture.unpackAlignment =
-      chunk.rowAlignmentBytes as TextureUnpackRowAlignment;
+    texture.unpackAlignment = chunk.rowAlignmentBytes;
 
     this.addObject(new Mesh(this.plane_, texture));
     this.setState("ready");
--- a/src/layers/image_layer.ts
+++ b/src/layers/image_layer.ts
@@ -4,6 +4,7 @@ import { PlaneGeometry } from "objects/geometry/plane_geometry";
 import { Region } from "data/region";
 import { ImageChunkSource } from "data/image_chunk";
 import { DataTexture2D } from "objects/textures/data_texture_2d";
+import { TextureUnpackRowAlignment } from "objects/textures/texture";
 
 // Loads data from an image source into renderable objects.
 export class ImageLayer extends Layer {
@@ -47,10 +48,18 @@ export class ImageLayer extends Layer {
     const texture = new DataTexture2D(
       chunk.data,
       chunk.shape.width,
-      chunk.shape.height,
-      chunk.rowStride,
-      chunk.rowAlignmentBytes
+      chunk.shape.height
     );
+
+    texture.dataFormat = "red_integer";
+    if (chunk.data instanceof Uint16Array) {
+      texture.datatType = "unsigned_short";
+    }
+
+    texture.unpackRowLength = chunk.rowStride;
+    texture.unpackAlignment =
+      chunk.rowAlignmentBytes as TextureUnpackRowAlignment;
+
     this.addObject(new Mesh(this.plane_, texture));
     this.setState("ready");
   }
--- a/src/objects/textures/texture.ts
+++ b/src/objects/textures/texture.ts
@@ -10,9 +10,15 @@ export type TextureDataType = "unsigned_byte" | "unsigned_short";
 
 export type TextureUnpackRowAlignment = 1 | 2 | 4 | 8;
 
+export function isTextureUnpackRowAlignment(
+  value: number
+): value is TextureUnpackRowAlignment {
+  return value === 1 || value === 2 || value === 4 || value === 8;
+}
+
 export abstract class Texture extends Node {
   public dataFormat: TextureDataFormat = "rgba";
-  public datatType: TextureDataType = "unsigned_byte";
+  public dataType: TextureDataType = "unsigned_byte";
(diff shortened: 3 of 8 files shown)


src/objects/textures/texture.ts
```diff
@@ -1,6 +1,31 @@
...
+
+export type TextureDataFormat = "rgb" | "rgba" | "red_integer";
+
+export type TextureDataType = "unsigned_byte" | "unsigned_short";
+
+export type TextureUnpackRowAlignment = 1 | 2 | 4 | 8;
+
+export abstract class Texture extends Node {
+  public dataFormat: TextureDataFormat = "rgba";
+  public datatType: TextureDataType = "unsigned_byte";
```
andy-sweet: You'll need to make the other updates so probably don't commit this.

```suggestion
  public dataType: TextureDataType = "unsigned_byte";
```

src/renderers/webgl_textures.ts
```diff
@@ -35,95 +39,158 @@ export class WebGLTextures {
...
   private configureTexture(texture: Texture) {
+    this.configureTextureParamaters(texture);
+    this.allocateTextureStorage(texture);
+
+    if (texture.data !== null) {
+      this.uploadTextureSubData(texture, 0, { x: 0, y: 0 });
+    }
+  }
+
+  private configureTextureParamaters(texture: Texture) {
```
andy-sweet:
```suggestion
  private configureTextureParameters(texture: Texture) {
```

src/layers/image_layer.ts
```diff
@@ -47,10 +48,18 @@ export class ImageLayer extends Layer {
...
     );
+
+    texture.dataFormat = "red_integer";
+    if (chunk.data instanceof Uint16Array) {
+      texture.datatType = "unsigned_short";
+    }
+
+    texture.unpackRowLength = chunk.rowStride;
+    texture.unpackAlignment =
+      chunk.rowAlignmentBytes as TextureUnpackRowAlignment;
```
andy-sweet: Optional

Given that `chunk.rowAlignmentBytes` currently comes from `TypedArray.BYTES_PER_ELEMENT`, I know this static cast is valid now. But it might be worth adding a run-time check for that or maybe better a refactor to share typing information across loader return types and texture data types.

src/renderers/webgl_textures.ts
```diff
@@ -35,95 +39,158 @@ export class WebGLTextures {
...
+      gl.TEXTURE_WRAP_R,
+      this.getGLWrapMode(texture.wrapR)
+    );
+  }
+
+  private allocateTextureStorage(texture: Texture) {
+    this.gl_.texStorage2D(
+      this.getGLTextureType(texture),
+      texture.mipmapLevels,
+      this.getGLInternalFormat(texture.dataFormat, texture.datatType) as GLenum,
```
andy-sweet: Nit: I don't think the `as GLenum` is needed here since the function returns a `GLenum` or throws.

Sorry for the lack of an explicit suggested change, but GitHub seems to have that taken that away from me for now.

src/renderers/webgl_textures.ts
```diff
@@ -35,95 +39,158 @@ export class WebGLTextures {
...
+    this.gl_.texSubImage2D(
+      this.getGLTextureType(texture),
+      mipmapLevel,
+      offset.x,
+      offset.y,
+      texture.width,
+      texture.height,
+      this.getGLFormat(texture.dataFormat),
+      this.getGLType(texture.datatType),
+      texture.data as HTMLImageElement
```
andy-sweet: I'm confused by the need for this static cast when I look at the overloads, but I can confirm I need it too.

I thought the error would be that if you pass an `HTMLImageElement`, then you cannot independently specify the width and height, but that does not seem to be the case?

src/objects/textures/texture.ts
```diff
@@ -1,6 +1,31 @@
...
+  public maxFilter: TextureFilter = "nearest";
+  public minFilter: TextureFilter = "nearest";
```
andy-sweet, line 23: Are these default values intended to be sane defaults for their corresponding parameters in WebGL2 (as our target graphics backend right now)? If so, I think both of these are "linear" by default.

If these are just intended to be our own sane defaults (where correspondences with WebGL2 are coincidental), I think "nearest" is fine as something that should always work.

src/renderers/webgl_textures.ts
```diff
@@ -35,95 +39,158 @@ export class WebGLTextures {
...
     }
-    if (texture.data instanceof Uint16Array) {
-      return {
-        internalFormat: gl.R16UI,
-        format: gl.RED_INTEGER,
-        type: gl.UNSIGNED_SHORT,
-      };
+  }
+
+  private getGLInternalFormat(
```
andy-sweet, line 183: I like how explicit and readable all these conversion utilities are now.

src/renderers/webgl_textures.ts
andy-sweet: This would ideally be fixed before merging, since the `HTMLImageElement` likely won't be too common in scientific applications.

andy-sweet (changes requested):
Looks great except for a few typos and the `texture.data as HTMLImageElement` static cast, which are the only two reasons I'm requesting changes.

I see how this is an improvement over `main` and that it will make supporting other pixel data types (e.g. float32) easier in the future.

src/objects/textures/texture.ts
shlomnissan: 😅 haha, this is really unlike me, but it's the second time typos have come up in my PRs in the last week

src/renderers/webgl_textures.ts
shlomnissan: This function has multiple overloads. The cast ensures the correct one is called, which handles various types including `TexImageSource`—an aggregate type encompassing `HTMLImageElement` and `ArrayBufferView`. I cast to `ArrayBufferView` and added an explanatory comment. Once we consolidate `Texture2D` and `DataTexture2D` (which I believe we should, separating the image loader from the texture), we could narrow the type.

src/objects/textures/texture.ts
shlomnissan: The default minification filter is `GL_NEAREST_MIPMAP_LINEAR`, which assumes mipmaps are always enabled. However, since we're not currently supporting mipmaps, the next best default would be `GL_NEAREST`. The magnification filter is `GL_LINEAR` by default (mipmaps have no impact on magnification) but since it doesn't seem like a suitable default for us, I decided to go with "nearest". To answer your question, I felt like this would be our own sane defaults but when I wasn’t sure, I used OpenGL’s defaults.

src/layers/image_layer.ts
shlomnissan: I added runtime checks. Let me know what you think.

src/data/image_chunk.ts
```diff
@@ -1,4 +1,5 @@
 import { Region } from "data/region";
+import { TextureUnpackRowAlignment } from "objects/textures/texture";
```
andy-sweet, line 2: I'm imagining that in the long term, we should avoid direct dependencies between data and textures. That's partly because the data loaders can be completely ignorant of textures and rendering. But also because the data loaders may have utility beyond rendering (e.g. for data processing in the browser).

No change needed for now, but something to think about.

andy-sweet (approved):

src/data/image_chunk.ts
shlomnissan: I agree. For the time being, at the very least, we have the necessary safeguards.

Issue #83, opened 2024-10-11
Provide an API for putting different objects and layers into a shared coordinate space


aganders3:
As of #79 (and #23) the `transform` property on layers (and renderable objects) is public.

Directly manipulating these transforms is probably not the public API we want.

There should be some ergonomic way to put layers and objects into a shared coordinate space. This may need to take into account real-world metadata such as sample orientation, units, anisotropy, etc.

andy-sweet:
I can think of least one example where I've needed to mutate the transform associated with a layer - which is manual registration/alignment of two or more layers.

I don't think that's a high priority for us, but I'd like to imagine that the API for the visualization library could support such a use case.

PR #84, opened 2024-10-11
Remove layer transform

aganders3:
### Summary
I merged #79 without @shlomnissan's review, and he provided some good feedback later. This PR opens that part for additional discussion.

Specifically, this:
* Removes the layer transform example
* Removes caching of `AffineMatrix.inverse`
* Adds a TODO (and issue #83) regarding making `Layer.transform` private

@shlomnissan - I don't think this is 100% what you were asking for. Specifically I'm a little confused about adding a transform inverse property to the camera object. Maybe I misunderstood but I think this was your suggestion.

I also have a couple questions about making `Layer.transform` private. I am curious if you feel the same way about the transform on `RenderableObject`? In any case for both, the transform needs to be accessible by the renderer, so it needs to be public in some way. Perhaps these properties should instead return only the computed `matrix`? This would prevent clients from mutating the transform directly.

### Related Issue
N/A

### Tests & Checks
N/A


src/core/layer.ts
```diff
@@ -24,6 +24,8 @@ export abstract class Layer {
     return this.state_;
   }
 
+  // TODO: make the layer transform private when we settle on an API for units/coordinates
+  // https://github.com/chanzuckerberg/imaging-active-learning/issues/83
   public get transform() {
```
andy-sweet: After this PR, I think we're in the strange state where `Layer.transform_` is always the identity and the only usage is in `Renderer` where we multiply with that identity matrix on every call to render.

As we're removing interesting usage of it in this PR, I think we just go the whole way and remove it and the corresponding usage in `Renderer`. That way we don't have to worry if this is public or private.

The previous PR still implemented support for `scale` and `rotate` and added tests to exercise usage, so it still has value.

andy-sweet (changes requested):
After this PR, `Layer.transform` feels vestigial (because we're not meaningfully using it or exercising it anywhere), so I think it should just be removed, which is why I'm requesting changes.

andy-sweet (approved):

shlomnissan (approved):
LGTM!

Issue #35, opened 2024-09-16
Use OME-Zarr metadata to define corner coordinates of plane


andy-sweet:
In #28, we added support for loading uint16 image data from OME-Zarr sources. A limitation of that implementation is that it uses an arbitrary plane geometry that works well with the default camera parameter values.

Instead, after opening the OME-Zarr group and retrieving its metadata, we can define the corner coordinates of its `PlaneGeometry` based on the physical extents of the image. This may need to be done in combination with modifications to the camera to easily allow us to define a 2D field of view to present.

Issue #52, opened 2024-09-23
Add a background color property to the Renderer class


andy-sweet:
This should be used when clearing the renderer (e.g. before a new render pass).

PR #85, opened 2024-10-15
Match orthographic camera coordinates with geometry coordinates 

### LINKED ISSUE #78: Use the orthographic camera in all 2D image presentations
Many of the 2D image presentation examples still use the perspective camera. Instead they should use the orthographic camera. As part of this, we may also wish to fix some of the hard coded coordinates associated with the image plane geometry (related to https://github.com/chanzuckerberg/imaging-active-learning/issues/35).

andy-sweet:
### Summary
This changes some internal details about the orthographic camera and geometry coordinates, so that users have some way to specify a frame or region of interest through the camera that is independent of the canvas' size.

It also updates the associated resizing logic so that the aspect ratio of the rendered geometries are maintained regardless of the aspect ratio of the canvas.

One major change here is to change the plane geometry to not flip the y-coordinates, which presents images in the same orientation as most other image visualization tools.

### Related Issue
Closes #78
Related to #35

### Tests & Checks
I manually tested my changes with the existing examples and the prototype application.


src/objects/cameras/orthographic_camera.ts
```diff
@@ -39,14 +45,46 @@ export class OrthographicCamera extends Camera {
...
+    const halfHeight = 0.5 * height;
     mat4.ortho(
       this.projectionMatrix_,
-      this.left_,
-      this.right_,
-      this.bottom_,
-      this.top_,
+      horizontalCenter - horizontalScale * halfWidth,
+      horizontalCenter + horizontalScale * halfWidth,
+      -(verticalCenter - verticalScale * halfHeight),
```
andy-sweet: Note that the image content in our visualization is flipped in y compared to every other image viewer I tried (e.g. try it for `public/texture_test.png`).

I absorbed the corrective factor in here, which works, but I'm not sure if that's correct.

The resulting issue of doing this is that I think we are implicitly defining a right-handed coordinate system, which may not be desirable because in most scientific image visualization positive-y goes down the canvas.

src/objects/cameras/orthographic_camera.ts
```diff
@@ -39,14 +45,46 @@ export class OrthographicCamera extends Camera {
...
+    const width = this.right_ - this.left_;
+    const height = this.top_ - this.bottom_;
+    const frameAspectRatio = width / height;
+
+    let horizontalScale = 1;
+    let verticalScale = 1;
+    // The viewport is wider than the frame, so scale the x-coordinates
+    // of the frame.
+    if (this.viewportAspectRatio_ > frameAspectRatio) {
+      horizontalScale = this.viewportAspectRatio_ / frameAspectRatio;
```
andy-sweet: The computations can be simplified here, but I left things unreduced for now to keep things more clear.

src/renderers/webgl_renderer.ts
```diff
@@ -73,11 +73,12 @@ export class WebGLRenderer extends Renderer {
   }
 
   protected resize(width: number, height: number) {
+    console.debug("Renderer::resize", width, height);
     this.gl.viewport(0, 0, width, height);
   }
 
   protected clear() {
-    this.gl.clearColor(0, 0, 0, 1.0);
+    this.gl.clearColor(1, 0, 0, 1.0);
```
andy-sweet: Debugging with a non-black color was especially helpful in this PR!

examples/image2d_from_omezarr5d_u16/main.ts
```diff
@@ -1,16 +1,16 @@
...
   OmeZarrImageSource,
+  OrthographicCamera,
 } from "@";
 
 const url =
   "https://public.czbiohub.org/royerlab/zebrahub/imaging/single-objective/ZSNS001.ome.zarr/";
 const layerManager = new LayerManager();
 const renderer = new WebGLRenderer("#canvas");
-const camera = new PerspectiveCamera(60, renderer.width / renderer.height);
+const camera = new OrthographicCamera(0, 457, 0, 457);
```
andy-sweet, line 13: These coordinates come from the size of the chunk that is returned when fetching the region defined below. That region is implicitly in units of microns according to the corresponding OME-Zarr axes and scale factors. We should aim to make this more explicit in the future.

src/objects/cameras/orthographic_camera.ts
```diff
@@ -39,12 +44,32 @@ export class OrthographicCamera extends Camera {
   }
 
   protected updateProjectionMatrix() {
+    // The following code ensures that the orthographic projection matrix
```
andy-sweet, line 47: I tried to name variables and structure this to make it fairly readable and performant, but still needed quite a few comments to help explain it. Open to suggestions and improvements!

src/renderers/webgl_renderer.ts
```diff
@@ -77,7 +77,7 @@ export class WebGLRenderer extends Renderer {
   }
 
   protected clear() {
-    this.gl.clearColor(0, 0, 0, 1.0);
+    this.gl.clearColor(1, 1, 1, 1);
```
andy-sweet, line 80: Don't want to churn on the bg color - but this is a nice compromise for the imaging data in the examples and the lack of a dark mode in SDS.

src/objects/cameras/orthographic_camera.ts
```diff
@@ -39,12 +44,32 @@ export class OrthographicCamera extends Camera {
...
+    const verticalCenter = 0.5 * (this.bottom_ + this.top_);
     mat4.ortho(
       this.projectionMatrix_,
-      this.left_,
-      this.right_,
-      this.bottom_,
-      this.top_,
+      horizontalCenter - viewportHalfWidth,
+      horizontalCenter + viewportHalfWidth,
+      -(verticalCenter - viewportHalfHeight),
```
andy-sweet: Note that the image content in our visualization is currently flipped in y (on `main`) compared to every other image viewer I tried (e.g. try it for public/texture_test.png).

I absorbed the corrective factor in here, which works, but I'm not sure if it's the best place for it. We also need to think if we need something similar for the perspective camera. 

The resulting issue of doing this is that I think we are implicitly defining a right-handed coordinate system, which may not be desirable because in most scientific image visualization positive-y goes down the canvas. The related effect on this PR is that the orthographic camera coordinates are defined in this right-handed coordinate system, which may be unnatural.

andy-sweet: The reason this was needed was because `PlaneGeometry` was flipping the y-coordinates, so this was flipping them back. I removed the flip in the geometry and so do not need this anymore.

src/core/renderer.ts
```diff
@@ -46,17 +49,17 @@ export abstract class Renderer {
...
+        this.activeCamera_.setAspectRatio(aspectRatio);
       }
       if (this.activeCamera_ instanceof OrthographicCamera) {
-        this.activeCamera_.setFrame(
-          -this.width / 2,
-          this.width / 2,
-          -this.height / 2,
-          this.height / 2
-        );
+        this.activeCamera_.setViewportAspectRatio(aspectRatio);
```
aganders3, line 62: It's out of scope for this PR, but I still think this should all be generic.

Both cameras here only care about the aspect ratio, but we're branching and calling differently named functions for no reason. We should instead add an abstract method to the base `Camera` type such that each camera may implement its own `onCanvasSizeChange(width, height)` (or whatever name makes sense).

src/objects/cameras/orthographic_camera.ts
```diff
@@ -39,12 +44,32 @@ export class OrthographicCamera extends Camera {
...
+    // When the viewport is wider than the camera frame, add horizontal
+    // padding such that the height is unchanged. Otherwise, add vertical
+    // padding such that the width is unchanged.
+    let viewportHalfWidth = 0.5 * width;
+    let viewportHalfHeight = 0.5 * height;
+    if (this.viewportAspectRatio_ > frameAspectRatio) {
+      viewportHalfWidth *= this.viewportAspectRatio_ / frameAspectRatio;
+    } else {
+      viewportHalfHeight *= frameAspectRatio / this.viewportAspectRatio_;
+    }
```
aganders3, line 63: I'm not sure it's any cleaner/clearer, but I'm pretty sure the `frameAspectRatio` just cancels out in these equations.

With this change you could also remove the `viewportAspectRatio_` property and just do this calculation in `setViewportAspectRatio`.

```suggestion
    // When the viewport is wider than it is tall, add horizontal
    // padding such that the height is unchanged.
    // Otherwise, add vertical padding such that the width is unchanged.
    let viewportHalfWidth = 0.5 * width;
    let viewportHalfHeight = 0.5 * height;
    if (this.viewportAspectRatio_ > 1) {
      viewportHalfWidth = viewportHalfHeight * this.viewportAspectRatio_;
    } else {
      viewportHalfHeight = viewportHalfWidth / this.viewportAspectRatio_;
    }
```

src/objects/geometry/plane_geometry.ts
```diff
@@ -21,15 +19,15 @@ export class PlaneGeometry extends Geometry {
...
       for (let ix = 0; ix < gridX1; ++ix) {
-        const x = ix * segmentW - halfWidth;
+        const x = ix * segmentW;
         const u = ix / gridX;
         const v = 1 - iy / gridY;
 
         // 'z = -5' is temporary until we add a transform to renderable objects
         // which will allow us to control the camera position
-        const position = [x, -y, -5];
+        const position = [x, y, -5];
```
aganders3: Optional: can we take this opportunity to remove the -5 here?

I guess we may then need to hard-code a default transform for the cameras, and at least change the near plane for the orthographic camera.

src/renderers/webgl_renderer.ts
aganders3: For what it's worth I think new SDS does have dark mode, so maybe we need an issue to update it.

I guess once we have #52 we can set the color based on the theme?

examples/image2d_from_omezarr5d_u16/main.ts
aganders3: I'm looking at the [.zattrs](https://public.czbiohub.org/royerlab/zebrahub/imaging/single-objective/ZSNS001.ome.zarr/.zattrs) to try to understand this. We request a region that is `800x800` but it comes back as `457x457` because it's accessing the coarsest resolution, which has pixel spacing in x and y of `1.756` and `800 / 1.756 ≈ 457`. Is that correct? Doing the math myself I get 455.6 but I just assume there's some different rounding.

Sorry, I think the code actually handling this is not new.

aganders3 (approved):
LGTM! The orthographic camera is much more intuitive with this change, especially on window resizing.

I also like that this is a step in the direction of #35, though I understand why this does not close that issue.

examples/image2d_from_omezarr5d_u16/main.ts
aganders3: This is another case where having something like `extent` would be helpful to set the camera based on the data.

andy-sweet: > I'm looking at the [.zattrs](https://public.czbiohub.org/royerlab/zebrahub/imaging/single-objective/ZSNS001.ome.zarr/.zattrs) to try to understand this. We request a region that is `800x800` but it comes back as `457x457` because it's accessing the coarsest resolution, which has pixel spacing in x and y of `1.756` and `800 / 1.756 ≈ 457`. Is that correct? Doing the math myself I get 455.6 but I just assume there's some different rounding.

Correct. I can add a comment here to help explain if that's helpful.

src/objects/cameras/orthographic_camera.ts
andy-sweet: While the calculations can be simplified as in the above, the condition cannot be.

You can see this in the OME-Zarr image example when resizing the window to have a roughly 1:1 aspect ratio will cause the intended camera frame to be clipped horizontally.

Note that you cannot see it in the single mesh layer example, because that image has an aspect ratio of 1:1.

> With this change you could also remove the viewportAspectRatio_ property and just do this calculation in setViewportAspectRatio.

I can do the calculation in `setViewportAspectRatio` and in `setFrame`, though the existing logic around `Camera::update` made me choose to do it this way.

src/core/renderer.ts
andy-sweet: Agreed. Given that the perspective camera doesn't have a frame (with an aspect ratio), I like the suggestion of `onCanvasSizeChange`.

src/objects/geometry/plane_geometry.ts
andy-sweet: Given the title of the PR, it is related, but I also wanted to avoid it for now!

I think I like specifying the default transform of the orthographic camera so that objects at a depth of 0 within its frame are displayed.

src/objects/cameras/orthographic_camera.ts
aganders3: > While the calculations can be simplified as in the above, the condition cannot be.

Okay that makes sense of course - you can't just divide one side of the inequality.

I don't have a preference about moving the calculation. What you have is fine!

examples/image2d_from_omezarr5d_u16/main.ts
aganders3: Thanks - no need to add a comment for this.

src/objects/cameras/orthographic_camera.ts
aganders3: I also think your suggestion in https://github.com/chanzuckerberg/imaging-active-learning/pull/71#issuecomment-2412425494 to limit the specification of the camera is a good one (at least for the common case).

src/objects/geometry/plane_geometry.ts
andy-sweet: One simple solution is to set this to 0 and then make the default orthographic camera near plane 0 too. Unlike the perspective projection matrix, the near plane coordinate can be 0.

PR #86, opened 2024-10-15
Add data updates for textures

### LINKED ISSUE #69: Allow sub-regions of textures to be updated

shlomnissan:
### Summary
This PR adds the ability to update texture data after the texture's creation.
It eliminates the need to create new textures and renderable objects each
time the time index changes in the image series layer.

### Notes
- This update doesn't support modifying sub-regions. It assumes we're always replacing the entire dataset. We can address sub-regions when there is a need for it.
- I consider this an intermediate step for the image series layer. Ideally, it will use a 2D texture array to load multiple layers simultaneously instead of loading layers individually when the time index changes.

### Related Issue
Closes #69

### Tests & Checks
- Visual verification (see video below.)
- I also manually verified that an "image series" allocates a texture only once and updates the data for each new time index.

https://github.com/user-attachments/assets/49197dc6-a314-4d96-869e-96e2233086ca

### DIFF
--- a/src/layers/image_series_layer.ts
+++ b/src/layers/image_series_layer.ts
@@ -14,6 +14,7 @@ export class ImageSeriesLayer extends Layer {
   private readonly region_: Region;
   private readonly timeInterval_: Interval;
   private readonly timeDimensionIndex_: number;
+  private texture_: DataTexture2D | null = null;
   private dataChunks_: ImageChunk[] = [];
 
   constructor(source: ImageChunkSource, region: Region, timeDimension: string) {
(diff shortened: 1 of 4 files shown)


andy-sweet (approved):
Great small PR that looks roughly like I expected, so no comments - merge it!

PR #87, opened 2024-10-17
Load multiple channels from data source

shlomnissan:
### Summary

This PR enhances the Zarr image loader and image series example to support multi-channel
data. It introduces a new texture type for 2D texture arrays, which is now the default for
multi-channel scenarios. Currently, the PR assumes 2D texture arrays represent textures
with three channels (RGB).

### Tests & Checks

- I verified that all examples still work as expected and that the image series example in particular renders the texture with RGB channels. See video below:

https://github.com/user-attachments/assets/4f81ad20-cdc1-41be-9555-e093c124b7c8

### DIFF
--- a/examples/image_series_from_omezarr5d_u8/main.ts
+++ b/examples/image_series_from_omezarr5d_u8/main.ts
@@ -13,13 +13,14 @@ const layerManager = new LayerManager();
 const renderer = new WebGLRenderer("#canvas");
 const camera = new OrthographicCamera(0, 1920, 0, 1440);
 
-// Source is 5D, so provide an interval in T and scalar indices in C (first of
-// three channels) and Z (first of only depth) to get a 2D image series.
+// Source is 5D, so provide an interval in T and C (all three channels)
+// and scalar indices in Z (first of only depth) to get a 2D image series.
 const source = new OmeZarrImageSource(url);
 const timeInterval = { start: 100, stop: 120 };
+const channels = { start: 0, stop: 3 };
 const region = [
   { dimension: "T", index: timeInterval },
-  { dimension: "C", index: 0 },
+  { dimension: "C", index: channels },
   { dimension: "Z", index: 0 },
 ];
 const layer = new ImageSeriesLayer(source, region, "T");
--- a/src/data/ome_zarr_image_loader.ts
+++ b/src/data/ome_zarr_image_loader.ts
@@ -89,15 +89,9 @@ export class OmeZarrImageLoader {
       );
     }
 
-    if (subarray.shape.length !== 2) {
+    if (subarray.shape.length !== 2 && subarray.shape.length !== 3) {
       throw new Error(
-        `Expected to receive a 2D subarray. Instead chunk has shape ${subarray.shape}`
-      );
-    }
-
-    if (subarray.stride[1] !== 1) {
-      throw new Error(
-        `Expected to find a column stride of 1. Instead found ${subarray.stride[1]}`
+        `Expected to receive a 2D or 3D subarray. Instead chunk has shape ${subarray.shape}`
       );
     }
 
@@ -110,8 +104,12 @@ export class OmeZarrImageLoader {
 
     const chunk = {
       data: subarray.data,
-      shape: { width: subarray.shape[1], height: subarray.shape[0] },
-      rowStride: subarray.stride[0],
+      shape: {
+        width: subarray.shape[subarray.shape.length - 1],
+        height: subarray.shape[subarray.shape.length - 2],
+        channels: subarray.shape.length === 3 ? subarray.shape[0] : 1,
+      },
+      rowStride: subarray.stride[subarray.stride.length - 2],
       rowAlignmentBytes: rowAlignment,
     };
     console.debug("loaded chunk ", chunk);
--- a/src/layers/image_series_layer.ts
+++ b/src/layers/image_series_layer.ts
@@ -3,7 +3,6 @@ import { Mesh } from "objects/renderable/mesh";
 import { PlaneGeometry } from "objects/geometry/plane_geometry";
 import { Interval, Region } from "data/region";
 import { ImageChunk, ImageChunkSource } from "data/image_chunk";
-import { DataTexture2D } from "objects/textures/data_texture_2d";
 import { Texture2DArray } from "objects/textures/texture_2d_array";
 
 // Loads 2D+t image data from an image source into renderable objects.
@@ -12,7 +11,7 @@ export class ImageSeriesLayer extends Layer {
   private readonly region_: Region;
   private readonly timeInterval_: Interval;
   private readonly timeDimensionIndex_: number;
-  private texture_: DataTexture2D | Texture2DArray | null = null;
+  private texture_: Texture2DArray | null = null;
   private dataChunks_: ImageChunk[] = [];
 
   constructor(source: ImageChunkSource, region: Region, timeDimension: string) {
@@ -104,19 +103,11 @@ export class ImageSeriesLayer extends Layer {
   }
 
   private initializeTexture(chunk: ImageChunk) {
-    if (chunk.shape.channels > 1) {
-      this.texture_ = new Texture2DArray(
-        chunk.data,
-        chunk.shape.width,
-        chunk.shape.height
-      );
-    } else {
-      this.texture_ = new DataTexture2D(
-        chunk.data,
-        chunk.shape.width,
-        chunk.shape.height
-      );
-    }
+    this.texture_ = new Texture2DArray(
+      chunk.data,
+      chunk.shape.width,
+      chunk.shape.height
+    );
 
     this.texture_.unpackRowLength = chunk.rowStride;
     this.texture_.unpackAlignment = chunk.rowAlignmentBytes;
--- a/src/layers/image_series_layer.ts
+++ b/src/layers/image_series_layer.ts
@@ -3,6 +3,7 @@ import { Mesh } from "objects/renderable/mesh";
 import { PlaneGeometry } from "objects/geometry/plane_geometry";
 import { Interval, Region } from "data/region";
 import { ImageChunk, ImageChunkSource } from "data/image_chunk";
+import { DataTexture2D } from "objects/textures/data_texture_2d";
 import { Texture2DArray } from "objects/textures/texture_2d_array";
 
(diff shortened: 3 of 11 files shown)


src/data/ome_zarr_image_loader.ts
```diff
@@ -110,8 +104,12 @@ export class OmeZarrImageLoader {
 
     const chunk = {
       data: subarray.data,
-      shape: { width: subarray.shape[1], height: subarray.shape[0] },
-      rowStride: subarray.stride[0],
+      shape: {
+        width: subarray.shape[subarray.shape.length - 1],
+        height: subarray.shape[subarray.shape.length - 2],
+        channels: subarray.shape.length === 3 ? subarray.shape[0] : 1,
```
andy-sweet, line 110: The definition of `ImageChunk` should be modified to include `channels` so that callers of this function can use that property with type checking.

andy-sweet (commented):
This PR is a little hard to evaluate/justify in isolation because it's not needed for anything on `main` and it complicates the typing and implementation. So this is a case where splitting up the PR into smaller chunks makes reviewing more difficult for me.

In the long term, I expect to define `ImageChunk` to mirror zarrita's array/subarray so that it can store any number of dimensions and their respective shape and stride. That way we don't need a separate type and implementation to fetch a 2D, 3D, 4D, or 5D chunk.

In the short term, I don't mind `ImageChunk` being whatever is most convenient. Plus I could be convinced that channels are a special type of image dimension that we might make a special case for. So if this helps move the multi-channel work forward in a meaningful way, go for it!

src/data/ome_zarr_image_loader.ts
aganders3: Maybe we should prioritize this TODO:
https://github.com/chanzuckerberg/imaging-active-learning/blob/540569ebb90fd9123a594acdd449082463b7e954/src/data/image_chunk.ts#L5

My concern here is mapping a dim to `channels` that does not represent channels. This feels like something that should be more explicit, like how the image series layer allows you to specify the time dimension.

src/objects/textures/texture_2d_array.ts
```diff
@@ -0,0 +1,43 @@
...
+  private readonly depth_: number;
+
+  constructor(data: ArrayBufferView, width: number, height: number) {
+    super();
+
+    this.data_ = data;
+    this.width_ = width;
+    this.height_ = height;
+    // We currently assume that each slice's size is equal to the image's area
+    this.depth_ = data.byteLength / (width * height);
```
shlomnissan, line 16: Technically, this is the only change compared to the "data texture 2D" type. We could consolidate them, but I suspect that any use case beyond our basic scenario here will require additional properties and configuration. Therefore, I'm inclined to keep it as a separate type, but I'm open to consolidating them.

examples/image_series_from_omezarr5d_u8/main.ts
```diff
@@ -15,13 +15,14 @@ const renderer = new WebGLRenderer("#canvas");
...
-// three channels) and Z (first of only depth) to get a 2D image series.
+// Source is 5D, so provide an interval in T and C (all three channels)
+// and scalar indices in Z (first of only depth) to get a 2D image series.
 const source = new OmeZarrImageSource(url);
 const timeInterval = { start: 100, stop: 120 };
+const channels = { start: 0, stop: 3 };
 const region = [
   { dimension: "T", index: timeInterval },
-  { dimension: "C", index: 0 },
+  { dimension: "C", index: channels },
```
andy-sweet, line 23: Since the "C" dimension in the region spans across all channels it can also be omitted (as for X and Y). But I don't mind keeping the explicit all channels as it gives us an easy way to modify the example to show 1 or 2 channels instead.

src/data/ome_zarr_image_loader.ts
andy-sweet: > My concern here is mapping a dim to `channels` that does not represent channels. This feels like something that should be more explicit, like how the image series layer allows you to specify the time dimension.

Agreed.

But I'm fine with how things are in done in this PR because it works and it provides more evidence that some technical design thought is needed in this area.

src/layers/image_series_layer.ts
```diff
@@ -14,7 +15,7 @@ export class ImageSeriesLayer extends Layer {
   private readonly region_: Region;
   private readonly timeInterval_: Interval;
   private readonly timeDimensionIndex_: number;
-  private texture_: DataTexture2D | null = null;
+  private texture_: DataTexture2D | Texture2DArray | null = null;
```
andy-sweet: Another thought here is to always use `Texture2DArray` and treat `DataTexture2D` as a special case where the array has a length/depth of 1. I'm not sure if there performance concerns there (I suspect not?), but it would help with the typing and related code branching.

That does not need to happen in this PR, but if we think that we'll handle multi-channel images using texture arrays in the future, I think it would be appropriate.

andy-sweet (approved):
Example works great! Obviously takes a little longer on the initial load (no cache), but given how the data is structured, I don't think there's much we can do there.

There's a little ugliness here, but some of that was present before and I think it's useful in terms of motivating some of the upcoming technical designs we want to propose.

andy-sweet:
This branch seems a little behind `main`, but I merged locally without conflict and didn't notice any issues with some of the improvements (image aspect ratio and orientation). If you merge it and push, it might also be nice to update the tracks example to present RGB images (it only requires removing one line!).

src/renderers/shaders/uint_image_array_frag.glsl
```diff
@@ -0,0 +1,17 @@
...
+precision mediump float;
+
+layout (location = 0) out vec4 fragColor;
+
+uniform mediump usampler2DArray texture0;
+
+in vec2 TexCoords;
+
+void main() {
+    float red = float(texture(texture0, vec3(TexCoords, 0)).r) / 256.0;
```
andy-sweet, line 12: One thing we'd want to support for multi-channel image presentation is to effectively turn off the visibility of some layers. How were you thinking about handling that?

One thought is to do it the same way as for pixel value normalization (#32), which is to pass through visualization configuration parameter values down from the layer to the shader.

src/renderers/webgl_textures.ts
```diff
@@ -32,7 +34,7 @@ export class WebGLTextures {
     if (texture.needsUpdate && texture.data !== null) {
       // Currently, we don't support mipmaps, so we always update the base level (0).
       const mipmapLevel = 0;
-      this.uploadTextureSubData(texture, mipmapLevel, { x: 0, y: 0 });
+      this.uploadTextureSubData(texture, mipmapLevel, { x: 0, y: 0, z: 0 });
```
aganders3: nit - could also add a comment about the offset always being 0.
```suggestion
      const offset = { x: 0, y: 0, z: 0 };
      this.uploadTextureSubData(texture, mipmapLevel, offset);
```

src/layers/image_series_layer.ts
aganders3: It's also likely this layer will always use `Texture2DArray` in the future as we've discussed storing the frames of the series in the array (interleaving frames + channels). I agree this can go in a separate PR though.

src/renderers/webgl_textures.ts
```diff
@@ -97,42 +99,79 @@ export class WebGLTextures {
...
+        texture.height
+      );
+    } else if (this.getGLTextureType(texture) === this.gl_.TEXTURE_2D_ARRAY) {
+      this.gl_.texStorage3D(
+        this.getGLTextureType(texture),
+        texture.mipmapLevels,
+        this.getGLInternalFormat(texture.dataFormat, texture.dataType),
+        texture.width,
+        texture.height,
+        (texture as Texture2DArray).depth
```
aganders3: Something doesn't sit well with me about this cast (and the same in `uploadTextureSubData` below). The branch is on `getGLTextureType` so the cast feels like leaking logic from `getGLTextureType`.

I don't have a great suggestion for this though, so it's not blocking. Maybe there's some more abstract type `Texture2DArray` can conform to that makes sure it includes a `depth` attribute?

aganders3 (approved):
Looks good and it's very exciting to see color images!

aganders3:
One more comment - I find the wrapping behavior a bit unintuitive. So modifying the example to use `const channels = { start: 0, stop: 2 };` we're actually seeing an RG displayed as RGR -> RGB. This was surprising to me. I'd also consider warning or error if there are more than three channels, because I think that information will just be discarded.

src/renderers/shaders/uint_image_array_frag.glsl
aganders3: Same with separate contrast limits per-channel, and setting color-per-channel.

For reference, check out the [vizarr example](https://hms-dbmi.github.io/vizarr/?source=https://minio-dev.openmicroscopy.org/idr/v0.3/idr0062-blin-nuclearsegmentation/6001240.zarr&viewState={%22zoom%22:1.6415460290875237,%22target%22:[135.5,137.5]}).

shlomnissan: > which is to pass through visualization configuration parameter values down from the layer to the shader.

This approach makes sense. While we work towards implementing dynamic features through preprocessor value injection, we can tackle these issues by managing separate fragment shaders and passing configuration parameters from the layer to the shader via uniform blocks.

src/renderers/webgl_textures.ts
shlomnissan: While I don't consider it a major issue, I've added type guards to eliminate the need for type casting.

shlomnissan:
> One more comment - I find the wrapping behavior a bit unintuitive. So modifying the example to use const channels = { start: 0, stop: 2 }; we're actually seeing an RG displayed as RGR -> RGB. This was surprising to me. I'd also consider warning or error if there are more than three channels, because I think that information will just be discarded.

@aganders3 I believe both issues will be addressed when we implement support for arbitrary channels. We can provide a more robust solution once the fragment shader is aware of the number of channels and blending method. For this reason, I'm inclined to leave it as is for now.

shlomnissan:
@aganders3 @andy-sweet I've updated the PR based on your feedback. Image series now consistently use `Texture2DArray`. The fragment shader didn't require changes due to the default wrapping behavior for 2D texture arrays. Specifically, when wrapping is set to repeat, the slice index doesn't overflow—it simply wraps around.

Issue #88, opened 2024-10-17
Render content to the canvas such that y-coordinates draw down the screen


andy-sweet:
In https://github.com/chanzuckerberg/imaging-active-learning/pull/85 we fixed some issues regarding the coordinate system and orthographic camera and also used the orthographic camera in all the 2D image examples.

One remaining issue with that work is that (0, 0) corresponds to the bottom left of an image.

In scientific 2D images, (0, 0) typically refers to the top-left corner, such that increasing y-coordinates go down the screen (assuming no extra camera transforms). For example, the coordinates of the tracks data refer to image coordinates that way which means that https://github.com/chanzuckerberg/imaging-active-learning/pull/71 needs to correct the coordinates for proper alignment.

One approach to fixing this is to include a vertical flip in y in the projection transform. In that case, I believe the UV coordinates must also be transformed by the corresponding flip to maintain the natural orientation of image content.

PR #91, opened 2024-10-18
[WIP] Half-baked ideas around basis vector directions

andy-sweet:
This is a very-much work-in-progress PR that captures some initial thoughts I had about how to achieve the goal described in #88 .

This approach makes the library opinionated about its coordinate system instead of following WebGL's. Importantly, it cares about two things.

- The (0, 1, 0) points to the bottom of the screen (with a default camera orientation).
- By default the top-left pixel of a 2D image at depth 0 should be (0, 0, 0).

The approach here would allow each rendering backend to make the appropriate adjustments to guarantee those, by defining a separate flip/direction transform that affects the projection matrix and the uv coordinates.

By making the flips/directions configurable, we could also allow users to choose their own coordinate system and constraints (e.g. maybe they want the bottom-left pixel of a 2D image to be at (0, 0)).

One strange consequence of this approach is that without any flips, the image content will appear upside down, but that is consistent with the constraints above (i.e. because we want the top-left pixel at (0, 0)).

Another thought here is that a flip in y could probably also be achieved by making the up-vector of camera (0, -1, 0), though I'm not sure how we would modify the uv-coordinates appropriately (or if that would even need to happen).


andy-sweet:
Closing in favor of the similar, but simpler #109

PR #92, opened 2024-10-21
Load tracks in prototype app

aganders3:
### Summary
This ports the example from #71 to the Ultrack prototype app. In the process I also converted the mock data in the app (still hard-coded, now with 10 "divison" tasks) to match the current format of the backend HITL server. As part of this process I created some types in the FE to match the API in the BE.

### Related Issue
Related to #74 (but not sufficient to close)

### Tests & Checks
Tested manually by running `npm run dev -- --mode prototype` and navigating though tasks, scrubbing time, playing/pausing.

https://github.com/user-attachments/assets/95322c64-893c-4b73-8f43-b28d4ee73717


ultrack-prototype/frontend/components/PlaybackControls.tsx
```diff
@@ -1,36 +1,45 @@
...
+        return nextTime;
+      });
+    }, playbackIntervalMs);
+
+    return () => clearInterval(interval);
+  }, [enabled, maxTime, minTime, playing, setCurTime]);
+
+  useEffect(() => {
+    setCurTime(minTime);
+  }, [minTime, setCurTime, task]);
```
andy-sweet: Is the reason you include `task` as a effect dependency here because you want to reset the `curTime` to the minimum in the case that the old and new task's share the same `minTime`? If so, it would be good to add a comment to that effect to prevent people removing `task` as a dependency (since the linter will not pick that up).

I guess that will also handle ensuring that the `curTime` is valid when the two task's `maxTime` values don't match.

ultrack-prototype/frontend/components/App.tsx
```diff
@@ -1,12 +1,47 @@
...
 export default function App() {
   const [playbackEnabled, setPlaybackEnabled] = useState(false);
-  const [curTime, setCurTime] = useState(imageSeriesTimeInterval.start);
+  const [tasks, setTasks] = useState<Task[]>([]);
+  const [currentTask, setCurrentTask] = useState(0);
+  const [curTime, setCurTime] = useState(0);
+
+  // TODO: we want to fetch new tasks more than just on mount
+  useEffect(() => {
+    const fetchOnMount = async () => {
```
andy-sweet, line 19: I'm guessing this is quick? One thought is to add some loading indication while we waiting for the tasks.

ultrack-prototype/frontend/components/Renderer.tsx
```diff
@@ -8,41 +8,61 @@ import {
...
+  const { curTime, setPlaybackEnabled, task } = props;
+  const [imageSeriesLayer, setImageSeriesLayer] =
+    useState<ImageSeriesLayer | null>(null);
 
   useEffect(() => {
     console.debug("Renderer::useEffect::curTime: ", curTime);
-    if (playbackEnabled) {
-      layer.setTimeIndex(curTime);
+    if (imageSeriesLayer !== null) {
+      setTimeIndexWhenReady(imageSeriesLayer, curTime);
```
andy-sweet, line 37: If #94 goes in before this, then you might want to inline this function so that you can clearly remove the layer state change callback as part of the effect's cleanup.

ultrack-prototype/frontend/components/PlaybackControls.tsx
aganders3: Yeah, exactly - when switching tasks I want to always reset to the minimum time. This would normally be triggered, unless the `minTime` of a new task is the same as the previous.

Another thought is that `minTime` and `maxTime` can just be properties of a `Task` if we change that to be a class instead of a type.

ultrack-prototype/frontend/components/App.tsx
aganders3: It's very quick now because it's hard-coded data. Fetching from the server is pretty quick, but we probably do want some loading indicator or maybe [`<Suspense>`](https://react.dev/reference/react/Suspense)?

ultrack-prototype/frontend/components/Renderer.tsx
```diff
@@ -8,41 +8,61 @@ import {
...
     }
-  }, [curTime, playbackEnabled]);
+    layerManager.layers.length = 0;
+    const { tracksLayer, imageSeriesLayer } = taskLayers(task, imageSource);
+    setImageSeriesLayer(imageSeriesLayer);
+    layerManager.add(tracksLayer);
+    layerManager.add(imageSeriesLayer);
+
+    // TODO: update the camera in-place instead of creating a new one (this will make zoom/pan callbacks easier to manage)
+    camera = tracksLayerCamera(tracksLayer);
```
andy-sweet: I don't understand when we should use a React ref for `camera` instead of the mutable module variable.  Do you have any thoughts?

At first, it was surprising to me this works correctly in conjunction with the animation effect - but then I had to re-learn that arrow functions capture bindings and not values/references.

ultrack-prototype/frontend/lib/tasks.ts
```diff
@@ -0,0 +1,108 @@
...
+
+import {
+  ImageSeriesLayer,
+  OmeZarrImageSource,
+  OrthographicCamera,
+  ProjectedLineLayer,
+} from "@";
+
+// TODO: create a function to validate tasks as they come from the server
+// TODO: possibly make this a class instead
```
andy-sweet: One minor benefit of this is that we could cache things like `minTime` and `maxTime` instead of recalculating them each time, assuming that the track data and time points are immutable.

ultrack-prototype/frontend/components/Renderer.tsx
```diff
@@ -8,41 +8,63 @@ import {
...
+    // TODO: need to remove observer as part of dismount function.
+    // https://github.com/chanzuckerberg/imaging-active-learning/issues/77
+    imageSeriesLayer.onStateChange((newState) => {
+      if (newState === "ready") {
+        setPlaybackEnabled(true);
+        layerManager.layers.length = 0;
+        layerManager.add(tracksLayer);
+        layerManager.add(imageSeriesLayer);
+        // TODO: update the camera in-place instead of creating a new one (this will make zoom/pan callbacks easier to manage)
+        camera = tracksLayerCamera(tracksLayer, 2.0);
```
andy-sweet: I tried the padding factor of 2 and generally find the tracks a little easier to verify, though I think we still need some contrast limit and other presentation improvements to help.

andy-sweet (commented):
Generally works well for me and looks good enough right now, though I still need to think through the `Renderer` component and if there are potential improvements we could address in this PR. Happy to review again when this is ready.

Also, on a probably unrelated note I started noticing that the lines occasionally flicker with a white outline. I wonder if this due to depth contention of the image plane and the lines?

aganders3:
> Also, on a probably unrelated note I started noticing that the lines occasionally flicker with a white outline. I wonder if this due to depth contention of the image plane and the lines?

Oh interesting - I don't think I've seen this. I guess it could be depth contention, or maybe an issue with anti-aliasing + blending?

edit: I have seen this now. I'll probably debug a bit but will create a new issue for it.

ultrack-prototype/frontend/components/Renderer.tsx
aganders3: Good question. I honestly forgot about `useRef`. Doing a bit of reading just now it seems like it probably doesn't matter for our use here, but `useRef` might still be a bit safer and more idiomatic.

ultrack-prototype/frontend/components/Renderer.tsx
aganders3: I made #95 to specifically capture tracks improvements. For what it's worth the color images are a big improvement even without contrast limits.

ultrack-prototype/frontend/components/Renderer.tsx
aganders3: I'm going to leave this unchanged for this PR, but will create an issue. I think this code is likely to change when pan/zoom is added (we can modify the camera in-place) and we may want `useRef` for some of the other components here (the layer manager and maybe the source).

ultrack-prototype/frontend/components/App.tsx
```diff
@@ -1,38 +1,46 @@
...
+    };
+    fetchOnMount();
+  }, []);
 
   const setTaskAnswer = (answer: Answer) => {
     setTasks((prevTasks) =>
-      prevTasks.map((task, index) =>
-        index === taskIndex ? { ...task, answer } : task
-      )
+      prevTasks.map((task, index) => {
```
andy-sweet, line 27: Is there a reason to prefer the new form? The old form is much more concise and doesn't require a method on `Task`.

ultrack-prototype/frontend/components/App.tsx
```diff
@@ -69,16 +77,19 @@ export default function App() {
         <Renderer
           playbackEnabled={playbackEnabled}
           setPlaybackEnabled={setPlaybackEnabled}
+          task={task}
           curTime={curTime}
         ></Renderer>
         <Question
+          disabled={!playbackEnabled}
           task={tasks[taskIndex]}
```
andy-sweet: For consistency with other components.

```suggestion
          task={task}
```

ultrack-prototype/frontend/lib/tasks.ts
```diff
@@ -0,0 +1,233 @@
...
+
+import {
+  ImageSeriesLayer,
+  OmeZarrImageSource,
+  OrthographicCamera,
+  ProjectedLineLayer,
+} from "@";
+
+type Track = {
+  track_id: number;
```
andy-sweet: I assume the reason these parameter names use `snake_case` is because of the fields in the JSON that the Python backend produces? Given that the rest of the code base uses `mixedCase` and we reserve a trailing `_` to indicate a class attribute, I'd prefer to stick with `mixedCase` if it's not much trouble.

ultrack-prototype/frontend/lib/tasks.ts
```diff
@@ -0,0 +1,233 @@
...
+  return (
+    typeof taskType === "string" && TASK_TYPES.includes(taskType as TaskType)
+  );
+}
+
+export type Answer = "Unanswered" | "Yes" | "No" | "Uncertain";
+
+// TODO: create a function to validate tasks as they come from the server
+export class Task {
+  task_id!: string;
```
andy-sweet: Should these have trailing underscores? I think the linter should have picked that up, but it might be misconfigured right now...

ultrack-prototype/frontend/lib/tasks.ts
```diff
@@ -0,0 +1,233 @@
...
+      case "division":
+        return "Is this a cell division event?";
+      default: {
+        const exhaustiveCheck: never = this.task_type;
+        throw new Error(`Unhandled task type: ${exhaustiveCheck}`);
+      }
+    }
+  }
+
+  tracksAs3DPaths(): vec3[][] {
```
andy-sweet: This is only intended for internal use, right?

```suggestion
  private tracksAs3DPaths(): vec3[][] {
```

Or do you think this could be useful elsewhere?

ultrack-prototype/frontend/lib/tasks.ts
```diff
@@ -0,0 +1,233 @@
...
+
+  public get minTime(): number {
+    return this.timeInterval.start;
+  }
+
+  imageSeriesLayer(
+    source: OmeZarrImageSource,
+    preLoad = true
+  ): ImageSeriesLayer {
+    const region = [
```
andy-sweet, line 197: Why is this not cached as with `tracksLayer()`?

ultrack-prototype/frontend/lib/tasks.ts
```diff
@@ -0,0 +1,233 @@
...
+          path,
+          color: COLOR_CYCLE[i % COLOR_CYCLE.length],
+          width: 0.01,
+        }))
+      );
+    }
+    return this.tracksLayer_;
+  }
+
+  public layers(imageSource: OmeZarrImageSource): {
```
andy-sweet, line 221: I would have assumed that the source would have been included in the task data. I guess it's always the same now, so it doesn't really matter, but I'd expect `OmeZarrImageSource` to be created in this module or in `mock_data.ts` instead of `Renderer.tsx`.

ultrack-prototype/frontend/components/Renderer.tsx
```diff
@@ -7,41 +7,68 @@ import {
...
+let camera = new OrthographicCamera(0, 1920, 0, 1440);
 const layerManager = new LayerManager();
-layerManager.add(layer);
 
 type RendererProps = {
+  curTime: number;
   playbackEnabled: boolean;
   setPlaybackEnabled: Dispatch<SetStateAction<boolean>>;
-  curTime: number;
+  task: Task | null;
```
andy-sweet, line 26: At some point in the future, I'd expect us to define a more generic `Renderer` component that might take in enough information to define layers and a camera. But we definitely don't have to get there in this PR, so this is fine.

andy-sweet (commented):
Still need to finish to approve/request changes, but generally looks good. And the functionality/behavior is great!

ultrack-prototype/frontend/lib/tasks.ts
aganders3: Good call - I think this was trickier when parsing the JSON as a `Task` _type_, but should be doable with the class.

ultrack-prototype/frontend/components/App.tsx
aganders3: I did this because `Task` is now a class, which I think provides some benefit. Copying the object with `{ ...task, answer }` creates a generic object with the same attributes, but it's missing the methods (and maybe some other stuff).

Without adding the `clone` method I think you need to do something weird like `Object.assign(Object.create(Object.getPrototypeOf(...` but I've only done some cursory searching about it.

ultrack-prototype/frontend/lib/tasks.ts
aganders3: I think the linter only enforces trailing underscores for private attributes.

aganders3: I don't love the usage of `!` here though, which from what I gather tells `TypeScript` "I promise I will set these even though they don't have faults and I don't explicitly set them in the constructor".

ultrack-prototype/frontend/lib/tasks.ts
aganders3: I agree it should come from the server. I was thinking of a separate endpoint but it makes sense to just include it with the task payload. I'd prefer to keep this as-is here and fix that (in server + client) in a separate PR.

ultrack-prototype/frontend/lib/tasks.ts
aganders3: No reason - I'm happy to add that. I cached the other layer because it's used in `task.camera()` but didn't have a specific need to re-access the image layer.

The image layers hold more memory (though this might be _more_ reason to cache) so we may want to have a way to expire the cache. For example we could clear it after answering the question.

aganders3: Ah I just went to implement this and remembered it's not cached because it takes the source.

I think this can be updated in a future PR with #103 if you're okay with that.

ultrack-prototype/frontend/components/Renderer.tsx
aganders3: This is pretty interesting and I agree it would be useful. In this case I might still expect it to be wrapped by a `TaskRenderer` or something.

Would you expect a generic `Renderer` component to be provided by the library, or perhaps as an optional package?

ultrack-prototype/frontend/lib/tasks.ts
aganders3: I got rid of the `!` but those top three attributes are still public.

ultrack-prototype/frontend/lib/tasks.ts
aganders3: I had to muck around a bit to get this to work with the `TaskData` and `Track` types here. So now there is a `fromJSON` function for each that will validate and return the proper type (converting from snake case). I'm not sure this is the proper method.

ultrack-prototype/frontend/components/Renderer.tsx
andy-sweet: I expect (at some point) to have an optional package that provides React components.

ultrack-prototype/frontend/components/App.tsx
andy-sweet: Another thought is to implement a builder pattern like method in `Task.withAnswer(answer: Answer)`, which would make the code here less verbose, but only implement it if you like.

ultrack-prototype/frontend/components/App.tsx
andy-sweet: I think we can leave this for now, unless we see something noticeable in real demos. We're shouldn't be too focused on the production quality of the app itself.

andy-sweet (approved):
Looks good to me now. And works well for me too.

PR #93, opened 2024-10-22
Add frontend components for question answering workflow

andy-sweet:
### Summary

This adds the frontend components required for a semi-functional question answering workflow in the prototype application.

Screenshot 2024-10-22 at 10 34 33 AM

The implementation is roughly based on the [early figma wireframes](https://www.figma.com/design/gPjEi85OeHkOGRHgtGk60D/Q3-Concept-Prototype-Wireframes?node-id=0-1&node-type=canvas&t=uxdz7Hlzk2DOWWnT-0).

This introduces application state to keep track of the current question and the answer to each question, but the actual task data is mocked rather than coming from the server in its payload format.

This also fixes some layout and resizing issues with the prototype.

### Related Issue
Related to #74

### Tests & Checks
I manually tested my changes by running through the basic prototype task question/answer workflow.


ultrack-prototype/frontend/components/TaskItem.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+function sdsIcon(answer: Answer | undefined) {
+  if (answer === undefined) return "DotsHorizontal";
+  switch (answer) {
+    case Answer.YES:
+      return "CheckCircle";
+    case Answer.NO:
+      return "XMarkCircle";
+    case Answer.UNCERTAIN:
+      return "ExclamationMarkCircle";
```
andy-sweet: Unfortunately, there is no `QuestionMarkCircle` in SDS' icon library.

ultrack-prototype/frontend/components/Renderer.tsx
```diff
@@ -64,16 +63,12 @@ export default function Renderer(props: RendererProps) {
...
 
   return (
-    <Box
-      sx={{
-        display: "flex",
-        flexDirection: "column",
-        flexGrow: 1,
+    <canvas
+      id={canvasId}
+      style={{
```
andy-sweet, line 68: From memory, I think that specifying the styled width and height of the canvas may not be what want because it applies a CSS-specific transform to the canvas with its own (anti-)aliasing independent from the canvas.

But I cannot make the outer `Box` work at all in this new layout - if someone has ideas/fixes, please help!

ultrack-prototype/frontend/components/TaskItem.tsx
aganders3: As an alternative you could try the `FlagOutline `, `FlagCheck`, `FlagQuestionMark`, and `FlagXMark` icons.

aganders3:
This looks great and makes the app really feel like it's taking shape.

I'm going to take a little more time to review if that's okay. Mostly I need to think a little more about the state here, and how each task object holds its own response, which means re-creating the `tasks` list whenever a response is added/changed. I think this is probably the correct pattern, but need to compare with my branch and let it sink in.

Specifically I'm thinking about:
* which components get re-rendered and which effects are re-run when `tasks` changes
* how to synchronize the local tasks with the server (and when that would be requested)

I also want to take a quick look at the layout (though I'm also okay if that's fixed later).

aganders3:
Here's a patch that _seems_ to fix or at least improve the layout issues for me. Some of the changes may not be strictly necessary.

I think it's okay to set the canvas width/height to 100% - I'm pretty sure this is handled correctly where the renderer updates on window resize.

```diff
diff --git a/ultrack-prototype/frontend/components/App.tsx b/ultrack-prototype/frontend/components/App.tsx
index dcc8746..0707cb0 100644
--- a/ultrack-prototype/frontend/components/App.tsx
+++ b/ultrack-prototype/frontend/components/App.tsx
@@ -34,7 +34,7 @@ export default function App() {
       <Box
         sx={{
           display: "flex",
-          flexGrow: 0,
+          flex: 0,
         }}
       >
         <TaskList
@@ -49,7 +49,7 @@ export default function App() {
           height: "100%",
           display: "flex",
           flexDirection: "column",
-          flexGrow: 1,
+          flex: 1,
           gap: "1em",
         }}
       >
diff --git a/ultrack-prototype/frontend/components/PlaybackControls.tsx b/ultrack-prototype/frontend/components/PlaybackControls.tsx
index 663fe75..1bb1b47 100644
--- a/ultrack-prototype/frontend/components/PlaybackControls.tsx
+++ b/ultrack-prototype/frontend/components/PlaybackControls.tsx
@@ -36,6 +36,7 @@ export default function PlaybackControls(props: PlaybackControlsProps) {
     <Box
       sx={{
         display: "flex",
+        flex: 0,
         flexDirection: "row",
         alignItems: "center",
         gap: "1em",
diff --git a/ultrack-prototype/frontend/components/Question.tsx b/ultrack-prototype/frontend/components/Question.tsx
index deb9891..ec19b0e 100644
--- a/ultrack-prototype/frontend/components/Question.tsx
+++ b/ultrack-prototype/frontend/components/Question.tsx
@@ -26,6 +26,7 @@ export default function Question(props: QuestionProps) {
     <Box
       sx={{
         display: "flex",
+        flex: 0,
         flexDirection: "column",
         alignItems: "center",
         gap: "1em",
diff --git a/ultrack-prototype/frontend/components/Renderer.tsx b/ultrack-prototype/frontend/components/Renderer.tsx
index ec88329..98a9d1d 100644
--- a/ultrack-prototype/frontend/components/Renderer.tsx
+++ b/ultrack-prototype/frontend/components/Renderer.tsx
@@ -7,6 +7,8 @@ import {
   WebGLRenderer,
 } from "@";
 
+import { Box } from "@mui/material";
+
 import { imageSeriesProps } from "../image_series_props";
 
 const canvasId = "canvas";
@@ -63,12 +65,14 @@ export default function Renderer(props: RendererProps) {
   }, [setPlaybackEnabled]);
 
   return (
-    <canvas
-      id={canvasId}
-      style={{
-        width: "100%",
-        height: "100%",
+    <Box
+      sx={{
+        display: "flex",
+        flex: 1,
+        minHeight: 0,
       }}
-    />
+    >
+      <canvas id={canvasId} style={{ width: "100%", height: "100%" }} />
+    </Box>
   );
 }

```

andy-sweet:
> Here's a patch that _seems_ to fix or at least improve the layout issues for me. Some of the changes may not be strictly necessary.

So much better - thank you!

It seems like `min-height: 0` is the most critical change here, but I left them all in except the outer box around the `canvas` in `Renderer.tsx` because I can just push the `flex` and `minHeight` attributes into the canvas's CSS. I think that `flexGrow: 0` and `flex: 0` should be equivalent, but the latter is encouraged because it encourages you to specify the other related flex params (shrink and basis) and makes it easier to add them later.

I still don't entirely understand the need for `min-height`, but at least understand that it's overriding the default [auto behavior of flex boxes](https://drafts.csswg.org/css-flexbox-1/#min-size-auto).

andy-sweet:
> Mostly I need to think a little more about the state here, and how each task object holds its own response, which means re-creating the `tasks` list whenever a response is added/changed. I think this is probably the correct pattern, but need to compare with my branch and let it sink in.

Since the task questions are immutable and their order is also immutable, I don't mind having a paired array of answers that is the thing that effectively gets mutated. It might mean doing a little more prop drilling to pass the two together as needed, but should not be too bad.

ultrack-prototype/frontend/components/TaskItem.tsx
andy-sweet: I do like having the consistent set, so i'm going for it

ultrack-prototype/frontend/components/App.tsx
```diff
@@ -3,28 +3,73 @@ import Renderer from "./Renderer";
...
 import { imageSeriesTimeInterval } from "../image_series_props";
+import TaskList from "./TaskList";
+import Question from "./Question";
+import { Task } from "../task";
+
+const defaultTasks: Task[] = [];
+for (let i = 0; i < 30; ++i) {
+  defaultTasks.push({
+    index: i,
+    question: `Is this Cell Division ${i + 1}?`,
```
shlomnissan, line 14: Do we know the actual shape of the data? I initially thought all tasks were categorized. For example, when reviewing cell division, all tasks would involve identifying divisions. In this case, wouldn't the data be modeled differently? At least, the question wouldn't need to be stored for every task since it's the same for each category.

ultrack-prototype/frontend/components/App.tsx
```diff
@@ -3,28 +3,73 @@ import Renderer from "./Renderer";
...
+  });
+}
 
 export default function App() {
   const [playbackEnabled, setPlaybackEnabled] = useState(false);
   const [curTime, setCurTime] = useState(imageSeriesTimeInterval.start);
+  const [taskIndex, setTaskIndex] = useState(0);
```
shlomnissan, line 21: If I understand the intent correctly, I think `currTask` is a more descriptive name.

ultrack-prototype/frontend/components/TaskItem.tsx
```diff
@@ -0,0 +1,38 @@
+import { Button, Icon } from "@czi-sds/components";
+import { Dispatch, SetStateAction } from "react";
+import { Answer } from "../task";
+
+export type TaskItemProps = {
+  index: number;
+  answer?: Answer;
```
shlomnissan, line 7: I generally avoid optional types, particularly in cases like this. Instead, consider adding an explicit "unanswered" state.

aganders3:
> I don't mind having a paired array of answers that is the thing that effectively gets mutated

I think I prefer it as-is that.

Most of the components only need a single task, it's only the `TaskList` that needs the whole list (for obvious reasons) and the `Question`. The `Question` is the one that feels strange because it only needs the list (and setter) because it's trying to mutate a single element.

Now that I'm writing this out I have an idea. How about passing a callback to `Question` that can only modify the current task? So something like:

_(edit: realizing you already have a better version of this callback in `Question` - so I guess I'm just suggesting to create that callback outside of the component and pass it in)_

```
const setTaskAnswer = useCallback((answer) => {
    tasks[taskIndex].answer = answer;
    setTasks([...tasks]);
}, [tasks, setTasks, taskIndex]);

...

<Question
    task={tasks[taskIndex]}
    setTaskAnswer={setTaskAnswer}
/>
```

Maybe it's not a big difference. I'm also happy to merge as-is and reconsider if there's a real problem.

aganders3 (approved):

andy-sweet:
> The `Question` is the one that feels strange because it only needs the list (and setter) because it's trying to mutate a single element.
> 
> Now that I'm writing this out I have an idea. How about passing a callback to `Question` that can only modify the current task? So something like:

I slightly prefer this approach because it limits what the `Question` component can do with the tasks in a meaningful way. Change incoming.

aganders3:
Looks good! Feel free to merge when ready and I'll rebase #92. I think once that's done I can mark it ready to review.

edit: too slow

ultrack-prototype/frontend/components/App.tsx
```diff
@@ -1,30 +1,88 @@
 import { Box } from "@mui/material";
 import Renderer from "./Renderer";
 import PlaybackControls from "./PlaybackControls";
-import { useState } from "react";
+import { useCallback, useState } from "react";
 import { imageSeriesTimeInterval } from "../image_series_props";
+import TaskList from "./TaskList";
+import Question from "./Question";
+import { Answer, Task } from "../task";
```
shlomnissan, line 8: - If you're exporting multiple types, I suggest renaming the module to `types` or something more generic.
- We should avoid relative paths. However, it's not a major issue to leave them as-is if it's necessary due to the Vite configuration.

ultrack-prototype/frontend/components/App.tsx
```diff
@@ -1,30 +1,88 @@
...
+      updatedTask.answer = answer;
+      const updatedTasks = Array(...tasks);
+      updatedTasks[taskIndex] = updatedTask;
+      setTasks(updatedTasks);
+      setTaskIndex((prevIndex) =>
+        prevIndex < tasks.length - 1 ? prevIndex + 1 : prevIndex
+      );
+    },
+    [tasks, setTasks, taskIndex, setTaskIndex]
+  );
```
shlomnissan, line 36: I find every use of `useCallback` suspicious unless it's clearly for optimizing performance. Can you clarify why it is needed? I think a cleaner and more efficient way to handle this would be:
```suggestion
  const setTaskAnswer = (answer: Answer) => {
    setTasks((prevTasks) =>
      prevTasks.map((task, index) =>
        index === taskIndex ? { ...task, answer } : task
      )
    );
    setTaskIndex((prevIdx) =>
      prevIdx < tasks.length - 1 ? prevIdx + 1 : prevIdx
    );
  };
```

ultrack-prototype/frontend/components/TaskItem.tsx
```diff
@@ -0,0 +1,38 @@
...
+      return "FlagCheck";
+    case Answer.NO:
+      return "FlagXMark";
+    case Answer.UNCERTAIN:
+      return "FlagQuestionMark";
+  }
+}
+
+export default function TaskItem(props: TaskItemProps) {
+  const { index, answer, active, setTaskIndex } = props;
```
shlomnissan, line 25: I don't see much value in destructuring props in this manner. Personally, I find it more beneficial to reference properties as `props.index`, particularly in larger components, as it clearly identifies input variables. If you're committed to destructuring props, consider doing it inline at the function's declaration.

ultrack-prototype/frontend/components/TaskList.tsx
```diff
@@ -0,0 +1,82 @@
+import { Box, Typography } from "@mui/material";
+import TaskItem from "./TaskItem";
+import { Button } from "@czi-sds/components";
+import { Dispatch, SetStateAction } from "react";
+import { Task } from "../task";
+
+type TaskListProps = {
+  tasks: Task[];
+  taskIndex: number;
+  setTaskIndex: Dispatch<SetStateAction<number>>;
```
shlomnissan, line 10: This reminds me of our discussion about playback controls and ownership. It's another instance where I believe the task list should internally track the index, rather than relying on a parent class to pass a mutator. Even if this approach leads to minor duplication, it will result in a more self-contained and "complete" component.

ultrack-prototype/frontend/task.ts
```diff
@@ -0,0 +1,11 @@
...
+  YES = "Yes",
+  NO = "No",
+  UNCERTAIN = "Uncertain",
+}
+
+export type Task = {
+  index: number;
+  question: string;
+  answer?: Answer;
+};
```
shlomnissan, line 11: I don't think exporting an enum is needed here.

```suggestion
export type Task = {
  index: number;
  question: string;
  answer: "Yes", "No", "Uncertain", "Unanswered";
};
```

shlomnissan (commented):
Overall, this PR looks good. I'm uncertain about how it fits the shape of the data, but I don't think that's a significant concern—we can easily adjust it if needed. I'd like to see the `setTaskAnswer` function simplified. While there's room for cleanup tasks, I don't see anything blocking from my perspective.

ultrack-prototype/frontend/components/App.tsx
andy-sweet: Yes, but the goal of this task was mostly around the front-end structure, so I used this as a placeholder. I expect that in #92 we'll store a task type instead and can derive the string from that in the `Question` component.

ultrack-prototype/frontend/components/App.tsx
andy-sweet: No need for `useCallback` other than avoiding function redefinition on some re-renders, so can remove.

I also like your more concise form for updating tasks, so will make a new PR with that.

ultrack-prototype/frontend/components/TaskList.tsx
andy-sweet: I'm still unconvinced by this argument, especially because state duplication is something I try to avoid at all costs. Still open to it, but I'd need more compelling reasons and examples.

ultrack-prototype/frontend/components/App.tsx
andy-sweet: I like the consistency with `curTime`, but dislike `curTask` because it implies (to me at least) that the assigned variable is probably of type `Task`, whereas instead it's a numeric index. So I'll probably leave this for now, and might consider renaming `curTime` to `timeIndex` for consistency.

ultrack-prototype/frontend/components/TaskItem.tsx
andy-sweet: Good idea. I'll make the change in a follow-up PR.

ultrack-prototype/frontend/components/App.tsx
andy-sweet: > If you're exporting multiple types, I suggest renaming the module to types or something more generic.

I don't love generic names like `types` or `utils`, but could be open to that here, especially if there are more than these two. I think #92 has something like that, so will likely wait for that PR to be merged before making any other updates.

> * We should avoid relative paths. However, it's not a major issue to leave them as-is if it's necessary due to the Vite configuration.

That's the reason it's like this. There's probably some way of massaging the config files to avoid the relative imports, but I'm not planning to spend time on that right now.

ultrack-prototype/frontend/components/TaskItem.tsx
andy-sweet: > Personally, I find it more beneficial to reference properties as `props.index`, particularly in larger components, as it clearly identifies input variable

That is appealing. I have a vague sense that react.dev or a similarly authoritative source recommended against this, but I cannot find that now and I don't remember why.

Given this pattern is used in a few places, I'll probably create an entirely separate PR to either use `props` through or de-structure inline consistently across all components. I'll create an issue, so we can get #92 in before I start that.

ultrack-prototype/frontend/task.ts
andy-sweet: I wanted to give typescript enums another shot, but they let me down in the `Question` component when I wanted to iterate over typed pairs of keys and values.

For that iteration, I think I need more than just this declaration - i.e. I need to export some type or export some array of type names. But that's probably still better than using enums. Since there aren't many valid values, maybe I should just hard code them though.

ultrack-prototype/frontend/components/App.tsx
aganders3: I can remove `useCallback` and in #92 - I have to update this anyway to account for `Task` becoming a class instead of a type.

I'm curious though why you say `useCallback` is suspicious? I've been using it freely when defining callbacks, particularly if they get passed to another component as a prop. Is it just that usually the performance benefit is small, and the downsides are noise in the code and a chance of forgetting a dependency causing weird issues?

PR #94, opened 2024-10-22
Allow layer state change callback to be removed

### LINKED ISSUE #77: Allow a layer state change callback to be removed
From Slack

> 2. Should we able to remove a callback (i.e. in a similar way to removeEventListener)?
I feel like the answer should be yes here. This would be useful for the React prototype where I want to add a callback in the mount effect, and remove it in the unmount, but I could imagine other usage too.

> 2. I think we definitely need a way to remove callbacks. They're also a big opportunity for memory leaks without this, especially when passing closures that may capture large objects.

andy-sweet:
### Summary
This adds the `Layer.removeStateChangeCallback` to allow a callback to be removed from those that will be executed on layer state changes. After adding that I decided to change the name of `Layer.onStateChange` to `Layer.addStateChangeCallback` to mirror the add/remove event listener API names.

### Related Issue
Closes #77

### Tests & Checks
I added an automated test to ensure that a callback is not executed after being added and then removed. I also tested the prototype manually to ensure that the callback is removed after the initial dev mount/dismount.


aganders3 (approved):
LGTM!

Issue #89, opened 2024-10-18
Simplify orthographic camera parameters


andy-sweet:
The orthographic camera currently allows us to specify left, right, bottom, top coordinates as well as the affine transform it gets from being a `RenderableObject`. These parameters seem to out-number the effective degrees of freedom it has in terms of the transforms it produces.

The goal of this issue is to reduce those parameters.

For example, a complete set of parameters for an orthographic camera could be a position, an up-vector, and a positive width and height.

As part of this issue, we may need to ensure that the perspective camera is as consistent with the orthographic camera as possible. For example, if the orthographic camera is no longer a renderable object with an `AffineTransform` (and instead has a position and up-vector), then the perspective camera should be similar.

andy-sweet:
As part of this issue, it would be good to modify at least one of the examples to show how to modify some of these new more intuitive parameters (e.g. pan and zoom). We may want to provide helpers to make these easy and concise in the examples.

aganders3:
> For example, a complete set of parameters for an orthographic camera could be a position, an up-vector, and a positive width and height.

This is true but I think it's also okay to have more than one way to specify this. In different applications it may be more intuitive to set parameters in a certain way.

> For example, if the orthographic camera is no longer a renderable object with an AffineTransform (and instead has a position and up-vector), then the perspective camera should be similar.

I don't have a strong preference on whether the camera remains a renderable object or not - this feels like a bigger question about how much to we protect clients from an attractive nuisance (modifying a `transform` property directly). For example a client application can already access/modify the transform on each renderable object, but I don't think we're encouraging it.

I think it would be useful for the `AffineTransform` to have some additional methods like `setScale`, `setRotation`, and `setTranslation`. Then cameras could have more intuitive methods/properties to `postion` and `up`, but ultimately they would just be more ergonomic wrappers for modifying transform and/or projection matrices. So you'd set `camera.positon` instead of calling `camera.transform.setTranslation`.

aganders3:
Related post from @shlomnissan [on Slack](https://czi-sci.slack.com/archives/C07GHGSJSQP/p1748362561052899):
> I wanted to share a few thoughts on how I think camera controls should fit into the runtime, and get your perspective:
[...]

PR #96, opened 2024-10-23
Add pan + zoom for orthographic and perspective cameras

### LINKED ISSUE #101: Line width is incorrect with PerspectiveCamera
> Why does the width need to be different for each camera? Is that due to the way the projected line's shader program? Or is there a simpler explanation?

_Originally posted by @andy-sweet in https://github.com/chanzuckerberg/imaging-active-learning/pull/96#discussion_r1815436015_

Simpler explanation: it's a bug :)

aganders3:
### Summary
As discussed each camera has two transformation matrices that ultimately just get composed. This also stores zoom separately and applies it when updating the projection matrix. For the `OrthographicCamera` it keeps the frame centered (0, 0) and pans by setting translation on the `camera.transform`, so I updated it to store just width and height.

I moved the camera controls to their own class in https://github.com/chanzuckerberg/imaging-active-learning/commit/2f7f7e6a5d579701fed8c4bc00095cdf88136c7d. This has some oddities as we discussed yesterday, mostly due to render.canvas being private and the "active" camera switching. I am open to reverting that (keeping the callbacks in the example for demo purposes) for to keep the PR a bit less intrusive.

It's an open question how to get the clip-space Z value for the object - necessary for natural panning with the `PerspectiveCamera`. Right now I calculate this in the client and just assume the object is at Z = 0 in world space (and that the camera is looking along Z). Ultimately picking or some other parameter (i.e. giving a world-space plane) might be a better solution.

### Related Issue
Related to #89 (but does not close)
Closes #101 

### Tests & Checks
Tested manually by adding pan/zoom mouse controls and a camera toggle to the `image2d_from_omezarr5d_u8` example.

https://github.com/user-attachments/assets/9a21c836-7039-48e9-9948-1e9b7bee4fbc

### DIFF
--- a/examples/image2d_from_omezarr5d_u8/main.ts
+++ b/examples/image2d_from_omezarr5d_u8/main.ts
@@ -1,9 +1,11 @@
+import { vec2, vec3 } from "gl-matrix";
 import {
   LayerManager,
   ImageLayer,
   WebGLRenderer,
   OmeZarrImageSource,
   OrthographicCamera,
+  PerspectiveCamera,
 } from "@";
 import { AxesLayer } from "@/layers/axes_layer";
 
@@ -11,7 +13,18 @@ const url =
   "https://public.czbiohub.org/royerlab/ultrack/multi-color/image.zarr/";
 const layerManager = new LayerManager();
 const renderer = new WebGLRenderer("#canvas");
-const camera = new OrthographicCamera(-2000, 2000, -2000, 2000);
+const orthoCam = new OrthographicCamera(-2000, 2000, -2000, 2000);
+const perspectiveCam = new PerspectiveCamera(60, 1, 0.1, 10000);
+const cameraPos = vec3.fromValues(0, 0, 5000);
+perspectiveCam.transform.translate(cameraPos);
+// project the (negative) camera position to clip space
+// to get the distance to the image plane (z = 0)
+const projectedCameraPos = vec3.transformMat4(
+  vec3.create(),
+  vec3.scale(vec3.create(), cameraPos, -1),
+  perspectiveCam.projectionMatrix
+);
+const clipDistance = projectedCameraPos[2];
 
 // Source is technically 5D (even though Z is unitary),
 // so provide indices at 3 dimensions to project to 2D.
@@ -22,13 +35,72 @@ const region = [
   { dimension: "Z", index: 0 },
 ];
 const layer = new ImageLayer(source, region);
-const axes = new AxesLayer({ length: 1920, width: 0.01 });
-layerManager.add(layer);
-layerManager.add(axes);
+
+let camera: PerspectiveCamera | OrthographicCamera;
+toggleCamera();
+
+// TODO: would be nice to provide these functions in the library
+const clientToWorld = (clientPos: vec2) => {
+  const clipPos = renderer.clientToClip(clientPos);
+  const d = camera == perspectiveCam ? clipDistance : 0;
+  return camera.clipToWorld(clipPos, d);
+};
+
+document.addEventListener("wheel", (event) => {
+  const clientPos = vec2.fromValues(event.clientX, event.clientY);
+  const preZoomPos = clientToWorld(clientPos);
+  if (event.deltaY < 0) {
+    camera.zoom *= 1.05;
+  } else {
+    camera.zoom /= 1.05;
+  }
+  // pan to zoom in on the mouse position
+  const postZoomPos = clientToWorld(clientPos);
+  const dWorld = vec2.sub(vec2.create(), postZoomPos, preZoomPos);
+  camera.pan(vec2.scale(vec2.create(), dWorld, -1));
+  console.log(camera);
+});
+
+document.addEventListener("mousedown", (event) => {
+  const clientStart = vec2.fromValues(event.clientX, event.clientY);
+  let worldStart = clientToWorld(clientStart);
+  console.log(clientStart, worldStart);
+
+  const onMouseMove = (event: MouseEvent) => {
+    const clientPos = vec2.fromValues(event.clientX, event.clientY);
+    const worldPos = clientToWorld(clientPos);
+    const dWorld = vec2.sub(vec2.create(), worldPos, worldStart);
+    camera.pan(vec2.scale(vec2.create(), dWorld, -1));
+    worldStart = worldPos;
+  };
+
+  const onMouseUp = () => {
+    document.removeEventListener("mousemove", onMouseMove);
+    document.removeEventListener("mouseup", onMouseUp);
+  };
+
+  document.addEventListener("mousemove", onMouseMove);
+  document.addEventListener("mouseup", onMouseUp);
+});
 
 function animate() {
   renderer.render(layerManager, camera);
   requestAnimationFrame(animate);
+}
+
+document.addEventListener("keydown", (event) => {
+  if (event.key === " ") {
+    toggleCamera();
+  }
+});
+
+function toggleCamera() {
+  camera = camera === orthoCam ? perspectiveCam : orthoCam;
+  const width = camera.type == "PerspectiveCamera" ? 20 : 0.01;
+  const axes = new AxesLayer({ length: 1920, width });
+  layerManager.layers.length = 0;
+  layerManager.add(layer);
+  layerManager.add(axes);
 }
 
 animate();
--- a/examples/image2d_from_omezarr5d_u8/main.ts
+++ b/examples/image2d_from_omezarr5d_u8/main.ts
@@ -1,4 +1,4 @@
-import { vec2, vec3 } from "gl-matrix";
+import { vec3 } from "gl-matrix";
 import {
   LayerManager,
   ImageLayer,
@@ -8,6 +8,7 @@ import {
   PerspectiveCamera,
 } from "@";
 import { AxesLayer } from "@/layers/axes_layer";
+import { PanZoomControls } from "@/objects/cameras/controls";
 
 const url =
   "https://public.czbiohub.org/royerlab/ultrack/multi-color/image.zarr/";
@@ -42,49 +43,13 @@ layerManager.add(axes);
 
 let camera: PerspectiveCamera | OrthographicCamera = perspectiveCam;
 
-// TODO: would be nice to provide these functions in the library
-const clientToWorld = (clientPos: vec2) => {
-  const clipPos = renderer.clientToClip(clientPos);
-  const d = camera == perspectiveCam ? clipDistance : 0;
-  return camera.clipToWorld(vec3.fromValues(clipPos[0], clipPos[1], d));
-};
+const orthoCamControls = new PanZoomControls(orthoCam);
+const perspectiveCamControls = new PanZoomControls(
+  perspectiveCam,
+  clipDistance
+);
 
-document.addEventListener("wheel", (event) => {
-  const clientPos = vec2.fromValues(event.clientX, event.clientY);
-  const preZoomPos = clientToWorld(clientPos);
-  if (event.deltaY < 0) {
-    camera.zoom *= 1.05;
-  } else {
-    camera.zoom /= 1.05;
-  }
-  // pan to zoom in on the mouse position
-  const postZoomPos = clientToWorld(clientPos);
-  const dWorld = vec3.sub(vec3.create(), postZoomPos, preZoomPos);
-  camera.pan(vec3.scale(vec3.create(), dWorld, -1));
-  console.log(camera);
-});
-
-document.addEventListener("mousedown", (event) => {
-  const clientStart = vec2.fromValues(event.clientX, event.clientY);
-  let worldStart = clientToWorld(clientStart);
-  console.log(clientStart, worldStart);
-
-  const onMouseMove = (event: MouseEvent) => {
-    const clientPos = vec2.fromValues(event.clientX, event.clientY);
-    const worldPos = clientToWorld(clientPos);
-    const dWorld = vec3.sub(vec3.create(), worldPos, worldStart);
-    camera.pan(vec3.scale(vec3.create(), dWorld, -1));
-    worldStart = worldPos;
-  };
-
-  const onMouseUp = () => {
-    document.removeEventListener("mousemove", onMouseMove);
-    document.removeEventListener("mouseup", onMouseUp);
-  };
-
-  document.addEventListener("mousemove", onMouseMove);
-  document.addEventListener("mouseup", onMouseUp);
-});
+renderer.setControls(perspectiveCamControls);
 
 function animate() {
   renderer.render(layerManager, camera);
@@ -99,6 +64,9 @@ document.addEventListener("keydown", (event) => {
 
 function toggleCamera() {
   camera = camera === orthoCam ? perspectiveCam : orthoCam;
+  renderer.setControls(
+    camera === orthoCam ? orthoCamControls : perspectiveCamControls
+  );
 }
 
 animate();
--- a/examples/image2d_from_omezarr5d_u8/main.ts
+++ b/examples/image2d_from_omezarr5d_u8/main.ts
@@ -15,9 +15,8 @@ const url =
 const layerManager = new LayerManager();
 const renderer = new WebGLRenderer("#canvas");
 const orthoCam = new OrthographicCamera(-2000, 2000, -2000, 2000);
-const perspectiveCam = new PerspectiveCamera(60, 1, 0.1, 10000);
 const cameraPos = vec3.fromValues(0, 0, 5000);
-perspectiveCam.transform.translate(cameraPos);
+const perspectiveCam = new PerspectiveCamera({ fov: 60, position: cameraPos });
 // project the (negative) camera position to clip space
 // to get the distance to the image plane (z = 0)
 const projectedCameraPos = vec3.transformMat4(
--- a/examples/image2d_from_omezarr5d_u8/main.ts
+++ b/examples/image2d_from_omezarr5d_u8/main.ts
@@ -17,14 +17,6 @@ const renderer = new WebGLRenderer("#canvas");
 const orthoCam = new OrthographicCamera(-2000, 2000, -2000, 2000);
 const cameraPos = vec3.fromValues(0, 0, 5000);
 const perspectiveCam = new PerspectiveCamera({ fov: 60, position: cameraPos });
-// project the (negative) camera position to clip space
-// to get the distance to the image plane (z = 0)
-const projectedCameraPos = vec3.transformMat4(
-  vec3.create(),
-  vec3.scale(vec3.create(), cameraPos, -1),
-  perspectiveCam.projectionMatrix
-);
-const clipDistance = projectedCameraPos[2];
 
 // Source is technically 5D (even though Z is unitary),
 // so provide indices at 3 dimensions to project to 2D.
@@ -43,10 +35,7 @@ layerManager.add(axes);
 let camera: PerspectiveCamera | OrthographicCamera = perspectiveCam;
 
 const orthoCamControls = new PanZoomControls(orthoCam);
-const perspectiveCamControls = new PanZoomControls(
-  perspectiveCam,
-  clipDistance
-);
+const perspectiveCamControls = new PanZoomControls(perspectiveCam);
 
 renderer.setControls(perspectiveCamControls);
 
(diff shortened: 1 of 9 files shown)


src/objects/cameras/orthographic_camera.ts
```diff
@@ -2,6 +2,7 @@ import { Camera } from "./camera";
 import { mat4 } from "gl-matrix";
 
 export class OrthographicCamera extends Camera {
+  // TODO: store width and height instead?
```
andy-sweet: I'm in favor because it makes this camera more consistent with `PerspectiveCamera` and it also removes the confusion about top/bottom when/if we choose to use a coordinate system where (0, 1, 0) points down.

src/objects/cameras/perspective_camera.ts
```diff
@@ -32,9 +32,12 @@ export class PerspectiveCamera extends Camera {
   }
 
   protected updateProjectionMatrix() {
+    // clamp the field of view and zoom to prevent degenerate behavior
+    const fov = Math.max(0.1, Math.min(179.9, this.fov_ / this.zoom_));
```
andy-sweet, line 60: It's a little weird that the clamping happens here and not when setting zoom or field-of-view.

Regardless, it probably makes sense to define an effective fov so the client has access to this information too.

examples/image2d_from_omezarr5d_u8/main.ts
```diff
@@ -1,17 +1,30 @@
...
 
 const url =
   "https://public.czbiohub.org/royerlab/ultrack/multi-color/image.zarr/";
 const layerManager = new LayerManager();
 const renderer = new WebGLRenderer("#canvas");
-const camera = new OrthographicCamera(-2000, 2000, -2000, 2000);
+const orthoCam = new OrthographicCamera(-2000, 2000, -2000, 2000);
+const perspectiveCam = new PerspectiveCamera(60, 1, 0.1, 10000);
+const cameraPos = vec3.fromValues(0, 0, 5000);
+perspectiveCam.transform.translate(cameraPos);
```
andy-sweet: It would be ideal to add a `position` argument that defaults to `(0, 0, 0)`.

Especially nice if we use structured keyword arguments for the camera constructors.

examples/image2d_from_omezarr5d_u8/main.ts
```diff
@@ -22,13 +35,72 @@ const region = [
...
+toggleCamera();
+
+// TODO: would be nice to provide these functions in the library
+const clientToWorld = (clientPos: vec2) => {
+  const clipPos = renderer.clientToClip(clientPos);
+  const d = camera == perspectiveCam ? clipDistance : 0;
+  return camera.clipToWorld(clipPos, d);
+};
+
+document.addEventListener("wheel", (event) => {
```
andy-sweet: At a minimum, it would be nice to move some of these callbacks and functions to a separate lib/helper module to help keep the example relatively concise and readable. But agreed with the PR description sentiment, that some/most of these belong in the visualization library as helpers.

examples/image2d_from_omezarr5d_u8/main.ts
```diff
@@ -22,13 +35,72 @@ const region = [
...
 
+document.addEventListener("keydown", (event) => {
+  if (event.key === " ") {
+    toggleCamera();
+  }
+});
+
+function toggleCamera() {
+  camera = camera === orthoCam ? perspectiveCam : orthoCam;
+  const width = camera.type == "PerspectiveCamera" ? 20 : 0.01;
```
andy-sweet: Why does the width need to be different for each camera? Is that due to the way the projected line's shader program? Or is there a simpler explanation?

Without this need, you would need to recreate the `AxesLayer` and reset the layers right? Alternatively, we could just add a setter for `AxesLayer.width`.

aganders3: Good catch - I had written this off as the line shrinking as it got farther from the camera (the PerspectiveCamera is much farther from the line) but that doesn't make sense because the line is projected in screen space.

I think it's a bug in the line shader and the offset needs to be multiplied by `currPos.w`. It's a tiny fix so I'll make an issue and fix here if that's okay with you.

src/objects/cameras/perspective_camera.ts
aganders3: The clamping is here such that we don't have to override the `zoom` setter from the base class, and right now FOV is private. But I agree this is weird. It's kind of weird to keep track of both FOV and zoom separately given they have the same effect.

I'll give it some thought but suggestions are welcome.

I like the idea of an effective FOV property.

aganders3: I think this also means adding a "effective" width and height properties to `OrthographicCamera`. Also I suppose a way to get the effective frame.

src/core/renderer.ts
```diff
@@ -85,4 +86,12 @@ export abstract class Renderer {
     }
     return this.activeCamera_;
   }
+
+  public clientToClip(position: vec2): vec2 {
```
andy-sweet: I would not expect clip-space to be 2D, so this feels like where I might provide a depth parameter to get into 3D.

src/objects/cameras/perspective_camera.ts
andy-sweet: > I think this also means adding a "effective" width and height properties to OrthographicCamera. Also I suppose a way to get the effective frame.

Agreed. I guess the reason many libraries with cameras define zoom separately is that there's some notion of a special fixed field of view, that is temporarily zoomed in on, and to which you might want to reset to.

We might also have that, but it's likelier trickier to handle with multiple layers. Therefore, I'd be interested to try to remove zoom as a parameter and just directly mutate the width/height or fov (not necessarily in this PR though).

src/core/renderer.ts
aganders3: I agree it makes sense for clip-space to be 3D, but the input is client-space (2D). What space would the depth parameter be? Would it just be assumed to already be clip-space and passed through to a `vec3` output?

I do think it makes sense to return a `vec3`. Long-term it would make sense to use picking, but until them I'm not sure what to set it to. Some ideas:
* pass through an input
* set it to 0
* estimate from the position of the `activeCamera`

`clipToWorld` below should take in a `vec3` instead (and also return one).

andy-sweet: Based on the implementation this clip-space seems like OpenGL's where x/y/z are each in [-1, 1] (i.e. GL normalized device coordinates). 

So given that and  that this is a 2->3D transform, I like requiring the user to pass a clip-space value for the depth and maybe defaulting to 0 as the mid-point of [-1, 1].

Do the z=-1 and z=1 in clip-space effectively correspond to the near and far planes of the camera? Or is my understanding not quite right?

aganders3: > So given that and that this is a 2->3D transform, I like requiring the user to pass a clip-space value for the depth and maybe defaulting to 0 as the mid-point of [-1, 1].

Right now this depth is passed to `clipToWorld`, which makes more sense to me, but I also don't mind moving it here. More generally I think `clientToClip` should perhaps be private (protected) but that may depend on how a `CameraControls` object works. I have some open questions about that we can discuss tomorrow.

> Do the z=-1 and z=1 in clip-space effectively correspond to the near and far planes of the camera? Or is my understanding not quite right?

I think this is correct. It's not linear between the planes in the case of the perspective camera, but I think the limits -1 and 1.

aganders3:
I think this is ready for a look. Thanks for the comments so far.

I moved the camera controls to their own class in https://github.com/chanzuckerberg/imaging-active-learning/pull/96/commits/2f7f7e6a5d579701fed8c4bc00095cdf88136c7d but I am open to reverting that for purposes of keeping the PR a bit less intrusive. This has some oddities as we discussed yesterday, mostly due to `render.canvas` being private and the "active" camera switching.

src/renderers/shaders/projected_line_vert.glsl
```diff
@@ -46,7 +46,7 @@ void main() {
         0.0,
         0.0
     );
-    gl_Position = currPos + offset;
+    gl_Position = currPos + offset * currPos.w;
```
andy-sweet, line 49: Is this related to the other changes this PR? Or did this PR help you debug what this is fixing?

src/core/renderer.ts
```diff
@@ -42,6 +46,32 @@ export abstract class Renderer {
...
+  }
+
+  private unbindControls() {
+    this.controlCallbacks_.forEach(([event, listener]) => {
+      this.canvas.removeEventListener(event, listener);
+    });
+  }
+
+  private bindControls() {
+    if (!this.controls_) {
```
andy-sweet: Optional

I'd consider inlining this and `unbindControls` to avoid the need for the error handling.

src/core/renderer.ts
```diff
@@ -85,4 +115,13 @@ export abstract class Renderer {
     }
     return this.activeCamera_;
   }
+
+  public clientToClip(position: vec2, depth: number = 0): vec3 {
+    const [x, y] = position;
+    return vec3.fromValues(
+      (2 * x) / this.canvas.clientWidth - 1,
+      1 - (2 * y) / this.canvas.clientHeight,
```
andy-sweet, line 121: I guess this inversion might need to be removed after merging https://github.com/chanzuckerberg/imaging-active-learning/pull/109?

I'm planning to wait to merge that after this goes in, so I can debug more easily.

examples/image2d_from_omezarr5d_u8/main.ts
```diff
@@ -1,17 +1,30 @@
...
 
 const url =
   "https://public.czbiohub.org/royerlab/ultrack/multi-color/image.zarr/";
 const layerManager = new LayerManager();
 const renderer = new WebGLRenderer("#canvas");
-const camera = new OrthographicCamera(-2000, 2000, -2000, 2000);
+const orthoCam = new OrthographicCamera(-2000, 2000, -2000, 2000);
+const cameraPos = vec3.fromValues(0, 0, 5000);
+const perspectiveCam = new PerspectiveCamera({ fov: 60, position: cameraPos });
+// project the (negative) camera position to clip space
```
andy-sweet: Does this need to be recalculated every time the camera position changes? Or do we assume that the z-position cannot change?

Also, could this be calculated on construction of the controls? It feels out of place in the example itself.

src/core/renderer.ts
```diff
@@ -42,6 +46,32 @@ export abstract class Renderer {
...
+  public setControls(controls: CameraControls) {
+    this.unbindControls();
+    this.controls_ = controls;
+    this.bindControls();
+  }
+
+  private unbindControls() {
+    this.controlCallbacks_.forEach(([event, listener]) => {
+      this.canvas.removeEventListener(event, listener);
+    });
```
andy-sweet, line 58: Should unbind also set `controlCallbacks_` to `[]`?

src/objects/cameras/controls.ts
```diff
@@ -0,0 +1,92 @@
...
+    const clientPos = vec2.fromValues(event.clientX, event.clientY);
+    const preZoomPos = clientToWorld(clientPos, this.depth_);
+    if (event.deltaY < 0) {
+      this.camera_.zoom *= 1.05;
+    } else {
+      this.camera_.zoom /= 1.05;
+    }
+    // pan to zoom in on the mouse position
+    const postZoomPos = clientToWorld(clientPos, this.depth_);
+    const dWorld = vec3.sub(vec3.create(), postZoomPos, preZoomPos);
```
andy-sweet: Consider renaming `dWorld` to `deltaWorld` if that's what's intended.

Also consider, swapping the arguments of the subtraction here instead of scaling by -1 below.

src/objects/cameras/controls.ts
```diff
@@ -0,0 +1,92 @@
...
+    this.camera_.pan(vec3.scale(vec3.create(), dWorld, -1));
+  }
+
+  private mousedown(
+    event: MouseEvent,
+    target: EventTarget,
+    clientToWorld: ClientToWorld
+  ): void {
+    console.log("mousedown", this.camera_.type);
```
andy-sweet: Should these be `debug` instead of `log`?

src/objects/cameras/controls.ts
```diff
@@ -0,0 +1,92 @@
+import { vec2, vec3 } from "gl-matrix";
+import { Camera } from "./camera";
+
+type ClientToClip = (clientPos: vec2, depth: number) => vec3;
```
andy-sweet, line 4: Just to double check that these are the same type now, but we want to allow them to diverge in the future?

andy-sweet: If not, this could be `clientTo3D` or similar?

src/objects/cameras/camera.ts
```diff
@@ -19,4 +20,31 @@ export abstract class Camera extends RenderableObject {
...
+    this.zoom_ = zoom;
+    this.updateProjectionMatrix();
+  }
+
+  public pan(vec: vec3) {
+    this.transform.translate(vec);
+  }
+
+  public clipToWorld(position: vec3): vec3 {
+    const screenPos = vec4.fromValues(position[0], position[1], position[2], 1);
```
andy-sweet: Why `screenPos` and not `clipPos`?

src/objects/cameras/perspective_camera.ts
```diff
@@ -31,10 +47,21 @@ export class PerspectiveCamera extends Camera {
...
+  public get fov() {
+    return this.fov_;
+  }
+
+  public get effectiveFov() {
+    return this.fov_ / this.zoom_;
+  }
+
   protected updateProjectionMatrix() {
+    // clamp the field of view and zoom to prevent degenerate behavior
```
andy-sweet, line 59: I also get some degenerate behavior for the orthographic camera if I zoom in far enough, though it is recoverable. Should we consider a similar clamping operation there? Although, I guess there's no way to reasonably clamp those values (other than as some multiple of the minimum absolute value of the precision).

andy-sweet (changes requested):
Generally works and looks great, but I had enough questions and suggestions along the way, where I can't approve this quite yet. Should be able to do so soon though!

src/renderers/shaders/projected_line_vert.glsl
aganders3: This fixes the issue raised in https://github.com/chanzuckerberg/imaging-active-learning/pull/96#discussion_r1815436015.

This PR helped debug by having multiple cameras in the same example, but it's otherwise unrelated and could be moved into a separate PR.

examples/image2d_from_omezarr5d_u8/main.ts
aganders3: Yeah, I agree it's out of place here and it would needs to be recalculated if the position changed. Maybe a good solution for now would be to give `PerspectiveCamera` a `target` point and the depth can be calculated from the distance from `position` to `target`?

src/objects/cameras/controls.ts
aganders3: I'm not sure they need to diverge. They happen to have the same signature but different meaning.

aganders3:
> Generally works and looks great, but I had enough questions and suggestions along the way, where I can't approve this quite yet. Should be able to do so soon though!

No worries - thanks! Some good suggestions from having fresh eyes on it.

src/objects/cameras/controls.ts
aganders3: Yes, or perhaps even removed.

src/core/renderer.ts
aganders3: Having them separate may be useful if we want also add `enableControls` and `disableControls`.

Another thought is to have the controls default to something like `NullControls` and avoid any error checking here. That would make `this.controls_` non-nullable, and fits with the concept of "nothing is something" in dependency injection.

src/objects/cameras/perspective_camera.ts
aganders3: Even with the perspective camera it's easy to lose the image if you zoom all the way out. Maybe the cameras could have configurable zoom limits somehow? I'm not sure what the best solution here is. I'll see if we can copy behavior from three.js or something.

andy-sweet: We can follow up in another PR

src/core/renderer.ts
andy-sweet: Enable/disable is a good point here, so I'm OK leaving as is. I also like `NullControls`, so feel free to implement that if you want.

src/objects/cameras/perspective_camera.ts
```diff
@@ -1,25 +1,48 @@
...
+  near?: number;
+  far?: number;
+  position?: vec3;
+  target?: vec3;
+};
+
 export class PerspectiveCamera extends Camera {
   private fov_: number;
   private aspectRatio_: number;
+  private position_: vec3;
```
andy-sweet: When is this used? And how is it different from `Camera.transform.translation_`?

src/objects/cameras/perspective_camera.ts
aganders3: For what it's worth three.js handles these limits in the controls:

https://github.com/mrdoob/three.js/blob/3728a0aafd5ee04a9a4bc66eb3f56f556bd8da15/examples/jsm/controls/OrbitControls.js#L60-L66

src/objects/cameras/perspective_camera.ts
aganders3: It's used to get the clip depth in the `PanZoomControls`. It's not different from `Camera.transform.translation_`, so I can remove the property but keep the setter/getter.

aganders3: `Camera.transform.translation_` is currently private, but I can add a getter or use     `mat4.getTranslation(vec3.create(), this.transform.matrix)`.

src/objects/cameras/perspective_camera.ts
```diff
@@ -1,25 +1,48 @@
...
+  private target_: vec3;
+
+  constructor(options: PerspectiveCameraOptions = {}) {
+    const {
+      fov = DEFAULT_FOV,
+      aspectRatio = DEFAULT_ASPECT_RATIO,
+      near = 0.1,
+      far = 10000,
+      position = vec3.create(),
```
andy-sweet: Based on code inspection, I don't think `create` is guaranteed to zero out the values, so use `vec3.zero(vec3.create())` or `vec3.fromValues(0, 0, 0)` instead.

src/objects/cameras/perspective_camera.ts
andy-sweet: Ah, I see it's in controls. But I'd still expect to be able to use `Camera.transform.translation_`.

src/objects/cameras/perspective_camera.ts
aganders3: Oh thanks - I had looked this up for `mat4` and saw it creates an identity, so I had just assumed this would make a zero vector!

aganders3: Is it confusing to have `position` and `target` default to the same point? Maybe `target` should be halfway between the near and far plane or something.

src/objects/cameras/perspective_camera.ts
aganders3: I agree this makes sense - just need to also override `pan` to move the target as well.

The more I'm working with this I think direct access to the camera transform could cause problems, but I am okay working that out later.

src/objects/cameras/perspective_camera.ts
```diff
@@ -1,25 +1,48 @@
...
+  far?: number;
+  position?: vec3;
+  target?: vec3;
+};
+
 export class PerspectiveCamera extends Camera {
   private fov_: number;
   private aspectRatio_: number;
+  private position_: vec3;
+  private target_: vec3;
```
andy-sweet: `target` is state I'd expect on controls, but not on a camera. Though `Camera.lookAt` seems fine - it's the extra state that's unexpected and undesirable. I think there's something a bit off about the `PanZoomControls` for controlling a perspective camera, so I don't want to block the PR based on this. We can always follow up by implementing more suitable controls for a perspective camera (e.g. arcball, orbit).

aganders3: > target is state I'd expect on controls, but not on a camera

This is interesting. The camera has a `position` and `direction`, but the `distance` is the part that is useful for the controls, and not for the camera itself. I'm not sure `target` is very intuitive on controls either in that case. Especially in the case of panning, where the position and target move together. I need to think about it a little more.

I agree pan/zoom may not be a very practical method for controlling a perspective camera.

andy-sweet (approved):
I'm happy enough with the state of this, so approving. I'll let you decide if you want to make some changes related to position and target.

aganders3:
Great - I'll merge today with the knowledge there is plenty of remaining work on cameras, controls, and transforms. 😅

PR #98, opened 2024-10-24
Use a string literal union Answer instead of an enum

andy-sweet:
### Summary
This addresses some post-merge feedback in #93.

Most importantly, it uses a string literal union for the `Answer` type instead of an enum, which makes the related code simpler and more readable. That change also helped fix a bug where a previously answered question did not show the answer as a highlighted button.

### Related Issue
Related to #74

### Tests & Checks
I manually tested the prototype workflow.


shlomnissan (approved):
LGTM 🚀

Issue #95, opened 2024-10-23
Improve visualization of tracks


aganders3:
Some proposed ideas for improving track visualization:
* Modify the layer as the current time index changes (make the tracks "crawl")
* Interpolate the paths to reduce line artifacts, especially with tight corners
* Use spline interpolation to make the lines smoother
* Encode time on the path (color and/or width tapering)
* Display the current time point location as an additional `Point`
* Visualize the "split" for dividing cells (e.g. dotted lines from the end of the parent tracklet to the start of the children)

aganders3:
I would include updating `ProjectedLineLayer` to be `TracksLayer` in this issue.

aganders3:
Some improvements are definitely still possible, but I think this is "done" for now as all these have been implemented:
> * Modify the layer as the current time index changes (make the tracks "crawl")
> * Interpolate the paths to reduce line artifacts, especially with tight corners
> * Use spline interpolation to make the lines smoother
> * Encode time on the path (~~color and/or~~ width tapering)
> * updating ProjectedLineLayer to be TracksLayer in this issue.

I tested this, but have not included it because implementation was a bit messy:
> * Visualize the "split" for dividing cells (e.g. dotted lines from the end of the parent tracklet to the start of the children)

PR #107, opened 2024-11-01
Initial refactor ProjectedLineLayer -> TracksLayer

aganders3:
### Summary
This adds a `TracksLayer` that provides visualization more specific for tracks, still building on `ProjectedLine` renderable object.

The additional features here are:
* tapering the line at both ends (future and past)
* setting the taper offset (allows the tracks to "crawl" along the track)
* bezier interpolation of the track paths (makes them look smoother, avoids join artifacts)

### Related Issue
#95

### Tests & Checks
Manually tested - here is a demo video. In this demo (example) I've included two "intermediate" tracks connecting the endpoint of the first track to the beginning of the next two.

https://github.com/user-attachments/assets/f9802fbe-eab5-4f49-82a5-911edade290d


examples/tracks/main.ts
```diff
@@ -21,24 +21,56 @@ const trackAPath: vec3[] = [
...
   [1844.0, 1440 - 1378.0, 0.0],
 ];
-const lineLayer = new ProjectedLineLayer([
-  { path: trackAPath, color: [1.0, 0.0, 0.0], width: 0.01 },
-  { path: trackBPath, color: [0.0, 1.0, 0.0], width: 0.01 },
-  { path: trackCPath, color: [0.0, 0.0, 1.0], width: 0.01 },
+const trackCTime = [34, 35, 36, 37, 38];
+const interpolation = { pointsPerSegment: 10, tangentFactor: 0.3 };
+
+// TODO: intermediate paths should be handled by the TracksLayer
```
andy-sweet: Are you planning to do this as part of this PR or not?

examples/tracks/main.ts
```diff
@@ -81,8 +114,10 @@ imageSeriesLayer.onStateChange((newState: LayerState) => {
     slider.addEventListener("input", (event) => {
       const value = (event.target as HTMLInputElement).valueAsNumber;
       imageSeriesLayer.setTimeIndex(value);
+      lineLayer.setTimeIndex(value);
```
andy-sweet, line 98: When I tested this, the imaging data was not being served, so I modified this example to only add the tracks layer. This meant I also had to add an event listener just for the line layer.

In this example, the image layer is acting like the driving layer because it is not immediately ready where the tracks layer is. Alternatively, we could just update each layer separately, but this has the disadvantage of potentially showing data at different time points.

I'm not sure I have a question here, but I think this example motivates defining the logic of indexing multiple layers in the visualization library. I'm not sure if we should always wait for all layers to be ready or have some slightly more flexible logic.

src/renderers/shaders/projected_line_vert.glsl
```diff
@@ -1,14 +1,18 @@
 #version 300 es
 
+const float PI = 3.14159265;
+
 layout (location = 0) in vec3 inPosition;
 layout (location = 3) in vec3 inPrevPosition;
 layout (location = 4) in vec3 inNextPosition;
 layout (location = 5) in float direction;
+layout (location = 6) in float distance;
```
andy-sweet: I'm not sure how GLSL name scopes work, but it might be a good idea to avoid colliding with the `distance` function: https://registry.khronos.org/OpenGL-Refpages/gl4/html/distance.xhtml

Looking at how the geometry value is calculated, maybe `proportion` could be an option?

src/renderers/shaders/projected_line_vert.glsl
```diff
@@ -39,10 +43,17 @@ void main() {
         diff = normalize(prevDiff) + normalize(nextDiff);
     }
 
+    // direction is + or -, but also encodes distance on path
```
andy-sweet: Is this comment out of date with the current implementation?

Thinking about it, why not encode the distance in direction?

andy-sweet (approved):
Looks good to me even though it's a draft, so feel free to merge once you are ready.

I did have a couple a few questions and suggestions that may be worth addressing though.

examples/tracks/main.ts
aganders3: I don't think so. I was even thinking to just remove these intermediate paths from the example. If we want this in the prototype it can come in a separate PR. I need to think a bit about the best way to represent the relationships in the constructor, for example.

src/renderers/shaders/projected_line_vert.glsl
aganders3: This is out of date. I did initially have this encoded in direction, but changed it because this is more semantic and it causes trouble with point 0. I was also thinking perhaps the `distance` should instead be `time`. Otherwise the issue is it treats each line as the same "length" so the taper could be a different length. I actually think this looks okay but eventually perhaps should be an option of some kind.

src/renderers/shaders/projected_line_vert.glsl
aganders3: Good call on collision with `distance`. I went with `path_proportion` for now.

examples/tracks/main.ts
aganders3: This is a good observation. I created #110 from this comment to track it. I think it could be useful to have some method to synchronize certain layer attributes...or something.

examples/tracks/main.ts
aganders3: I removed these intermediate paths here. I think they are helpful but can be added later. I'm going to keep #95 open for now, and the idea is at least recorded there.

andy-sweet (approved):
Looks good to me still!

Issue #106, opened 2024-10-25
Sync answers with HITL server


aganders3:
Task answers should sync to the server. Previously we discussed each answer having a uuid and timestamp, such that a task could be answered multiple times. It will be the responsibility of the prioritization model to take multiple answers into account (or not).

It would be good to have a UI indication showing the sync progress/status.

PR #108, opened 2024-11-01
Connect client to HITL server

aganders3:
### Summary
This makes the client (prototype frontend) talk to the HITL server (prototype backend).

This is WIP so feel free to take a look, but I'd like to split it up a bit to make it easier to review and merge.

What's in here now:
* added `tasks` table to the HITL server DB
* added `answers` table to the HITL server DB
* added an `answer` endpoint to the HITL server
* changed `Answer` in FE to a type that contains a UUID, timestamp, and sync status
* added effects to sync answers with the server
* icon color in `TaskList` reflects sync status
* control whether using hard-coded data (and mock `answer` endpoint) via `VITE_MOCK_ULTRACK` env var (allows developing FE without running BE)
* make the HITL server very CORS-permissive (required if FE and BE are hosted on different origin)
* sort the tracks by ID within a task - this makes every division `red -> (blue, green)` but I'm not sure it's completely correct (that is, can a parent have an ID larger than its child?). Maybe we just need to make sure parent is first.

TODO (will make an issue):
Currently you get a new UUID for each `Answer` when the page loads. Answers can then be overwritten if you go back to a given task and re-answer, it will overwrite the previous answer in the DB. If the page is reloaded or refreshed, you will get new UUIDs and can re-submit _new_ answers. It's up to the server to do what it wants with multiple answers to one task. An improvement may be adding a concept of a user or session ID (no need to be secure), such that answers can be changed within the same session, and perhaps sessions could be resumed

### Related Issues
#105
#106

### Tests & Checks
Manual testing

### DIFF
--- a/ultrack-prototype/backend/src/ultrack_learns/db.py
+++ b/ultrack-prototype/backend/src/ultrack_learns/db.py
@@ -1,9 +1,19 @@
 import os
 from contextlib import contextmanager
 
-from ultrack_learns.models import TrackData
+from ultrack_learns.models import TaskData, TaskType, TrackData
 
-from sqlalchemy import create_engine, select, Column, Integer, Float
+from sqlalchemy import (
+    create_engine,
+    select,
+    Column,
+    DateTime,
+    Integer,
+    Float,
+    ForeignKey,
+    String,
+    UUID,
+)
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.orm import sessionmaker, Session
 
@@ -18,11 +28,14 @@ SessionLocal = sessionmaker(autocommit=False, autoflush=False, bind=engine)
 
 
 def get_session():
-    db = SessionLocal()
-    try:
-        yield db
-    finally:
-        db.close()
+    with SessionLocal() as session:
+        try:
+            yield session
+            session.commit()
+        except Exception:
+            # TODO: log the exception, but don't expose it
+            session.rollback()
+            raise
 
 
 session_context = contextmanager(get_session)
@@ -41,6 +54,39 @@ class TrackPoint(Base):
     x = Column(Float)
 
 
+class Task(Base):
+    __tablename__ = "tasks"
+
+    id = Column(Integer, primary_key=True, index=True)
+    task_id = Column(UUID, index=True, unique=True)
+    # these can be used to query track_points to get the task_data
+    # this is much simpler than storing this relationship directly in the DB
+    # and allows more flexibility (e.g. changing the time window)
+    task_type = Column(String, nullable=False)
+    node_id = Column(Integer, ForeignKey("track_points.id"))
+
+    def get_task_data(self, db: Session, time_window: int = 10) -> TaskData:
+        return TaskData(
+            node_id=self.node_id,
+            tracks_data=track_points_around_node(
+                db,
+                self.node_id,
+                time_window=time_window,
+                include_children=self.task_type == TaskType.DIVISION,
+            ),
+        )
+
+
+class Answer(Base):
+    __tablename__ = "answers"
+
+    id = Column(Integer, primary_key=True, index=True)
+    answer_id = Column(UUID, index=True, unique=True)
+    task_id = Column(UUID, ForeignKey("tasks.task_id"))
+    created_at = Column(DateTime(timezone=True))
+    answer = Column(String, nullable=False)
+
+
 def track_points_around_node(
     db: Session,
     node_id: int,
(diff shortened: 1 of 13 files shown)


ultrack-prototype/backend/src/ultrack_learns/db.py
```diff
@@ -41,6 +54,39 @@ class TrackPoint(Base):
...
+                include_children=self.task_type == TaskType.DIVISION,
+            ),
+        )
+
+
+class Answer(Base):
+    __tablename__ = "answers"
+
+    id = Column(Integer, primary_key=True, index=True)
+    answer_id = Column(UUID, index=True, unique=True)
```
aganders3: We don't _really_ need both of these columns

ultrack-prototype/backend/src/ultrack_learns/mock_data.py
```diff
@@ -136,14 +137,17 @@ def _get_divisions(db: Session, n_samples: int):
 def all_tasks(
     rng_seed: int = 42,
     num_tasks: int = 10,
+    time_window: int = 16,
```
aganders3, line 140: `time_window` was added as a parameter here for the hacky caching on this function.

The slightly wider time window looks better to me, but this is otherwise unrelated.

ultrack-prototype/backend/src/ultrack_learns/db.py
```diff
@@ -82,7 +128,7 @@ def track_points_around_node(
                 if tp.track_id == track_id
             ],
         )
-        for track_id in track_ids
+        for track_id in sorted(track_ids)
```
aganders3, line 131: Sorting here ensures the tracks in a division task come in a consistent order.

ultrack-prototype/frontend/lib/data.ts
aganders3, line 1: `data.ts` and `mock_data.ts` are meant to implement the same functions. The idea here is to allow development of the FE without the BE running. `data.ts` is the _real_ functionality, though for now it's still fetching from a "mock" endpoint on the server.

ultrack-prototype/frontend/lib/tasks.ts
```diff
@@ -126,16 +149,21 @@ export class Task {
...
+      this.taskId,
+      this.taskType,
+      {
+        nodeId: this.taskData.nodeId,
+        tracksData: this.taskData.tracksData.map((track) => ({
+          trackId: track.trackId,
+          time: [...track.time],
+          position: track.position.map((pos) => [
+            ...pos,
+          ]) as typeof track.position,
```
andy-sweet, line 162: This is not new, by why isn't this just the following?

```suggestion
          position: track.position,
```

andy-sweet: Oh, sorry this is `clone`. Forget it.

ultrack-prototype/frontend/components/App.tsx
```diff
@@ -38,8 +60,45 @@ export default function App() {
...
+          if (task.answer.synced === "not_synced") {
+            const newTask = task.clone();
+            newTask.answer.synced = "pending";
+            return newTask;
+          }
+          return task;
+        })
+      );
+    };
+    setPending();
```
andy-sweet, line 84: Given that you only call this arrow function here, and it's not async, is there a reason not to inline it? Do you think it helps readability?

andy-sweet (approved):
Looks good to me. There are lots of TODOs, some of which might be higher priority than others. But nothing blocking for me. I was able to use the real and mock server.

ultrack-prototype/frontend/components/App.tsx
```diff
@@ -38,8 +60,45 @@ export default function App() {
     );
   };
 
-  const task = tasks[taskIndex] ?? null;
-  console.debug(`App::taskIndex: ${taskIndex}/${tasks.length}, task:`, task);
+  useEffect(() => {
```
andy-sweet, line 63: I generally found this effect a little hard to reason about. Partly because it's quite long. But maybe more importantly because it's async, so I could not easily think clearly about `tasks`, `synced`, and `prevTasks` and if there was a chance for interleaving that state in a bad way - I don't think there is and I also don't have great ideas for improvement.

aganders3: I'm _pretty sure_ there's not a problem, but I agree it's confusing. My only idea to improve it is to abstract it out to a custom hook. Something like `const [tasks, setTasks] = useSyncingList<Task>(getEndpoint, putEndpoint)`.

In a real app perhaps this is where something like [TanStack Query](https://tanstack.com/query) (formerly React-Query) would help?

aganders3:
Thanks! I think I'll make a tracking issue to prioritize and keep track of the outstanding TODOs here.

ultrack-prototype/frontend/components/App.tsx
aganders3: I can go either way - I thought it helped readability but probably not a lot, and it does have some overhead. I guess it's basically serving the role of a comment.

Looking now the part I don't like is re-running the map with the same test as above. I think it may help to refactor and use `answersToSync` instead.

I think part of the problem is this pattern we need to use in `setTasks` that takes a very similar callback each time to map/clone the list.

PR #109, opened 2024-11-05
Make the positive y-axis go down the screen

### LINKED ISSUE #88: Render content to the canvas such that y-coordinates draw down the screen

andy-sweet:
### Summary
This adds an extra transform to the projection to make the y-axis go down to the screen. This approach makes the library opinionated about its coordinate system instead of following WebGL's. Importantly, it cares about two things.

1. The vector (0, 1, 0) points to the bottom of the screen (with a default camera orientation).
2. The top-left pixel of a 2D image at depth 0 should start at (0, 0, 0).

Each renderer backend must specify this extra transform to go from our coordinate system to its own.

### Related Issue
Closes #88

### Tests & Checks
I manually tested all the existing examples and prototype app to check that they were working as before.


aganders3 (approved):
This makes sense to me and I like how it cleans up some of the code. I'm not totally sold on using a LH coordinate system (Neuroglancer is RH for what it's worth) but I think we can adjust that via the same mechanism, right?

The most confusing part to me is the reversal of "bottom" and "top" nomenclature in the orthographic camera, but I think it's an acceptable tradeoff and something we can change later if necessary.

The additional matrix multiply on render is a little unfortunate but I'm sure it can be optimized if it becomes a problem.

andy-sweet:
> This makes sense to me and I like how it cleans up some of the code. I'm not totally sold on using a LH coordinate system (Neuroglancer is RH for what it's worth) but I think we can adjust that via the same mechanism, right?

Yes. I actually like the idea of using WebGPU's coordinate system why is right-handed, but with y going down and z going into the screen. Will leave that change for later though.

> The most confusing part to me is the reversal of "bottom" and "top" nomenclature in the orthographic camera, but I think it's an acceptable tradeoff and something we can change later if necessary.

Agreed. But this should be resolved if we re-factor to use width and height. Alternatively, if we want to allow specification of the frame in world coordinates, then this could be `yMin` and `yMax` to reduce confusion about the camera's orientation.

PR #112, opened 2024-11-12
Use pan-zoom controls in tracks example and prototype

andy-sweet:
### Summary
This uses the recently added `PanZoomControls` in the tracks layer example and in the HITL prototype app.

### Tests & Checks
I manually tested this by using the tracks layer example and the HITL prototype in its standard workflow.

### DIFF
--- a/ultrack-prototype/frontend/components/Renderer.tsx
+++ b/ultrack-prototype/frontend/components/Renderer.tsx
@@ -71,9 +71,9 @@ export default function Renderer(props: RendererProps) {
         layerManager.layers.length = 0;
         layerManager.add(tracksLayer);
         layerManager.add(imageSeriesLayer);
-        const { xMin, xMax, yMin, yMax } = task.camera(2.0);
-        camera.setFrame(xMin, xMax, yMax, yMin);
-        camera.update();
+        const extent = tracksLayer.extent;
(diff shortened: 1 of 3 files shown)


ultrack-prototype/frontend/components/Renderer.tsx
```diff
@@ -64,9 +66,10 @@ export default function Renderer(props: RendererProps) {
         layerManager.layers.length = 0;
         layerManager.add(tracksLayer);
         layerManager.add(imageSeriesLayer);
-        // TODO: update the camera in-place instead of creating a new one
-        // (this will make zoom/pan callbacks easier to manage)
-        camera = task.camera(2.0);
+        const { xMin, xMax, yMin, yMax } = task.camera(2.0);
+        camera.setFrame(xMin, xMax, yMax, yMin);
+        camera.update();
```
aganders3: I think this can be changed to just use `tracksLayer.extent` and `camera.zoom = 0.25` (or similar) and we can get rid of `task.camera`.

aganders3 (approved):
Looks good - thanks!

I had added a `useRef` here for the `renderer` to hack this to work, but I prefer this approach for now as it's a smaller change.

ultrack-prototype/frontend/components/Renderer.tsx
andy-sweet: Ah, good point. Done.

Issue #34, opened 2024-09-16
Support loading and rendering many OME-Zarr chunks to many textures


andy-sweet:
In #28, we added support for loading uint16 image data from OME-Zarr sources. This uses zarrita's `get` to fetch multiple chunks and wait for all of them to arrive before creating a texture from all of them.

This has a few limitations.

- All the chunks may be bigger than the maximum support 2D texture size. We could downsample, but that is lossy and also requires non-trivial client-side compute.
- We have to wait for all the chunks to arrive before rendering anything.

This ultimate goal of this issue is to progressively load and render 2D chunks of OME-Zarr data to remove both of these limitations. (e.g. using zarrita's `getChunk`). This requires us to split up chunk fetching into separate tasks and also return the region of the chunk, so that it can be used to update the appropriate region of a texture.

This task is potentially large, so will likely need to be split up further.

We may also not need it for the prototype, so may not want to start this immediately.

andy-sweet:
Closing this issue in favor of the ongoing work around chunk streaming because that already meets the goals described here.

Issue #14, opened 2024-08-21
Technical overview of data loading


andy-sweet:
The goal of this issue is to write a technical overview of what data loading means in the context of the viewer component. This covers the span of reading data from persistent storage and transferring it to some memory for rendering.

The document will focus on libraries, API design, and implementation patterns. The document should not be a strict specification of the viewer's data loading API, but should help inform that. It should also distinguish separate components within data loading that could be worked on independently.

aganders3:
I think this is the relevant document?

https://docs.google.com/document/d/1b8NhTejl7LIKUO0g4tCuIgo0_Ygk2oeQcJjKs_vDDLM/edit?tab=t.0#heading=h.wx60qweyuyn9

I just wanted a link here to make it easier to  find.

PR #113, opened 2024-11-15
[WIP] Progressive loading of image data

andy-sweet:
### Summary
This contains some very rough and early progress on progressive loading of chunked image data sources.

Here is an example of progressive tiled image loading done the wrong way by using a separate mesh for each tile.

https://github.com/user-attachments/assets/575a1057-3591-429d-b7d0-bb549305640c

Notice there are significant artifacts at the boundaries of the tiles. Instead of this approach, we should use a single mesh and populate the texture data progressively instead (currently a work-in-progress).

### Related Issue
#14, #34, #35

### Tests & Checks
I manually tested my changes on the relevant existing and newly added examples.


aganders3:
> Notice there are significant artifacts at the boundaries of the tiles. Instead of this approach, we should use a single mesh and populate the texture data progressively instead (currently a work-in-progress).

I'm surprised by these artifacts. I think this is how Neuroglancer and the new Allen Institute vis library work. Maybe something else is going on here? I'm not necessarily advocating we use this approach.

andy-sweet:
Closing this in favor of more recent and better developed efforts: https://docs.google.com/document/d/1b1vHIZb01m3Pc9YUt6IpEtQz_aV_hUeHq1jq6VQPy7M

PR #117, opened 2024-11-21
Add instructions and helper script for running the prototype on Bruno

### LINKED ISSUE #114: Deploy Ultrack Prototype on Bruno and document process
Our goal for this quarter was to deploy the prototype application on Bruno (BioHub HPC).

I have access to Bruno and will attempt to run it there. I don't think this should be too difficult, but it's important for us to document the process and/or do some light automation (i.e. a launch script).

Below are some CLI incantations @JoOkuma suggested may be helpful for launching jobs and opening tunnels.

Launch an interactive job session (`srun` runs a job via slurm):
```
srun --pty --partition=interactive --cpus-per-task=1 --mem=20G bash -l
```

Create a tunnel to localhost by jumping through a login node to the job node (`gpu-sm01-19` in this case):
```
ssh -J $USER@login01 -L 24000:localhost:4000 $USER@gpu-sm01-19
```

aganders3:
### Summary
This adds a helper script and some instructions for running the prototype app on Bruno (BioHub HPC). This was our stated goal deployment state for Q4 2024.

Hopefully most of this is pretty clear. I was happy to see no necessary code changes. I tried to include as much setup as possible in the slurm script, so it sets up node version manager, node, and uv.

Future improvements include:
* First check for node version instead of blindly installing it
* Run production versions of the servers
* Document process for _users_ of the app, and look into simplifying this - right now one person needs to launch the slurm job, and others need to set up the SSH tunnel to connect to the servers.
* Consider splitting jobs for FE + BE servers - FE will be very cheap, but BE might be co-located with compute and shorter lived. Alternatively the BE could kick off its own additional slurm jobs (still probably good to split FE + BE).

### Related Issue
Closes #114 

### Tests & Checks
With this configuration I can interact with the application from my laptop, and later inspect the SQLite file on Bruno to confirm the answers have synced properly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,9 +86,9 @@ directly with the backend from your computer.
 
 Now you should be able to access the prototype application on
 `http://localhost:5173`. Likewise the backend server is accessible on
-`http://localhost:8000` as if you were running it locally. Other uses can
+`http://localhost:8000` as if you were running it locally. Other users can
 access the same running instance by just opening the tunneling SSH session.
 
-`run-full-stack.slurm` is also just a bash script, so you can also use it to
+`run-full-stack.slurm` is just a bash script, so you can also use it to
 run both front- and back-end servers locally.
 
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ VITE_MOCK_ULTRACK=true npm run dev -- --mode prototype
 
 The backend server for the prototype is in the `ultrack-prototype/backend` directory. This is a
 Python FastAPI server that serves the data and provides an API for the frontend to interact with.
-See the README in that directory for more information on installing and running the server.
+See `ultrack-prototype/backend/README.md` for more information on installing and running the server.
 
 ### Running on Bruno
 
@@ -70,7 +70,7 @@ provided script in `ultrack-prototype/run-full-stack.slurm`:
 Submitted batch job 1234567  # job ID will be unique
 ❯ squeue --user $USER
 JOBID PARTITION     NAME     USER ST       TIME  NODES NODELIST(REASON)
-17117490       cpu ultrack- ashley.a  R      28:23      1 cpu-c-1
+1234567       cpu ultrack- your.name  R      28:23      1 cpu-c-1
 ```
 
 Take the node ID from the output of `squeue` (`cpu-c-1` in this case) and open
--- a/README.md
+++ b/README.md
@@ -54,3 +54,41 @@ VITE_MOCK_ULTRACK=true npm run dev -- --mode prototype
 The backend server for the prototype is in the `ultrack-prototype/backend` directory. This is a
 Python FastAPI server that serves the data and provides an API for the frontend to interact with.
 See the README in that directory for more information on installing and running the server.
+
+### Running on Bruno
+
+First, clone the repo into your home directory on Bruno.
+
+The full-stack prototype can be run as a slurm batch job on Bruno with the
+provided script in `ultrack-prototype/run-full-stack.slurm`:
+
+```shell
+❯ ssh $USER@login01.czbiohub.org
+❯ # ...you will need to confirm login using Duo 2FA
+❯ cd src/imaging-active-learning  # or whatever your local path to the repo
+❯ sbatch ./ultrack-prototype/run-full-stack.slurm
+Submitted batch job 1234567  # job ID will be unique
+❯ squeue --user $USER
+JOBID PARTITION     NAME     USER ST       TIME  NODES NODELIST(REASON)
+17117490       cpu ultrack- ashley.a  R      28:23      1 cpu-c-1
+```
+
+Take the node ID from the output of `squeue` (`cpu-c-1` in this case) and open
+a second SSH connection (for example in a new terminal) to set up local tunnels
+to the FE and BE servers. Both servers need to be mapped because the frontend
+is a purely client-side application, and it needs to be able to communicate
+directly with the backend from your computer.
+
+```shell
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -53,4 +53,42 @@ VITE_MOCK_ULTRACK=true npm run dev -- --mode prototype
...
+directly with the backend from your computer.
+
+```shell
+❯ ssh -L 5173:localhost:5173 -L 8000:localhost:8000 -J $USER@login01.czbiohub.org $USER@cpu-c-1
+❯ # ...again you will need to confirm login using Duo 2FA
+```
+
+Now you should be able to access the prototype application on
+`http://localhost:5173`. Likewise the backend server is accessible on
+`http://localhost:8000` as if you were running it locally. Other uses can
```
andy-sweet:
```suggestion
`http://localhost:8000` as if you were running it locally. Other users can
```

README.md
```diff
@@ -53,4 +53,42 @@ VITE_MOCK_ULTRACK=true npm run dev -- --mode prototype
...
+access the same running instance by just opening the tunneling SSH session.
+
+`run-full-stack.slurm` is also just a bash script, so you can also use it to
```
andy-sweet: Nit: remove repeated `also`

```suggestion
`run-full-stack.slurm` is just a bash script, so you can also use it to
```

ultrack-prototype/run-full-stack.slurm
```diff
@@ -0,0 +1,52 @@
...
+REPO_ROOT=$(git rev-parse --show-toplevel)
+
+if command -v uv &> /dev/null; then
+    echo "uv already installed..."
+else
+    echo "installing uv..."
+    curl -LsSf https://astral.sh/uv/install.sh | sh
+fi
+
+if [ -s "NVM_DIR/nvm.sh" ]; then
```
andy-sweet:
```suggestion
if [ -s "$NVM_DIR/nvm.sh" ]; then
```

ultrack-prototype/backend/README.md
```diff
@@ -49,7 +49,7 @@ options:
 
 ```
 
-Otherwise, mock data is avaialable via the development server at `http://localhost:8000/mock-data`.
+Otherwise, mock data is avaialable via the development server at `http://localhost:8000/mock_data`.
```
andy-sweet:
```suggestion
Otherwise, mock data is available via the development server at `http://localhost:8000/mock_data`.
```

ultrack-prototype/run-full-stack.slurm
```diff
@@ -0,0 +1,52 @@
...
+    echo "nvm already installed..."
+else
+    echo "installing nvm..."
+    curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.40.1/install.sh | bash
+fi
+[ -s "$NVM_DIR/nvm.sh" ] && source "$NVM_DIR/nvm.sh"
```
andy-sweet, line 35: Even if that script exports the `NVM_DIR` environment variable, I don't think it's possible for the `bash` subprocess to mutate's this calling process' environment variables.

The simplest fix is just to define that environment variable here

```suggestion
export NVM_DIR="$([ -z "${XDG_CONFIG_HOME-}" ] && printf %s "${HOME}/.nvm" || printf %s "${XDG_CONFIG_HOME}/nvm")"
[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh" # This loads nvm
```

as given in the nvm docs: https://github.com/nvm-sh/nvm?tab=readme-ov-file#install--update-script

Alternatively, we could assume that `NVM_DIR` is `$HOME/.nvm`, which it should be by default.

ultrack-prototype/run-full-stack.slurm
andy-sweet: Actually, this won't work because if $NVM_DIR is not defined, the script will fail. Why not just use `command` as above?

```suggestion
if command -v nvm; then
```

andy-sweet (approved):
Looks good to me once the minor install issues with the slurm shell script are fixed. After that, I was able to use the front-end and back-end served from a small compute node by following the instructions in the readme. Thanks for setting this up!

ultrack-prototype/run-full-stack.slurm
aganders3: This is what I tried first. I think `nvm` is a function defined by sourcing `nvm.sh`, so if it hasn't been sourced it can't be found, even though `nvm` is installed. I believe running via `sbatch` does not use a login shell by default, so often this would be the case.

I think you're right that defining `NVM_DIR` first makes sense. The invocation from the nvm docs is a good find for that! We could set if not set:

```shell
export NVM_DIR="${NVM_DIR:-$([ -z "${XDG_CONFIG_HOME-}" ] && printf %s "${HOME}/.nvm" || printf %s "${XDG_CONFIG_HOME}/nvm")}"
```
or since that's pretty ugly, something like:
```shell
DEFAULT_NVM_DIR="$([ -z "${XDG_CONFIG_HOME-}" ] && printf %s "${HOME}/.nvm" || printf %s "${XDG_CONFIG_HOME}/nvm")"
export NVM_DIR=${NVM_DIR:-$DEFAULT_NVM_DIR}
```

aganders3:
Thanks for debugging and catching my typos! I'll make some tweaks and try to test again by first uninstalling `uv` and `nvm` in my account.

It's also worth noting these are both still running in dev mode. I think this is a good enough start, but I can look into packaging and/or creating a Docker image. With a little more effort we could host both servers on the same port to make the tunnel setup simpler.

aganders3:
I updated the script and tested fresh by uninstalling `uv` and `nvm` in my account. This should be good enough for now!

PR #119, opened 2024-11-22
Include image source URL and region information in task data

### LINKED ISSUE #103: Include data source URL in task data from server
> I would have assumed that the source would have been included in the task data. I guess it's always the same now, so it doesn't really matter, but I'd expect `OmeZarrImageSource` to be created in this module or in `mock_data.ts` instead of `Renderer.tsx`.

_Originally posted by @andy-sweet in https://github.com/chanzuckerberg/imaging-active-learning/pull/92#discussion_r1816915000_

andy-sweet:
### Summary
This adds the imaging data to each task in the database schema, along with the corresponding mock data. This allows the hard-coded imaging metadata to be removed from the application itself. While the image URL, time dimension, and slice indices are constant for the current mock data, these changes should support tasks from other image sources in the future. 

### Related Issue
Closes #103

### Tests & Checks
I ran the following commands from the root of this repository

```shell
cd ultrack-prototype/backend
rm ultrack_tmp.db
uv run write-mock-data-json
mv mock_data.json ../data
uv run fastapi dev src/ultrack_learns/server.py
```

and in a different terminal from the root of this repository
```shell
npm run format --fix  // to format the new JSON file
npm run dev -- --mode prototype
```

I then manually went through the basic prototype workflow successfully.

I also tested with the static mock data by running

```shell
VITE_MOCK_ULTRACK=true npm run dev -- --mode prototype
```

and checking that the prototype provided a consistent experience.


ultrack-prototype/backend/src/ultrack_learns/models.py
```diff
@@ -22,9 +22,16 @@ class TrackData(BaseModel):
     position: Path2D | Path3D
 
 
+class ImageData(BaseModel):
+    url: str = "https://public.czbiohub.org/royerlab/ultrack/multi-color/image.zarr/"
```
andy-sweet: Now might be a good time to use the `normalized.zarr` image, though I know there are issues with that, so I lean towards doing it in a different PR.

ultrack-prototype/frontend/lib/tasks.ts
```diff
@@ -47,17 +48,66 @@ function trackFromJSON(trackJSON: unknown): Track {
   return { trackId: track_id, time, position };
 }
 
+type ImageData = {
+  url: string;
+  timeDimension: string;
+  sliceIndices: Map<string, number>;
+};
+
+function imageFromJSON(imageJSON: unknown): ImageData {
```
aganders3: nit: I might name this `imageDataFromJSON`

ultrack-prototype/backend/src/ultrack_learns/models.py
```diff
@@ -22,9 +22,16 @@ class TrackData(BaseModel):
...
+    time_dimension: str
+    slice_indices: dict[str, int]
+
+
 class TaskData(BaseModel):
     node_id: int
     tracks_data: list[TrackData]
+    image_data: ImageData
```
aganders3, line 35: Bringing my comment from Slack for visibility/durability:

I think we should also store this image data in the database here (or perhaps in a separate images table referenced by each task with a key):

https://github.com/chanzuckerberg/imaging-active-learning/blob/fd083328e1dcb5b8283d685552af58e832c4330e/ultrack-prototype/backend/src/ultrack_learns/db.py#L57

This could come in a follow-up PR though. It doesn't really matter as long as it's still hard-coded to be the same for all tasks. This wouldn't affect the application at all currently, but it would be useful when saving off the DB it creates.

ultrack-prototype/backend/src/ultrack_learns/models.py
aganders3: I agree it would be nice to change to `normalized.zarr` ASAP, but I don't think it's ready without re-chunking.

Can the default be removed here? It's duplicated in ultrack-prototype/backend/src/ultrack_learns/mock_data.py where I think it makes more sense to keep it.

ultrack-prototype/frontend/lib/tasks.ts
```diff
@@ -202,15 +255,16 @@ export class Task {
...
   }
 
-  imageSeriesLayer(
-    source: OmeZarrImageSource,
-    preLoad = true
-  ): ImageSeriesLayer {
-    const region = [
-      { dimension: "T", index: this.timeInterval },
-      { dimension: "Z", index: 0 },
+  imageSeriesLayer(preLoad = true): ImageSeriesLayer {
```
aganders3, line 258: Now that this no longer takes the `OmeZarrImageSource` I think we can cache it. Currently switching back to a previous task relies on the browser cache to speed up re-fetching and is still pretty slow for me.

This could be somewhat large though so maybe I'd hold off. Any thoughts?

ultrack-prototype/frontend/lib/tasks.ts
```diff
@@ -237,12 +291,12 @@ export class Task {
...
+  public layers(): {
     imageSeriesLayer: ImageSeriesLayer;
     tracksLayer: TracksLayer;
   } {
     const layers = {
-      imageSeriesLayer: this.imageSeriesLayer(imageSource),
+      imageSeriesLayer: this.imageSeriesLayer(),
       tracksLayer: this.tracksLayer(),
     };
     this.tracksLayer_ = layers.tracksLayer;
```
aganders3:
```suggestion
```
This is left over from me, but I think this line is redundant since it's done in the `tracksLayer()` call above.

aganders3 (approved):
Thanks for taking this on! I have some fairly minor suggestions/ideas but this is already a big improvement so I'm happy to merge and save further improvements for later.

aganders3:
> I ran the following commands from the root of this repository
> ```
> cd ultrack-prototype/backend
> rm ultrack_tmp.db
> uv run write-mock-data-json
> mv mock_data.json ../data
> uv run fastapi dev src/ultrack_learns/server.py
> ```

This should probably be monitored in CI to ensure it's in sync, but I don't think it's worth the effort right now.

ultrack-prototype/backend/src/ultrack_learns/models.py
andy-sweet: > Can the default be removed here? It's duplicated in ultrack-prototype/backend/src/ultrack_learns/mock_data.py where I think it makes more sense to keep it.

Agreed. Done.

ultrack-prototype/frontend/lib/tasks.ts
andy-sweet: We could cache the instance of `OmeZarrImageSource` based on the URL, but I'm unsure it will have much of a difference on performance of the usage you described because it only fetches the OME-NGFF JSON metadata, while the `ImageSeriesLayer` (and other attributes in `ImageData`) fetches the chunks that correspond to particular time points.

So I'll probably leave it for now in this PR, unless I misunderstood something.

This does motivate having some somewhat-general-purpose chunk cache/store though.

ultrack-prototype/frontend/lib/tasks.ts
andy-sweet: Removed.

ultrack-prototype/frontend/lib/tasks.ts
andy-sweet: Done. 

I also considered keeping this name and changing the local type name to `Image`, but that clashes with a built-in.

ultrack-prototype/frontend/lib/tasks.ts
aganders3: Sorry, I was not clear. I meant to cache the whole `ImageSeriesLayer` for the task. This came up in a [review comment on #92](https://github.com/chanzuckerberg/imaging-active-learning/pull/92/files#r1816910624) because we cache the `TracksLayer` already.

I still think this can be pushed to another PR because I think it needs some cache eviction consideration. It also may be related to pre-fetching upcoming tasks, so could be considered at the same time.

ultrack-prototype/backend/src/ultrack_learns/models.py
andy-sweet: Done. Let me know if this is what you were thinking - it feels a little arbitrary when there is only one image, but agreed it's nice for consistency.

ultrack-prototype/frontend/lib/tasks.ts
andy-sweet: The memory usage of a single layer in the prototype is not trivial (i.e. a few hundred MB), so this needs a little care, so leaving it for a separate PR.

PR #120, opened 2024-11-25
Add keyboard shortcuts for answering questions

### LINKED ISSUE #118: Add keyboard shortcuts for choosing answer to current question
This has come up a few times and is also a feature in the napari plugin. The shortcuts can be hard-coded and should ideally be ergonomic for all users (e.g. using 1, 2, 3 which should be close regardless of keyboard layout). We could also just use whatever the napari plugin uses if compatible with existing shortcuts.

andy-sweet:
### Summary
This adds hard-coded keyboard shortcuts for selecting an answer to the current question. The shortcut keys are consistent with Ultrack's napari plugin, and should be ergonomic for most keyboard layouts.

### Related Issue
Closes #118

### Tests & Checks
I manually tested my changes by running through the standard workflow on the prototype.

### DIFF
--- a/ultrack-prototype/frontend/components/Question.tsx
+++ b/ultrack-prototype/frontend/components/Question.tsx
@@ -8,6 +8,9 @@ type Answer = {
   shortcut: string;
 };
 
+// Shortcuts "1", "2", and "3" are used to be ergonomic on most
+// keyboard layouts and to be consistent with Ultrack's napari plugin:
+// https://github.com/royerlab/ultrack/blob/ff3bf242e395bde3236153317802b5aff8dbb6a6/ultrack/validation/link_validation.py#L159
 const answers: Answer[] = [
   { type: "Yes", shortcut: "1" },
   { type: "No", shortcut: "2" },
--- a/ultrack-prototype/frontend/components/Question.tsx
+++ b/ultrack-prototype/frontend/components/Question.tsx
@@ -1,8 +1,18 @@
 import { Button } from "@czi-sds/components";
 import { Box, Typography } from "@mui/material";
 import { AnswerType, Task } from "../lib/tasks";
+import { useEffect } from "react";
 
-const answers: AnswerType[] = ["Yes", "No", "Uncertain"];
+type Answer = {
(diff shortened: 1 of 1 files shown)


ultrack-prototype/frontend/components/Question.tsx
```diff
@@ -1,8 +1,21 @@
 import { Button } from "@czi-sds/components";
 import { Box, Typography } from "@mui/material";
 import { AnswerType, Task } from "../lib/tasks";
+import { useEffect } from "react";
 
-const answers: AnswerType[] = ["Yes", "No", "Uncertain"];
+type Answer = {
+  type: AnswerType;
+  shortcut: string;
+};
```
aganders3: nit: an `Answer` type is also exported from `tasks.ts`, just not imported here. I don't think the collision is a big problem but this could be `AnswerShortcut` or something.

ultrack-prototype/frontend/components/Question.tsx
```diff
@@ -13,6 +26,21 @@ export type QuestionProps = {
...
+  useEffect(() => {
+    const selectAnswer = (event: KeyboardEvent) => {
+      const answer = answers.find(
+        (answer: Answer) => answer.shortcut === event.key
+      );
+      if (answer !== undefined) {
+        setTaskAnswer(answer.type);
+      }
+    };
+    document.addEventListener("keydown", selectAnswer);
```
aganders3, line 39: This is more broad and we shouldn't change it here, but I'd be curious to look up or ask someone if there are accessibility rules/conventions about which element should register the listener.

aganders3 (approved):
Looks good and works well for me. Thanks!

ultrack-prototype/frontend/components/Question.tsx
andy-sweet: I used `AnswerOption` to avoid the double naming of `AnswerShortcut.shortcut`.

andy-sweet:
I also added a check against `disabled` to prevent the shortcut from doing something until the `Question` is fully presented and enabled.

PR #122, opened 2024-11-27
[WIP] Concurrent fetching with limited resources

andy-sweet:
### Summary
Using the `normalized.zarr` sample image can cause resource issues, that are likely related to the number of concurrent requests.

This PR addresses those issues by using an executor that limits the number of concurrent OME-Zarr chunk-loading tasks. That executor also allows those tasks to be cancelled, which the app does when swapping layers.

The OME-Zarr loader also caches any previously opened arrays, so that the array does not need to be re-opened for each call to `loadChunks`.

In the future, we likely want to limit the number of concurrent requests across all layers.

This is a WIP because the task executor needs some unit tests and I still need to consider the best way to introduce cancellation logic in the app.

### Tests & Checks
I manually tested my changes and I added some automated tests for the executor.


andy-sweet:
Closing in favor of #131 and #132

Issue #104, opened 2024-10-25
Pre-fetch data to optimize UX


aganders3:
We can start fetching data for the "next" task (for example) rather than waiting for the user to answer or advance.

PR #125, opened 2024-12-06
[WIP] Cache image series layer on task

### LINKED ISSUE #121: Final task is re-loaded after answering
This is a minor bug - the prototype will re-load the final task when it is answered. This is a result of the task advancing when answered.

andy-sweet:
### Summary
This caches the image series layer on the `Task` object and also passes all cached values to the new instance of the `Task` create in `Task.clone`.

In this case, the layers will still be re-added, but their state will already be `"ready"` so we need to explicitly check for that.

### Related Issue
Closes #121

### Tests & Checks
I manually tested the prototype in a workflow that reproduces the bug in #121.

### DIFF
--- a/ultrack-prototype/frontend/components/Renderer.tsx
+++ b/ultrack-prototype/frontend/components/Renderer.tsx
@@ -62,17 +62,23 @@ export default function Renderer(props: RendererProps) {
     imageSeriesLayer.update();
     setImageSeriesLayer(imageSeriesLayer);
     setTracksLayer(tracksLayer);
+    const onReady = () => {
+      setPlaybackEnabled(true);
+      // TODO: update the data on the layers instead of creating new ones
+      layerManager.layers.length = 0;
+      layerManager.add(imageSeriesLayer);
+      layerManager.add(tracksLayer);
+      const extent = tracksLayer.extent;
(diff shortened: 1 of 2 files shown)


andy-sweet:
This is a draft because I'm concerned about memory usage here. After moving through all the tasks, that Chrome tab was using about 2GB of RAM. Instead of caching layers for each task, we may need a global cache.

An alternative fix for #121 could be to check if everything but the answer changed (which does not affect what we render), in which case we could skip changing what is rendered.

andy-sweet:
I favor #126 to fix #121. This PR might be independently useful to avoid the small latency associated with reloading/rendering the layers from the browser cache.

aganders3:
> I favor https://github.com/chanzuckerberg/imaging-active-learning/pull/126 to fix https://github.com/chanzuckerberg/imaging-active-learning/issues/121.

I agree, but thanks for exploring both options!

> This PR might be independently useful to avoid the small latency associated with reloading/rendering the layers from the browser cache.

This still seems worth adding depending on complexity. I think we should have some way to bound the usage (for example the global cache you mention) before merging this.

aganders3:
I'm going to close this. If we want to reopen I think it should be in the context of also pre-loading the next task in the list (#104). Conversations with the BioHub team suggest re-visiting a previous task is not something we need to optimize UX for.

Issue #102, opened 2024-10-25
Consider useRef


aganders3:
> I don't understand when we should use a React ref for `camera` instead of the mutable module variable.  Do you have any thoughts?
> At first, it was surprising to me this works correctly in conjunction with the animation effect - but then I had to re-learn that arrow functions capture bindings and not values/references.

_Originally posted by @andy-sweet in https://github.com/chanzuckerberg/imaging-active-learning/pull/92#discussion_r1813097298_

I don't have a strong sense for this right now. `useRef` seems a bit safer and more idiomatic, maybe? What we have is working though. I think it makes a difference when you're using a component more than once, but we're not (especially not the `Renderer`).

PR #126, opened 2024-12-09
Skip reload when task ID is unchanged

### LINKED ISSUE #121: Final task is re-loaded after answering

andy-sweet:
### Summary
This skips reloading the layers when the task ID does not change, which prevents unnecessary reloads, renders, and loading state indicators. It also addresses the possible case when the image series layer loads and transitions to the ready state before effect finishes, which is another possible cause of the bug described in #121.

### Related Issue
Closes #121

### Tests & Checks
I manually tested the prototype in a workflow that reproduces the bug in #121.

### DIFF
--- a/ultrack-prototype/frontend/components/Renderer.tsx
+++ b/ultrack-prototype/frontend/components/Renderer.tsx
@@ -1,4 +1,4 @@
-import { Dispatch, SetStateAction, useEffect, useState } from "react";
+import { Dispatch, SetStateAction, useEffect, useRef, useState } from "react";
 import {
   ImageSeriesLayer,
   LayerManager,
@@ -20,7 +20,6 @@ const canvasId = "canvas";
 const camera = new OrthographicCamera(0, 1920, 0, 1440);
 const controls = new PanZoomControls(camera, camera.position);
 const layerManager = new LayerManager();
-let lastTaskId = "";
 
 type RendererProps = {
   curTime: number;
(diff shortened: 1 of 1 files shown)


ultrack-prototype/frontend/components/Renderer.tsx
```diff
@@ -54,25 +55,31 @@ export default function Renderer(props: RendererProps) {
...
+      // TODO: update the data on the layers instead of creating new ones
+      layerManager.layers.length = 0;
+      layerManager.add(imageSeriesLayer);
+      layerManager.add(tracksLayer);
+      const extent = tracksLayer.extent;
+      camera.setFrame(extent.xMin, extent.xMax, extent.yMax, extent.yMin);
+      camera.zoom = 0.25;
+      controls.panTarget = camera.position;
+    };
+    if (imageSeriesLayer.state == "ready") {
```
aganders3: I'm still a little confused by JS equality, but should this be `===`?

ultrack-prototype/frontend/components/Renderer.tsx
```diff
@@ -20,6 +20,7 @@ const canvasId = "canvas";
 const camera = new OrthographicCamera(0, 1920, 0, 1440);
 const controls = new PanZoomControls(camera, camera.position);
 const layerManager = new LayerManager();
+let lastTaskId = "";
```
aganders3: Should this be state private to this component? I think if it's module-level scope there could be issues if we end up with a page with two `Renderer` components. Similar to #102 I think it's not really a problem in this app and can be considered later.

aganders3 (approved):
I have a couple small comments, but this makes sense and fixed this issue for me. Thanks!

ultrack-prototype/frontend/components/Renderer.tsx
andy-sweet: Yes, it should. Thanks for the catch!

ultrack-prototype/frontend/components/Renderer.tsx
andy-sweet: Probably, yes. I think this is similar to #102, but I don't mind using `useRef` here to avoid it, so I've done that.

PR #127, opened 2024-12-09
Use bounding rect offset of canvas in client-to-clip transform

### LINKED ISSUE #124: Zoom doesn't follow mouse
This feels like possibly a CSS issue that the canvas is a different size than it appears, like it's extending under the other controls.

Zoom works as expected in the example that supports it.

andy-sweet:
### Summary
This fixes the zoom centering bug described in #124 by using the canvas element's [`getClientBoundingRect`](https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect) in the `Renderer.clientToClip` transform. This accounts for the more complex layout in the prototype app, which has components to the left of the canvas.

It also fixes a pan bug in the prototype app by removing controls on the dismount of the renderer component.

### Related Issue
Closes #124

### Tests & Checks
I manually tested my changes with the tracks example and the prototype.

### DIFF
--- a/src/core/renderer.ts
+++ b/src/core/renderer.ts
@@ -116,9 +116,10 @@ export abstract class Renderer {
 
   public clientToClip(position: vec2, depth: number = 0): vec3 {
     const [x, y] = position;
+    const rect = this.canvas.getBoundingClientRect();
     return vec3.fromValues(
-      (2 * x) / this.canvas.clientWidth - 1,
-      (2 * y) / this.canvas.clientHeight - 1,
+      (2 * (x - rect.x)) / this.canvas.clientWidth - 1,
+      (2 * (y - rect.y)) / this.canvas.clientHeight - 1,
       depth
     );
   }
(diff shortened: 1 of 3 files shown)


andy-sweet:
This is still a WIP because while this mostly fixes things, I still feel like there are small errors when very zoomed in and need to confirm if `offsetX/Y` is what we should be using (or maybe doing some calculations with `event.target.getClientBoundingRect` instead).

andy-sweet:
Another thought is to use `getClientBoundingRect` in `Renderer.clientToClip` since that is more consistent with the current naming.

aganders3:
The diagram from [this stackoverflow answer](https://stackoverflow.com/a/74946571) suggests `offsetX`/`offsetY` are what we want.
image

I tested these changes and zooming feels right to me. Pan/zoom still feels a little off, but it's much more usable.

aganders3:
Ah, pan/zoom is still using `client` - I'll try changing it.

aganders3 (approved):
Looks good - thanks! Nice debugging and I like the clean fixes.

Issue #58, opened 2024-09-27
Start leveraging destructured parameters


aganders3:
> Note: I think we could start leveraging destructured parameters. This would make it easier to understand what these values represent without having to navigate to the layer's signature.
> ```ts
> new SingleLineLayer({path, color: [0.0, 0.7, 0.0], width: 0.5});
> ```

_Originally posted by @shlomnissan in https://github.com/chanzuckerberg/imaging-active-learning/pull/46#discussion_r1775419144_

aganders3:
Related:

> * For consistency, let's stick with type for plain data structures.
> * We could use vec3[] for paths and potentially for colors too (though we might want to create a dedicated Color type at some point).

_Originally posted by @shlomnissan in https://github.com/chanzuckerberg/imaging-active-learning/pull/46#discussion_r1778931454_

PR #128, opened 2024-12-12
Destructure React component props inline

### LINKED ISSUE #99: Reconsider prop de-structuring in the prototype application
I don't see much value in destructuring props in this manner. Personally, I find it more beneficial to reference properties as `props.index`, particularly in larger components, as it clearly identifies input variables. If you're committed to destructuring props, consider doing it inline at the function's declaration.

_Originally posted by @shlomnissan in https://github.com/chanzuckerberg/imaging-active-learning/pull/93#discussion_r1815429926_

andy-sweet:
### Summary
This uses inline destructuring for the React component functions in the prototype application, which avoids the need for explicitly defining a props type and destructuring that on the first line of the component functions.

### Related Issue
Closes #99

### Tests & Checks
I manually tested the application using the standard workflow, and also relied on the linter/formatter to confirm these changes.

### DIFF
--- a/ultrack-prototype/frontend/components/PlaybackControls.tsx
+++ b/ultrack-prototype/frontend/components/PlaybackControls.tsx
@@ -6,15 +6,17 @@ import { Task } from "../lib/tasks";
 const playbackFPS = 8;
 const playbackIntervalMs = 1000 / playbackFPS;
 
-type PlaybackControlsProps = {
+export default function PlaybackControls({
+  curTime,
+  enabled,
+  setCurTime,
+  task,
+}: {
   curTime: number;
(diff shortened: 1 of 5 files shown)


andy-sweet:
We might also want to address #58 here, which would spread a similar pattern across the library to effectively allow for keyword arguments.

aganders3 (approved):
I don't really like how this looks, but I agree with your analysis of the tradeoffs in the issue. In the end I like keyword args, and I prefer consistency over any particular style. So I'm happy to merge this and reconsider later as-needed.

As for #58 I leave it up to you if you want to do it here or in a separate PR.

PR #129, opened 2024-12-12
Support system-dependent theme in prototype

### LINKED ISSUE #52: Add a background color property to the Renderer class

### LINKED ISSUE #100: Support dark mode in prototype
[SDS now supports dark mode](https://sds.czi.design/009eaf17b/v/0/p/169911-theming), but I don't really know how to use it.

I tried briefly to just set the theme to dark, but it didn't look right (background of components were still white, some text was still black).

andy-sweet:
### Summary
This adds a wrapper `ThemedApp` component to add a system-dependent dark/light theme. I created a wrapper app based on the [SDS docs](https://sds.czi.design/009eaf17b/v/0/p/169911-theming/b/36245e) and moved that to a separate module to satisfy the default requirements of the [fast refresh React plugin](https://github.com/ArnaudBarre/eslint-plugin-react-refresh).

I also made the background color of the renderer a mutable attribute. That's not necessary for this PR as I just use a default of (0, 0, 0, 0) to make it transparent so that it works well in the application in light and dark mode. But I think it seems generally useful.

I started making changes to the application to use a non-transparent background color that matches the light/dark theme, which might be needed to support certain types of blending. But those set of changes are much more complicated.

https://github.com/user-attachments/assets/3ea18aec-faa0-4cd2-8e74-977a9a6c70ca

### Related Issue
Closes #52
Closes #100 

### Tests & Checks
I manually tested the prototype in both system dark and light mode after these changes. I also tested the examples to ensure there were no regressions there either.

### DIFF
--- a/src/core/renderer.ts
+++ b/src/core/renderer.ts
@@ -6,10 +6,13 @@ import { PerspectiveCamera } from "objects/cameras/perspective_camera";
 import { OrthographicCamera } from "objects/cameras/orthographic_camera";
 import { CameraControls, NullControls } from "objects/cameras/controls";
 
+type Color = readonly [number, number, number, number];
+
 export abstract class Renderer {
   private readonly canvas_: HTMLCanvasElement | null;
   private width_ = 0;
   private height_ = 0;
+  private backgroundColor_: Color = [0, 0, 0, 0];
   private activeCamera_: Camera | null = null;
   private controls_: CameraControls = new NullControls();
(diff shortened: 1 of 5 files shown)


aganders3 (approved):
Looks good and works well for me! One nitpick is that the loading indicator probably doesn't need to change appearance because it's overlaid on the (generally dark) image, but it's still readable in both modes.

PR #131, opened 2024-12-23
Run limited number of concurrent promises to load image series chunks

andy-sweet:
### Summary
This allows an optional promise scheduler to be passed to `loadChunks`, so that the number of concurrent promises can be limited. This scheduler is used to implement a queue that can be passed through to zarrita's `Array.get` function, so that the number of underlying concurrent requests can be limited.

Allowing too many concurrent requests has been observed to cause unhandled errors for some datasets using Chrome. Since most browsers can actually handle a limited number of requests anyway, limiting the number to something like 16 should have no adverse performance effects, and also allows us to cancel pending tasks. Using an `AbortController` and passing its `AbortSignal` down to `Array.get` should also allow us to cancel requests that have already started.

### Tests & Checks
I manually tested my changes in the prototype and I added some automated tests for the scheduler.

### DIFF
--- a/src/layers/image_series_layer.ts
+++ b/src/layers/image_series_layer.ts
@@ -4,6 +4,7 @@ import { PlaneGeometry } from "objects/geometry/plane_geometry";
 import { Interval, Region } from "data/region";
 import { ImageChunk, ImageChunkSource } from "data/image_chunk";
 import { Texture2DArray } from "objects/textures/texture_2d_array";
+import { PromiseScheduler } from "@/data/promise_scheduler";
 
 // Loads 2D+t image data from an image source into renderable objects.
 export class ImageSeriesLayer extends Layer {
@@ -13,6 +14,7 @@ export class ImageSeriesLayer extends Layer {
   private readonly timeDimensionIndex_: number;
   private texture_: Texture2DArray | null = null;
   private dataChunks_: ImageChunk[] = [];
+  private scheduler_: PromiseScheduler = new PromiseScheduler(16);
 
   constructor(source: ImageChunkSource, region: Region, timeDimension: string) {
     super();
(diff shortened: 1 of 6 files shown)


andy-sweet:
@aganders3 : I separated this out from https://github.com/chanzuckerberg/imaging-active-learning/pull/122 to keep this PR relatively self-contained and more reviewable.

src/layers/image_series_layer.ts
```diff
@@ -13,6 +14,7 @@ export class ImageSeriesLayer extends Layer {
   private readonly timeDimensionIndex_: number;
   private texture_: Texture2DArray | null = null;
   private dataChunks_: ImageChunk[] = [];
+  private scheduler_: PromiseScheduler = new PromiseScheduler(16);
```
aganders3, line 17: For my understanding - this could eventually/potentially be a shared instance across loaders, right?

aganders3 (approved):
This looks good - thanks for being thorough and I appreciate the tests!

I still don't understand 100% of this, especially the parts around `shutdown` and the `AbortController` and what kind of control they give. Regardless, this seems like a good foundation and testing #132 shows it sufficiently solves our problems at the moment.

src/layers/image_series_layer.ts
andy-sweet: Yes, though we'd need to make sure the cancellation and shutdown logic works as intended for that purpose.

andy-sweet:
> especially the parts around `shutdown` and the `AbortController` and what kind of control they give.

I think that `signal` should be passed down through to `fetch` by zarrita via the [`RequestInit` type](https://developer.mozilla.org/en-US/docs/Web/API/RequestInit#signal). I'm not sure how eagerly that can/will abort a fetch, but it was easy enough to add.

aganders3:
Thanks - I think this means that the cancellation logic is not particularly granular. I'm trying to understand the potential applications. My guess is this would be helpful for something like deleting a layer, but not for example when zooming in (chunks moving out of the viewport).

andy-sweet:
> Thanks - I think this means that the cancellation logic is not particularly granular. I'm trying to understand the potential applications. My guess is this would be helpful for something like deleting a layer, but not for example when zooming in (chunks moving out of the viewport).

That's right. If we want something more granular, then we'll need a new `AbortController` for at least each task.

PR #132, opened 2024-12-23
Use the normalized zarr image in the prototype mock data

andy-sweet:
### Summary
This uses the `normalized` zarr image in the prototype mock data, which has much higher color contrast, so makes the tasks easier to complete. It also requires more data to be loaded for each task because of its chunking structure, so loading typically takes longer.

### Tests & Checks
I manually tested the prototype with the mock data.


aganders3 (approved):
I didn't test performance against main, but this works for me and the image is much nicer looking. Thanks!

PR #133, opened 2025-01-06
Adding pixi support

JoOkuma:
### Summary

If we decide to support OSX and Windows we will need to switch the build system from `uv` to `pixi` because some dependencies are mainly on available on `conda-forge`.

### Tests & Checks

I haven't tested if ultrack is still working with `pixi` I need to advance with the modeling, and I'll come back to this if we decide to support other platforms.


JoOkuma:
Hi, @andy-sweet,

This is ready for review. It works on my Linux desktop.

I added some dependencies I'm already using. 

Should I remove the `uv` lock file and their additional config on the `pyproject.toml`?

JoOkuma:
@andy-sweet I'm adding the other platforms, I forgot about them

JoOkuma:
@andy-sweet, done.

For documentation purposes, `pixi` searches `[tool.pixi.dependencies]` over conda, so I had to add them even though they are specified on `ultrack` because they were looking (and not finding) these compiled libraries on pypi.

ultrack-prototype/backend/.gitignore
```diff
@@ -165,3 +165,6 @@ cython_debug/
 # SQLite database files
 *.sqlite3
 *.db
+# pixi environments
+.pixi
```
andy-sweet: If you put this in the top level `.gitignore` then prettier also ignores this when formatting, which is desirable because running `npm run format` processes far fewer files. I'm not sure why that is the case - maybe because prettier only looks at the top-level `.gitignore`? Maybe @aganders3 can confirm?

As with other comments in this review, I think you should just replace the equivalent uv `.venv` currently there.

ultrack-prototype/backend/pyproject.toml
```diff
@@ -11,12 +11,35 @@ requires-python = ">=3.11"
...
+    "ultrack",
 ]
 
 [project.scripts]
 write-mock-data-json = "ultrack_learns.mock_data:write_mock_data_json"
 
+[project.optional-dependencies]
+dev = ["pandas>=2.2.3"]
+
 [tool.uv]
```
andy-sweet: I think you should just remove the `uv` mentions here and elsewhere, like you already did in the docs.

ultrack-prototype/backend/.gitignore
```diff
@@ -165,3 +165,6 @@ cython_debug/
...
+*.egg-info
```
andy-sweet: I think this is already handled above.

```suggestion
*.egg-info
```

ultrack-prototype/backend/pyproject.toml
```diff
@@ -11,12 +11,35 @@ requires-python = ">=3.11"
...
+
+[tool.pixi.dependencies]
+numba = "*"
+higra = "*"
+
+[tool.pixi.pypi-dependencies]
+ultrack-learns = { path = ".", editable = true }
+mip = ">=0.16.0rc0"  # ultrack clunky dependency
+
+[tool.pixi.tasks]
```
andy-sweet: Does this need to be present when empty?

andy-sweet (approved):
I was able to run the two commands in the docs with pixi on linux and on macOS (after installation), so I'm approving.

I think you should remove all the `uv` mentions and files. I'm not sure if that includes `.python-version` which I think is generated by `uv`.

ultrack-prototype/backend/.gitignore
aganders3: > I'm not sure why that is the case - maybe because prettier only looks at the top-level `.gitignore`?

I thought this was the case, but I just looked it up and it's slightly different:

> Prettier will also follow rules specified in the ".gitignore" file if it exists in the same directory from which it is run.

I think that has pretty much the same effect for us as CI runs the commands from the repo root.

ultrack-prototype/backend/pyproject.toml
JoOkuma: Removed

ultrack-prototype/backend/.gitignore
JoOkuma: Indeed, removed

ultrack-prototype/backend/pyproject.toml
JoOkuma: Removed

ultrack-prototype/backend/.gitignore
JoOkuma: Moved to root .gitignore

andy-sweet:
Still works for me after your changes. Please delete the `uv.lock` and, if it's not needed for pixel, the `.python-version`. Also update the PR description to remove the empty `Resolves issue` section. Then feel free to merge when ready!

JoOkuma:
Thanks, @andy-sweet; I forgot about these files.
 Done!

PR #139, opened 2025-01-10
Make region coordinates and geometry vertices consistent

### LINKED ISSUE #35: Use OME-Zarr metadata to define corner coordinates of plane

### LINKED ISSUE #38: Handle more than just a list of scales in OME-Zarr image loader
In #28, we added support for loading uint16 image data from OME-Zarr sources. One limitation of that implementation is that it assumes the coordinate transformations are all scale transformations defined by lists.

The goal of this task is to provide support beyond that. At a minimum, we should also handle a list of translation values. We may also want to handle [scales and translations defined by paths](https://ngff.openmicroscopy.org/latest/#trafo-md), though I've never seen that used in the wild.

andy-sweet:
### Summary
This makes the loading input region coordinates and the output geometry vertex coordinates consistent for image layers. This is demonstrated in the u16 example, which defines a sub-region in the x-y plane which is used for both loading and for the orthographic camera.

This also uses OME-Zarr dataset translation values if they are present and generally simplifies reading the dataset transformation values based on the restricted specification described in OME-NGFF.

### Related Issue
Closes #35
Closes #38

### Tests & Checks
I tested the development examples.

### DIFF
--- a/src/data/image_chunk.ts
+++ b/src/data/image_chunk.ts
@@ -15,13 +15,13 @@ export type ImageChunk = {
   rowStride: number;
   rowAlignmentBytes: TextureUnpackRowAlignment;
   scale: {
-    x: number,
-    y: number,
-  },
+    x: number;
+    y: number;
+  };
   offset: {
-    x: number,
-    y: number,
-  },
+    x: number;
+    y: number;
+  };
 };
 
 export type ImageChunkSource = {
--- a/src/data/image_chunk.ts
+++ b/src/data/image_chunk.ts
@@ -14,6 +14,8 @@ export type ImageChunk = {
   };
   rowStride: number;
   rowAlignmentBytes: TextureUnpackRowAlignment;
+  scale: [number, number];
+  offset: [number, number];
 };
 
 export type ImageChunkSource = {
--- a/src/data/image_chunk.ts
+++ b/src/data/image_chunk.ts
@@ -14,8 +14,14 @@ export type ImageChunk = {
   };
   rowStride: number;
   rowAlignmentBytes: TextureUnpackRowAlignment;
-  scale: [number, number];
-  offset: [number, number];
+  scale: {
+    x: number,
+    y: number,
+  },
+  offset: {
+    x: number,
+    y: number,
+  },
 };
 
 export type ImageChunkSource = {
--- a/src/data/image_chunk.ts
+++ b/src/data/image_chunk.ts
@@ -8,9 +8,9 @@ import { PromiseScheduler } from "./promise_scheduler";
 export type ImageChunk = {
   data: Uint8Array | Uint16Array;
   shape: {
-    width: number;
-    height: number;
-    channels: number;
+    x: number;
+    y: number;
+    c: number;
   };
   rowStride: number;
   rowAlignmentBytes: TextureUnpackRowAlignment;
--- a/src/layers/image_layer.ts
+++ b/src/layers/image_layer.ts
@@ -44,12 +44,12 @@ export class ImageLayer extends Layer {
     const shape = chunk.shape;
     const texture = new DataTexture2D(chunk.data, shape.width, shape.height);
     const plane = new PlaneGeometry(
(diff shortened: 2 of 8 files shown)


src/layers/image_layer.ts
```diff
@@ -42,8 +42,15 @@ export class ImageLayer extends Layer {
...
     const shape = chunk.shape;
-    const texture = new DataTexture2D(chunk.data, shape.width, shape.height);
-    const plane = new PlaneGeometry(shape.width, shape.height, 1, 1);
+    const texture = new DataTexture2D(chunk.data, shape.x, shape.y);
+    const plane = new PlaneGeometry(
+      chunk.scale.x * shape.x,
+      chunk.scale.y * shape.y,
+      1,
+      1,
+      chunk.offset.x,
```
andy-sweet: My other idea was to modify `Mesh.transform` with the chunk's scale and offset, but that did not have the desired effect for reasons I don't understand yet.

andy-sweet: Understood the issue now. Making the change described above gives `Mesh.transform` a non-identity scale and then this multiplies with the inverse camera object (not projection matrix) in the WebGL renderer implementation (which is used to define the position of the camera): https://github.com/chanzuckerberg/imaging-active-learning/blob/c0c1377a22475c3ac51c517b685761edceb7aa7f/src/renderers/webgl_renderer.ts#L50

Changing the order of multiplication to form the `ModelView` matrix fixes this, so I'm wondering if that should be made regardless. Looking at the mesh vertex shader, that order makes more sense to me because the vertices then get multiplied by the mesh object transform first and the camera object transform last.

@aganders3 : any thoughts?

aganders3: > Changing the order of multiplication to form the ModelView matrix fixes this, so I'm wondering if that should be made regardless. Looking at the mesh vertex shader, that order makes more sense to me because the vertices then get multiplied by the mesh object transform first and the camera object transform last.

This makes sense to me! I think it's just an oversight since models have just had the identity matrix so far, right?

src/data/image_chunk.ts
```diff
@@ -8,12 +8,20 @@ import { PromiseScheduler } from "./promise_scheduler";
 export type ImageChunk = {
   data: Uint8Array | Uint16Array;
   shape: {
-    width: number;
-    height: number;
-    channels: number;
```
aganders3, line 13: I'm not totally against this change, but I kind of appreciated the image chunk having explicit `width` and `height` to disambiguate from the `X` and `Y` dimensions in the source. Maybe this too becomes arbitrary as the camera movement becomes more sophisticated.

andy-sweet: The main reason I changed this was to achieve consistent naming across shape, scale, and offset. An alternative is to define a `transform: mat2d` (or similar) instead of `scale` and `offset`, though I'd need to a mechanism to use that in the renderable object's transform (e.g. by making `RenderableObject.transform` settable or by passing it through the constructor).

src/layers/image_layer.ts
```diff
@@ -53,7 +53,10 @@ export class ImageLayer extends Layer {
     texture.unpackRowLength = chunk.rowStride;
     texture.unpackAlignment = chunk.rowAlignmentBytes;
 
-    this.addObject(new Mesh(plane, texture));
+    const mesh = new Mesh(plane, texture);
```
andy-sweet: Unsure if I should also do this for the `ImageSeriesLayer`. I think I probably should and at least define a utility function to create a `Mesh` for an image.

aganders3 (approved):
This is a nice improvement to me. Thanks!

PR #148, opened 2025-01-22
Add support for loading and rendering float32 data

### LINKED ISSUE #146: Support loading and rendering float32 image chunks
The goal of this issue is to support loading float32 image data chunks (likely from OME-Zarr as our only existing source), then use the loaded chunks to populate a texture that is rendered.

We should at least add an example to test this works.

The implementation does not need to be particularly sophisticated and could just follow the skeleton in https://github.com/chanzuckerberg/imaging-active-learning/pull/138

andy-sweet:
### Summary
This adds support for loading a 32-bit floating point image chunk and using it as a texture. The definitions and logic follow the existing style and make no attempt to abstract the repetition and redundancy. It also adds an example that uses a CryoET OME-Zarr tomogram as a data source.

This PR does does not add support for the equivalent texture array, mostly because that would need another example and/or other tests.

### Related Issue
Closes #146

### Tests & Checks
I manually tested the new example to check consistency with napari (up to contrast limits/scaling), and also checked the existing examples continue to work as expected.

### DIFF
--- a/src/layers/image_utils.ts
+++ b/src/layers/image_utils.ts
@@ -21,6 +21,9 @@ function updateImageTexture(texture: Texture, chunk: ImageChunk) {
   texture.dataFormat = "red_integer";
   if (chunk.data instanceof Uint16Array) {
     texture.dataType = "unsigned_short";
+  } else if (chunk.data instanceof Float32Array) {
+    texture.dataType = "float";
+    texture.dataFormat = "red";
   }
   texture.unpackRowLength = chunk.rowStride;
   texture.unpackAlignment = chunk.rowAlignmentBytes;
--- a/src/layers/image_utils.ts
+++ b/src/layers/image_utils.ts
@@ -1,6 +1,6 @@
-import { DataTexture2D } from "@/objects/textures/data_texture_2d";
+import { DataTexture2D } from "objects/textures/data_texture_2d";
 import { Texture } from "objects/textures/texture";
-import { Texture2DArray } from "@/objects/textures/texture_2d_array";
+import { Texture2DArray } from "objects/textures/texture_2d_array";
 import { ImageChunk } from "data/image_chunk";
 import { PlaneGeometry } from "objects/geometry/plane_geometry";
 import { Mesh } from "objects/renderable/mesh";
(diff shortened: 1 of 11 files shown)


src/layers/image_utils.ts
```diff
@@ -21,6 +21,9 @@ function updateImageTexture(texture: Texture, chunk: ImageChunk) {
   texture.dataFormat = "red_integer";
   if (chunk.data instanceof Uint16Array) {
     texture.dataType = "unsigned_short";
+  } else if (chunk.data instanceof Float32Array) {
+    texture.dataType = "float";
+    texture.dataFormat = "red";
```
aganders3: Do we need to also set `dataType` as `unsigned_byte` for u8 data? It's the default but this is updating the texture in-place and we don't know what it was before. This function is not exported and always called right after the constructor, so I don't think there's a bug here but the logic was a bit confusing for me at first.

This is more a thought for outside this PR: would it make sense to do more in a Texture constructor? Should some of the texture properties instead be `readonly` and/or handled in a constructor? It feels like it could get confusing to update values when data changes.

src/renderers/shaders/float_image_frag.glsl
aganders3, line 1: This is near identical to `mesh_frag` - I prefer this name for this purpose. Maybe the idea is that `mesh_frag` will eventually be used for solid color meshes (including other shading effects)? Obviously an image is drawn on a (very simple) mesh, but I feel like we should define them as distinct concepts eventually.

src/objects/textures/texture.ts
```diff
@@ -4,9 +4,9 @@ export type TextureFilter = "nearest" | "linear";
 
 export type TextureWrapMode = "repeat" | "clamp_to_edge";
 
-export type TextureDataFormat = "rgb" | "rgba" | "red_integer";
```
aganders3, line 7: What do you think of consolidating `red` and `red_integer` as `scalar` here? This would be more intuitive to me. The current implementation seems highly coupled to WebGL specific semantics and we're already using the format/type properties to assign WebGL constants later. `WebGLTextures.getGLFormat` would need to be updated to take both format and type.

aganders3 (approved):
This looks good to me! I have a couple minor comments, mostly thoughts for future work, but otherwise I think this can go in to move forward towards an OrganelleBox component.

src/objects/textures/texture.ts
andy-sweet: I like that. Even though WebGL is our only renderer, I'm also bothered by the WebGL-specificity here. Change incoming.

src/renderers/shaders/float_image_frag.glsl
andy-sweet: Yes, I agree with your thinking. I've considered defining `Image` as a more specific type of `Mesh` before - that also improves the logic of setting uniforms that are image-specific (e.g. [contrast limits](https://github.com/chanzuckerberg/imaging-active-learning/pull/67/files#diff-0cae18dac850d6b04443a017e285a15dcb6c0e5b21dd48543d62a080f7534576R76)). I'll avoid it in this PR though, since all the image fragment shaders are near identical.

src/layers/image_utils.ts
andy-sweet: I explicitly set the `dataType` as you suggested.

Agreed that I'd like textures to be less mutable than they are now and setting/deriving some of these properties on construction is an idea I like. One minor problem is that `Texture2DArray.data` is settable, so we'd need to do similar validation/derivation there - but that should fine and cheap (compared to pushing the actual data over to the GPU).

andy-sweet: https://github.com/chanzuckerberg/imaging-active-learning/issues/149

PR #151, opened 2025-01-27
Add codegen'd zarr types and metadata validation

jgadling:
### Related Issue
Closes #<issue-number>


src/data/ome_zarr_hcs_metadata_loader.ts
```diff
@@ -1,81 +1,47 @@
...
+// TODO: we can initialize ajv and compile schemas at any point during application startup,
+// it's just here as a quick example of how to work with it.
+const ajv = new Ajv2020();
+const well_schema: JSONSchemaType<OMENGFFWellSchema> =
+  JSON.parse(wellSchemaJson);
+const plate_schema: JSONSchemaType<OMENGFFWellSchema> =
+  JSON.parse(plateSchemaJson);
+// We could cache these compiled schemas as part of the `ajv` object and export that
+// object for use by other modules, if we want: https://ajv.js.org/guide/managing-schemas.html#using-ajv-instance-cache
+// Alternatively, we could codegen validators to reduce dependencies for this app: https://ajv.js.org/standalone.html
```
jgadling, line 17: I know we're trying to keep dependencies to a minimum -- is ajv a reasonable dependency, or should I look into codegen for validations?

src/data/zarr/0.4/README.md
```diff
@@ -0,0 +1,12 @@
...
+These files were generated by the [json-schema-to-typescript](https://github.com/bcherny/json-schema-to-typescript) utility.
+
+Regenerating these files looks something like this:
+
+```
+npm install -g json-schema-to-typescript
+git clone git@github.com:ome/ngff.git /tmp/ngff
+cp /tmp/ngff/0.4/schemas/*.schema ./schemas
+for i in `ls ./schemas/*.schema | grep -v strict | xargs basename | cut -d '.' -f 1`; do json2ts schemas/$i.schema > $i.ts; done
+```
```
jgadling, line 12: I left a breadcrumb for how to regenerate these types, but I can convert this to an `npm run` target or something if there's a preference!

src/data/ome_zarr_hcs_metadata_loader.ts
andy-sweet: I haven't used it before, but based on the `package-lock.json` it doesn't actually add much, so I'd be inclined to depend on it. Codegen for the validation and thus only a dev-dependency would be ideal, but don't invest too much time.

aganders3: I'm also open to adding a dependency for this, especially after writing the validation code in the ultrack prototype. Longer-term we can perhaps move it into a separate ome-zarr package or something.

I don't know much about TS validation libraries, but a quick search suggests Zod is a good one as well. It has 0 dependencies on NPM and there is a json-schema-to-zod tool as well. Maybe we can compare these two?

I need to think about it a bit, but if we do codegen for the types anyway, I might just prefer codegen for the validators as well (the ajv "standalone" mode).

aganders3: Actually here's another one that looks pretty cool. I think it can codegen types + validators in one step: https://quicktype.io/. They have a nice playground for it as well:
https://app.quicktype.io/.

jgadling: thanks @aganders3 - the outputted. code from that tool looks much shinier, I'll try it out!

jgadling: Ok, I ran that experiment here: #153 but imop it's a dead-end since that library does *not* currently support value validation, such as min/max values or regexes.

aganders3: That sounds much less useful! Thanks for checking it out.

In that case I think it makes sense to stick with the current implementation here of codegen for types (quicktype or json2ts, probably doesn't matter) and ajv for runtime validation.

Issue #141, opened 2025-01-21
Improve multi-channel support for image layer


aganders3:
https://github.com/chanzuckerberg/imaging-active-learning/pull/87 added support for multiple channels, but assumes RGB data. For the OrganelleBox widget we will need to support 6+ (arbitrary) channels with additive or translucent blending (later we may want configurable blending, but a hard-coded mode should work for now).

Some additional features we will want (may need separate issues):
* Per-channel contrast limits/normalization (see #32)
* Per-channel colors and/or colormaps
* Show/hide each channel

andy-sweet:
> Per-channel contrast limits/normalization (see https://github.com/chanzuckerberg/imaging-active-learning/issues/32)

@phoenixAja : I have a rough solution for the per-channel contrast limits in https://github.com/chanzuckerberg/imaging-active-learning/pull/152 that should naturally extend to the per-channel color and visibility

I am still assuming the 3 RGB channels for now though, so if you can focus on that improvement first, that would be ideal

andy-sweet:
I also built on top of #152 to provide visibility and RGB color definition: https://github.com/chanzuckerberg/imaging-active-learning/compare/andy-sweet/normalize-pixel-values-multichannel...andy-sweet/normalize-pixel-values-multichannel-props?expand=1

phoenixAja:
great thanks I'll take a look!

aganders3:
Two more notes:
1. I believe channels are only supported at all in the `ImageSeries` layer so far
2. Multi-channel support is only available for `uint` images, but we also need it for `float`

PR #152, opened 2025-01-27
Allow contrast limits for multi-channel images

### LINKED ISSUE #32: Control the normalized value of scalar pixel types

andy-sweet:
### Summary
This allows the contrast limits of an image to be configured on layer creation. These limits are defined by a low and high value per channel, which the current fragment shader uses to rescale the values.

The contrast limits are passed down to the supporting `Texture` objects and used by the renderer to set them as a uniform in the related fragment shaders.

I also changed the image and image-series layer constructors to use de-structuring, which is not required, but feels more ergonomic in the long term and helps readability here too.

### Related Issue
Closes #32
Related to #141 

### Tests & Checks
I manually tested the existing image examples and the Ultrack prototype.


src/renderers/webgl_renderer.ts
```diff
@@ -78,7 +80,28 @@ export class WebGLRenderer extends Renderer {
...
+          );
+          break;
+        }
+        case "Texture2DArray": {
+          const texture2DArray = texture as Texture2DArray;
+          const flatContrastLimits = new Array<number>();
+          for (const channel of texture2DArray.channels) {
+            flatContrastLimits.push(...channel.contrastLimits);
+          }
+          program.setUniform("ContrastLimits[0]", flatContrastLimits);
```
andy-sweet: I found this on a forum somewhere and don't have a better reference to confirm this is the correct way to set an array uniform - but it seems to work?

examples/image2d_from_omezarr5d_u8/main.ts
```diff
@@ -26,8 +26,8 @@ const region = [
   { dimension: "C", index: 0 },
   { dimension: "Z", index: 0 },
 ];
-const layer = new ImageLayer(source, region);
-
+const channelProps = { contrastLimits: [0, 128] as [number, number] };
```
andy-sweet: Is there a better way to keep the TS compiler happy here? Maybe with `readonly` or similar?

src/renderers/shaders/float_image_frag.glsl
```diff
@@ -5,15 +5,13 @@ precision mediump float;
...
 void main() {
-    // The scaling factor gives the float32 CryoET example data some
-    // contrast when rendered.
-    // TODO: instead normalization of the value should be controlled by
-    // variable parameters (e.g. contrast limits).
-    // https://github.com/chanzuckerberg/imaging-active-learning/issues/32
-    float value = 2e5 * texture(texture0, TexCoords).r;
+    float texel = texture(texture0, TexCoords).r;
+    float range = ContrastLimits.y - ContrastLimits.x;
+    float value = (texel - ContrastLimits.x) / range;
```
andy-sweet: I think the following should be more performant, though needs a bit more work on the typescript side.

```suggestion
    float texel = texture(texture0, TexCoords).r;
    float value = (texel + offset) / scale;
```

src/objects/textures/channel.ts
```diff
@@ -0,0 +1,47 @@
+import { Texture } from "objects/textures/texture";
+
+export type ChannelProps = {
+  contrastLimits?: [number, number];
+};
+
+export type Channel = {
+  contrastLimits: [number, number];
+};
```
andy-sweet, line 9: My goal here is to define the type that a user can pass through as part of the public API (`ChannelProps`) without an import and a post-validation type. I considered folding these into a class, where the constructor arguments effectively define the props, but that has the downsides that it would need to be imported and that the instance needs to be mutated and re-validated when the texture is available.

src/renderers/webgl_renderer.ts
aganders3: This looks correct. A couple comments for future (I'll turn this into an issue):
* Consider setting values for the whole uniform array - I think this is relying on or using uninitialized uniforms (which on my machine seem to be zero, but I don't know if that's defined behavior)
* Consider using [`uniformBlockBinding`](https://developer.mozilla.org/en-US/docs/Web/API/WebGL2RenderingContext/uniformBlockBinding) to reduce the number of gl calls.

examples/image2d_from_omezarr5d_u8/main.ts
aganders3: It seems to not complain if it's set inline, e.g.:

```typescript
const layer = new ImageLayer({
    source,
    region,
    channelProps: [
      // contrast limits manually looked up in the zarr omero metadata
      { contrastLimits: [110, 800] },
      { contrastLimits: [110, 250] },
      { contrastLimits: [110, 800] },
    ],
  });
```

I suppose another option is to use an object with explicit min and max instead of a 2-element array.

aganders3 (approved):
LGTM

One note - this has the effect of single-channel images showing up as red. I think this can be fixed in the next PR which will add per-channel color configuration.

PR #156, opened 2025-01-30
Zarr typing/validation via zod

jgadling:
### Summary
This is another version of codegen'ing types and validations, this time using the zod library. Imop this is the best of the bunch I've worked with (ava, quicktype, and zod)

Zod implements value validation, generates very nice simple declarative code, and generally behaves nicely with the TS ecosystem by raising exceptions when it runs into trouble with validation.

### Related Issue
Closes #<issue-number>


package-lock.json
```diff
@@ -3692,11 +3717,29 @@
...
+    "node_modules/eslint/node_modules/ajv": {
+      "version": "6.12.6",
+      "resolved": "https://registry.npmjs.org/ajv/-/ajv-6.12.6.tgz",
+      "integrity": "sha512-j3fVLgvTo527anyYyJOGTYJbG+vnnQYvE0m5mmkc1TK+nxAppkCLMIL0aZ4dblVCNoGShhm+kzE4ZUykBoMg4g==",
+      "dev": true,
+      "peer": true,
+      "dependencies": {
+        "fast-deep-equal": "^3.1.1",
+        "fast-json-stable-stringify": "^2.0.0",
+        "json-schema-traverse": "^0.4.1",
```
andy-sweet: `ajv` and `json-schema-traverse` can be uninstalled now, right?

package-lock.json
```diff
@@ -29,7 +30,6 @@
         "@vitejs/plugin-react": "^4.3.1",
         "@vitest/browser": "^2.0.5",
         "@vitest/coverage-istanbul": "^2.0.5",
-        "eslint": "^9.9.0",
```
andy-sweet: Why was this removed?

vite.config.js
```diff
@@ -31,6 +31,7 @@ export default defineConfig(({ mode }) => {
   return {
     plugins,
     root: modeToRoot(mode),
+    assetsInclude: [path.resolve(_dirname, "src/data/zarr/0.4/schemas/*.schema")],
```
andy-sweet: This is not needed anymore, right?

src/data/zarr/0.4/README.md
```diff
@@ -0,0 +1,11 @@
+# OMEZarr 0.4 type definitions
+
+These files were generated by the [zod](https://zod.dev/) utility.
+
+Regenerating these files looks something like this:
+
+```
+git clone git@github.com:ome/ngff.git /tmp/ngff
+# ZSH only:
+for i in $(ls /tmp/ngff/0.4/schemas/*.schema | grep -v strict | xargs basename | cut -d '.' -f 1); do json-schema-to-zod -i /tmp/ngff/0.4/schemas/$i.schema -o $i.ts --name ${(C)i} --withJsdocs --type ${(C)i}; done
```
andy-sweet, line 10: The author of `json-schema-to-zod` discourages using it for runtime validation because JSON Schema and zod are not entirely compatible: https://github.com/StefanTerdell/json-schema-to-zod/blob/a1773d64634d93f090e62a520fe96c23c5315b90/README.md?plain=1#L137

I guess this is meant to generate a starting point that we edit manually?

package-lock.json
```diff
@@ -1145,11 +1149,29 @@
         "url": "https://opencollective.com/eslint"
       }
     },
+    "node_modules/@eslint/eslintrc/node_modules/ajv": {
```
andy-sweet: This and `json-schema-traverser` are not needed now right?

src/data/ome_zarr_hcs_metadata_loader.ts
```diff
@@ -1,73 +1,12 @@
...
+import { Plate } from "data/zarr/0.4/plate";
+import { Well } from "data/zarr/0.4/well";
 
 export async function loadOmeZarrPlate(url: string): Promise<Plate> {
   const store = new zarr.FetchStore(url);
   const group = await zarr.open.v2(store, { kind: "group" });
-  // TODO: validate attributes
-  return group.attrs as Plate;
+  // Will throw validation exceptions that we can catch if we want.
+  return Plate.parse(group.attrs) as Plate;
```
andy-sweet: I assume the cast can be removed?

```suggestion
  return Plate.parse(group.attrs);
```

src/data/ome_zarr_hcs_metadata_loader.ts
```diff
@@ -76,6 +15,7 @@ export async function loadOmeZarrWell(
 ): Promise<Well> {
   const store = new zarr.FetchStore(url + "/" + path);
   const root = await zarr.open.v2(store, { kind: "group" });
-  // TODO: validate attributes
-  return root.attrs as Well;
+
+  // Will throw validation exceptions that we can catch if we want.
+  return Well.parse(root.attrs) as Well;
```
andy-sweet: I assume the cast can be removed?

andy-sweet:
```suggestion
  return Well.parse(root.attrs);
```

src/data/zarr/0.4/README.md
jgadling: Hrmm, that's understandable - basically zero schema enforcement tools are 100% compatible with each other 😭 

I think we can decide whether we thing zod's validations are "good enough" for our use case, or whether we want our validation to always be strictly correct. 

I'm inclined to lean towards thinking of zod as "good enough" for the moment. Zod generates nice, declarative, idiomatic TS code, and supports a pretty reasonable array of validations. I don't think our viewer needs to be a strict-metadata-enforcement tool, but it just needs to sanity-check its input.

If we run into issues with zod or permissive validation, replacing this implementation with ajv down the road is pretty straightforward.

andy-sweet: That argument is good enough for me.

I just need to careful to not use `zod` for defining data-classes in other parts of the library, because it's pretty tempting to do that!

jgadling: The problem is with the json-schema-to-TS conversion, not just zod itself, so it should be safe to use in other cases, right?

andy-sweet (approved):
This looks good to me after some small cleanup (e.g. packaging). I'm down to try out zod for now and we can replace as needed (e.g. with ajv) later. Also feel free to merge and I'll clean up my branch.

src/data/zarr/0.4/README.md
andy-sweet: Yes, but if we use it for other purposes, then it'll be harder to replace/remove as a dependency - maybe for good reason though!

package-lock.json
jgadling: Still a learning curve for npm :'(

src/data/zarr/0.4/README.md
```diff
@@ -0,0 +1,11 @@
+# OMEZarr 0.4 type definitions
+
+These files were generated by the [zod](https://zod.dev/) utility.
+
+Regenerating these files looks something like this:
+
+```
+git clone git@github.com:ome/ngff.git /tmp/ngff
```
andy-sweet, line 8: This would be ideal as a `npm run` target (e.g. `npm run codegen`), but I'm fine to just have the comment for now (especially with the zsh assumption).

src/data/zarr/0.4/README.md
aganders3: I nearly a comment on #152 to this effect! I think it makes sense to hold off for a bit on going all-in on Zod. It's clearly useful and has momentum, but deserves a separate consideration of whether we use it more broadly.

PR #157, opened 2025-01-31
Use complete NGFF types in OME Zarr loader

### LINKED ISSUE #145: Validate OME-NGFF metadata
Currently we define OME-NGFF metadata types (based on the v0.4 specification) and do some limited validation of that metadata to check that it conforms to the specification.

The goal is this issue to completely validate the contents of that metadata according to the types/specs.

One potentially attractive option here is to automatically generate that validation code based on the types. Or even better to generate both from the [OME-NGFF JSON schema definitions](https://github.com/ome/ngff/tree/main/0.4/schemas). This approach could generalize well to other data sources/formats and is likely to be less error prone.

andy-sweet:
### Summary
This uses the newly generated NGFF types in the existing OME-Zarr loader.

It is an early work in progress, because type information is not being generated for the referenced parts of the schema (e.g. coordinateTransformations, axes) and because validation fails with the uint8 data used for the Ultrack prototype (need to check if that is actually a data error).

### Related Issue
Closes #145

### DIFF
--- a/src/data/ome_ngff/0.4/README.md
+++ b/src/data/ome_ngff/0.4/README.md
@@ -7,5 +7,5 @@ Regenerating these files looks something like this:
 ```
 git clone git@github.com:ome/ngff.git /tmp/ngff
 # ZSH only:
-for i in $(ls /tmp/ngff/0.4/schemas/*.schema | grep -v strict | xargs basename | cut -d '.' -f 1); do json-schema-to-zod -i /tmp/ngff/0.4/schemas/$i.schema -o $i.ts --name ${(C)i} --withJsdocs --type ${(C)i}; done
+for i in $(ls /tmp/ngff/0.4/schemas/*.schema | grep -v strict | xargs basename | cut -d '.' -f 1); do json-refs resolve /tmp/ngff/0.4/schemas/$i.schema | json-schema-to-zod -o $i.ts --name ${(C)i} --type ${(C)i}; done
 ```
--- a/src/data/ome_ngff/0.4/README.md
+++ b/src/data/ome_ngff/0.4/README.md
@@ -7,5 +7,5 @@ Regenerating these files looks something like this:
 ```
 git clone git@github.com:ome/ngff.git /tmp/ngff
 # ZSH only:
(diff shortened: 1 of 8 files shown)


andy-sweet:
@jgadling : please take a look at this when you can and see if you can remove the need for `Axis` definition and static cast to it.

Some version of the silly workaround for the top-level coordinateTransformation may be needed, so if you think of something better, please change it!

aganders3:
> please take a look at this when you can and see if you can remove the need for Axis definition and static cast to it.

I am not @jgadling but took a look here. One option is to still define `Axis` locally, but based on the `Image` type. This at least gets rid of the cast, but requires some a priori knowledge of the schema.

```diff
git diff ome_zarr_image_loader.ts
diff --git a/src/data/ome_zarr_image_loader.ts b/src/data/ome_zarr_image_loader.ts
index 53678f2..07a694f 100644
--- a/src/data/ome_zarr_image_loader.ts
+++ b/src/data/ome_zarr_image_loader.ts
@@ -6,12 +6,8 @@ import { ImageChunk } from "data/image_chunk";
 import { isTextureUnpackRowAlignment } from "objects/textures/texture";
 import { PromiseScheduler } from "./promise_scheduler";
 
-type Axis = {
-  name: string;
-  type?: string;
-};
-
 import { Image as OmeNgffImage } from "data/ome_ngff/0.4/image";
+type Axis = OmeNgffImage["multiscales"][number]["axes"][number];
 
 const dataTypes = [Uint8Array, Uint16Array, Float32Array] as const;
 const dataTypeNames = dataTypes.map((DataType) => DataType.name);
@@ -82,7 +78,7 @@ export class OmeZarrImageLoader {
     const dataset = image.datasets[lowestResolutionIndex];
     const scale = dataset.coordinateTransformations[0].scale;
     // TODO: fix zod to generate this type information.
-    const axes = image.axes as Array<Axis>;
+    const axes = image.axes;
     const translation =
       dataset.coordinateTransformations.length === 2
         ? dataset.coordinateTransformations[1].translation
```

aganders3 (approved):

PR #161, opened 2025-02-07
Refactor programName to be provided by the renderable object

aganders3:
### Summary
This has been a `TODO` in the code for a while. I found it hard to reason about when trying to load OrganelleBox data. I think it's easier to understand as part of the renderable object. This also makes sense to me as you can now observe the possible programs for each renderable object to ensure the object provides the required uniforms, etc.

### Related Issue
N/A, but removes a TODO from the code

### Tests & Checks
I manually checked the examples and the ultrack prototype to ensure they still work


phoenixAja (approved):
nice!

PR #164, opened 2025-02-10
Basic multi-channel support for image layer + float data

aganders3:
### Summary
This PR adds support for float32 image chunks in the Texture2DArray class, which is used in the ImageSeriesLayer for multi-channel support. This also extends that basic multi-channel support to the regular ImageLayer.

### Related Issues
#146 
#141 

### Tests & Checks
Manual testing of examples and ultrack prototype
Testing the HCS example with data provided by BioHub SF for the OrganelleBox project

### DIFF
--- a/src/data/image_chunk.ts
+++ b/src/data/image_chunk.ts
@@ -2,11 +2,28 @@ import { Region } from "data/region";
 import { TextureUnpackRowAlignment } from "objects/textures/texture";
 import { PromiseScheduler } from "./promise_scheduler";
 
+const imageChunkDataTypes = [Uint8Array, Uint16Array, Float32Array] as const;
+type ImageChunkData = InstanceType<(typeof imageChunkDataTypes)[number]>;
+export function isImageChunkData(value: unknown): value is ImageChunkData {
+  if (
+    imageChunkDataTypes.some(
+      (ImageChunkData) => value instanceof ImageChunkData
+    )
+  ) {
+    return true;
+  }
+  const supportedDataTypeNames = imageChunkDataTypes.map((dtype) => dtype.name);
+  console.debug(
+    `Unsupported image chunk data type: ${value}. Supported data types: ${supportedDataTypeNames}`
+  );
+  return false;
+}
+
 // One 2D chunk of n-dimensional image data.
 // TODO: include the region of this chunk.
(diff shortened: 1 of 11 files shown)


src/data/image_chunk.ts
```diff
@@ -2,11 +2,28 @@ import { Region } from "data/region";
...
+  ) {
+    return true;
+  }
+  const supportedDataTypeNames = imageChunkDataTypes.map((dtype) => dtype.name);
+  console.debug(
+    `Unsupported image chunk data type: ${value}. Supported data types: ${supportedDataTypeNames}`
+  );
+  return false;
+}
+
```
aganders3, line 21: This was moved in from the ome zarr image loader, because the datatypes completely overlap. I kept the _texture_ data types separate, but the image loader is an implementation of a chunk loader so I tink it's okay for them to share a datatype. If a specific layer only supports a subset of these types I think that's still fine.

src/objects/textures/texture.ts
```diff
@@ -10,6 +10,8 @@ export type TextureDataType = "unsigned_byte" | "unsigned_short" | "float";
 
 export type TextureUnpackRowAlignment = 1 | 2 | 4 | 8;
 
+export type DataTextureTypedArray = Uint8Array | Uint16Array | Float32Array;
```
aganders3, line 13: This replaces `ArrayBufferView` as the number of bytes per element is important but not available on that more generic class.

src/objects/textures/texture_2d_array.ts
```diff
@@ -1,26 +1,26 @@
...
-  constructor(data: ArrayBufferView, width: number, height: number) {
+  constructor(data: DataTextureTypedArray, width: number, height: number) {
     super();
 
     this.data_ = data;
     this.width_ = width;
     this.height_ = height;
     // We currently assume that each slice's size is equal to the image's area
-    this.depth_ = data.byteLength / (width * height);
+    this.depth_ = data.length / (width * height);
```
aganders3, line 16: Here using `byteLength` caused issues with a dataset I had with `uint16` data. I think it was effectively allocating the wrong size texture.

phoenixAja (approved):
lg

PR #170, opened 2025-02-12
feat: multichannel up to 16

### LINKED ISSUE #141: Improve multi-channel support for image layer

phoenixAja:
### Summary

modified the image viewer to support an arbitrary number of channels (up to a maximum of 16) by updating the fragment shader to use a preprocessor-defined MAX_CHANNELS constant instead of hardcoding 3 channels. In the WebGL renderer, we now pad the uniform arrays (visibility, colors, contrast limits) to MAX_CHANNELS length, where extra slots are set to visible: false and default values. The changes make the viewer more flexible for different types of multichannel data while maintaining clean shader code.

### Related Issue
Closes #141

### Tests & Checks
tested with examples


src/renderers/shaders/float_image_array_frag.glsl
```diff
@@ -5,16 +5,18 @@ precision mediump float;
 layout (location = 0) out vec4 fragColor;
 
 uniform mediump sampler2DArray texture0;
-uniform bool Visible[3];
-uniform vec3 Color[3];
-uniform float ValueOffset[3];
-uniform float ValueScale[3];
+// Define a maximum number of channels
+#define MAX_CHANNELS 32
```
aganders3, line 9: This second hard-code can be removed if/when we add a shader preprocessor step.

src/renderers/shaders/float_image_array_frag.glsl
aganders3, line 1: I think we need to do the same thing in `uint_image_array_frag.glsl`.

src/renderers/webgl_renderer.ts
```diff
@@ -10,6 +10,7 @@ import { ProjectedLine } from "objects/renderable/projected_line";
 import { mat4 } from "gl-matrix";
 import { DataTexture2D } from "objects/textures/data_texture_2d";
 import { Texture2DArray } from "objects/textures/texture_2d_array";
+import { MAX_CHANNELS } from "../constants";
```
aganders3: nit: I'm not sure how strict we've been but I think our convention has been to use relative imports only for items next to or under the current file, not above it.

src/renderers/webgl_renderer.ts
```diff
@@ -106,15 +107,27 @@ export class WebGLRenderer extends Renderer {
...
+              // Pad remaining slots with default values
+              visible.push(false);
+              color.push(0, 0, 0);
+              valueOffset.push(0);
+              valueScale.push(1);
+            }
           }
           // TODO: we should be careful of uninitialized values here, though it appears they
           // are zero-initialized in the shaders.
+
```
aganders3, line 128: I think with your code we can remove the above TODO!

aganders3 (approved):
Looks great - thank you!

src/renderers/shaders/float_image_array_frag.glsl
phoenixAja: yeah i didn't know how to add interpolation to the glsl files, but i agree it would be nice to use the constant here

src/renderers/shaders/float_image_array_frag.glsl
phoenixAja: done! thanks for pointing that out

src/renderers/webgl_renderer.ts
phoenixAja: fixed

src/renderers/shaders/float_image_array_frag.glsl
phoenixAja: i looked into setting a uniform for this, but there was an issue with webgl needing to know the exact size of the array prior to uniform injection.

PR #171, opened 2025-02-13
replace mesh with ImageRenderable

phoenixAja:
# Replace Mesh with ImageRenderable for Image Handling

## Changes
- Added `ImageRenderable` class to handle image-specific rendering logic
- Moved uniform generation for image shaders from `WebGLRenderer` into `ImageRenderable`
- Updated `ImageLayer` and `ImageSeriesLayer` to use `ImageRenderable` instead of `Mesh`
- Removed `makeImageMesh` and replaced with `makeImageRenderable` in `image_utils.ts`

## Why
Previously, image rendering used a generic `Mesh` class with texture-specific uniform handling in the renderer. This created a tight coupling between the renderer and texture types, and scattered image rendering logic across multiple files.

The new `ImageRenderable` class:
- Encapsulates all image-specific rendering logic in one place
- Handles both single-channel (`DataTexture2D`) and multi-channel (`Texture2DArray`) cases
- Manages channel properties and uniform generation internally
- Reduces complexity in `WebGLRenderer` by moving texture-specific code out

## Testing
- Tested with both single-channel and multi-channel images
- Verified contrast and color controls still work
- Checked both float and uint image types
- Confirmed example apps still work correctly

## Notes
This is part of an effort to make the rendering system more modular and type-safe. Each renderable object should handle its own shader uniforms and rendering specifics, rather than putting that logic in the renderer.


src/renderers/webgl_renderer.ts
```diff
@@ -78,66 +75,21 @@ export class WebGLRenderer extends Renderer {
 
     this.bindings_.bind(object);
 
+    // Bind texture if present
```
phoenixAja:
```suggestion
```

aganders3 (approved):
This is a nice improvement. I especially like how much logic moves out of the Renderer.

PR #172, opened 2025-02-13
Reorganize repo into workspaces

aganders3:
### Summary
This reorganizes the repo into separate workspaces. This will allow us to create separate packages for our core repo, as well as a React component library that depends on the core.

Testing and help updating the README are appreciated!

### Related Issue
Closes #154

### Tests & Checks
I tested I was able to
* build the library with `npm run build` (repo root)
* run the examples with `npm run examples` (repo root)
* run tests in the core workspace
* run the ultrack prototype with `npm run ultrack` (repo root) and `npm run dev` (ultrack workspace) using `VITE_MOCK_ULTRACK=true`

### Known issues
There are some problems with _types_ being exported properly for our library such that they help in the other workspaces.
Also this makes it a little harder to auto-reload during development. Right now I have to `npm run build` the library (and maybe restart the ultrack dev server) for changes to take up in the ultrack workspace, for example.

PR #155, opened 2025-01-30
[WIP] Organelle box component

aganders3:
### Summary
There's a lot left to do here, so definitely WIP, but at least this puts something in place.

Basically this works off @jgadling's branch (#151) and ports @andy-sweet's example (#138) into a React component. Next steps are to actually think about the component design. The idea is to have a (mostly?) declarative interface for a renderer, as well as the NGFF HCS data loader.

### Related Issue
This PR is related to #142 and #154, but not sufficient to close either.

### Tests & Checks
Manual testing - run this with `npm run dev -- --mode organellebox`

Issue #154, opened 2025-01-29
Update repo and package structure


aganders3:
As our intent is to use this repo as a library for many projects, we eventually need to restructure the repo.

We can keep the examples in this repo (a la three.js).

Prototypes (e.g. ultrack, organellebox) should be moved out of the repo.

I think we will however want to provide some basic React components, and they can live in this repo. They should be in a separate package, so React does not become a dependency of the library.

Issue #142, opened 2025-01-21
Create OrganelleBox Component (widget)


aganders3:
~~We're waiting on wireframes from @Janeece, but~~ I believe we will at least need the following features (look/feel/arrangement TBD):

- Range sliders for contrast limits (per channel)
- Show/hide channels
- Select color/colormap for channel
- Slider for Z (+ possibly t)
- OME-Zarr HCS controls/navigation (well/FOV selector)
- Loading indicator

Feel free to reach out to @Janeece for consulting on design, but we are not planning to wait for wireframes.

andy-sweet:
I'm picking this up today to drive it forward with some of the other features I've been working on (e.g. multi-channel contrast limits).

aganders3:
I'm still not totally clear on whether our component needs to have affordances for loading/navigating HCS metadata. It sounds like that may be handled by the portal UI.

So our component here probably only needs to take a URL of the multiscale image. It should check for default channel properties in the zarr omero data.

aganders3:
Good ideas from Janeece:

> I had a couple of ideas, but none would be blockers:
> * To make it clearer that hidden channels are hidden, change slider to disabled state while hidden
> * nice to have - Maybe have a "reset" button for channels to reset to default levels?
> * nice to have - add UI zoom controls to provide alternative ways to zoom
> Here's some rough visuals: https://www.figma.com/design/gPjEi85OeHkOGRHgtGk60D/HITL%2FViewer-Concept-Prototype-Wireframes?node-id=596-27556&t=uCpaixNnTrxsObHd-1

PR #173, opened 2025-02-13
Initial React component library (workspace)

aganders3:
Merge after #172 

### Summary
This copies the demo component from #155 into a new workspace for react components. This gives us a place to put React components - e.g. to be used on the OrganelleBox portal.

This still leaves open actually making/improving those components, and there may be additional work to package this (and core) workspace for use in an external project. Hopefully with this merged we can at least start testing that next week.

### Related issues
#154 
#142 

### Known issues
I'm still having trouble properly using types from `core` in `react`. The `react` workspace seems aware of the main exported types/classes, but not for example the type signatures of constructors, functions, etc.


packages/react/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "@idetik/react",
```
aganders3, line 2: Maybe should be `@idetik/components`? I don't know if this matters.

phoenixAja: i think components is better, but not a huge deal

phoenixAja (approved):
good starting point!

PR #182, opened 2025-02-19
Complete refactor to image renderable

aganders3:
### Summary
This finishes the refactor from #171 

* Remove channel stuff from Textures
* Use validateChannelProps (etc.)
* Validate channel props in one place only
* Add `mesh` program back for `Texture2D`
* Update some logic in Renderer such that the program will only set necessary uniforms

### Related Issue
N/A

### Tests & Checks
Tested examples
Rebuild library and tested react demo


packages/core/src/renderers/webgl_renderer.ts
```diff
@@ -45,50 +44,44 @@ export class WebGLRenderer extends Renderer {
...
-        program.setUniform("Resolution", [
-          this.canvas.width,
-          this.canvas.height,
-        ]);
-        program.setUniform("LineColor", line.color);
-        program.setUniform("LineWidth", line.width);
-        program.setUniform("TaperOffset", line.taperOffset);
-        program.setUniform("TaperPower", line.taperPower);
-        break;
+    const resolution = [this.canvas.width, this.canvas.height];
```
aganders3, line 57: these uniforms could be calculated in the switch below to save cpu cycles

packages/core/src/objects/textures/channel.ts
```diff
@@ -59,3 +67,23 @@ function validateContrastLimits(
   }
   return contrastLimits;
 }
+
+export function validateChannelProps(
```
aganders3:
```suggestion
export function validateChannels(
```

Maybe more consistent with the above function?

packages/core/src/objects/textures/channel.ts
```diff
@@ -59,3 +67,23 @@ function validateContrastLimits(
...
+): Channel[] {
+  if (channelProps.length > MAX_CHANNELS) {
+    throw new Error(`Maximum number of channels is ${MAX_CHANNELS}`);
+  }
+
+  if (texture?.type === "Texture2DArray") {
+    const depth = (texture as Texture2DArray).depth;
+    if (channelProps.length !== depth) {
+      throw new Error(
+        `Number of channels must match depth of texture: ${depth}.`
```
aganders3:
```suggestion
        `Number of channels (${channelProps.length}) must match depth of texture (${depth}).`
```

packages/core/src/vite-env.d.ts
aganders3, line 1: I don't know why this was deleted, but I also don't know what this file does

packages/core/src/objects/renderable/image_renderable.ts
```diff
@@ -84,8 +73,7 @@ export class ImageRenderable extends RenderableObject {
     }
 
     if (texture.type === "DataTexture2D") {
-      const dataTexture = texture as DataTexture2D;
-      const { color, contrastLimits } = dataTexture.channel;
+      const { color, contrastLimits } = this.channels_[0];
```
aganders3:
```suggestion
      const { color, contrastLimits } = this.channels_[0] ?? {};
```
This code doesn't sit well with me. Maybe this is better?

packages/core/src/layers/image_series_layer.ts
```diff
@@ -23,6 +24,7 @@ export class ImageSeriesLayer extends Layer {
   private dataChunks_: ImageChunk[] = [];
   private scheduler_: PromiseScheduler = new PromiseScheduler(16);
   private channelProps_?: ChannelProps[];
+  private renderable_?: ImageRenderable;
```
aganders3, line 27: I don't love holding this reference when it's already in the array of objects, but this was easiest. We could add a TODO but I'm not even sure what it would say.

aganders3:
I still need to resolve some of my own comments/questions, but input is appreciated.

I also need to test the ultrack prototype, but otherwise I think this is good to go.

phoenixAja (approved):
works for me and looks good, thanks for your help on this

aganders3:
[e9f65f5](https://github.com/chanzuckerberg/imaging-active-learning/pull/182/commits/e9f65f5925fc28f0122a1017881e31a06f1669a3) is a fix for a problem I discovered testing with the ultrack prototype.

Issue #175, opened 2025-02-14
Create channel properties component


aganders3:
This can possibly be split further:
* Contrast limits per-channel (range slider)
* Visibility per-channel (icon button)
* Color per-channel (color indicator)
* Color picker
* Contrast controls list (collapsable?)

aganders3:
I think this is pretty close, but we need a few more things to close this:
* add channel label
* add contrast slider min/max props

aganders3:
Closed by #201 for now (with the bulk of the work done in #183)

PR #183, opened 2025-02-20
feat: channel controls components

phoenixAja:
## Add Channel Controls Components

This PR adds a set of React components to control channel properties (visibility, color, and contrast) for multi-channel images. (issue chanzuckerberg/idetik#175 )

### Components Added
Channel Controls
* **ChannelControlsList**: Container component that manages channel properties for all channels and syncs with ImageLayer's channel properties using React state
* Handles updates to channel properties and propagates them to the layer
* **ChannelControl**: Individual channel control component that combines:
  * **VisibilityToggle**: Toggle button for channel visibility
  * **ColorPicker**: Color input for channel tint
  * **ContrastSlider**: Slider for adjusting contrast limits

### Tests & Checks

https://github.com/user-attachments/assets/8dea3586-493d-4b15-93f2-886094ba5249

### Future TODOs
* migrate to SDS
* migrate styling to tailwind or emotion?


packages/core/src/objects/textures/channel.ts
```diff
@@ -90,3 +90,23 @@ function validateContrastLimits(
...
+    const depth = (texture as Texture2DArray).depth;
+    if (channelProps.length !== depth) {
+      throw new Error(
+        `Number of channels must match depth of texture: ${depth}.`
+      );
+    }
+  }
+
+  return channelProps.map((props) => validateChannel(texture, props));
+}
```
aganders3:
```suggestion
```

aganders3: This snuck back in but was renamed above `validateChannels`

packages/react/src/components/Renderer.tsx
```diff
@@ -44,37 +44,42 @@ const region = [
 // http://localhost:8080/20200812-CardiomyocyteDifferentiation14-Cycle1_mip.zarr/B/03/0/.zattrs
 // ...but they look bad? especially the third channel is very bright and yellow
 const channelProps = [
-  { color: [0, 1, 1], contrastLimits: [0, 800] },
-  { color: [1, 0, 1], contrastLimits: [0, 250] },
+  { color: [0, 1, 1], contrastLimits: [0, 800], visible: true  },
+  { color: [1, 0, 1], contrastLimits: [0, 250], visible: true  },
   { color: [1, 1, 0], contrastLimits: [0, 800], visible: false },
```
aganders3:
```suggestion
  { color: [0, 1, 1], contrastLimits: [110, 800] },
  { color: [1, 0, 1], contrastLimits: [110, 250] },
  { color: [1, 1, 0], contrastLimits: [110, 800] },
```
Test it first, but this should have been fixed by #182

The default is that channels are visible if not specified. The fix in the shader makes it so the image looks okay with all three channels visible (can also delete the comment above about it looking bad). These are the correct contrast limits for our current dataset.

packages/react/src/components/Renderer.tsx
```diff
@@ -44,37 +44,42 @@ const region = [
...
 layerManager.add(layer);
 
-export default function Renderer() {
-  // Use the mount-effect so that the renderer can find the corresponding
-  // element by its ID.
+interface RendererProps {
+  onLayerReady?: (layer: ImageLayer) => void;
+}
+
+export default function Renderer({ onLayerReady }: RendererProps) {
```
aganders3, line 45: In the ultrack prototype we had settled on not specifying props as a separate interface but just doing it all in the declaration. For example:

```typescript
export default function Renderer({
  curTime,
  playbackEnabled,
  setPlaybackEnabled,
  task,
}: {
  curTime: number;
  playbackEnabled: boolean;
  setPlaybackEnabled: Dispatch<SetStateAction<boolean>>;
  task: Task | null;
}) {
  ...
```

I kind of prefer it as a separate interface object but I'm also in favor of consistency. I think we should match the previous style and open an issue if we want to change it (and then change it everywhere).

packages/react/src/components/Renderer.tsx
```diff
@@ -44,37 +44,42 @@ const region = [
...
   useEffect(() => {
     console.debug("Renderer::mount");
     let lastRequestId = 0;
     const renderer = new WebGLRenderer(`#${canvasId}`);
     const controls = new PanZoomControls(camera, camera.position);
     renderer.setControls(controls);
+
+    // Notify parent about the layer
+    onLayerReady?.(layer);
```
aganders3: 🤯 I did not know you could use optional chaining like this

packages/react/src/components/controls/ColorPicker.tsx
```diff
@@ -0,0 +1,39 @@
...
+  };
+
+  return (
+    <Box sx={{ display: "flex", alignItems: "center" }}>
+      <input
+        type="color"
+        value={rgbToHex(color)}
+        onChange={(e) => onChange(hexToRgb(e.target.value))}
+        style={{ width: 40, height: 40 }}
+      />
```
aganders3, line 36: TIL about `<input type="color" />`

packages/react/src/components/controls/ChannelControlsList.tsx
```diff
@@ -0,0 +1,74 @@
...
+      updatedChannelProps[index] = {
+        ...channelProps[index],
+        visible: !currentVisible
+      };
+    } else {
+      updatedChannelProps[index] = {
+        ...channelProps[index],
+        ...updates
+      };
+    }
```
aganders3: I don't understand why this is needed...

packages/react/src/components/controls/ChannelControlsList.tsx
```diff
@@ -0,0 +1,74 @@
...
+        <AccordionDetails>
+          <Box sx={{ display: "flex", flexDirection: "column", gap: 1 }}>
+            {channelProps.map((props: ChannelProps, index: number) => {
+              return (
+                <ChannelControl
+                  key={index}
+                  channelIndex={index}
+                  color={props.color}
+                  contrastLimits={props.contrastLimits}
+                  visible={props.visible}
```
aganders3:
```suggestion
                  visible={props.visible === undefined ? true : props.visible}
```

I'm not sure this is the proper fix, but I think the default should be visible.

I don't know of an easy way to keep this in sync if the library changed this for some reason. The only thing I can think of is to sync `Channel[]` instead of `ChannelProps[]` here, but that would mean exposing it from the `ImageLayer` as well (another pass-through from its renderable).

aganders3 (changes requested):
Looks great with a few minor changes. Thank you!

packages/react/src/components/controls/ColorPicker.tsx
phoenixAja: I learned about it only when researching for this lol

packages/react/src/components/Renderer.tsx
phoenixAja: noooo

packages/react/src/components/controls/ChannelControlsList.tsx
phoenixAja: yeah it was tricky

so we need the explicit visibility toggle becasue otherwise react was updating visibility on stale state (react state batching was limiting state updates). The Dom wasn't sensing that the channelState changed and therefore re-rendering wasn't occurring. 

Because we need the current value of disability to toggle it (unlike the other channel props like color, contrast where new values were being passed in) we needed to explicitly toggle and reset state

I'm adding a comment to explain more

packages/react/src/components/Renderer.tsx
phoenixAja: i personally like the props being a separate interface, it makes it easier to share prop interfaces and just feels like we have less code in the consrtructor making it easier to read. plus it's more scalable - if the number of props grows it's easier to manage in a constructor

aganders3: For reference see original considerations in https://github.com/chanzuckerberg/imaging-active-learning/issues/99. I'm also copying my thoughts from Slack here:

> My impulse opinion is:
> * use an interface
> * use `props: Interface` in the construtor
> * use `props.foo` for the most part
> * destructure or reassign `const foo = props.foo` just before the hook if you need to use a prop in a hook + hook dependencies (I think this will satisfy the linter, but did not test)


Feel free to leave as-is in the PR here and create an issue for changing it everywhere else.

packages/react/src/components/controls/ChannelControlsList.tsx
phoenixAja: removed this!

packages/react/src/components/Renderer.tsx
phoenixAja: here's an issue: https://github.com/chanzuckerberg/imaging-active-learning/issues/190

packages/react/src/components/Renderer.tsx
```diff
@@ -16,8 +16,7 @@ const camera = new OrthographicCamera(0, 825, 0, 500);
 const layerManager = new LayerManager();
 
 // TODO: use props to pass in most of this config
-const plateUrl =
-  "http://localhost:8080/20200812-CardiomyocyteDifferentiation14-Cycle1_mip.zarr";
+const plateUrl = "http://localhost:8080/";
```
aganders3: nit: this is now different from the URL for the same dataset in `core/examples` but that's fine with me for now. It would be great to find a publicly-hosted HCS dataset for this.

phoenixAja: agreed, i'll change this back now though

packages/react/src/components/Renderer.tsx
```diff
@@ -38,32 +37,40 @@ const channelProps = [
...
     let lastRequestId = 0;
+
+    // Initialize renderer if not already done
     if (renderer.current === null) {
       renderer.current = new WebGLRenderer(`#${canvasId}`);
       const controls = new PanZoomControls(camera, camera.position);
       renderer.current.setControls(controls);
+
+      // Notify parent about the layer
+      onLayerReady?.(layer);
```
aganders3, line 59: This might need to be something like this to prevent a race, but it doesn't really matter and can change in #186.
```
layer.addStateChangeCallback((state) => {
   if (state === "ready") { onLayerReady(layer); };
});
```

aganders3 (approved):
LGTM - thanks!

packages/react/src/components/Renderer.tsx
phoenixAja: good catch! I think 186 would be a good place to add 😄

Issue #181, opened 2025-02-14
Q1 2025 OrganelleBox tracking issue


aganders3:
This issue is meant to capture our outstanding work to complete the OrganelleBox requirements for Q1 (OrganelleBox portal V1).

- [x] chanzuckerberg/idetik#141 
- [x] chanzuckerberg/idetik#154 (#172, chanzuckerberg/idetik#173)
- [x] Coordinate with BioHub SF to test installation of library/components
- [x] chanzuckerberg/idetik#142
- [x] Coordinate with BioHub on component boundary/API
- [x] Coordinate with BioHub SF to pre-process data
- [x] chanzuckerberg/idetik#193 
- [x] chanzuckerberg/idetik#195
- [ ] chanzuckerberg/idetik-react#29 (nice-to-have)

aganders3:
I think this is pretty much done, but will wait to close until the public launch of OrganelleBox.

andy-sweet:
Closing due to launch.

Issue #174, opened 2025-02-14
Library does not bundle complete types


aganders3:
I am surprised to find vite doesn't do this, and apparently doesn't want to (https://github.com/vitejs/vite/issues/3461). There are some workarounds, but I haven't found them to work.

As an alternative we may be able to switch the library to build with [tsup](https://tsup.egoist.dev/), which seems more geared towards library building. Apparently we could still use the vite dev server for the examples. I will probably try this and continue looking into other options. I feel we should not need complex tooling for what we're building here.

aganders3:
Other options:
* rollup (vite uses it under the hood)
* esbuild
* https://github.com/qmhc/vite-plugin-dts

aganders3:
The same issue exists with the @idetik/react workspace

aganders3:
I'm able to get some amount of this working, but it requires a ton of changes to imports and we should discuss it or consult someone who knows what they are doing.

PR #184, opened 2025-02-20
Changes to make it work in organelle-box-portal project

aganders3:
### Summary
I tested our `Renderer` component in the organelle-box-portal project and ran into a few issues.

* Exclude react and react-dom in rollup packaging
* Can't use `await` at top-level in a next.js app
* Temporarily export the source map and skip minification while we're debugging

### Related Issue
#181 
#174 
#185 

### Tests & Checks
I created a test page in a local version of the organelle-box-portal project and added the `@idetik/react` `Renderer` component to the page.

Screenshot 2025-02-20 at 12 13 04 PM

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -25,6 +25,12 @@
       "import": "./dist/index.js",
       "require": "./dist/index.umd.cjs"
     },
+    "./*": {
+      "types": "./dist/*/index.d.ts"
+    },
+    "./**/*": {
+      "types": "./dist/**/*.d.ts"
+    },
     "./dist/*.js.map": "./dist/*.js.map",
     "./dist/*.d.ts.map": "./dist/*.d.ts.map"
   },
--- a/packages/react/src/components/Renderer.tsx
+++ b/packages/react/src/components/Renderer.tsx
@@ -1,4 +1,4 @@
-import { useEffect } from "react";
+import { useEffect, useRef } from "react";
 import {
   ImageLayer,
   OmeZarrImageSource,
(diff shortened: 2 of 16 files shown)


packages/core/package.json
```diff
@@ -25,6 +25,12 @@
       "import": "./dist/index.js",
       "require": "./dist/index.umd.cjs"
     },
+    "./*": {
+      "types": "./dist/*/index.d.ts"
+    },
+    "./**/*": {
+      "types": "./dist/**/*.d.ts"
+    },
```
aganders3, line 33: This helps TypeScript find types in downstream projects, because while our code gets all rolled up into `index.js` the types are still in individual `.d.ts` files in a tree.

phoenixAja (approved):

packages/react/src/components/Renderer.tsx
```diff
@@ -20,19 +18,7 @@ const layerManager = new LayerManager();
...
-  throw new Error("No wells found in plate");
-}
-const well = await loadOmeZarrWell(plateUrl, wellPaths[0]);
-//@ts-expect-error TODO: export more types?
-const imagePaths = well.well?.images.map((image) => image.path);
-if (!imagePaths) {
-  throw new Error("No images found in well");
-}
-const imageUrl = plateUrl + "/" + wellPaths[0] + "/" + imagePaths[0];
+const imageUrl = plateUrl + "/B/03/0";
```
phoenixAja, line 21: why did you take this out?

aganders3: Oh sorry I saw this comment after merging.

I removed this just to simplify integration. I think they're going to handle selecting the plate and well (hijacked to represent organelle and condition, I think?) on the portal side, so we'll just get the full URL as input.

PR #186, opened 2025-02-21
OmeZarrImageViewer component with declarative Renderer

### LINKED ISSUE #176: Create declarative Renderer widget
This should provide a straightforward interface to the renderer.

As an initial idea, it could handle the renderer/camera, and require a layer list as a prop.

aganders3:
### Summary
This refactors the `Renderer` component to be more declarative. For now it's still basically intended for 2D, but that's mostly a limitation of our camera controls.

This also then creates a new `OmeZarrImageViewer` component that is intended to be the component used on the organellebox portal. This component takes a URL for a Zarr and a `Region`. The component handles creating the source/loader and setting up an `ImageLayer` to be passed to the renderer. It also attempts to read/set channel information (color, contrast limits) from the `omero` section of the zattrs for the image. Finally, it attempts to set the initial camera frame based on the image size once the region is loaded.

Main changes in `core` are:
* Add `extent` to `ImageLayer`
* Add `metadata` property to `OmeZarrImageLoader` to provide access to omero channel data
* Some orthographic camera implementation cleanup

### Related Issue
Closes #176
(Integration with #183 is TODO and required for #142 to be closed)

### Tests & Checks
Tested manually using 20200812-CardiomyocyteDifferentiation14-Cycle1_mip.zarr

### DIFF
--- a/packages/core/src/objects/cameras/orthographic_camera.ts
+++ b/packages/core/src/objects/cameras/orthographic_camera.ts
@@ -3,8 +3,8 @@ import { mat4 } from "gl-matrix";
 
 export class OrthographicCamera extends Camera {
   // width_ and height_ should always be defined by constructor (see setFrame)
-  private width_: number | undefined;
-  private height_: number | undefined;
+  private width_: number = 128;
+  private height_: number = 128;
   private viewportAspectRatio_: number = 1;
 
   constructor(
@@ -40,11 +40,6 @@ export class OrthographicCamera extends Camera {
   }
 
   protected updateProjectionMatrix() {
-    if (!this.width_ || !this.height_) {
-      throw new Error(
-        "OrthographicCamera frame must be defined before updating projection matrix"
-      );
-    }
     // The following code ensures that the orthographic projection matrix
     // is updated so that the aspect ratio of renderable objects is respected
     // (e.g. image pixels are isotropic) by padding the camera frame to form
--- a/packages/core/src/objects/cameras/orthographic_camera.ts
+++ b/packages/core/src/objects/cameras/orthographic_camera.ts
@@ -2,8 +2,9 @@ import { Camera } from "./camera";
 import { mat4 } from "gl-matrix";
 
 export class OrthographicCamera extends Camera {
-  private width_: number;
-  private height_: number;
+  // width_ and height_ should always be defined by constructor (see setFrame)
+  private width_: number | undefined;
+  private height_: number | undefined;
   private viewportAspectRatio_: number = 1;
 
   constructor(
@@ -15,19 +16,9 @@ export class OrthographicCamera extends Camera {
     far = 100.0
   ) {
     super();
-
-    // this keeps the camera frame centered at the origin
-    // use camera.pan or set its *position* (this.transform) to move the center
-    this.width_ = Math.abs(right - left);
-    this.height_ = Math.abs(top - bottom);
-    const centerX = 0.5 * (left + right);
-    const centerY = 0.5 * (bottom + top);
-    this.transform.setTranslation([centerX, centerY, 0]);
-    this.setFrame(left, right, bottom, top);
-
     this.near_ = near;
     this.far_ = far;
-
+    this.setFrame(left, right, bottom, top);
     this.updateProjectionMatrix();
   }
 
@@ -49,6 +40,11 @@ export class OrthographicCamera extends Camera {
   }
 
   protected updateProjectionMatrix() {
+    if (!this.width_ || !this.height_) {
+      throw new Error(
+        "OrthographicCamera frame must be defined before updating projection matrix"
+      );
+    }
     // The following code ensures that the orthographic projection matrix
     // is updated so that the aspect ratio of renderable objects is respected
     // (e.g. image pixels are isotropic) by padding the camera frame to form
--- a/packages/react/src/components/App.tsx
+++ b/packages/react/src/components/App.tsx
@@ -5,7 +5,6 @@ const plateUrl =
   "http://localhost:8080/20200812-CardiomyocyteDifferentiation14-Cycle1_mip.zarr";
 const imageUrl = plateUrl + "/B/03/0";
 const region = [
-  { dimension: "c", index: { start: 0, stop: 3 } },
   { dimension: "z", index: 0 },
 ];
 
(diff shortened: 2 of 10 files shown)


aganders3:
Screenshot after rebasing on main to get #183 
Screenshot 2025-02-24 at 8 44 08 PM

packages/react/src/components/App.tsx
```diff
@@ -1,11 +1,14 @@
 import { Box } from "@mui/system";
-import Renderer from "./Renderer";
-import { ChannelControlsList } from "./controls/ChannelControlsList";
-import { useState } from "react";
-import { ImageLayer } from "@idetik/core";
+import OmeZarrImageViewer from "./OmeZarrImageViewer";
+
+const plateUrl =
+  "http://localhost:8080/20200812-CardiomyocyteDifferentiation14-Cycle1_mip.zarr";
+const imageUrl = plateUrl + "/B/03/0";
```
aganders3, line 6: We could put plate/well selectors back in for the demo now.

aganders3: I have this working locally, but don't really want to spend much time on it. It is useful for debugging with the actual organellebox data though.

aganders3: Screenshot 2025-02-25 at 1 30 50 PM
Code can be found here if we want it, but I think it should be a separate PR:
https://github.com/chanzuckerberg/imaging-active-learning/tree/aganders3/hcs-component-selectors

packages/core/src/objects/cameras/orthographic_camera.ts
```diff
@@ -2,8 +2,9 @@ import { Camera } from "./camera";
 import { mat4 } from "gl-matrix";
 
 export class OrthographicCamera extends Camera {
-  private width_: number;
-  private height_: number;
+  // width_ and height_ should always be defined by constructor (see setFrame)
```
phoenixAja, line 5: why did you add undefined as a type then?

packages/react/src/components/OmeZarrImageViewer.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+type OmeroChannel = OmeNgffImage["omero"]["channels"][number];
+
+export default function OmeZarrImageViewer({
+  sourceUrl,
+  region,
+}: {
+  sourceUrl: string;
+  region: Region;
+}) {
```
phoenixAja: remember this needs to be a separate interface now since we agreed yesterday

packages/react/src/components/OmeZarrImageViewer.tsx
```diff
@@ -0,0 +1,107 @@
...
+    const source = new OmeZarrImageSource(sourceUrl);
+    setSource(source);
+  }, [sourceUrl]);
+
+  useEffect(() => {
+    setLoading(true);
+    const getLayer = async () => {
+      if (!source) return;
+      const loader = await source?.open();
+      // TODO: should getting channel properties from the source go in the library?
```
phoenixAja: i think so, since that's how channelProperties is working?

packages/react/src/components/Renderer.tsx
```diff
@@ -1,77 +1,61 @@
...
+  layerManager,
+  camera,
+  cameraControls = "panzoom",
+  canvasId = "renderer",
+}: {
+  layerManager: LayerManager,
+  camera: OrthographicCamera,
+  cameraControls: ControlType,
+  canvasId?: string,  // allows for multiple renderers on the page
+}) {
```
phoenixAja: same thing here with the separate interface

packages/react/src/components/controls/ColorPicker.tsx
```diff
@@ -1,4 +1,5 @@
 import { Box } from "@mui/material";
+import { hexToRgb } from "../../lib/color";
```
phoenixAja: remove rel import?

phoenixAja (changes requested):
looks good besides the interfaces

packages/core/src/objects/cameras/orthographic_camera.ts
aganders3: Hm, I probably should just not change this here. The linter complains that these values are not set in the constructor, because they're set within `setFrame` not directly in the constructor. Maybe a better alternative is to just give them some default.

packages/react/src/components/OmeZarrImageViewer.tsx
aganders3: I'm not quite sure what you mean, but I moved this into a function in the library.

I think those ngff helper functions need better organization eventually but that's okay.

packages/core/src/objects/cameras/orthographic_camera.ts
aganders3: I changed this to set them to a default for now.

packages/react/src/components/OmeZarrImageViewer.tsx
```diff
@@ -0,0 +1,105 @@
...
+        if (layer.state === "ready") {
+          setLoading(false);
+          camera.setFrame(0, layer.extent.x, 0, layer.extent.y);
+          camera.update();
+        }
+      });
+      setImageLayer(layer);
+    };
+    getLayer();
+  }, [source, sourceUrl, region, camera]);
```
phoenixAja, line 63: source is updated already when sourceUrl changes so maybe we don't need sourceUrl here?

packages/react/src/components/OmeZarrImageViewer.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+  useEffect(() => {
+    if (imageLayer) {
+      layerManager.layers.length = 0;
+      layerManager.add(imageLayer);
+    }
+  }, [imageLayer, layerManager]);
+
+  console.log("OmeZarrImageViewer::render", loading);
```
phoenixAja: should this be removed? I personally don't like having a bunch of console statements because i end up commenting them all out when i end up debugging something to keep the clutter down

phoenixAja (approved):
looks good!

packages/react/src/components/OmeZarrImageViewer.tsx
aganders3:
```suggestion
```

packages/react/src/components/OmeZarrImageViewer.tsx
aganders3: Yeah not strictly necessary but both are used directly here so I think it's best to include both. I think this will eventually get refactored as well - these ngff helper functions right now each re-fetch the metadata I think and we should instead have the source just parse it once.

PR #189, opened 2025-02-24
Make types work downstream

### LINKED ISSUE #174: Library does not bundle complete types

aganders3:
Merge after #184 

### Summary
This is the result of some annoying investigations into building, packaging, bundling, typescript, etc. I'm not sure this is optimal, but I believe it's better than what we were doing before.

Some example things that were wrong:
* `const layer = new ImageLayer(123);` would _not_ produce an error when it should (wrong args)
* `layerManager.add(123)` would similarly _not_ produce an error
* after some adjustments (exporting types, but not updating imports in `@idetik/core`) `layerManager.add(layer)` _would_ produce an error that `ImageLayer` did not conform to `Layer` (the base class that `layerManager.add` expects)

Things I did here:
* Export types using wildcard syntax so we don't have to do rollup of type declarations (.d.ts files)
* Export a lot more stuff - exporting more types seems to help typescript downstream, especially when checking functions defined on base classes
* Use path-based imports for everything, otherwise typescript does not understand that `ImageLayer` is a subclass of `Layer`, even if both are exported. I don't love the `../..` required here and there's probably a better way, but I'd rather have types. I think using absolute imports _without_ relying on the tsconfig `baseURL` or something might also work. I think we could also use the alias `@` but there was some opposition to this before.
* Export ngff stuff explicitly, so we can keep those exports separate and versioned

At the moment everything is exported from the root `@idetik/core` but I think we can change this if we're explicit (see ngff stuff). For example if we want to export layers from `@idetik/core/layers` I think we can. Doing this with wildcards was tempting but only half-worked and caused more confusion in the end.

### Related Issue
Closes #174 (the @idetik/react workspace may need similar changes though)

### Tests & Checks
Tested with `npm run build:all` which is now happy (good). It's also sad when I add a deliberately bad line such as `const layer = new ImageLayer(123);` or `layerManager.add(123)`, where previously these lines would _not_ be caught by typescript and would cause hard-to-diagnose issues at runtime.


aganders3:
I think @ehoops-cz is going to look at this with us next week.
Also @tihuan may look this week.
Thanks! Reach out to discuss please :).

We can try to hold of merging until then if it's not blocking.

aganders3:
Comments from @tihuan here: https://docs.google.com/document/d/1TdK5qxbYwuq4GGZVw_PWIW942eZNAtEnoRQDRNwERFA/edit?usp=sharing

TLDR:
* We should try a couple more vite/rollup plugins that may help by combining type declarations into a single .d.ts file, as this should let us use a mix of absolute/relative paths as we have been.
* Relative imports everywhere (as in this PR) also work and we can leave the type declarations in separate files - is there a lint we can turn on to enforce this? I don't love the look of some of these (especially referencing significantly "up" the tree), but oh well.
* Is there another form of absolute import, like adding a path with `@idetik` so the imports look the same in our project as they would in a downstream project? I still find it a bit weird that our "absolute" imports are from `src/foo/bar` and indistinguishable from an installed package named `src` (as odd as that would be).

aganders3:
Closing in favor of #205

PR #198, opened 2025-02-26
Update tailwind to v4 and update styles

aganders3:
### Summary
Tailwind config and other styling changes.

(Copied from Slack)
I have learned some things poking around the config/styling:
1. ~~we need to point to the tailwindcss compiled output css. Because the output is in ./dist I found this easiest to do by putting it in index.html instead of the Main.tsx~~
    * This was just not true! It seems to have been fixed by updating tailwind to v4, which matches the version of the @tailwind/vite plugin
3. ~~For tailwind to generate the class I had to use a weird name - <span className="text-dark-sds-color-primitive-green-400"> - I don't know how to reference the light/dark based on the current theme. Emotion has useTheme so we can get it from there and dynamically set the class but that's kind of annoying. Even with that I guess there can be some issues with dynamically setting class names because the tailwind compiler may not be able to know the options it needs to generate.~~
    * I can't get this to work now even with the very long name here, so I think we can keep that open
4. I think the [SDS Accordion is intentionally transparent](https://github.com/chanzuckerberg/sci-components/blob/19c0a39b2d1d7bd667b0bdab2c7be04500b15161/packages/components/src/core/Accordion/style.ts#L43) (and square) which kind of sucks. Replacing the main Accordion with the MUI version seems to still work and fix this.

### Tests & Checks
Tested in light and dark mode.

Updated screenshots:
Screenshot 2025-02-27 at 11 29 38 AM
Screenshot 2025-02-27 at 11 29 28 AM


phoenixAja (approved):

packages/react/package.json
```diff
@@ -4,13 +4,13 @@
...
   "scripts": {
-    "dev": "vite",
+    "dev": "npm run build:tailwind -- --watch & vite",
     "compile": "tsc",
     "build": "vite build && tsc --project tsconfig.build.json",
     "format": "prettier --write './**/*.{ts,tsx,html,json}'",
     "format-check": "prettier --check './**/*.{ts,tsx,html,json}'",
     "lint": "eslint .",
-    "build:tailwind": "tailwindcss -i ./src/index.css -o ./dist/index.css --minify",
+    "build:tailwind": "tailwindcss -i ./src/index.css -o ./dist/index.css",
```
phoenixAja: why remove minify?

aganders3: just for debugging - I'll put it back to merge

PR #199, opened 2025-02-27
Fix channel blending by clamping extracted texture value

### LINKED ISSUE #195: Adjust multi-channel blending
Right now the blending of multiple channels is purely additive. I think instead we want to do some alpha-blending such that the low values for each channel are transparent rather than black. You can see some issues with this when blending channels over a phase image (mostly gray) and getting surprising colors in the result.

For example this should appear as a blue overlay on a grayscale image, but instead it tints the background yellow.

Image

aganders3:
### Summary
When contrast limits are applied in the shader, we can actually end up with a negative value. This messes with the colors of other channels.

### Related Issue
Closes #195 

### Tests & Checks
Tested with organelle box data and no longer seeing yellow tinted background when scaling the blue channel


phoenixAja (approved):
looks good, my only thought is maybe to add a comment on why this is needed?

PR #201, opened 2025-02-27
Add label and contrast range props for channel controls

aganders3:
### Summary
This gets the label, initial visibility, and contrast limits from the omero data.

### Related Issue
Closes #193 

### Tests & Checks
Screenshot 2025-02-27 at 10 16 32 PM

PR #202, opened 2025-02-27
[WIP] Fixes to get working in Organelle Box Portal/Next.js 15

kyleawayan:
#### todo

- [ ] add `@idetik/core` to `external`

### Related Issue
Closes #<issue-number>

### DIFF
--- a/packages/react/src/index.ts
+++ b/packages/react/src/index.ts
@@ -1,2 +1,2 @@
-import Renderer from "components/Renderer";
-export { Renderer };
+export * from "components/Renderer";
+export * from "components/OmeZarrImageViewer";
--- a/packages/react/vite.config.js
+++ b/packages/react/vite.config.js
@@ -31,7 +31,7 @@ export default defineConfig(() => {
         fileName: "index",
       },
       rollupOptions: {
-        external: ['react', 'react-dom'],
+        external: ['react', 'react/jsx-runtime', 'react-dom', 'react-dom/client'],
         output: {
           globals: {
             'react': 'React',
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -10,7 +10,7 @@
     "format": "prettier --write './**/*.{ts,tsx,html,json}'",
     "format-check": "prettier --check './**/*.{ts,tsx,html,json}'",
     "lint": "eslint .",
-    "build:tailwind": "tailwindcss -i ./src/index.css -o ./dist/index.css --minify",
+    "build:tailwind": "npx @tailwindcss/cli -i ./src/index.css -o ./dist/index.css --minify",
     "build:css": "mkdir -p dist && npm run build:tailwind"
   },
   "dependencies": {
--- a/packages/react/src/components/App.tsx
+++ b/packages/react/src/components/App.tsx
@@ -1,5 +1,5 @@
 import cns from "classnames";
-import OmeZarrImageViewer from "./OmeZarrImageViewer";
+import { OmeZarrImageViewer } from "./OmeZarrImageViewer";
 
 const plateUrl =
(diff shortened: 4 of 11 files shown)


aganders3 (approved):
Looks great - thanks!

Ahh, just saw this is still in draft so maybe you've got more to add. Feel free to ping me when ready.

packages/react/vite.config.js
```diff
@@ -31,7 +31,7 @@ export default defineConfig(() => {
         fileName: "index",
       },
       rollupOptions: {
-        external: ['react', 'react-dom'],
+        external: ['react', 'react/jsx-runtime', 'react-dom', 'react-dom/client'],
```
aganders3, line 34: I think perhaps we want `@idetik/core` here as well?

My understanding of how this works is pretty low, though.

packages/react/src/index.ts
```diff
@@ -1,2 +1,2 @@
-import Renderer from "components/Renderer";
-export { Renderer };
+export * from "components/Renderer";
+export * from "components/OmeZarrImageViewer";
```
aganders3, line 2: Testing locally we may also need to change the components to named exports instead of `export default`

aganders3:
I pulled in these changes along with #204, #201, and #203 and replaced the component in the portal with `OmeZarrImageViewer` from this library. There are definitely still some style issues, but hopefully not too difficult to resolve.

One thing I noticed was you're still using Tailwind 3 but we updated this repo to use Tailwind 4. Updating seemed to resolve some issues for me, but I'm still a bit confused on the configuration. Also I feel like especially as a dev/bild dependency it maybe doesn't matter in the final version if we're using different versions.

The video below also shows a change to channel blending we should discuss.

https://github.com/user-attachments/assets/000d2460-477d-4819-8a2d-209c033f3c6b

kyleawayan:
Yep. I think since we're compiling the Tailwind into CSS anyway it doesn't matter. I'll bring the styles I made in the portal to here.

packages/react/vite.config.js
kyleawayan: I think we should be good -- it worked without any modifications to the core library. I think it was just confusing React dependency issues

packages/react/src/index.ts
kyleawayan: Yep will change that here too

kyleawayan:
> Yep. I think since we're compiling the Tailwind into CSS anyway it doesn't matter. I'll bring the styles I made in the portal to here.

Nevermind take that back. I'm having issues putting it in the portal 😅. Might have to update the portal to Tailwind 4 as well then.

kyleawayan:
Ok so upgrading the portal to Tailwind 4 may be a huge task... and I can't figure out how to get the styling to work without it.

For now I just copied and pasted the updated code into the portal: https://github.com/czbiohub-sf/organelle-box-portal/pull/17

packages/react/vite.config.js
aganders3: I guess what I mean is that if we _don't_ put `@idetik/core` here I think you need to install both in the downstream project. I think you've done this already anyway, but it means you could end up with two versions of core in your tree/bundle.

aganders3:
Hmm...thanks for taking a look. I'll keep looking at the styling as well but at worst we can fallback on copy/paste for v1.

packages/react/vite.config.js
kyleawayan: Ah. Yes I think we actually may have to put `@idetik/core` in this `external` array. From my experience using Rollup, when you put dependencies in `external` the library is expected to be "externally" provided where it's being used. Sort of like a concept of a peer dependency. I'll add it in for this PR

aganders3:
Closing as stale and obviated by #220

PR #204, opened 2025-02-28
Add optional scale prop to OmeZarrImageViewer (+source +loader)

### LINKED ISSUE #200: Add configuration to ImageLayer and ImageSeriesLayer to select which scale is loaded
Eventually this should be handled by proper multiscale loading, but for now it would be nice to just specify which scale to load.

It would be nice to speciy:
* highest
* lowest
* specific scale level

For the "specific scale level" case, I'm not sure if it should error if the scale does not exist, or select the nearest (lowest, basically).

Also perhaps specifying a negative number would reverse index (a la Python) so you could specify `-1` to get the lowest scale.

aganders3:
### Summary
The organelle box folks have provided us with 4-deep multiscale images. This provides a way to allow them to specify a scale to display. The only option they have without this is to re-process their data such that the lowest scale is the size they want to load in our viewer (that's how our viewer currently works).

I'm not totally sure I love this implementation, so I'm wide open to alternatives, but I think this is mostly "good enough" for now. The API preserves the idea that the default (undefined) option will intelligently load the proper scale in the future.

This allows indexing with `>=0` to select a specific scale, or `<0` to select starting at the smallest scale. It will then clamp to the bounds so it will compensate for lack of a-priori knowledge of the number of scales and not throw errors. The default behavior remains the same and will load the _last_ scale (equivalent to passing `-1`, typically the smallest image / coarsest resolution) for now.

### Tests & Checks
Added a `scale` prop to `OmeZarrImageViewer` and you can select different scales as described above by changing it.

### DIFF
--- a/packages/core/src/data/ome_zarr_image_loader.ts
+++ b/packages/core/src/data/ome_zarr_image_loader.ts
@@ -32,9 +32,10 @@ export class PromiseQueue<T> {
 // https://ngff.openmicroscopy.org/0.4/#image-layout
 export class OmeZarrImageLoader {
   root_: zarr.Group<zarr.FetchStore>;
+  scaleIndex_: number;
   metadata_: OmeNgffImage;
 
-  constructor(root: zarr.Group<zarr.FetchStore>) {
+  constructor(root: zarr.Group<zarr.FetchStore>, scaleIndex?: number) {
     this.root_ = root;
     const attrs = this.root_.attrs;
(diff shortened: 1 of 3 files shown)


aganders3:
> I'm not totally sure I love this implementation

Specifically I feel this should perhaps not be a property of the source, but instead of the layer (+ still the loader).

phoenixAja (approved):

PR #205, opened 2025-03-05
Make types work downstream (take 2)

### LINKED ISSUE #174: Library does not bundle complete types

aganders3:
### Summary
Take 2 of #189 

Based on recommendations from @tihuan this uses rollup to combine all our type declarations into a single `index.d.ts` file that matches our rolled-up code (`index.js`). This seems to resolve all the path-related typescript issues, and makes types work as I would expect downstream.

Thus, this PR also includes fixes for all the typescript issues in `@idetik/react`.

This also bundles `style.css` in the `@idetik/react` package, which needs to be imported in the downstream project CSS.

### Related Issue
Closes #174 

### Tests & Checks
Tested locally with the [organelle-box-portal](https://github.com/czbiohub-sf/organelle-box-portal) repo
Screenshot 2025-03-06 at 2 43 00 PM


aganders3:
@kyleawayan can you try installing from this branch when you have a chance? For me it fixes some the `@ts-expect-error` notes in the organelle box portal project.

As noted above this still doesn't fix the style issues.

aganders3:
Thanks @tihuan and @masoudmanson for all the help on this!

packages/react/package.json
```diff
@@ -6,15 +6,12 @@
   "scripts": {
     "dev": "vite",
     "compile": "tsc",
-    "build": "vite build && tsc --project tsconfig.build.json && npm run build:css",
+    "build": "vite build && tsc && rollup -c rollup.config.js",
     "format": "prettier --write './**/*.{ts,tsx,html,json}'",
     "format-check": "prettier --check './**/*.{ts,tsx,html,json}'",
-    "lint": "eslint .",
-    "build:tailwind": "tailwindcss -i ./src/index.css -o ./dist/index.css --minify",
-    "build:css": "mkdir -p dist && npm run build:tailwind"
```
aganders3, line 14: vite seems to just handle this, but it might be useful to keep a command to build the output for debugging, etc.

phoenixAja (approved):
👏 👏 👏

PR #206, opened 2025-03-07
[PoC] Chunk-caching zarr loader

aganders3:
### Summary
This is a drop-in replacement for the `OmeZarrImageLoader` that transparently caches chunks (source chunks), keeping them uncompressed. This reduces our previous reliance on the browser cache, where fetching multiple slices within the same chunk had a significant overhead for formulating the HTTP request and decompressing the (cached) response.

This is still a pretty simple mechanism that wraps the underlying zarrita object to override it's `getChunk` method with a cached version. I think we can come up with a more robust approach, but this may be worth keeping in the meantime. 

### Related Issue
Motivated by performance issues with the example introduced in #207 

### Tests & Checks
Tested all examples. Note in particular when testing the example introduced in #207 you can watch the dev tools to see significant delays in loading chunks even when hitting the disk cache. This PR eliminates that bottleneck.

### DIFF
--- a/packages/core/src/data/ome_zarr_image_loader.ts
+++ b/packages/core/src/data/ome_zarr_image_loader.ts
@@ -5,9 +5,9 @@ import { Region } from "data/region";
 import { ImageChunk, isImageChunkData } from "data/image_chunk";
 import { isTextureUnpackRowAlignment } from "objects/textures/texture";
 import { PromiseScheduler } from "./promise_scheduler";
+import { CachedZarrArray } from "./zarr_chunk_cache";
 
 import { Image as OmeNgffImage } from "data/ome_ngff/0.4/image";
-type Axis = OmeNgffImage["multiscales"][number]["axes"][number];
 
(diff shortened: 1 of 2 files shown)


packages/core/src/data/ome_zarr_image_loader.ts
```diff
@@ -34,6 +34,8 @@ export class OmeZarrImageLoader {
   root_: zarr.Group<zarr.FetchStore>;
   scaleIndex_: number;
   metadata_: OmeNgffImage;
+  // Single CachedZarrArray instance for the entire loader
+  private cachedZarrArray_: CachedZarrArray | null = null;
```
aganders3, line 38: I considered splitting this into a different class/file, but I think we can keep it here with a way to choose whether to use/not use the cache.

aganders3:
I'm going to close this as I'm not currently pushing it forward. I think we may want something like it when using logical chunks for progressive chunk streaming, but we can revisit then. The performance improvement when loading from the same chunk repeatedly is significant.

Issue #192, opened 2025-02-25
Consider and unify API for layer.extent


aganders3:
I have found the need for a layer extent in a couple places - mainly to set the camera focused on the data in a layer. This is implemented differently in the `TracksLayer` and `ImageLayer` (+ `ImageSeriesLayer`) but we should consider the interface a bit and unify the API.

PR #207, opened 2025-03-11
Add ImageStackLayer

aganders3:
### Summary
`ImageSeriesLayer` expects unscaled indices in its `timeDimension`. This is not sufficient for displaying a stack of images and slicing through Z, for example, where the Z-dimension has an associated scale.

This introduces instead `ImageStackLayer` which takes into account the scale of the selected Z-dimension. It supports a 4D multi-channel region, and as well supports omitting the Z-dimension from this region to load a whole stack without knowing the extent.

This new layer also will mark itself `ready` and display the first slice it can. This causes a bit of an issue if an application needs to determine when all loading complete. I can think of two solutions:
* the `Renderer` can render objects for layers in the `loading` state ([currently it only renders `ready` layers](https://github.com/chanzuckerberg/imaging-active-learning/blob/7827fe7f387b568cf154b264f97bc7671e15e9fb/packages/core/src/core/renderer.ts#L44))
* we can add a new state (such as `ready-partial` or `ready-complete`) to disambiguate ready-to-draw from finished loading

This PR also adds functionality to the `OmeZarrImageViewer` component used by the organelle box portal. If you provide a `zDimension` prop it will attempt to load the region as a stack.

### Related Issue
Because this code attempts to load each slice individually, #206 enhances performance significantly with some datasets. I don't think that PR is necessary for the organelle box data but it really helps with the example introduced here with the CryoET data.

### Tests & Checks
Added a new example `Z-Stack from 3D OMEZarr with f32 data`
Tested with current organelle box data in the components workspace


shlomnissan:
> the Renderer can render objects for layers in the loading state ([currently it only renders ready layers](https://github.com/chanzuckerberg/imaging-active-learning/blob/7827fe7f387b568cf154b264f97bc7671e15e9fb/packages/core/src/core/renderer.ts#L44))
we can add a new state (such as ready-partial or ready-complete) to disambiguate ready-to-draw from finished loading

My intuition is that maybe a new state is not needed in the long-term. If the library focuses on large-scale multi-dimensional data, we should probably never load everything — and _ready_ will always mean “ready to render”. What do you think?

packages/core/examples/image_stack_from_omezarr3d_f32/index.html
```diff
@@ -0,0 +1,97 @@
+<!doctype html>
```
shlomnissan, line 1: Note: We need to create an examples landing page to replace the copy/paste code approach. This ties into improving documentation overall. The goal is to build a landing page like Three.js that combines interactive examples with auto-generated documentation. Perhaps I should add a GitHub issue?

packages/core/examples/image_stack_from_omezarr3d_f32/main.ts
```diff
@@ -0,0 +1,159 @@
...
+  Region,
+  WebGLRenderer,
+} from "@";
+import { PanZoomControls } from "@/objects/cameras/controls";
+
+const url =
+  "https://files.cryoetdataportal.cziscience.com/10444/24apr23a_Position_12/Reconstructions/VoxelSpacing4.990/Tomograms/100/24apr23a_Position_12.zarr";
+const layerManager = new LayerManager();
+const renderer = new WebGLRenderer("#canvas");
+const camera = new OrthographicCamera(0, 128, 0, 128);
```
shlomnissan, line 16: Note: While having the y-basis vector point downward feels intuitive (top 0 bottom 128), I'm not fully convinced it's the best approach.

packages/core/examples/image_stack_from_omezarr3d_f32/main.ts
```diff
@@ -0,0 +1,159 @@
...
+
+// Source is 3D with axes (z, y, x), so we provide an interval in z
+const source = new OmeZarrImageSource(url, -1);
```
shlomnissan: Why are we passing `-1` here?

packages/core/examples/image_stack_from_omezarr3d_f32/main.ts
```diff
@@ -0,0 +1,159 @@
...
+const loader = await source.open();
+const attributes = await loader.loadAttributes();
+const zDimName = "z";
+const zAxisIndex = attributes.dimensions.findIndex((dim) => dim === zDimName);
+const zMin = 0;
+const zMax = attributes.shape[zAxisIndex];
+const region: Region = [
+  // empty Z dimension interval will load the entire Z stack
+];
```
shlomnissan: I generally prefer explicit definitions rather than assuming that an empty value implies loading the entire stack. Is there a reason we shouldn't require users to always specify a valid region? In this case, would it make sense to update the code to something like:

```ts
const region: Region = [
  [zMin, zMax],
  // ...
];
```

packages/core/examples/image_stack_from_omezarr3d_f32/main.ts
```diff
@@ -0,0 +1,159 @@
...
+const maxValueEl = document.querySelector<HTMLSpanElement>("#max-value");
+
+// Check that all elements exist
+if (!zSlider) throw new Error("Z slider not found.");
+if (!zIndexEl) throw new Error("Z index element not found.");
+if (!zTotalEl) throw new Error("Z total element not found.");
+if (!minSlider) throw new Error("Min slider not found.");
+if (!maxSlider) throw new Error("Max slider not found.");
+if (!minValueEl) throw new Error("Min value element not found.");
+if (!maxValueEl) throw new Error("Max value element not found.");
```
shlomnissan: Nit: It's just an example, right? I wouldn't worry too much about validating DOM objects.

packages/core/examples/image_stack_from_omezarr3d_f32/main.ts
```diff
@@ -0,0 +1,159 @@
...
+minSlider.max = "0";
+minSlider.step = "0.000001";
+minSlider.value = initialMinValue.toString();
+
+maxSlider.min = "0";
+maxSlider.max = "0.00005";
+maxSlider.step = "0.000001";
+maxSlider.value = initialMaxValue.toString();
+
+minValueEl.textContent = initialMinValue.toString();
```
shlomnissan: Nit: I generally prefer template literals for consistency with modern JS and flexibility in case additional text is needed, but it's not a big deal either way — you can argue that explicit conversion is more important.
```ts
minValueEl.textContent = `${initialMinValue}`;
maxValueEl.textContent = `${initialMaxValue}`;
```

packages/core/examples/image_stack_from_omezarr3d_f32/main.ts
```diff
@@ -0,0 +1,159 @@
...
+
+  // Update channel properties
+  layer.setChannelProps([
+    {
+      visible: true,
+      color: [1, 1, 1],
+      contrastLimits: [minValue, maxValue],
+    },
+  ]);
+});
```
shlomnissan:
```suggestion
const sliderHandler = () => {
  const minValue = minSlider.valueAsNumber;
  const maxValue = maxSlider.valueAsNumber;

  if (minValue >= maxValue) minSlider.value = `${(maxValue - 0.01)}`;
  if (maxValue <= minValue) maxSlider.value = `${(minValue + 0.01)}`;

  minValueEl.textContent = minValue.toFixed(2);
  maxValueEl.textContent = maxValue.toFixed(2);

  layer.setChannelProps([{
    visible: true,
    color: [1, 1, 1],
    contrastLimits: [minValue, maxValue],
  }]);
}

minSlider.addEventListener("input", sliderHandler);
maxSlider.addEventListener("input", sliderHandler);
```

packages/core/examples/image_stack_from_omezarr3d_f32/main.ts
```diff
@@ -0,0 +1,159 @@
...
+
```
shlomnissan, line 124: There may be an issue with the slider implementation:
- The minimum slider appears to be greater than the maximum slider.
- The text values are not displaying correctly.

Screenshot 2025-03-13 at 1 30 40 PM

packages/core/examples/image_stack_from_omezarr3d_f32/main.ts
```diff
@@ -0,0 +1,159 @@
...
+layer.addStateChangeCallback((newState: LayerState) => {
+  if (newState === "ready") {
+    // Set up slider event handler
+    zSlider.addEventListener("input", (event) => {
+      const value = (event.target as HTMLInputElement).valueAsNumber;
+      try {
+        layer.setZIndex(value);
+        zIndexEl.textContent = value.toString();
+      } catch {
+        console.debug("Tried to set Z index out of bounds");
```
shlomnissan: In what scenarios would this behavior occur?

packages/core/examples/image_stack_from_omezarr3d_f32/main.ts
```diff
@@ -0,0 +1,159 @@
...
+      }
+    });
+
+    if (layer.extent !== undefined) {
+      camera.setFrame(0, layer.extent.x, 0, layer.extent.y);
+      renderer.setControls(new PanZoomControls(camera, camera.position));
+      camera.update();
+    }
```
shlomnissan, line 143: Nit: consider rewriting it as:
```suggestion
    if (layer.extent) {
      camera.setFrame(0, layer.extent.x, 0, layer.extent.y);
      renderer.setControls(new PanZoomControls(camera, camera.position));
      camera.update();
    }
```
This makes the code more readable and flexible, as it will also handle other falsy cases like null.

packages/core/examples/index.html
```diff
@@ -32,6 +32,11 @@ <h1>Examples</h1>
           >2D Image Series from 5D OMEZarr with uint8 data</a
         >
       </li>
+      <li>
+        <a href="/image_stack_from_omezarr3d_f32/index.html"
+          >Z-Stack from 3D OMEZarr with f32 data</a
```
shlomnissan: Nit: Consider renaming this to `2D Image Stack from 3D OMEZarr with float32 data` so it's consistent with our existing example naming convention.

packages/core/src/data/image_chunk.ts
```diff
@@ -45,6 +45,13 @@ export type ImageChunkSource = {
...
 
+export type LoaderAttributes = {
+  dimensions: string[];
+  shape: readonly number[];
+  scale: readonly number[];
+};
+
 export type ImageChunkLoader = {
   loadChunk(input: Region, scheduler?: PromiseScheduler): Promise<ImageChunk>;
+  loadAttributes(): Promise<LoaderAttributes>;
```
shlomnissan, line 56: Wouldn't it be more appropriate for `ImageChunkLoader` to be an interface rather than a type, given that loaders like `OmeZarrImageLoader` are class-based? Using an interface would ensure that these classes explicitly adhere to a contract while allowing for better extensibility and inheritance.

packages/core/src/data/ome_zarr_image_loader.ts
```diff
@@ -106,7 +101,7 @@ export class OmeZarrImageLoader {
 
     if (!isImageChunkData(subarray.data)) {
       throw new Error(
-        `Subarray has an unsupported data type ${subarray.data.constructor.name}`
+        `Subarray has an unsupported data type, data=${subarray.data}`
```
shlomnissan: I would have assumed this would output `[object Object]` in certain cases. What is the reason for this change?

packages/core/src/data/ome_zarr_image_loader.ts
```diff
@@ -147,36 +149,57 @@ export class OmeZarrImageLoader {
...
-    return this.metadata_;
+  public async loadAttributes(): Promise<LoaderAttributes> {
+    const image = this.metadata_.multiscales[0];
+    console.log("loading attributes for image", image);
+    const dimensions = image.axes.map((axis) => axis.name);
+    const dataset = image.datasets[this.scaleIndex_];
+    const array = await zarr.open.v2(this.root_.resolve(dataset.path), {
+      kind: "array",
+      attrs: false,
+    });
```
shlomnissan: It seems like there's some duplication in this class when fetching attributes. Both loadChunk and regionToIndices access similar data attributes, such as:

- Retrieving the image from `this.metadata_.multiscales[0]`
- Fetching the scale from `dataset.coordinateTransformations[0].scale`
- Etc.

Is there any difference between these values when used in loadChunk versus loadAttributes? If not, we might be able to store them as class member variables and initialize them in the constructor to reduce redundancy.

I think making `regionToIndices` a member function is a significant improvement (unless there's a reason it wasn't a member function to begin with) but there's some opportunity for deduplication and improved clarity.

This is not blocking.

packages/core/src/data/ome_zarr_image_loader.ts
```diff
@@ -147,36 +149,57 @@ export class OmeZarrImageLoader {
...
+    region: Region,
+    datasetIndex: number = 0
+  ): Array<Slice | number> {
+    const axes = this.metadata_.multiscales[0].axes;
+    const dataset = this.metadata_.multiscales[0].datasets[datasetIndex];
+    const scale = dataset.coordinateTransformations[0].scale;
+    const translation =
+      dataset.coordinateTransformations.length === 2
+        ? dataset.coordinateTransformations[1].translation
+        : new Array(axes.length).fill(0);
```
shlomnissan, line 181: Directly related to the previous comment -- all these values should probably be computed elsewhere IMO.

packages/core/src/layers/image_stack_layer.ts
```diff
@@ -0,0 +1,183 @@
...
+};
+
+// Loads 2D+z image data (Z-stack) from an image source into renderable objects.
+export class ImageStackLayer extends Layer {
+  private readonly source_: ImageChunkSource;
+  private readonly region_: Region;
+  private readonly zDimension_: string;
+  private texture_: Texture2DArray | null = null;
+  private dataChunks_: ImageChunk[] = [];
+  private scheduler_: PromiseScheduler = new PromiseScheduler(16);
```
shlomnissan: This can likely be marked as `readonly` and moved up to align with the other `readonly` member variables for consistency.

packages/core/src/layers/image_stack_layer.ts
```diff
@@ -0,0 +1,183 @@
...
+  private channelProps_?: ChannelProps[];
+  private renderable_?: ImageRenderable;
+  private extent_?: { x: number; y: number };
```
shlomnissan: Can you remind me what extent represents in this context? I know we discussed this a few days ago, but I’ve already forgotten. Is it a predefined "viewport" over the data? If so, maybe we should consider a more descriptive name to make its purpose clearer.

Also, is it likely this will be needed in other layers, e.g. `ImageSeriesLayer`?

packages/core/src/layers/image_stack_layer.ts
```diff
@@ -0,0 +1,183 @@
...
+
+  constructor({
```
shlomnissan: I’m not convinced that introducing a separate layer is the right approach. Instead, I think we should generalize the existing “image series layer” to support multiple scrollable dimensions (e.g., time and the z-axis). My main concerns are:

Most of the code is identical, and the differences seem like they could be abstracted unless I’m overlooking something.
Even though both layers serve the same purpose, they might diverge over time with features that should ideally apply to both.
I believe we should be cautious about adding new layers unless a distinct rendering paradigm is necessary. Additionally, tying “image series” exclusively to time feels overly restrictive. If there’s a compelling reason to keep them separate, we should probably rename both layers to make their behavior explicit.

I’d like to discuss this further to better understand the trade-offs. That said, I don’t want this to be a blocker—if you disagree, I’m happy to merge both layers and do a quick follow-up if we align on that direction.

packages/core/src/layers/image_stack_layer.ts
```diff
@@ -0,0 +1,183 @@
...
+        this.load();
+        break;
+      case "loading":
+      case "ready":
+        break;
+      default: {
+        const exhaustiveCheck: never = this.state;
+        throw new Error(`Unhandled LayerState case: ${exhaustiveCheck}`);
+      }
+    }
```
shlomnissan:
```suggestion
    if (this.state === "initialized") {
      this.load();
    }
```

packages/core/src/layers/image_stack_layer.ts
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  public setZIndex(index: number) {
+    if (this.state !== "ready") {
+      // TODO: error instead?
+      console.warn(`Trying to set Z index before ready: ${this.state}`);
```
shlomnissan: Is this an expected error? If so, throwing an exception (TODO) may not be the best approach. Ideally, the user should see a loading spinner when requesting a Z-index that isn’t ready. Alternatively, we could prevent them from making the request in the first place by restricting the UI, in which case an exception may be appropriate.

packages/core/src/layers/image_stack_layer.ts
```diff
@@ -0,0 +1,183 @@
...
+    }
+    if (index < 0 || index >= this.dataChunks_.length) {
+      throw new Error(
+        `Z index ${index} is out of bounds [0, ${this.dataChunks_.length - 1}]`
+      );
+    }
+    this.setZIndex_(index);
+  }
+
+  private setZIndex_(index: number) {
```
shlomnissan: You might not need both `setZIndex` and `setZIndex_` as separate functions. Since `setZIndex_` is only called internally, it could likely be merged into `setZIndex`, unless there's a strong reason to keep them apart. Otherwise, consolidating them would simplify the API and reduce unnecessary indirection. Thoughts?

packages/core/src/layers/image_stack_layer.ts
```diff
@@ -0,0 +1,183 @@
...
+      const zLoc = zStart + slice * zScale;
+      sliceRegion[sliceRegion.length - 1].index = zLoc;
+      // Queue loading this slice
+      loadPromises.push(
+        loader.loadChunk(sliceRegion, this.scheduler_).then((chunk) => {
+          // Store the chunk at the correct index
+          this.dataChunks_[slice] = chunk;
+          console.debug(`Loaded Z slice ${zLoc} (array index ${slice})`);
+
+          // If this is the first slice to load, we mark the layer as ready
```
shlomnissan: When you say "the first slice to load," do you mean the first slice in sequence or simply the first one that finishes loading? If we need to ensure a specific starting slice, should we enforce ordering here? It might be worth explicitly handling this case if predictable behavior is important.

packages/core/src/layers/image_stack_layer.ts
```diff
@@ -0,0 +1,183 @@
...
+
+    // Wait for all slices to finish loading
+    await Promise.all(loadPromises).catch((error) => {
+      if (error instanceof AbortError) {
+        console.debug("Loading aborted.");
+        return;
+      }
+      console.error("Error loading Z stack", error);
+      throw error;
+    });
```
shlomnissan: Should we handle partial failures more gracefully? Right now, if any slice fails (except for an aborted load), we log the error and throw, potentially preventing any successfully loaded slices from being used. Depending on the use case, we might want to allow partial loading or retry failed slices instead of failing the entire operation. What do you think?

packages/react/src/components/App.tsx
```diff
@@ -3,12 +3,14 @@ import OmeZarrImageViewer from "./OmeZarrImageViewer";
...
+  // "http://localhost:8000";
+  "https://public.czbiohub.org/organelle_box/datasets/A549/organelle_box_crop_v1.zarr";
 // const imageUrl = plateUrl + "/B/03/0";
-const imageUrl = plateUrl + "/GOLGA2/Live/000000";
+const imageUrl = plateUrl + "/GOLGA2/Live/000002";
 const region = [
   { dimension: "T", index: 0 },
-  { dimension: "Z", index: 0 },
+  // { dimension: "Z", index: 10.4 },
+  // { dimension: "Z", index: { start: 0, stop: 11 } },
```
shlomnissan: I assume this component is temporary, but before merging, let's remove any unnecessary comments to keep the code clean and focused.

packages/react/src/components/OmeZarrImageViewer.tsx
```diff
@@ -15,31 +16,37 @@ import {
...
 import { ChannelControlsList } from "./controls/ChannelControlsList";
 import { ChannelControlProps } from "./controls/ChannelControl";
+import { ZControl } from "./controls/ZControl";
 import { hexToRgb } from "lib/color";
 
 interface OmeZarrImageViewerProps {
   sourceUrl: string;
   region: Region;
   scale?: number;
+  zDimension?: string;
```
shlomnissan, line 26: Is this prop meant to be a label representing the z-dimension? If so, should we rename it to something clearer, like `zDimensionLabel`? I wouldn’t typically expect a prop named `zDimension` to store an arbitrary string—it sounds more like it would hold a numerical value or an index.

packages/react/src/components/OmeZarrImageViewer.tsx
```diff
@@ -55,7 +62,17 @@ export default function OmeZarrImageViewer({
       const omeroChannels = await loadOmeroChannels(sourceUrl);
       const channelProps = omeroToChannelProps(omeroChannels);
       setControlProps(omeroToControlProps(omeroChannels));
-      const layer = new ImageLayer({ source, region, channelProps });
+      let layer;
+      if (zDimension === undefined) {
```
shlomnissan, line 66: Would `if (!zDimension)` be a more robust check here? It would catch cases where `zDimension` is null, undefined, or an empty string, making the condition more resilient. What do you think?

packages/react/src/components/OmeZarrImageViewer.tsx
```diff
@@ -77,6 +94,20 @@ export default function OmeZarrImageViewer({
...
+        const loader = await source.open();
+        const attributes = await loader.loadAttributes();
+        const zAxisIndex = attributes.dimensions.findIndex(
+          (dim) => dim === zDimension
+        );
+        setZRange([0, attributes.shape[zAxisIndex] - 1]);
+      };
+      setZRangeFromData();
+    }
+  }, [source, zDimension]);
```
shlomnissan, line 110: A little bit of refactoring to make this function more robust:

- Prevents state updates if the component unmounts.
- Includes try/catch and error handling.

Also, I think we need to provide a type for `dim`? It might be implicitly typed as `any`?

```suggestion
useEffect(() => {
    if (!zDimension || !source) return;

    let isMounted = true;
    const setZRangeFromData = async () => {
      try {
        const loader = await source.open();
        const attributes = await loader.loadAttributes();
        const zAxisIndex = attributes.dimensions.findIndex((dim) => dim === zDimension);

        if (zAxisIndex === -1) {
          console.warn(`zDimension "${zDimension}" not found in dimensions:`, attributes.dimensions);
          return;
        }

        if (isMounted) {
          setZRange([0, attributes.shape[zAxisIndex] - 1]);
        }
      } catch (error) {
        console.error("Error loading Z range:", error);
      }
    };

    setZRangeFromData();

    return () => isMounted = false;
  }, [source, zDimension]);
```

packages/react/src/components/OmeZarrImageViewer.tsx
```diff
@@ -104,10 +135,28 @@ export default function OmeZarrImageViewer({
...
+      {zDimension && (
+        <div
+          className={cns(
+            "absolute",
+            "bottom-0",
+            "right-3",
+            "w-1/2",
+            "px-5",
+            "py-3"
+          )}
```
shlomnissan, line 152: I think this functional component should not include any CSS. Instead, the styling should be handled by the component that imports it (likely a composition component). This keeps the component more reusable and ensures styling concerns are properly separated. The same applies to other embedded classes in React components throughout this PR. Would you agree?

packages/react/src/components/controls/ContrastSlider.tsx
```diff
@@ -18,12 +18,18 @@ export function ContrastSlider({
     onChange([limits[0], limits[1]]);
   };
 
+  let step = (max - min) / 512.0;
```
shlomnissan, line 21: Why 512.0? It seems a bit arbitrary. Could you assign it to a well-named constant to clarify its purpose? That would make the intent more explicit and improve maintainability.

```suggestion
  const isInteger = Number.isInteger(value[0]) && Number.isInteger(value[1]);
  const step = isInteger ? Math.max(1, Math.floor((max - min) / 512)) : (max - min) / 512.0;
```

packages/react/src/components/controls/ZControl.tsx
```diff
@@ -0,0 +1,41 @@
+import cns from "classnames";
+import { InputSlider } from "@czi-sds/components";
+
+interface ZControlProps {
+  zRange: [number, number];
+  onChange: (zIndex: number) => void;
+  disabled: boolean;
+}
+
+export function ZControl({ zRange, onChange, disabled }: ZControlProps) {
```
shlomnissan, line 10: I’m not sure what value this component adds. Similar to my previous comment, I don’t think we should include components that embed CSS-related classes—the client should handle styling. If possible, I’d suggest using `InputSlider` directly and removing this component to keep things simpler and more maintainable. What do you think?

packages/core/examples/image_stack_from_omezarr3d_f32/main.ts
```diff
@@ -0,0 +1,159 @@
...
+        layer.setZIndex(value);
+        zIndexEl.textContent = value.toString();
+      } catch {
+        console.debug("Tried to set Z index out of bounds");
+      }
+    });
+
+    if (layer.extent !== undefined) {
+      camera.setFrame(0, layer.extent.x, 0, layer.extent.y);
+      renderer.setControls(new PanZoomControls(camera, camera.position));
```
shlomnissan, line 141: I think this API introduces unnecessary coupling between rendering and interaction. The renderer shouldn’t be responsible for handling user controls—this logic should be encapsulated within the camera, similar to how Three.js structures camera controls.

shlomnissan (approved):
Great job! This PR is looking solid—most of my comments are just nits.

My main concern is whether we really need a separate layer. I’d like us to explore generalizing the image series layer to cover both use cases instead. I’m not blocking this PR, but if you agree, we should prioritize a quick patch to merge them before moving on to new features.

Also, the React components feel out of place in this repository. Even if we decide to keep them here for now, we should clearly distinguish between functional components (handling logic) and composition components (handling styles). We discussed this in our last web viewer eng meeting, but I’ve pointed it out in the review as well.

Let me know what you think!

packages/core/examples/image_stack_from_omezarr3d_f32/main.ts
aganders3: lol fair enough - this just keeps typescript happy without putting ? all over.

packages/core/examples/image_stack_from_omezarr3d_f32/main.ts
aganders3: I agree, especially given what I would say "nothing is something" in this case. I do think as "region" is using scaled units it's valuable to have something to say "load this whole dimension" without having to do calculations in advance. This is also why `[zMin, zMax]` doesn't work - you need something like `[zMin * zScale, zMax * zScale]`.

Right now region is expects a list of indices specified as `{ dimension: "X", index: 0 }` or `{ dimension: "X", index: { start: 0, stop: 10.2 } }`. I could imagine something like `{ dimension: "Z", index: Region.FULL }` or something. The idiomatic approach if we follow numpy, zarrita, etc. would be to pass something nullish for the index.

packages/core/examples/image_stack_from_omezarr3d_f32/main.ts
aganders3: I can remove this - it's from debugging.

packages/core/examples/image_stack_from_omezarr3d_f32/main.ts
aganders3: We can revisit, but the renderer owns the DOM element where the callbacks are attached. This is why Three.js takes the renderer's DOM element anyway, which I find kind of clunky.

packages/core/examples/image_stack_from_omezarr3d_f32/index.html
aganders3: I agree and you've made this comment before, so we already have an issue for it :)

https://github.com/chanzuckerberg/imaging-active-learning/issues/61

packages/react/src/components/OmeZarrImageViewer.tsx
aganders3: Yeah you made a similar comment about checking `layer.extent !== undefined` as well. I sort of like the explicit check because if someone passes `zDimension: null` or `zDimension: ""` I'm not really sure what they're expecting, and I'd probably prefer to get an error from the result.

That said this doesn't explicitly throw an error for that and I'm not sure the error handling on `ImageStackLayer` is appropriately robust for it either.

Part of this could also be my relative inexperience with JS, I still haven't really developed strong intuition around falsy and nullish values.

packages/react/src/components/OmeZarrImageViewer.tsx
aganders3: I'm really unfamiliar with best practices here, so I trust you on this. I understand the value of how flexible this would be, but I'm confused by how to understand the API for setting all the required formatting in an importing component, but maybe that's just a documentation issue?

packages/react/src/components/OmeZarrImageViewer.tsx
aganders3: I'm not sure what the try/catch is expecting to catch here. I think the check for `zAxisIndex === -1` is reasonable but should perhaps throw an error as well.

My immediate reaction is the check for `isMounted` feels like noise and premature optimization, but maybe I'm misunderstanding. I'm fairly new to react so don't quite understand when components mount/unmount. This seems like it would prevent one state update if the component unmounts shortly after updating `source` or `zDimension` after this effect is called. Would you recommend this `isMounted` check for any `useEffect` that updates state?

packages/react/src/components/controls/ContrastSlider.tsx
aganders3: It is arbitrary - do you have a recommendation? The idea is to just divide to provide a sufficiently small step size to make the slider feel "smooth" rather than notched.  I guess we could set `const numSteps = 512;` or something?

Good idea to make sure the step is non-zero though 👍.

packages/react/src/components/controls/ZControl.tsx
aganders3: Yeah, if the styling goes somewhere else that makes sense, otherwise I think it makes the parent quite crowded.

packages/core/src/layers/image_stack_layer.ts
aganders3: Yeah, I don't think we should start implementing retry logic right now but it seems reasonable to use `allSettled` and log the errors instead of throwing.

packages/core/src/layers/image_stack_layer.ts
aganders3: Extent represents the coordinates of the layer's bounding box in world space, such that you can use it to set/reset an initial camera frame for example.

I agree it will be useful on all layers. #192 is open to consider this API further because right now it's not particularly considered.

packages/core/src/layers/image_stack_layer.ts
aganders3: This is part of resolving the definition of the states. Even in the case of having a loading spinner I think an exception might be the right feedback, since I presume that would happen in the UI and not the rendering context.

With that in mind I would perhaps expect different exceptions for a zIndex that's not ready and a zIndex that is out of the expected range.

packages/core/src/layers/image_stack_layer.ts
aganders3: I think the way to change this is to factor out the block under `this.texture_ === null` so the private method is the only one ever initializing the texture.

packages/core/src/layers/image_stack_layer.ts
aganders3: So far this is just the first that finishes loading.

Do you have an idea for an API to prioritize certain slices?

One thing this brings up for me (I had the same question with ImageSeriesLayer) is that we have the zIndex held as implicit state, and additionally as the slices load the indices may be available at different times with no indication of which are "ready". This means `setZIndex` may result in uncaught errors.

packages/core/examples/image_stack_from_omezarr3d_f32/main.ts
aganders3: It can be removed here because it's effectively the default right now. I added this to provide some control over which scale is loaded. Previously this only supported loading the lowest-resolution from a multiscale image.

packages/core/src/data/ome_zarr_image_loader.ts
aganders3: I think this is left from #206 which I started in the same branch. I'm pretty sure it's needed there because my types are not correct in that branch and typescript complains about accessing `constructor.name` for some reason.

packages/core/examples/image_stack_from_omezarr3d_f32/main.ts
aganders3: Thanks - I'll check this along with the suggested code changes in this file.

packages/core/src/layers/image_stack_layer.ts
aganders3: I'd like to start with clear definition of what a layer is to guide us. It's not quite so clear to me that these should not be separate layers, but I also recognize the overlap right now. Similarly I think `SingleMeshLayer` and `Imagelayer` should be considered at the same time if we're going to unify layers of 2D image representations.

Mostly this was a faster way to get to shipping the OrganelleBox features without concern of breaking the Ultrack prototype, but I appreciate that's generally not the way we want to work! I think the changes here are improvements so I'd expect to merge the two implementations and make sure the Ultrack prototype still works.

aganders3 (commented):
Thanks! This is all great input. I'm going to integrate some of the suggested changes particularly in the example and the loader. I think for a lot of the other things we should create issues and/or add them to our agenda to discuss in person.

aganders3:
Closing in favor of #217

PR #208, opened 2025-03-13
[WIP] Add two-pass loading to OmeZarrImageViewer component

aganders3:
Based on #207 - merge after! [Review the diff from that brach.](https://github.com/chanzuckerberg/imaging-active-learning/compare/aganders3/image-stack-layer...aganders3/progressive-loading-punt)

### Summary
This implements an (in my opinion) kind of hacky version of progressive loading purely in the React space.

If the component is provided with a non-zero `scale` input, it will first load that scale. When that is finished it will kick off loading the zero-scale (highest resolution) data in the background. When the high-res layer is "ready", it will snap into place replacing the low-res layer. The loading indicator will remain (overlaid on the image) until all data are loaded.

There is still an issue with camera responsiveness while loading I can look into.

One more small bug - it can flicker briefly when switching from the low to high res image.

https://github.com/user-attachments/assets/0b80b6aa-84bb-45f5-a74f-580ff08ba210

### Related Issue
Closes #<issue-number>


kyleawayan:
Hi Ashley, this looks great. Implemented it into the portal and it looks like my browser is still downloading 2GB+ worth of zarr files, might be a bug (at least with GOLGA2) https://organelle-box-portal-git-latest-idetik-czbiohub.vercel.app?_vercel_share=8bXeFbzXzzMOotbUDnkF16AD9fQzqBxM

Other than that, I'm worried that dumping 100MB+ without some sort of warning may be bad. Is it possible that we can do the first pass (low resolution) loading, but also initially have a button that says "Load High Resolution (200MB)" or something like that?

aganders3:
lol 2GB+ without warning is pretty bad. I think we're using different sources. It looks like you're using 
https://public.czbiohub.org/organelle_box/datasets/A549/organelle_box_v1.zarr but you can use https://public.czbiohub.org/organelle_box/datasets/A549/organelle_box_crop_v1.zarr which should be much smaller.

We can definitely move this behind a button as well - likely even easier as we don't have to deal with triggering off the signals from the first load!

kyleawayan:
Sounds great! Didn't notice the difference, will update the base URL

aganders3:
Closing to rebase work on main with #217

I will base a new component on Janeece's design including the "load higher res" button + moved loading indicator.

The only specific piece we may need from this PR is making SDS/MUI "external" in the vite config.

PR #210, opened 2025-03-20
(Easy) Remove SingleMeshLayer and Texture2D

shlomnissan:
### Summary

This PR removes `SingleMeshLayer` and `Texture2D`, both of which were originally
used for rendering images in the early stages of development. While a
"single mesh layer" will be necessary in the future for supporting arbitrary
geometry, this specific implementation is no longer needed.

### Note
Regarding a dedicated 2D texture module, we already have `DataTexture2D`, which serves the intended purpose. I suggest renaming it to `Texture2D` in a quick follow-up PR for consistency.

### Tests & Checks


aganders3 (approved):
good lookin'

PR #211, opened 2025-03-21
refactor(react): reorganize components with barrel files

phoenixAja:
# Restructure React Components into Library Format

Having this structure in place will make it easier to implement more control components needed for the segmentation viewer.

## Changes
- Reorganized the React components into a proper component library structure
- Created dedicated directories for viewer components and their subcomponents
- Added index.tsx files for each component directory to provide clean public APIs and avoid deep imports
- Removed default exports in favor of named exports for better tree-shaking

### Related Issue
Closes #<issue-number>

### Tests & Checks
- All existing functionality remains unchanged
- Components can now be imported through their respective index files
- Viewer continues to work with remote Zarr sources


packages/react/src/components/App.tsx
```diff
@@ -1,11 +1,10 @@
...
-import OmeZarrImageViewer from "./OmeZarrImageViewer";
+import { OmeZarrImageViewer } from "components/viewers/OmeZarrImageViewer";
 
-const plateUrl =
-  // "http://localhost:8080/20200812-CardiomyocyteDifferentiation14-Cycle1_mip.zarr";
-  "http://localhost:8000";
-// const imageUrl = plateUrl + "/B/03/0";
+const plateUrl = "https://public.czbiohub.org/organelle_box/datasets/A549/organelle_box_crop_v1.zarr";
 const imageUrl = plateUrl + "/GOLGA2/Live/000000";
+// const plateUrl = "http://localhost:8080/20200812-CardiomyocyteDifferentiation14-Cycle1_mip.zarr";
```
phoenixAja, line 7: left this for now in case we want to test with it in the future?

packages/react/src/components/viewers/OmeZarrImageViewer/OmeZarrImageViewer.tsx
```diff
@@ -0,0 +1,110 @@
+import { useEffect, useState } from "react";
```
phoenixAja: not sure why gh is treating this like a new file, it was moved from the original `OmeZarrImageViewer/index.tsx` file - the only change with this was moving `omeroToChannelProps` and `omeroToControlProps` to a util file

packages/react/src/components/App.tsx
aganders3: I think it's okay to remove this, but I don't mind either way.

packages/react/README.md
```diff
@@ -0,0 +1,111 @@
...
+The development server will automatically load sample data from a publicly hosted Zarr file at:
+`https://public.czbiohub.org/organelle_box/datasets/A549/organelle_box_crop_v1.zarr`
+
+No need to host any data locally to start development.
+
+### Development with Core Package
+
+If you make changes to the `@idetik/core` package, you'll need to:
+1. Rebuild the core package: `npm run build --workspace=@idetik/core`
+2. Rebuild this package: `npm run build --workspace=@idetik/react`
```
aganders3, line 27: I think you're right that you have to rebuild core, but that should trigger a refresh if running `npm run dev` here, right?

Also I think we may want to make `core` external in `react/package.json` but we don't have to do that here.

packages/react/README.md
```diff
@@ -0,0 +1,111 @@
...
+React components for image viewers - wrapping idetik/core functionality
+
+## Getting Started
+
+```bash
+# Install dependencies from root directory
+npm install
+
+# Build dependencies (core package)
+npm run build
```
aganders3: Maybe make it clear this needs to be run from root as well?

packages/react/README.md
```diff
@@ -0,0 +1,111 @@
...
+
+### Development with Core Package
+
+If you make changes to the `@idetik/core` package, you'll need to:
+1. Rebuild the core package: `npm run build --workspace=@idetik/core`
+2. Rebuild this package: `npm run build --workspace=@idetik/react`
+
+This ensures your React components use the latest version of the core package.
+
+## Component Development Guidelines
```
aganders3, line 31: Maybe add a note that pure TypeScript can/should go in `lib`?

aganders3 (approved):
Thanks! This generally looks good to me.

Shlomi has some thoughts on organizing/separating component style from logic, but I don't understand it well enough (will let him comment). Regardless that can wait for another PR.

package.json
```diff
@@ -35,4 +35,4 @@
     "vite-plugin-glsl": "^1.3.0",
     "vitest": "^2.1.9"
   }
-}
+}
```
shlomnissan, line 38: Is this change adding or removing whitespace? You might already be doing this, but If you're using VSCode, you can use the following to make sure whitespaces are trimmed automatically.
```
"files.trimTrailingWhitespace": true,
"files.trimFinalNewlines": true,
"files.insertFinalNewline": true
```

packages/react/README.md
```diff
@@ -0,0 +1,111 @@
...
+
+### File Structure
```
shlomnissan, line 35: Documenting "component structure" in README files isn’t the most effective way to enforce patterns, in my experience. On a previous team, we added a script that scaffolded new components following the exact structure you proposed (which is solid, by the way -- I used to have strong opinions about it 😂), and found that developers were much more likely to stick to it when tooling supported it. I can't remember what tool we used for scaffolding, but there are a ton of them out there.

This is a no-op -- I don’t think it’s necessary to include it in this PR or project for that matter.

packages/react/README.md
```diff
@@ -0,0 +1,111 @@
...
+Each component should have its own directory with the following structure:
+```
+ComponentName/
+├── index.tsx         # Public API exports
+├── ComponentName.tsx # Main implementation
+├── utils.ts         # Component-specific utilities (if needed)
+└── components/      # Sub-components (if needed)
+```
+
+### Best Practices
```
shlomnissan, line 45: I appreciate the intention behind documenting “best practices,” but I’m not sure a README is the right place for it. We're currently working on establishing engineering standards across teams, so this might be better handled at that level. I'm not sure if SDS already has a React style guide or a scaffolding tool — if not, it could be a great opportunity for someone to take on. If they do have one, we should just link to it here instead.

packages/react/README.md
```diff
@@ -0,0 +1,111 @@
...
+
+1. **Exports**
+   - Use named exports instead of default exports
```
shlomnissan, line 48: I generally agree that named exports are preferable, but when writing style guides, it’s important to explain the reasoning behind the recommendation. I’d also suggest avoiding “good/bad” labels — especially when the difference is more about preference than a strong technical advantage. Using terms like “prefer” and “avoid” tends to be clearer and more balanced.

packages/react/README.md
```diff
@@ -0,0 +1,111 @@
...
+   - Use descriptive prop names
+   ```typescript
+   interface MyComponentProps {
+     sourceUrl: string;
+     onUpdate: (data: UpdateData) => void;
+     isLoading?: boolean;
+   }
+   ```
+
+4. **State Management**
```
shlomnissan: This particular “best practice” feels a bit too general. State management across components is a nuanced topic, and "keep state as local as possible" might be oversimplifying it. Happy to share some of my thoughts on this if you're interested!

packages/react/src/components/viewers/OmeZarrImageViewer/utils.ts
```diff
@@ -0,0 +1,32 @@
+import { OmeroChannel, ChannelProps } from "@idetik/core";
+import type { ChannelControlProps } from "./components/ChannelControlsList/components/ChannelControl";
+import { hexToRgb } from "../../../lib/color";
```
shlomnissan: I prefer using absolute paths over relative ones whenever possible — it improves readability, avoids messy and error-prone `../../../` chains, and makes refactoring simpler. I believe we can set up path aliases in Vite or TypeScript to support this, but I’m a bit rusty on the exact setup since it’s been a while since I worked on client code.

packages/react/src/components/viewers/OmeZarrImageViewer/utils.ts
```diff
@@ -0,0 +1,32 @@
...
+      color: hexToRgb(channel.color),
+      contrastLimits: [Math.max(start, min), Math.min(end, max)],
+    };
+  });
+};
+
+export const omeroToControlProps = (
+  omeroChannels: OmeroChannel[]
+): Partial<ChannelControlProps>[] => {
+  return omeroChannels.map((channel: OmeroChannel, index: number) => {
```
shlomnissan: No need to manually annotate the types here — TypeScript should be able to infer them automatically.
```ts
return omeroChannels.map((channel, index) => {
```

packages/react/src/components/viewers/OmeZarrImageViewer/utils.ts
```diff
@@ -0,0 +1,32 @@
...
+      visible: channel.active,
+      color: hexToRgb(channel.color),
+      contrastLimits: [Math.max(start, min), Math.min(end, max)],
+    };
+  });
+};
+
+export const omeroToControlProps = (
+  omeroChannels: OmeroChannel[]
+): Partial<ChannelControlProps>[] => {
```
shlomnissan, line 23: Structural typing within a static type system feels so wrong to me 😅

shlomnissan (approved):
Looks great overall! The only thing I’d consider is removing the “best practices” section — but I’ll leave that up to your judgment.

packages/react/README.md
phoenixAja: oh interesting, i like the idea of a tool to automatically scaffold

packages/react/README.md
phoenixAja: i agree, i'll remove this section

packages/react/README.md
phoenixAja: searching around I can't find any CZI wide react style guides - including on SDS. I agree that it would be nice to have a more central react best practices official documentation somewhere.

packages/react/src/components/viewers/OmeZarrImageViewer/utils.ts
phoenixAja: yep i changed the import

packages/react/src/components/viewers/OmeZarrImageViewer/utils.ts
phoenixAja: In this case, using Partial<ChannelControlProps> makes sense to me since we're only populating a subset of the props. And with the structural typing we can satisfy the expected shape without having to declare a whole new interface.

PR #212, opened 2025-03-25
Require fully-specified region when loading data

aganders3:
### Summary
This updates the `Region` type to be a tagged union that can be an interval, point, or "full" (to span an entire dimension of unknown size). However you now need to specify all dimensions, rather than implicitly loading the full span of any omitted dimension. This does require more a priori knowledge of the saved data, but is more explicit.

In practice you generally needed to know this information anyway, and flexibility for this is better handled at the application layer.

### Related Issue
N/A

### Tests & Checks
Tested all examples, react demo, and ultrack prototype.


packages/core/src/layers/image_series_layer.ts
```diff
@@ -45,9 +45,9 @@ export class ImageSeriesLayer extends Layer {
       );
     }
     const timeIndex = this.region_[this.timeDimensionIndex_].index;
-    if (typeof timeIndex === "number") {
+    if (timeIndex.type === "point" || timeIndex.type === "full") {
       throw new Error(
-        `Time index is a number (${timeIndex}). It should be an interval.`
+        `Time index is a number (${timeIndex}) should be an interval.`
       );
     }
```
aganders3: This layer will support "full" in a future PR.

packages/react/src/components/App.tsx
```diff
@@ -1,15 +1,17 @@
...
+const sourceUrl =
+  "https://public.czbiohub.org/organelle_box/datasets/A549/organelle_box_crop_v1.zarr";
+const wellPath = "GOLGA2/Live";
+const imagePath = "000002";
+const imageUrl = `${sourceUrl}/${wellPath}/${imagePath}`;
+const region: Region = [
+  { dimension: "T", index: { type: "point", value: 0 } },
+  { dimension: "Z", index: { type: "point", value: 0 } },
+  { dimension: "Y", index: { type: "full" } },
+  { dimension: "X", index: { type: "full" } },
```
phoenixAja, line 15: I'm getting this error in the console when trying to run locally:

`Error: Region does not contain an index for dimension C, must provide an index for all dimensions`

aganders3: Oops! I "fixed" something in [aa41589](https://github.com/chanzuckerberg/imaging-active-learning/pull/212/commits/aa415895409e9c9ae60b14c162c755f43ed1d8de) and didn't re-check the react example. The good news is that means it's working.

aganders3: Thanks for testing - can you try again? I also fixed the Ultrack prototype that had a similar issue.

phoenixAja: it works now!

phoenixAja (approved):
looks good!

PR #217, opened 2025-03-26
ImageSeriesLayer refactor

aganders3:
### Summary
This replaces #207 and instead of adding a new layer, improves the `ImageSeriesLayer` to support scaled dimensions.

I also changed the API to allow sync calls to set the slice either by index or by "position" (scaled units within the specified region) with `setIndex` and `setPosition`. I prefer this because it replaces the need to put `setIndex` in a callback that listens for a state change.

This uses a loading token to make sure that the most recent `setIndex` or `setPosition` call is respected if data needs to be loaded asynchronously. This enables lazy-loading slices as they are selected. To mimic the previous behavior you can use `preloadSeries` - this will still show slices as they are ready instead of waiting for all to load, which I think is an improvement.

I added a new example for this, but it pretty much replaced the old CryoET example so I removed that one. The example (especially "Load all slices") is still _way_ slower than it needs to be without #206 or something like it.

### Related Issue
Obviates #207 
Merge after #212 ([compare](https://github.com/chanzuckerberg/imaging-active-learning/compare/aganders3/full-region...aganders3/image-series-refactor))

### Tests & Checks
Tested all examples, react demo, and ultrack prototype


packages/core/src/data/image_chunk.ts
```diff
@@ -45,6 +45,13 @@ export type ImageChunkSource = {
   open(): Promise<ImageChunkLoader>;
 };
 
+export type LoaderAttributes = {
+  dimensions: string[];
+  shape: readonly number[];
+  scale: readonly number[];
+};
```
aganders3: This is minimum data I needed for now, but I suspect this will change in the future (i.e. for multiscale data).

packages/core/src/layers/image_series_layer.ts
```diff
@@ -62,81 +79,214 @@ export class ImageSeriesLayer extends Layer {
...
   public close(): void {
     this.scheduler_.shutdown();
   }
 
-  private async load() {
-    if (this.state !== "initialized") {
-      throw new Error(`Trying to open chunk loader more than once.`);
+  private async loadSeriesAttributes() {
+    if (this.seriesAttributes_) {
+      return this.seriesAttributes_;
```
aganders3, line 148: For the time being `seriesAttributes` does not change, but I'm not sure if that will hold in the future (with multiscale).

packages/react/src/components/viewers/OmeZarrImageViewer/OmeZarrImageViewer.tsx
```diff
@@ -1,8 +1,10 @@
 import { useEffect, useState } from "react";
 import cns from "classnames";
 import CircularProgress from "@mui/material/CircularProgress";
+import { InputSlider } from "@czi-sds/components";
 import {
   ImageLayer,
+  ImageSeriesLayer,
```
aganders3, line 6: This supports both single-image and z-stack, but I think we can simplify it for organellebox as they always want a stack.

(in a future PR)

packages/ultrack-prototype/frontend/components/Renderer.tsx
```diff
@@ -86,6 +75,7 @@ export default function Renderer({
     const onStateChange = (newState: LayerState) => {
       if (newState === "ready") {
         onReady();
+        imageSeriesLayer.removeStateChangeCallback(onStateChange);
```
aganders3, line 77: "ready" state is now triggered more often - this prevents resetting the camera when changing slices

packages/core/examples/image_series_from_omezarr5d_u8/main.ts
```diff
@@ -47,24 +46,22 @@ const channelProps = [
...
   channelProps,
 });
 layerManager.add(layer);
 
 const slider = document.querySelector<HTMLInputElement>("#slider");
 if (slider === null) throw new Error("Time slider not found.");
-slider.min = timeInterval.start.toString();
-slider.max = (timeInterval.stop - 1).toString();
+slider.min = `${timeInterval.start}`;
+slider.max = `${timeInterval.stop - 1}`;
```
phoenixAja, line 57: much nicer 😄

packages/core/examples/image_stack_from_omezarr3d_f32/main.ts
```diff
@@ -0,0 +1,164 @@
...
+  OrthographicCamera,
+  Region,
+  WebGLRenderer,
+} from "@";
+import { PanZoomControls } from "@/objects/cameras/controls";
+
+const url =
+  "https://files.cryoetdataportal.cziscience.com/10444/24apr23a_Position_12/Reconstructions/VoxelSpacing4.990/Tomograms/100/24apr23a_Position_12.zarr";
+const layerManager = new LayerManager();
+const renderer = new WebGLRenderer("#canvas");
```
shlomnissan, line 16: I remember you mentioned that Three.js injects a `<canvas>` element if one isn’t provided — seems like a nice, small improvement we could include. Obviously not directly related to this PR.

packages/core/examples/image_stack_from_omezarr3d_f32/main.ts
```diff
@@ -0,0 +1,164 @@
...
+const camera = new OrthographicCamera(0, 128, 0, 128);
```
shlomnissan, line 17: Just curious — are these dimensions arbitrary, or do they align with the underlying data?

packages/core/examples/image_stack_from_omezarr3d_f32/main.ts
```diff
@@ -0,0 +1,164 @@
...
+const attributes = await loader.loadAttributes();
+const zDimName = "z";
+const zAxisIndex = attributes.dimensions.findIndex((dim) => dim === zDimName);
+const zMin = 0;
+const zMax = attributes.shape[zAxisIndex];
+const region: Region = [
+  { dimension: zDimName, index: { type: "full" } },
+  { dimension: "x", index: { type: "full" } },
+  { dimension: "y", index: { type: "full" } },
+];
```
shlomnissan, line 33: This API is a big improvement — much cleaner!

packages/core/examples/image_stack_from_omezarr3d_f32/main.ts
```diff
@@ -0,0 +1,164 @@
...
+
+// Initial contrast limits for grayscale electron microscopy data
+const initialMinValue = -0.00001;
+const initialMaxValue = 0.00001;
+
+// This dataset is grayscale electron microscopy data
+const channelProps = [
+  {
+    visible: true,
+    color: [1, 1, 1] as [number, number, number],
```
shlomnissan: I think we should introduce a `Color` type at a more global level — even if it’s currently identical to `RgbColor`. (I don’t think `RgbColor` is exported right now.) That way, we have a clear, centralized type that we can extend in the future if needed. I’d also avoid casting into user-defined types — instead, just export and import the type directly where it’s needed. It keeps things clearer and avoids unnecessary assertions.

packages/core/examples/image_stack_from_omezarr3d_f32/main.ts
```diff
@@ -0,0 +1,164 @@
...
+    contrastLimits: [initialMinValue, initialMaxValue] as [number, number],
```
shlomnissan: Do we have a shared `Range` type (like [min, max]) that we can use for these min/max pairs? Might be worth introducing one if not — it could help standardize things across the codebase.

packages/core/examples/image_stack_from_omezarr3d_f32/main.ts
```diff
@@ -0,0 +1,164 @@
...
+  },
+];
+
+const layer = new ImageSeriesLayer({
+  source,
+  region,
+  seriesDimensionName: zDimName,
```
shlomnissan, line 51: I think `seriesAxis` or `scrollAxis` would be clearer — they emphasize that this is the "scrollable axis" of the series. That said, you probably have a better sense of naming conventions in this domain, so I’ll defer to your judgment.

packages/core/examples/image_stack_from_omezarr3d_f32/main.ts
```diff
@@ -0,0 +1,164 @@
...
+  !zSlider ||
+  !zIndexEl ||
+  !zTotalEl ||
+  !minSlider ||
+  !maxSlider ||
+  !minValueEl ||
+  !maxValueEl
+) {
+  throw new Error("Could not find all necessary elements");
+}
```
shlomnissan: If you want to satisfy TypeScript without boilerplate, you can use the non-null assertion operator (`!`).

```suggestion
const zSlider = document.querySelector<HTMLInputElement>("#z-slider")!;
const zIndexEl = document.querySelector<HTMLSpanElement>("#z-index")!;
const zTotalEl = document.querySelector<HTMLSpanElement>("#z-total")!;
const minSlider = document.querySelector<HTMLInputElement>("#min-slider")!;
const maxSlider = document.querySelector<HTMLInputElement>("#max-slider")!;
const minValueEl = document.querySelector<HTMLSpanElement>("#min-value")!;
const maxValueEl = document.querySelector<HTMLSpanElement>("#max-value")!;
const stateEl = document.querySelector<HTMLSpanElement>("#layer-state")!;
const loadAllButton = document.querySelector<HTMLButtonElement>("#load-all")!;
```

packages/core/examples/image_stack_from_omezarr3d_f32/main.ts
```diff
@@ -0,0 +1,164 @@
...
+
+// Initialize sliders
+zSlider.min = `${zMin}`;
+zSlider.max = `${zMax - 1}`;
+zSlider.value = "0";
+zTotalEl.textContent = `${zMax - zMin - 1}`;
+
+minSlider.min = "-0.000010";
+minSlider.max = "0";
+minSlider.step = "0.000001";
```
shlomnissan, line 74: I think you could use the `initialMinValue` and `initialMaxValue` you defined at the top — unless this slider is referring to a different range.

packages/core/examples/image_stack_from_omezarr3d_f32/main.ts
```diff
@@ -0,0 +1,164 @@
...
+  { dimension: "x", index: { type: "full" } },
+  { dimension: "y", index: { type: "full" } },
+];
+
+// Initial contrast limits for grayscale electron microscopy data
+const initialMinValue = -0.00001;
+const initialMaxValue = 0.00001;
+
+// This dataset is grayscale electron microscopy data
+const channelProps = [
```
shlomnissan: Could you specify the underlying type (`ChannelProps`) directly to eliminate the need for casting later on? It’ll help make the intent clearer and keep things type-safe.

packages/core/examples/image_stack_from_omezarr3d_f32/main.ts
```diff
@@ -0,0 +1,164 @@
...
+      visible: true,
+      color: [1, 1, 1],
+      contrastLimits: [minValue, maxValue],
+    },
+  ]);
+});
+
+// set up event handler with debouncing
+let debounce: number;
+zSlider.addEventListener("input", (event) => {
```
shlomnissan, line 116: Is debouncing necessary in the context of this example? Also, it might be a good idea to introduce a shared debouncing utility so we have a standardized approach across the codebase.

packages/core/examples/image_stack_from_omezarr3d_f32/main.ts
```diff
@@ -0,0 +1,164 @@
...
+    layer.setIndex(value);
+    zIndexEl.textContent = `${value}`;
+  }, 20);
+});
+
+layer.setIndex(zSlider.valueAsNumber);
+const setCameraFrame = (newState: LayerState) => {
+  if (newState === "ready" && layer.extent !== undefined) {
+    camera.setFrame(0, layer.extent.x, 0, layer.extent.y);
+    renderer.setControls(new PanZoomControls(camera, camera.position));
```
shlomnissan, line 129: I know I mentioned this in the previous PR, but I may have missed your response. Decoupling the camera controls from the renderer makes a lot of sense to me. If you're on the same page, maybe we can create an issue for it. Otherwise, I’d love to hear more about your reasoning.

packages/core/examples/image_stack_from_omezarr3d_f32/main.ts
```diff
@@ -0,0 +1,164 @@
...
+    camera.update();
+    // remove the callback to only set the camera frame once
+    layer.removeStateChangeCallback(setCameraFrame);
+  }
+};
+layer.addStateChangeCallback(setCameraFrame);
```
shlomnissan, line 135: This callback feels awkward because it registers itself only to immediately remove itself, creating an implicit one-time behavior that's easy to overlook. It tightly couples camera and renderer logic to a specific layer state, making the flow harder to reason about.

I don't know what the right solution is, maybe something like `layer.onceStateChange("ready", ...)`. It's not a blocker, let me know what you think.

packages/core/examples/image_stack_from_omezarr3d_f32/main.ts
```diff
@@ -0,0 +1,164 @@
...
+layer.addStateChangeCallback((newState: LayerState) => {
+  stateEl!.textContent = newState;
+});
+
+loadAllButton?.addEventListener("click", () => {
```
shlomnissan: If you use the non-null assertion operator as mentioned above, you could remove the optional operator.

packages/core/examples/tracks/main.ts
```diff
@@ -115,17 +115,17 @@ const slider = document.querySelector<HTMLInputElement>("#slider");
...
-  if (newState === "ready") {
-    slider.addEventListener("input", (event) => {
-      const value = (event.target as HTMLInputElement).valueAsNumber;
-      imageSeriesLayer.setTimeIndex(value);
-      lineLayer.setTimeIndex(value);
-    });
-    imageSeriesLayer.setTimeIndex(slider.valueAsNumber);
-    lineLayer.setTimeIndex(slider.valueAsNumber);
-  }
+imageSeriesLayer.preloadSeries({
```
shlomnissan: This feels a little awkward for me. The name `preloadSeries` suggests it loads the entire series, but the `initialIndex` parameter implies a priority or focus within that preload. This makes the intent a little unclear. I can't think of a better pattern atm, but renaming things can make the intent clearer. For example:

```ts
imageSeriesLayer.preloadSeriesWithPriority({
  priorityIndex: [slider.valueAsNumber - timeInterval.start],
});
```

packages/core/src/data/ome_zarr_image_loader.ts
```diff
@@ -123,6 +118,13 @@ export class OmeZarrImageLoader {
       );
     }
 
+    const axes = image.axes;
+    const scale = dataset.coordinateTransformations[0].scale;
+    const translation =
+      dataset.coordinateTransformations.length === 2
+        ? dataset.coordinateTransformations[1].translation
+        : new Array(axes.length).fill(0);
```
shlomnissan: These variables are currently defined in both `loadChunk` and `regionToIndices`. Could we extract this logic into a dedicated member function that returns the transform information (axes, scale, translation)? If there's a reason for the differences in how these values are calculated, I’d recommend using more descriptive variable names to make those distinctions clear.

packages/core/src/data/ome_zarr_image_loader.ts
```diff
@@ -147,44 +149,58 @@ export class OmeZarrImageLoader {
     return chunk;
   }
 
-  public get metadata(): OmeNgffImage {
-    return this.metadata_;
+  public async loadAttributes(): Promise<LoaderAttributes> {
```
shlomnissan: What’s the distinction between “reading metadata” and “loading attributes”? The terminology feels a bit ambiguous here—are these separate steps, or just different names for the same thing?

Also, this member function includes a computed property that appears multiple times in this file
```ts
const scale = dataset.coordinateTransformations[0].scale;
```

It’s calculated three times, which might indicate an opportunity to extract it into a computed property member.

For context, I had started ramping up on OME-Zarr before going on family leave, but much of that knowledge has faded. That said, on a quick pass, I do get the sense that there’s some duplication or at least naming that could be clarified to better reflect the underlying responsibilities.

One thing I’m trying to understand is whether reading metadata and loading attributes can be merged into a single step that’s shared across the class’s member functions. If they’re conceptually distinct, that’s fine—but if there’s overlap, consolidating them could simplify the code and reduce redundancy.

packages/core/src/data/ome_zarr_image_loader.ts
```diff
@@ -147,44 +149,58 @@ export class OmeZarrImageLoader {
...
-  region: Region,
-  axes: Array<Axis>,
-  scale: number[],
-  translation: number[]
-): Array<Slice | number> {
-  const indices: Array<Slice | number> = [];
-  for (const [i, axis] of axes.entries()) {
-    const match = region.find((s) => s.dimension == axis.name);
-    if (match) {
+  // Converts a region to indices within an OME-Zarr image array.
```
shlomnissan: Nit: I don’t think this comment adds much value—the method is already named regionToIndices, and it's defined within the context of an OME-Zarr image loader module. The purpose is clear from the name and location alone. While not a big deal, I personally find redundant comments to be more than just noise—they can create subtle forms of coupling.

packages/core/src/layers/image_series_layer.ts
```diff
@@ -10,45 +14,58 @@ import { ImageRenderable } from "objects/renderable/image_renderable";
...
+// Loads 2D+z image data (Z-stack) from an image source into renderable objects.
 export class ImageSeriesLayer extends Layer {
   private readonly source_: ImageChunkSource;
   private readonly region_: Region;
-  private readonly timeInterval_: Interval;
-  private readonly timeDimensionIndex_: number;
+  private readonly seriesDimensionName_: string;
+  private readonly seriesDimensionalIndex_: DimensionalIndex;
+  private readonly scheduler_: PromiseScheduler = new PromiseScheduler(16);
+  private loader_: ImageChunkLoader | null = null;
```
shlomnissan, line 40: Can this be marked `readonly` too?

packages/core/src/layers/image_series_layer.ts
```diff
@@ -62,81 +79,214 @@ export class ImageSeriesLayer extends Layer {
...
+    if (this.state === "initialized") {
+      this.loadSeriesAttributes();
     }
   }
 
-  public setTimeIndex(index: number) {
-    if (this.state !== "ready") {
-      console.warn(`Trying to set time index before ready: ${this.state}`);
+  public setPosition(position: number) {
+    // this calls async from sync, but it is handled in setIndex by the loading token
```
shlomnissan: It's generally risky to hide async behavior inside sync methods—it can lead to surprising control flow. This comment is also vague -- or at least, it doesn't make clear why this pattern is needed. I would suggest rewriting this comment to make it clear, or better yet, refactor the function so the async flow is more explicit.

packages/core/src/layers/image_series_layer.ts
```diff
@@ -62,81 +79,214 @@ export class ImageSeriesLayer extends Layer {
...
+      this.loadingToken_.index === index &&
+      !this.loadingToken_.canceled
+    ) {
+      console.debug("Ignoring duplicate active load index request");
       return;
+    } else if (this.loadingToken_) {
+      console.debug(
+        `Cancelling load for index ${this.loadingToken_.index}, requested index is ${index}`
+      );
+      this.loadingToken_.canceled = true;
```
shlomnissan: I think this code can be cleaned up for readability. Maybe something along the lines of:
```ts
const token = this.loadingToken_;
if (token) {
  if (token.index === index && !token.canceled) {
    console.debug("Ignoring duplicate active load index request");
    return;
  }
  console.debug(`Cancelling load for index ${token.index}, requested index is ${index}`);
  token.canceled = true;
}
```

packages/core/src/layers/image_series_layer.ts
```diff
@@ -62,81 +79,214 @@ export class ImageSeriesLayer extends Layer {
...
-    for (let t = start; t < stop; ++t) {
-      const region = structuredClone(this.region_);
-      region[this.timeDimensionIndex_].index = { type: "point", value: t };
-      loadPromises.push(
-        loader
-          .loadChunk(region, this.scheduler_)
-          .then((chunk) => (this.dataChunks_[t - start] = chunk))
+    const loader = await this.getLoader();
+
+    const attributes = await loader.loadAttributes();
```
shlomnissan, line 153: Since we always load attributes immediately after getting the loader, should we consider preloading attributes as part of `getLoader()`?

packages/core/src/layers/image_series_layer.ts
```diff
@@ -62,81 +79,214 @@ export class ImageSeriesLayer extends Layer {
...
-      }
-    });
+    const seriesDimScale = attributes.scale[seriesIndex];
+    const seriesMax = attributes.shape[seriesIndex] * seriesDimScale;
+
+    if (this.seriesDimensionalIndex_.index.type === "point") {
+      throw new Error(
+        "Series dimension index in region must be a range or 'full', not a point value"
+      );
+    }
```
shlomnissan: Should we throw this exception before extracting `seriesDimScale` and `seriesMax`?

packages/core/src/layers/image_series_layer.ts
```diff
@@ -62,81 +79,214 @@ export class ImageSeriesLayer extends Layer {
...
+    let seriesStart;
+    let seriesStop;
+    if (this.seriesDimensionalIndex_.index.type === "full") {
+      seriesStart = 0;
+      seriesStop = seriesMax;
+    } else {
+      seriesStart = this.seriesDimensionalIndex_.index.start;
+      seriesStop = this.seriesDimensionalIndex_.index.stop;
+    }
```
shlomnissan: This is a bit of a nervous tic from writing a lot of C—where you should never define variables without initializing them. In this case, it's just slightly cleaner.

```suggestion
    let seriesStart = 0;
    let seriesStop = seriesMax;
    if (this.seriesDimensionalIndex_.index.type !== "full") {
      seriesStart = this.seriesDimensionalIndex_.index.start;
      seriesStop = this.seriesDimensionalIndex_.index.stop;
    }
```

packages/core/src/layers/image_series_layer.ts
```diff
@@ -62,81 +79,214 @@ export class ImageSeriesLayer extends Layer {
...
+    };
     this.setState("ready");
+    return this.seriesAttributes_;
+  }
+
+  private async loadAndSetIndex(index: number, token?: LoadingToken) {
+    if (token && !token.canceled && this.state !== "loading") {
+      // if there is no token, we're only loading in the background
+      this.setState("loading");
+    }
```
shlomnissan: You can "encode" the intent behind this comment in code. For example:
```ts
const isForegroundLoad = token && !token.canceled;
if (isForegroundLoad && this.state !== "loading") {
  this.setState("loading");
}
```
Or even better, create a helper like `shouldSetLoadingState(token)` or something more declarative.

packages/core/src/layers/image_series_layer.ts
```diff
@@ -62,81 +79,214 @@ export class ImageSeriesLayer extends Layer {
...
+    const {
+      start: seriesStart,
+      scale: seriesDimScale,
+      length,
+    } = await this.loadSeriesAttributes();
+    if (index < 0 || index >= length) {
+      throw new Error(
+        `Requested index ${index} is out of bounds [0, ${length - 1}]`
+      );
+    }
```
shlomnissan: Nit: Prefer to delay destructuring here—no need to initialize `start` and `scale` if the index is out of bounds.

packages/core/src/layers/image_series_layer.ts
```diff
@@ -62,81 +79,214 @@ export class ImageSeriesLayer extends Layer {
...
+    const chunk = await loader.loadChunk(pointRegion, this.scheduler_);
+
+    // Store the chunk at the correct index
+    this.dataChunks_[index] = chunk;
+    console.debug(
+      `Loaded position ${position} (array index ${index}) for dim ${this.seriesDimensionName_}`
+    );
+
+    // set the data and mark the layer as ready if the token is valid
+    if (token && !token.canceled) {
```
shlomnissan: I think this block can be refactored a little
```ts
if (token) {
  if (token.canceled) {
    console.debug(`...`);
  } else {
    console.debug(`...`);
    this.setData(chunk);
    this.loadingToken_ = null;
  }
}
```
Regarding this log
```
Not setting data for position ${position} (array index ${index}) due to no token
```
It’s currently a debug message, but are we actually expecting the token to be missing in any valid scenario? If not, this might indicate an error case. And if missing the token is valid, the message could be confusing as it implies something went wrong.

Not a big deal—just something that felt a bit unclear on first read.

packages/core/src/layers/image_series_layer.ts
```diff
@@ -62,81 +79,214 @@ export class ImageSeriesLayer extends Layer {
...
+        `Not setting data for position ${position} (arry index ${index}) due to cancellation`
+      );
+    } else {
+      console.debug(
+        `Not setting data for position ${position} (arry index ${index}) due to no token`
+      );
+    }
+  }
+
+  public async preloadSeries({ initialIndex = 0 }: { initialIndex: number }) {
```
shlomnissan: As mentioned in a previous comment, I think the scope of this member function could be clarified or broken down into separate functions for preloading, and preloading with a priority index.

packages/react/src/components/App.tsx
```diff
@@ -28,7 +28,12 @@ export default function App() {
           "gap-4"
         )}
       >
-        <OmeZarrImageViewer sourceUrl={imageUrl} region={region} />
+        <OmeZarrImageViewer
+          sourceUrl={imageUrl}
+          region={region}
+          scale={-2}
```
shlomnissan: What does a `-2` scale mean in this context?

packages/react/src/components/viewers/OmeZarrImageViewer/OmeZarrImageViewer.tsx
shlomnissan: I think simplifying is worth it here. Right now, I treat these components as temporary and don’t think too deeply about them—so it’s better if they stay focused and minimal. If we do want to move toward stable, generalized components, we should take a step back and define a clear strategy first.

packages/core/src/layers/image_series_layer.ts
```diff
@@ -62,81 +79,214 @@ export class ImageSeriesLayer extends Layer {
...
+      `Loaded position ${position} (array index ${index}) for dim ${this.seriesDimensionName_}`
+    );
+
+    // set the data and mark the layer as ready if the token is valid
+    if (token && !token.canceled) {
+      console.debug(
+        `Setting data for position ${position} (array index ${index}) for dim ${this.seriesDimensionName_}`
+      );
+      this.setData(chunk);
+      this.loadingToken_ = null;
```
shlomnissan: Is it possible for the token to be canceled between checks, causing this block to run on a canceled token? I find JavaScript’s async model a bit confusing in this regard. In a multithreaded environment, if `token.canceled` were shared mutable state, this would definitely be a race condition—but that might not apply here. Not entirely sure.

shlomnissan (approved):
First pass looks great! I don’t feel like I have a solid mental model yet of our current loading patterns or the OME-Zarr file format, so this isn’t as deep a review as I’d like—but I also don’t want to block progress. I’ll set aside time next week to dig into the loading logic more thoroughly as part of ramping up for chunking exploration. Most of my comments are minor nits or no-ops. Overall, these changes look solid, and we can revisit broader loading patterns and abstractions later. 🚀

packages/core/examples/image_stack_from_omezarr3d_f32/main.ts
aganders3: These are arbitrary because the camera gets re-set based on the layer extent once data is loaded. I don't like initializing a camera with arbitrary values only to change it, but I also don't love the idea of constructing a camera with an uninitialized state we need to test for. Maybe the compromise is to have some default frame?

packages/core/examples/image_stack_from_omezarr3d_f32/main.ts
aganders3: Yeah the naming is hard here and especially I'm aware of debate around dimension vs. axis, but I tend to use them interchangeably. I was trying to be somewhat consistent with our `Region` type where we use `dimension`.

The OME-NGFF spec refers to "axes" but even they waffle in the description:
> "axes" describes the dimensions of a physical coordinate space. It is a list of dictionaries, where each dictionary describes a dimension (axis)

I know the variable is long but I felt adding "name" implied it was a string. I'll do a little research so we can document a decision and try to be consistent across our code.

packages/core/examples/image_stack_from_omezarr3d_f32/main.ts
aganders3: I totally agree! This is super awkward. I like the idea of another API for a one-time callback. I'll turn this into an issue.

aganders3: #218

packages/core/examples/image_stack_from_omezarr3d_f32/main.ts
aganders3: It's not really necessary for an example. It helps prevent the queue from getting clogged with requests when dragging the slider (vs. clicking to some point). Right now in the loader I think requests get effectively pushed to the back of a queue, but we might want new requests to go to the front (or allow both behaviors).

I agree we should have utilities for both debounce and throttle.

packages/core/examples/image_stack_from_omezarr3d_f32/main.ts
aganders3: This could use more work for sure, so I'm happy to create an issue and we can put discussion there.

aganders3: #219

packages/core/src/layers/image_series_layer.ts
aganders3: I'm also not 100% sure but my current understanding is that in a single-threaded async/await system flow is not interrupted without yielding with `await`. So I think we're safe from a TOCTOU issue as long as we don't `await` between check and use.

As a general comment this is why I find async/await useful in single-threaded context but confusing in multi-threaded.

packages/core/src/layers/image_series_layer.ts
aganders3: Right now the token is used to control two things, depending on cancellation timing. The _request_ could be canceled (if early enough) and _setting the data_ could be canceled (if later). So to load something purely in the background I omit the token, then the request is _never_ canceled, but also the data is _never_ set.

Maybe these should be separated - like by adding another attribute on the token:
```
type LoadingToken = {
  index: number;
  shouldFetchData: boolean;
  shouldSetData: boolean;
};
```

packages/core/src/layers/image_series_layer.ts
aganders3: Good idea - this would also allow preload with a priority _position_ which would be awkward to add here.

packages/core/src/layers/image_series_layer.ts
aganders3: Yeah I was uncomfortable hiding the async behavior but thought it improved the API from the client side. Now I'm not so sure...I think if clients care about waiting for the result they can figure out a way to await. If not they can call the async function and let it run async.

So maybe `setPosition` and `setIndex` should just be async?

packages/core/src/layers/image_series_layer.ts
aganders3: I don't think so, because `source.open()` (which returns the loader) is async. Maybe I don't understand `readonly` though?

packages/core/src/layers/image_series_layer.ts
aganders3: Actually this type can be narrowed in the constructor. Thanks!

packages/core/src/layers/image_series_layer.ts
aganders3: Actually I think I can keep them const instead:
```suggestion
    const indexIsFull = this.seriesIndex_.type === "full";
    const seriesStart = indexIsFull ? 0 : this.seriesIndex_.start;
    const seriesStop = indexIsFull ? seriesMax : this.seriesIndex_.stop;
```

packages/core/src/layers/image_series_layer.ts
aganders3: Actually never mind. I think this `initialIndex thing can just be removed. If people need to do this they can do:
```
layer.setIndex(initialIndex);
layer.preloadSeries();
```

packages/core/src/data/ome_zarr_image_loader.ts
aganders3: Yes, this needs work. Right now it's pretty ad-hoc. I refactored a bit to remove duplicates, but I can be much improved.

The difference to me is that the metadata here is OME-Zarr specific (based on the schema/spec), but the attributes are more generic. The image layer(s) right now are not OME-Zarr specific, but instead take an `ImageChunkSource` which can provide an `ImageChunkLoader`. I think we need to determine which attributes we need from this loader. For `OmeZarrImageLoader` it's likely most will come from the metadata. The naming is one thing we can improve for sure because the semantic difference between attributes/metadata is meaningless.

The practical issue at the moment is that getting metadata is async, and the zarr structure is nested. So we get the first level metadata when the loader is opened, but then we need more information from the individual scale images (arrays). I guess we could do this recursively up front.

packages/react/src/components/App.tsx
aganders3: This means "second lowest resolution available".

PR #220, opened 2025-03-31
OrganelleBox V1 component

aganders3:
### Summary
Introduces a new "load high res" button based on BioHub SF requests and @Janeece's designs:
Screenshot 2025-03-13 at 12 02 19 PM

I also added a "reset channels" button to the channel controls. @phoenixAja I simplified the code in the channel control list to not do full two-way syncing of the controls with the layer. I think this may be "good enough" for now but lmk what you think.

I'm still concerned a "reset camera" button would clutter the viewer, so I left it out. Previous discussions suggested a page refresh may be sufficient. Otherwise maybe it can be implemented as a button on the portal.

Here is a recording of this branch:

https://github.com/user-attachments/assets/f028964b-d5e8-49da-9dba-8eede7b6f738

### Related Issue
Replaces #208

### Tests & Checks
Tested in react workspace
Added a button to set a new url to test how the carousel will work

### DIFF
--- a/packages/core/src/objects/cameras/controls.ts
+++ b/packages/core/src/objects/cameras/controls.ts
@@ -55,9 +55,9 @@ export class PanZoomControls implements CameraControls {
     const clientPos = vec2.fromValues(event.clientX, event.clientY);
     const preZoomPos = clientToWorld(clientPos, this.clipDepth);
     if (event.deltaY < 0) {
-      this.camera_.zoom *= 1.1;
+      this.camera_.zoom *= 1.05;
     } else {
-      this.camera_.zoom /= 1.1;
+      this.camera_.zoom /= 1.05;
     }
     // pan to zoom in on the mouse position
     const postZoomPos = clientToWorld(clientPos, this.clipDepth);
--- a/packages/core/src/objects/cameras/controls.ts
+++ b/packages/core/src/objects/cameras/controls.ts
@@ -55,9 +55,9 @@ export class PanZoomControls implements CameraControls {
     const clientPos = vec2.fromValues(event.clientX, event.clientY);
     const preZoomPos = clientToWorld(clientPos, this.clipDepth);
     if (event.deltaY < 0) {
-      this.camera_.zoom *= 1.05;
+      this.camera_.zoom *= 1.1;
     } else {
-      this.camera_.zoom /= 1.05;
+      this.camera_.zoom /= 1.1;
     }
     // pan to zoom in on the mouse position
     const postZoomPos = clientToWorld(clientPos, this.clipDepth);
--- a/packages/core/vite.config.js
+++ b/packages/core/vite.config.js
@@ -44,7 +44,7 @@ export default defineConfig(({ mode }) => {
       outDir: 'dist',
       // TODO: set these by build mode or something
       sourcemap: true,
-      minify: false,
+      // minify: false,
       lib: {
         entry: path.resolve(_dirname, 'src/index.ts'),
         name: 'idetik-core',
--- a/packages/react/src/components/viewers/OmeZarrImageViewer/OmeZarrImageViewer.tsx
+++ b/packages/react/src/components/viewers/OmeZarrImageViewer/OmeZarrImageViewer.tsx
@@ -173,6 +173,8 @@ export function OmeZarrImageViewer({
     } catch {
       console.debug("Load 3D high-res aborted - likely selected new condition");
       return;
+    } finally {
+      setLoading(false);
     }
     setAllSlicesLoaded(true);
   }, [imageLayer]);
--- a/packages/react/src/components/viewers/OmeZarrImageViewer/OmeZarrImageViewer.tsx
+++ b/packages/react/src/components/viewers/OmeZarrImageViewer/OmeZarrImageViewer.tsx
@@ -1,7 +1,6 @@
 import { useEffect, useState } from "react";
 import cns from "classnames";
-import CircularProgress from "@mui/material/CircularProgress";
-import { InputSlider } from "@czi-sds/components";
+import { Button, InputSlider, LoadingIndicator } from "@czi-sds/components";
 import {
   ImageSeriesLayer,
   LayerManager,
@@ -20,14 +19,25 @@ interface OmeZarrImageViewerProps {
   sourceUrl: string;
   region: Region;
   seriesDimensionName: string;
-  scale?: number;
+  initialScale?: number;
+  loadHighResButton?: boolean;
+  highResSizeEstimate?: string;
+}
+
+// consolidated state to prevent effects from firing multiple times
+interface SourceState {
+  source: OmeZarrImageSource | null;
+  url: string;
+  isHighRes: boolean;
 }
 
 export function OmeZarrImageViewer({
   sourceUrl,
   region,
-  scale,
   seriesDimensionName,
+  initialScale,
+  loadHighResButton,
+  highResSizeEstimate,
 }: OmeZarrImageViewerProps) {
   const [layerManager, _setLayerManager] = useState<LayerManager>(
     new LayerManager()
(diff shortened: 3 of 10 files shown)


aganders3:
@kyleawayan - can you please post a link to a preview of the organellebox portal with this branch installed for the viewer?

kyleawayan:
@aganders3 Here it is! https://organelle-box-portal-git-latest-idetik-20240401-czbiohub.vercel.app/gene/2025_03_31_A549_GOLGA2?condition=Live

Switching between conditions is much smoother!

packages/core/src/objects/cameras/controls.ts
```diff
@@ -55,9 +55,9 @@ export class PanZoomControls implements CameraControls {
     const clientPos = vec2.fromValues(event.clientX, event.clientY);
     const preZoomPos = clientToWorld(clientPos, this.clipDepth);
     if (event.deltaY < 0) {
-      this.camera_.zoom *= 1.05;
+      this.camera_.zoom *= 1.1;
```
phoenixAja: why these specific values? Also these might be nice as constants somewhere

packages/core/vite.config.js
```diff
@@ -44,7 +44,7 @@ export default defineConfig(({ mode }) => {
       outDir: 'dist',
       // TODO: set these by build mode or something
       sourcemap: true,
-      minify: false,
+      // minify: false,
```
phoenixAja, line 47: did you mean to add this?

packages/core/src/objects/cameras/controls.ts
aganders3: Yes, I think it should either be configurable, or we should look into some convention. I just felt like the zooming was sluggish so I increased it a bit.

It also is a fixed number per wheel event, but I think we might be able to use the actual `deltaY` value to do something more natural feeling.

packages/core/vite.config.js
aganders3: Yes, I'm pretty sure we want to minify the production build. I'm not sure if there is some value in setting it depending on the build mode (maybe build time?). From what I understand the source map should be providing the same thing, but I don't have a complete understanding of it.

packages/react/src/components/viewers/OmeZarrImageViewer/OmeZarrImageViewer.tsx
```diff
@@ -20,89 +25,158 @@ interface OmeZarrImageViewerProps {
...
+    } catch {
+      console.debug("Load 3D high-res aborted - likely selected new condition");
+      return;
     }
-  }, [source, seriesDimensionName]);
+    setAllSlicesLoaded(true);
+  }, [imageLayer]);
 
-  imageLayer?.setIndex(zIndex);
+  const zIndex = Math.round(zValue * (zRange[1] - zRange[0]) + zRange[0]);
```
phoenixAja: not that big of a deal but this line is identical with line 145.

packages/core/src/objects/cameras/controls.ts
aganders3: I reverted this here. The sensitivity was really high at 1.1 when I tried it on my trackpad (instead of mouse).

Edit: I created #222 to track

packages/react/src/components/viewers/OmeZarrImageViewer/OmeZarrImageViewer.tsx
```diff
@@ -20,89 +25,159 @@ interface OmeZarrImageViewerProps {
...
+      shouldSetLayer = false;
+      setNeedChannelsReset(true);
+      setImageLayer(null);
+    };
+  }, [source, region, seriesDimensionName]);
 
   useEffect(() => {
     if (imageLayer) {
       layerManager.layers.length = 0;
       layerManager.add(imageLayer);
```
phoenixAja: this is outside of this PR but since we're mutating internal state here i think this won't trigger re-renders when the LayerManager changes.

aganders3: I think you're right! It probably doesn't actually affect anything in this app, but I can change it here to just create a new layer manager instead (which is pretty much what this does). Does that seems like the right fix?

packages/react/src/components/viewers/OmeZarrImageViewer/OmeZarrImageViewer.tsx
```diff
@@ -20,89 +25,159 @@ interface OmeZarrImageViewerProps {
...
-        if (layer.extent !== undefined) {
-          camera.setFrame(0, layer.extent.x, 0, layer.extent.y);
-          camera.update();
-          layer.removeStateChangeCallback(setCamera);
+      if (!shouldSetLayer) {
+        return;
+      }
+      setImageLayer(layer);
+      await setZRangeFromData();
+      const zoomToFit = () => {
```
phoenixAja: i feel like getLayer() is getting a bit long and is doing  a bunch of stuff now. can we maybe extract ZoomToFit?

packages/react/src/components/viewers/OmeZarrImageViewer/OmeZarrImageViewer.tsx
phoenixAja: yes sounds great!

packages/react/src/components/viewers/OmeZarrImageViewer/components/ChannelControlsList/ChannelControlsList.tsx
```diff
@@ -1,44 +1,44 @@
...
-    const updatedChannelProps = initialLayerChannelProps.map(
-      (layerChannel: ChannelProps, index: number) => ({
-        visible:
-          controlProps?.[index]?.visible ?? layerChannel?.visible ?? true,
-        color: controlProps?.[index]?.color ?? layerChannel?.color,
-        contrastLimits:
-          controlProps?.[index]?.contrastLimits ?? layerChannel?.contrastLimits,
-      })
-    );
+    if (reset || layer.channelProps?.length !== channelProps.length) {
```
phoenixAja: we might miss other channelProps updates if we're just checking that the length is the same?

packages/react/src/components/viewers/OmeZarrImageViewer/OmeZarrImageViewer.tsx
aganders3: I agree - I'll see if I can clean it up.

packages/react/src/components/viewers/OmeZarrImageViewer/components/ChannelControlsList/ChannelControlsList.tsx
aganders3: The logic in this component could certainly be improved. I think what you had before was more robust, but I'm not sure we need two-way sync for now - we can just consider the controls the only way to adjust the layer channel properties.

I need to think about this block. It's really meant to reset things when `reset` is passed, but it also serves the purpose of not trying to set a bad value if the length is different (in which case it will fail).

I'm not super happy with `reset` in any case though. Given the changed OrganelleBox requirements I don't really think it's necessary anymore, so I can take another crack at this.

packages/react/src/components/viewers/OmeZarrImageViewer/OmeZarrImageViewer.tsx
aganders3: It's still a bit messy, but I think a bit more semantic. I pulled `zoomToFit` and `setZRangeFromData` out, but I for now I think we need to register the state change callback here, otherwise I believe it could be too late.

Shlomi called it out before but we need a cleaner API to say something like "run this only once - either right now, or when the layer is ready". We have this pattern or similar in a few places and it's clunky. Maybe something async would work.

phoenixAja (approved):

packages/react/src/components/viewers/OmeZarrImageViewer/OmeZarrImageViewer.tsx
phoenixAja: i don't think the pattern you have now of unsubscribing from the zoomToFit callback is the worst. But it might be nice to see how it would work in core to have the "run only once" option.  Especially since i imagine there might be a fair amount of "run only once" cases?

PR #221, opened 2025-04-04
(Small) Rename DataTexture2D to Texture2D

shlomnissan:
### Summary
This PR is a follow-up to #210.  
It renames `DataTexture2D` to `Texture2D` for consistency.

### Tests & Checks

### DIFF
--- a/packages/core/src/layers/image_utils.ts
+++ b/packages/core/src/layers/image_utils.ts
@@ -1,4 +1,4 @@
-import { Texture2D } from "@/objects/textures/texture_2d";
+import { Texture2D } from "objects/textures/texture_2d";
 import { Texture } from "objects/textures/texture";
 import { Texture2DArray } from "objects/textures/texture_2d_array";
 import { ImageChunk } from "data/image_chunk";
--- a/packages/core/src/layers/image_utils.ts
+++ b/packages/core/src/layers/image_utils.ts
@@ -1,4 +1,4 @@
-import { DataTexture2D } from "objects/textures/data_texture_2d";
+import { Texture2D } from "@/objects/textures/texture_2d";
 import { Texture } from "objects/textures/texture";
 import { Texture2DArray } from "objects/textures/texture_2d_array";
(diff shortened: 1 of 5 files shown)


packages/core/src/layers/image_utils.ts
```diff
@@ -1,4 +1,4 @@
-import { DataTexture2D } from "objects/textures/data_texture_2d";
+import { Texture2D } from "@/objects/textures/texture_2d";
```
aganders3:
```suggestion
import { Texture2D } from "objects/textures/texture_2d";
```

packages/core/src/renderers/webgl_textures.ts
```diff
@@ -6,7 +6,7 @@ import {
   TextureDataFormat,
 } from "objects/textures/texture";
 
-import { DataTexture2D } from "objects/textures/data_texture_2d";
+import { Texture2D } from "@/objects/textures/texture_2d";
```
aganders3:
```suggestion
import { Texture2D } from "objects/textures/texture_2d";
```

packages/core/src/objects/renderable/image_renderable.ts
```diff
@@ -112,8 +112,6 @@ export class ImageRenderable extends RenderableObject {
     if (!texture) {
       throw new Error("un-textured image not implemented");
     } else if (texture.type == "Texture2D") {
-      this.programName = "mesh";
-    } else if (texture.type == "DataTexture2D") {
```
aganders3, line 116: Does this mean the "mesh" shader pair can be removed?

aganders3 (approved):
A couple nits for consistency, but LGTM!

packages/core/src/objects/renderable/image_renderable.ts
shlomnissan: I considered that when working on this PR but decided to leave it since the mesh vertex shader is still in use. After revisiting it, though, removing the pair does make sense. I’ll update it before merging.

PR #223, opened 2025-04-08
Refactor utilities structure

shlomnissan:
### Summary

This PR moves the UUID generation into its own module and separates out the
image utilities into a dedicated module. The motivation behind this change was
to create a logger utility class, and I felt it made sense to organize all
utility modules under a common utils folder.

While reviewing, I noticed the existing image utilities module and saw an
opportunity to improve its structure. In particular, the factory functions
should be tied to their respective classes to keep responsibilities more
clearly defined.

I also spotted a potential issue: the image layer appears to be using a
2D texture array instead of a 2D texture. Let me know if that was intentional.

### Tests & Checks


packages/core/src/layers/image_series_layer.ts
```diff
@@ -276,4 +273,16 @@ export class ImageSeriesLayer extends Layer {
       this.setState("loading");
     }
   }
+
+  private createImage(
```
phoenixAja, line 276: this is identical with what's in image_layer.ts. could we extract the core logic to be somewhere shared?

aganders3: This feels like it's getting at the fact that `ImageLayer` and `ImageSeriesLayer` are quite similar, and perhaps don't need to be separate layers after all.

phoenixAja: tbh i am confused about how they are different

shlomnissan: @phoenixAja You're right that the methods look similar, but they're not quite the same — I'm passing in a 2D texture rather than a 2D texture array. Technically, I could change the parameter type to accept the texture's base class, but in my experience, this kind of generalization often leads to subtle bugs or confusing abstractions down the road. In this case, I think the duplication is acceptable, especially since the boundaries and responsibilities of each layer aren't fully defined yet.

@aganders3 As for the `ImageLayer` and `ImageSeriesLayer`, I agree they currently look quite similar — the series layer could render a single image. But I see them as conceptually distinct: `ImageLayer` is about rendering a single static image, while `ImageSeriesLayer` is oriented around temporal or indexed navigation across a stack. That distinction matters because keeping the concepts separate gives us the flexibility to evolve the layers independently as their roles become clearer. For example, we may want to add scrubbing, interpolation, or caching behavior that only makes sense in a "series" context. Having dedicated layers lets us grow those capabilities without overcomplicating a single generic layer.

shlomnissan: @phoenixAja To add some context — one of the challenges I've seen in OOP designs (especially in languages without direct memory access or value semantics) is that passing around objects via base-class interfaces can lead to fragile abstractions. Over time, these interfaces tend to accumulate assumptions that don't hold for all subclasses, or obscure important type-specific behavior. That's part of why I’m leaning toward keeping this explicit and duplicating for now — it keeps the data flow clear and avoids future footguns.

> tbh i am confused about how they are different

One thing to keep in mind is that the scope of a layer goes beyond just rendering elements — it also includes functionality that's specific to that layer type. I expect these layers to diverge further in terms of behavior as their responsibilities grow over time.

aganders3: > Having dedicated layers lets us grow those capabilities without overcomplicating a single generic layer.

I agree with this. I really don't like the idea of simple classes with extraneous methods that don't do anything depending on the rest of their configuration.

phoenixAja: that makes sense, I agree with your points. Thanks for clarifying

phoenixAja (approved):
looks good to me!

shlomnissan:
@phoenixAja did you get a chance to verify that using a texture array in this case was unnecessary?

https://github.com/chanzuckerberg/imaging-active-learning/pull/223/files#diff-1fd87f239eb9627e2f7b1c14444175eacea8d0161473ba0a5741c970eb09ebd4L71

aganders3:
There is a problem, but it's in existing code. You can see this (silently) breaks the "2D Image from 4D OMEZarr HCS" example.

One part of the problem is in `validateChannels`, which doesn't really do anything if it's not given a texture array. It should assume the "depth" of a `Texture2D` is 1, and fail for mis-matched number of channels/depth of texture (or only work for `Texture2DArray`).
https://github.com/chanzuckerberg/imaging-active-learning/blob/5da3f79ea07c9063ec4a783d1f39a7df48cdbd7c/packages/core/src/objects/textures/channel.ts#L41-L59

The other part of the problem is that creating a `Texture2D` from the `ImageChunk` should fail if it's not 2D. It probably just kind of works by accident, but likely we're effectively getting the first channel from the `ImageChunk` (but still storing all three channels in the texture's `data_` attribute).

I don't think this all needs to be changed in this PR, but I would propose:
* Keep `Texture2D` and `Texture2DArray` separate (I'm 50/50 on this, they could be combined but as we discussed above about `ImageLayer` vs. `ImageSeriesLayer` I think they are likely to diverge)
* Don't access these `Texture` classes directly from the `Layer` classes at all. This keeps the code in the layers more semantic. To me this implies a tiered API comprising:
  * _Layers_ - generally the API for downstream users
  * _Renderable objects_ - objects used within layer code (image, mesh, point cloud, projected line, etc.), so you use these if you're building a new layer
  * _Low-level objects_ (geometry, textures, shaders) - objects used within renderable object code

So instead of (or more likely in addition to) making static factories on `Texture2D` and `Texture2DArray`, we would have a static factory on `ImageRenderable.createWithImageChunk`, which would handle the geometry and texture creation outside of the layer. I think this gets at what @phoenixAja was suggesting above.

packages/core/src/layers/image_layer.ts
```diff
@@ -1,10 +1,10 @@
 import { Layer } from "core/layer";
 import { Region } from "data/region";
-import { ImageChunkSource } from "data/image_chunk";
-import { Texture2DArray } from "objects/textures/texture_2d_array";
-import { makeImageTextureArray, makeImageRenderable } from "layers/image_utils";
+import { ImageChunk, ImageChunkSource } from "data/image_chunk";
 import { ChannelProps } from "objects/textures/channel";
 import { ImageRenderable } from "objects/renderable/image_renderable";
+import { Texture2D } from "objects/textures/texture_2d";
```
aganders3: For this PR I would just revert this to use `Texture2DArray`

packages/core/src/layers/image_layer.ts
```diff
@@ -68,13 +66,14 @@ export class ImageLayer extends Layer {
...
     };
-    this.texture_ = makeImageTextureArray(chunk);
-    this.renderable_ = makeImageRenderable(
+
+    this.image_ = this.createImage(
       chunk,
-      this.texture_,
+      Texture2D.createWithImageChunk(chunk),
       this.channelProps_
     );
```
aganders3: I would move `Texture2D.createWithImageChunk(chunk)` into the `createImage` function and remove it as an argument.

packages/core/src/layers/image_series_layer.ts
```diff
@@ -37,13 +37,13 @@ export class ImageSeriesLayer extends Layer {
   private readonly seriesDimensionName_: string;
   private readonly seriesIndex_: Interval | Full;
   private readonly scheduler_: PromiseScheduler = new PromiseScheduler(16);
-  private loader_: ImageChunkLoader | null = null;
+  private readonly loader_: ImageChunkLoader | null = null;
```
aganders3: I don't understand how this can be `readonly` if we're setting it `null` here but assigning to it in `getLoader()`

Oh wait - I see now that we _don't_ actually assign it in `getLoader()` but we should (my mistake).

packages/core/src/layers/image_series_layer.ts
```diff
@@ -37,13 +37,13 @@ export class ImageSeriesLayer extends Layer {
...
   private seriesAttributes_?: SeriesAttributes;
   private loadingToken_: LoadingToken | null = null;
   private texture_: Texture2DArray | null = null;
```
aganders3, line 43: I think we should remove `texture_` here (as in `ImageLayer`) and instead add an affordance on `ImageRenderable` to update its texture's data. This can be a separate PR if you'd like.

aganders3 (changes requested):
In general this is a big improvement. I approve of the direction but we need to hammer out some of these details.

shlomnissan:
@aganders3 

Thanks for the great feedback!
- I'm keeping Texture2D and Texture2DArray as separate types.
- I really like your suggestion to treat textures as low-level objects that layers shouldn't need to know about.

I am concerned about the two silent errors you highlighted:
1. `validateChannels` doesn't do anything if a flat texture is passed in.
2. It's possible to create a Texture2D from an ImageChunk even when it's not actually 2D.

I'll work on addressing your feedback and these concerns in a follow-up commit.

packages/core/src/layers/image_layer.ts
shlomnissan: I'm a bit confused here, and I suspect it's due to a leaky abstraction. At first glance, I'd expect an "image layer" to work exclusively with `Texture2D`, and an "image series layer" to work exclusively with `Texture2DArray`.

The confusion seems to stem from the fact that channels are stored as `Texture2DArray`, which makes sense for supporting multiple channels—even in a single image. So maybe it's totally fine for an image layer to use `Texture2DArray`. But if that's the case, shouldn't it also support single-channel images in a clean and consistent way?

On top of that, I don’t really get the purpose of the "image renderable" abstraction. It seems to assume the presence of multiple channels by default. If that’s the case, wouldn’t it be clearer to name it something like `MultiChannelImage`? And do we even need the "renderable" suffix in the name?

Should we even consider "image" a renderable object? I don't think it's a fundamental primitive.

This seems like a place where naming and intent need to be absolutely clear.

shlomnissan: @aganders3 @phoenixAja

shlomnissan: @aganders3 

> You can see this (silently) breaks the "2D Image from 4D OMEZarr HCS" example.

I can't get this example to load on main either?

packages/core/src/layers/image_series_layer.ts
shlomnissan: I'll handle this in a follow-up PR, since it ties into a broader discussion about the "image renderable" abstraction.

shlomnissan:
@aganders3 @phoenixAja I'm going to be working on a quick follow-up for this PR. My goals for it are:
- Reevaluate the "image renderable" abstraction—this might just come down to a rename.
- Move texture creation logic into the renderable itself.
- Add a safeguard to ensure `Texture2D` throws an error if initialized with a non-2D image chunk.
- Fix the "2D Image from 4D OMEZarr HCS" example if it is broken on `main`.

packages/core/src/layers/image_layer.ts
aganders3: > I can't get this example to load on main either?

Eek fair enough - we should fix that!

> This seems like a place where naming and intent need to be absolutely clear.

I agree, I think this warrants some thinking + discussion.

aganders3 (approved):
Looks good and I agree with the items you have listed for the follow-on PR.

PR #224, opened 2025-04-10
feat: [Blending Phase 1] Add transparent, opacity, and blendingMode to Layer class

phoenixAja:
## Summary

This PR lays some of the groundwork for upcoming blending support in the Viz renderer.

- Adds `transparent`, `opacity`, and `blendingMode` fields to the `Layer` base class
- Uses safe defaults and optional constructor params to preserve backward compatibility
- Includes ESLint suppression for temporarily-unused fields (these will be used in later phases; the suppression will be removed accordingly)

## Notes

These fields are not yet used in the rendering logic. They will be integrated in:

- **Phase 2**: Opaque/transparent layer partitioning
- **Phase 3**: Update the renderer to receive and apply blending-related metadata

## Testing

tested all examples

### DIFF
--- a/packages/core/src/core/layer.ts
+++ b/packages/core/src/core/layer.ts
@@ -1,16 +1,40 @@
 import { RenderableObject } from "./renderable_object";
 
 export type LayerState = "initialized" | "loading" | "ready";
+export type BlendingMode = "normal" | "additive" | "subtractive" | "multiply";
+
 
 type StateChangeCallback = (
   newState: LayerState,
   prevState?: LayerState
 ) => void;
 
+export interface LayerOptions {
+  transparent?: boolean;
+  opacity?: number;
+  blendingMode?: BlendingMode;
+}
+
 export abstract class Layer {
   private objects_: RenderableObject[] = [];
   private state_: LayerState = "initialized";
   private callbacks_: StateChangeCallback[] = [];
+
+  public readonly transparent: boolean;
+  public readonly opacity: number;
+  public readonly blendingMode: BlendingMode;
+
+  /* eslint-disable @typescript-eslint/no-unused-vars -- these fields are scaffolded for upcoming blending support */
+  constructor({
+    transparent = false,
+    opacity = 1.0,
+    blendingMode = "normal"
+  }: LayerOptions = {}) {
+    this.transparent = transparent;
+    this.opacity = opacity;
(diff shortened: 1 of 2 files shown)


packages/core/src/core/layer.ts
```diff
@@ -1,17 +1,49 @@
 import { RenderableObject } from "./renderable_object";
+import { clamp01 } from "./utils";
 
 export type LayerState = "initialized" | "loading" | "ready";
+export type BlendingMode = "normal" | "additive" | "subtractive" | "multiply";
```
phoenixAja, line 5: went with a string union instead of an enum here — felt like using an enum would be overkill since we're just representing a fixed set of simple string options

packages/core/src/core/layer.ts
```diff
@@ -1,17 +1,49 @@
...
   private objects_: RenderableObject[] = [];
   private state_: LayerState = "initialized";
   private callbacks_: StateChangeCallback[] = [];
 
+  public readonly transparent: boolean;
+  /**
+   * For layer opacity, value is clamped to the range [0.0, 1.0].
+   */
+  public readonly opacity: number;
+  public readonly blendingMode: BlendingMode;
```
aganders3: Will these be left `readonly` or is that part of the preliminary implementation?

packages/core/src/core/layer.ts
```diff
@@ -1,17 +1,49 @@
...
+  opacity?: number;
+  blendingMode?: BlendingMode;
+}
+
 export abstract class Layer {
   private objects_: RenderableObject[] = [];
   private state_: LayerState = "initialized";
   private callbacks_: StateChangeCallback[] = [];
 
+  public readonly transparent: boolean;
```
aganders3: nit: I feel like we've generally been using an `is` prefix for booleans. What do you think of `isTransparent` (or even `isOpaque`) instead?

aganders3 (approved):
LGTM!

packages/core/src/core/layer.ts
phoenixAja: yes i like that more!

packages/core/src/core/layer.ts
phoenixAja: i thought left readonly, but if you think these values could change in the future i can remove thata

PR #228, opened 2025-04-17
feat: layer-level transparency, opacity, and blending mode support

phoenixAja:
## Summary

This PR completes the core implementation of layer-level blending support, as outlined in the [blending technical specification](blending technical specification).

**Key features include:**

- Updated `ImageSeriesLayer` to accept and forward blending metadata
- Modified `WebGLRenderer` to:
  - Use `renderObject(layer, index)` signature
  - Enable and configure `gl.BLEND` based on layer `blendingMode`
  - Pass `u_opacity` uniform to fragment shaders
  - Use `gl.depthMask(false)` and `gl.polygonOffset` for transparent layers
- Updated all relevant fragment shaders to support `u_opacity`
- Added `LayerManager.partitionLayers()` to separate opaque and transparent layers
- Transparent layers are rendered after opaque layers.
- Back-to-front sorting is implemented for transparent layers (for correct visual stacking of transparent layers)
- Front-to-back sorting for opaque layers. (for performance)

**Testing:**
- changes are backwards compatible
- tested all existing examples
- added a new `layer_blending` example for testing stacked transparent layers with real-time opacity and blend mode controls.  (By adding a transparent yellow overlay layer.)

https://github.com/user-attachments/assets/3f37cdcb-7a5d-48e7-98c9-9cadd5cb5fa8


packages/core/src/core/layer.ts
```diff
@@ -64,7 +84,7 @@ export abstract class Layer {
   protected setState(newState: LayerState) {
     const prevState = this.state_;
     this.state_ = newState;
-    console.debug(`${this.constructor.name} state change: ${newState}`);
+    // console.debug(`${this.constructor.name} state change: ${newState}`);
```
phoenixAja:
```suggestion
```

packages/core/src/data/ome_zarr_image_loader.ts
```diff
@@ -68,10 +68,10 @@ export class OmeZarrImageLoader {
     region: Region,
     scheduler?: PromiseScheduler
   ): Promise<ImageChunk> {
-    console.debug("loading chunk with region", region, this.scaleIndex_);
+    // console.debug("loading chunk with region", region, this.scaleIndex_);
```
phoenixAja:
```suggestion
```

packages/core/src/data/ome_zarr_image_loader.ts
```diff
@@ -68,10 +68,10 @@ export class OmeZarrImageLoader {
...
     const { datasetPath, scale, translation } = this.getImageAttributes();
     const indices = this.regionToIndices(region);
-    console.debug("loading dataset with indices", datasetPath, indices);
+    // console.debug("loading dataset with indices", datasetPath, indices);
```
phoenixAja:
```suggestion
```

packages/core/src/data/ome_zarr_image_loader.ts
```diff
@@ -125,7 +125,7 @@ export class OmeZarrImageLoader {
       scale: { x: scale[indices.length - 1], y: scale[indices.length - 2] },
       offset: { x: xOffset, y: yOffset },
     };
-    console.debug("loaded chunk ", chunk);
+    // console.debug("loaded chunk ", chunk);
```
phoenixAja:
```suggestion
```

packages/core/src/data/ome_zarr_image_source.ts
```diff
@@ -14,7 +14,7 @@ export class OmeZarrImageSource {
   async open(): Promise<OmeZarrImageLoader> {
     const store = new zarr.FetchStore(this.url_);
     const root = await zarr.open.v2(store, { kind: "group" });
-    console.debug("opened root ", root, root.attrs);
+    // console.debug("opened root ", root, root.attrs);
```
phoenixAja:
```suggestion
```

packages/core/src/renderers/webgl_renderer.ts
```diff
@@ -91,6 +118,11 @@ export class WebGLRenderer extends Renderer {
     } else {
       this.gl.drawArrays(type, 0, object.geometry.itemSize);
     }
+
+    if (layer.isTransparent) {
+      // Apply polygon offset to avoid z-fighting for overlapping transparent layers
+      this.gl.disable(this.gl.POLYGON_OFFSET_FILL);
+    }
```
phoenixAja:
```suggestion
```

packages/core/examples/layer_blending/index.html
```diff
@@ -0,0 +1,57 @@
...
+      </select>
+      <label for="opacitySlider">Overlay Opacity</label>
+      <input
+        type="range"
+        id="opacitySlider"
+        min="0"
+        max="1"
+        step="0.01"
+        value="0.5"
+      />
```
shlomnissan, line 53: One suggestion as we start thinking about revamping our examples: consider using something like [lil-gui](https://github.com/georgealways/lil-gui) — it's what three.js uses and could help improve the look and feel (e.g. fullscreen layout per example) while reducing some of the HTML boilerplate. No need to make this change now, but feel free to experiment if you're curious — would be great to see!

packages/core/examples/layer_blending/main.ts
```diff
@@ -0,0 +1,137 @@
...
+    visible: false,
+    color: [0, 0, 0] as [number, number, number],
+    contrastLimits: [0, 255] as [number, number],
+  },
+  {
+    visible: false,
+    color: [0, 0, 0] as [number, number, number],
+    contrastLimits: [128, 255] as [number, number],
+  },
+];
```
shlomnissan: Nit: You could potentially simplify this by using `structuredClone()`, since both structures are mostly the same, and then modifying just the necessary fields. For example:

```ts
const overlayChannelProps = structuredClone(channelProps);
overlayChannelProps[0].visible = true;
overlayChannelProps[1].visible = false;
overlayChannelProps[2].visible = false;
```
In this case, you may need to modify the first object in the original `channelProps` array.

packages/core/examples/layer_blending/main.ts
```diff
@@ -0,0 +1,137 @@
...
+
+const overlayLayer = new ImageSeriesLayer({
+  source,
+  region,
+  seriesDimensionName: "T",
+  channelProps: overlayChannelProps,
+  isTransparent: true,
+  opacity: 0.5,
+  blendingMode: "normal", // just for now; you can try "additive", etc. later
```
shlomnissan: If these properties had mutators exposed, we might not need to include them in the layer’s constructor. In general, adding values to the constructor makes sense when they’re essential for the layer’s initial setup or identity, while using mutators can be better for optional or changeable configuration. A good rule of thumb is, if these constructor parameters are optional, they should probably not be included.

packages/core/src/core/layer.ts
```diff
@@ -13,21 +13,24 @@ export interface LayerOptions {
   isTransparent?: boolean;
   opacity?: number;
   blendingMode?: BlendingMode;
```
shlomnissan: See my earlier comment about potentially removing these parameters from the constructor.

packages/core/src/core/layer.ts
```diff
@@ -13,21 +13,24 @@ export interface LayerOptions {
...
+  zIndex?: number;
```
shlomnissan: I’m not convinced we should support a z-index in this way at this point. I may not have been clear in the spec, but when I mentioned sorting back-to-front (or front-to-back for opaque objects), I was referring to depth based on camera distance — not a manual z-index. To clarify, I do think we’ll need to support z-indexing for 2D images eventually, but given that we currently have a fixed drawing order via the layer manager, we can defer figuring out the best way to implement that. This current approach might not be the right fit.

I don’t have a strong opinion on this at this point, so I’m open to discussion if you have a clear sense of the trade-offs involved.

packages/core/src/core/layer.ts
```diff
@@ -13,21 +13,24 @@ export interface LayerOptions {
...
 export abstract class Layer {
   private objects_: RenderableObject[] = [];
   private state_: LayerState = "initialized";
   private readonly callbacks_: StateChangeCallback[] = [];
 
-  public readonly isTransparent: boolean;
-  public readonly opacity: number;
-  public readonly blendingMode: BlendingMode;
+  public zIndex: number = 0;
+  public isTransparent: boolean;
```
shlomnissan: Nit: I’d suggest naming this transparent instead — it’s more in line with idiomatic JavaScript/TypeScript conventions (like `element.hidden` or `object.visible`).

packages/core/src/core/layer.ts
```diff
@@ -37,6 +40,23 @@ export abstract class Layer {
     this.isTransparent = isTransparent;
     this.opacity = clamp(opacity, 0.0, 1.0);
     this.blendingMode = blendingMode;
+    this.zIndex = zIndex;
+  }
+
+  public setZIndex(z: number): void {
```
shlomnissan: You don’t need to explicitly specify the return type here — TypeScript can infer it for you. Letting the compiler infer return types keeps the code more concise and reduces duplication. It also ensures the return type always stays in sync with the actual implementation, which can help prevent bugs during refactors.

packages/core/src/core/layer.ts
```diff
@@ -37,6 +40,23 @@ export abstract class Layer {
...
+    this.blendingMode = mode;
+  }
+
+  public setOpacity(opacity: number): void {
+    this.opacity = clamp(opacity, 0.0, 1.0);
+  }
+
+  public setIsTransparent(isTransparent: boolean): void {
+    this.isTransparent = isTransparent;
   }
```
shlomnissan: A couple of notes here:
1. If the properties are public and the mutators don’t introduce any side effects, there's usually no need to define trivial setter methods. Avoiding them helps reduce boilerplate and keeps the code simpler and easier to read.
2. That said, if there were a reason to introduce mutators (which might not apply here), our typical pattern is to use `set` accessors. For example:
```ts
public set transparent(value: boolean) {
  this.transparent = value;
}
```
This approach keeps the external API clean and intuitive (`layer.transparent = true`), while still allowing you to introduce logic or side effects internally if needed later. It also encapsulates state without leaking internal implementation details.

Just to clarify — we do have some trivial accessors and mutators in the codebase already (and I’m likely responsible for a few of those). But it's something I'd like to gradually refactor out as we go.

packages/core/src/core/layer_manager.ts
```diff
@@ -3,6 +3,30 @@ import { Layer } from "./layer";
...
+      } else {
+        opaque.push(layer);
+      }
+    }
+
+    // Sort layers by zIndex:
+    // - Opaque: front-to-back (ascending) to optimize early depth rejection
+    // - Transparent: back-to-front (descending) to ensure correct alpha compositing
+    opaque.sort((a, b) => a.zIndex - b.zIndex);
+    transparent.sort((a, b) => b.zIndex - a.zIndex);
```
shlomnissan: If you remove the `z-index` property, you can delete this block as well.

packages/core/src/data/ome_zarr_image_loader.ts
```diff
@@ -68,10 +68,8 @@ export class OmeZarrImageLoader {
     region: Region,
     scheduler?: PromiseScheduler
   ): Promise<ImageChunk> {
-    console.debug("loading chunk with region", region, this.scaleIndex_);
```
shlomnissan, line 71: Thanks for removing all these debug logs :+1:

packages/core/src/layers/image_layer.ts
```diff
@@ -1,4 +1,4 @@
-import { Layer } from "core/layer";
+import { BlendingMode, Layer } from "core/layer";
```
shlomnissan: Nit: I think `BlendMode` is more idiomatic for an exported type and class property. It’s more widely recognized in the graphics/rendering context and aligns with usage.

packages/core/src/layers/image_layer.ts
```diff
@@ -10,6 +10,10 @@ export type ImageLayerProps = {
   source: ImageChunkSource;
   region: Region;
   channelProps?: ChannelProps[];
+  isTransparent?: boolean;
+  opacity?: number;
+  blendingMode?: BlendingMode;
+  zIndex?: number;
```
shlomnissan: Same concerns about having these optional values in the constructor. This applies to the image series layer as well.

packages/core/src/layers/image_series_layer.ts
```diff
@@ -51,8 +55,12 @@ export class ImageSeriesLayer extends Layer {
     region,
     seriesDimensionName,
     channelProps,
+    isTransparent = false,
+    opacity = 1.0,
+    blendingMode = "normal",
+    zIndex = 0,
```
shlomnissan: I think these default values could be set in the base `Layer` class — it looks like you’ve already set `zIndex` to 0 there.

packages/core/src/core/renderer.ts
```diff
@@ -39,14 +39,27 @@ export abstract class Renderer {
...
+
+    // Render transparent layers last
+    for (const layer of transparent) {
+      layer.update();
+      if (layer.state === "ready") {
+        for (let i = 0; i < layer.objects.length; i++) {
+          this.renderObject(layer, i);
+        }
+      }
+    }
```
shlomnissan: Nit: Here's a more idiomatic way to write these loops:
```ts
for (const layer of transparent) {
  layer.update();

  if (layer.state !== "ready") continue;

  layer.objects.forEach((_, i) => this.renderObject(layer, i));
}
```
1. Uses `continue` to reduce nesting.
2. Uses `forEach` for clarity, since you're just iterating indices.

packages/core/src/renderers/webgl_renderer.ts
```diff
@@ -41,9 +41,33 @@ export class WebGLRenderer extends Renderer {
...
+          break;
+        case "normal":
+        default:
+          this.gl.blendFunc(this.gl.SRC_ALPHA, this.gl.ONE_MINUS_SRC_ALPHA);
+      }
+    } else {
+      this.gl.disable(this.gl.BLEND);
+      this.gl.depthMask(true);
+    }
+
```
shlomnissan: 1. We should consider disabling the depth mask in the renderer for now to reduce context switching. We'll also want to make sure it's reset afterward—perhaps something like:

```ts
if (transparent.size() > 0) this.gl.depthMask(false);

for (const layer of transparent) {
  layer.update();
  if (layer.state !== "ready") continue;
  layer.objects.forEach((_, i) => this.renderObject(layer, i));
}

this.gl.depthMask(true);
```

2. This could probably be addressed in a separate PR, but it might be worth introducing a `WebGLState` class to wrap all `this.gl.*` state interactions. In general, we want to avoid making redundant WebGL state changes — calls like `gl.enable` and `gl.depthMask` can be surprisingly expensive, especially in tight render loops. By centralizing and tracking state changes, we can ensure we only update WebGL state when necessary, which can help reduce driver overhead and improve GPU performance — especially on mobile and low-power devices.

In this case, it looks like we're calling `gl.enable` and `gl.depthMask` even when the state hasn't changed, which is something we could optimize.

This is a very common approach when building OpenGL applications:

https://github.com/mrdoob/three.js/blob/dev/src/renderers/webgl/WebGLState.js

packages/core/examples/layer_blending/main.ts
```diff
@@ -0,0 +1,137 @@
...
+// Source is 5D, so provide an interval in T a scalar index in Z
+// (first of only depth) to get a 2D image series.
+const source = new OmeZarrImageSource(url);
+
+// Get the default Z index from the zattrs
+let zIndex = 0;
+(async () => {
+  try {
+    zIndex = await loadOmeroDefaultZ(url);
+    console.log("Default Z index:", zIndex);
```
shlomnissan: Once you rebase this PR onto main, you can use the new Logger utility here.

shlomnissan (changes requested):
This PR looks great overall — really well structured, nice work Phoenix!!!

The only blocking comment I have is around introducing a `z-index`. As I mentioned, it’s not a strong opinion, but I do think it’s worth discussing further if you see it differently.

I'd also like to gradually clean up the API by removing some properties from constructors, though that’s more of a nice-to-have than a pressing issue. The rest of my comments are just nits.

Awesome job on this!

packages/core/src/core/renderer.ts
```diff
@@ -39,14 +39,27 @@ export abstract class Renderer {
...
-          this.renderObject(obj);
-        });
+        for (let i = 0; i < layer.objects.length; i++) {
+          this.renderObject(layer, i);
+        }
       }
-    });
+    }
+
+    // Render transparent layers last
```
shlomnissan: I’d suggest avoiding this type of comment. Comments like this can introduce a subtle form of coupling. In this case, the comment doesn’t add much clarity or context that isn’t already expressed by the code itself, so it’s probably better to leave it out.

packages/core/examples/layer_blending/index.html
phoenixAja: interesting! I think adding something like lil-gui would be really nice moving forwards. I agree that our examples are currently a little cumbersome

packages/core/examples/layer_blending/main.ts
phoenixAja: added

packages/core/src/core/layer.ts
phoenixAja: i feel torn on this. I feel like isTransparent makes more sense, but i think it's important to be idiomatic. I'll rename to transparent

packages/core/src/core/layer.ts
phoenixAja: cool thank you! I removed the other void return types in the code base while i was at it

packages/core/src/core/layer.ts
phoenixAja: great i removed the not needed setters and added public set for opacity since we have clamping logic there

packages/core/src/data/ome_zarr_image_loader.ts
phoenixAja: no problem, they were driving me crazy lol

packages/core/src/layers/image_layer.ts
phoenixAja: done

packages/core/src/core/renderer.ts
phoenixAja: replaced

packages/core/examples/layer_blending/main.ts
phoenixAja: whoops i meant to remove this console!

packages/core/src/core/renderer.ts
phoenixAja: removed

packages/core/src/renderers/webgl_renderer.ts
phoenixAja: That's good feedback! i added beginTransparentPass() and endTransparentPass() protected methods in renderer to toggle the state behavior for disable depthMask and enable blend when rendering transparent layers for now.

packages/core/src/core/layer.ts
phoenixAja: done

packages/core/examples/layer_blending/main.ts
phoenixAja: yep! removed them

packages/core/src/renderers/webgl_renderer.ts
phoenixAja: i created an [issue](https://github.com/chanzuckerberg/imaging-active-learning/issues/234) for adding WebGLState class

packages/core/examples/layer_blending/index.html
phoenixAja: created ticket [here](https://github.com/chanzuckerberg/imaging-active-learning/issues/235) to look into this further.

packages/core/src/core/layer.ts
phoenixAja: I removed zIndex for now, since we're still figuring out the best way to handle sorting (manual vs. depth-based). Thanks for the clarification — we'll revisit this when we need more explicit control over layer draw order for overlapping images.

packages/core/src/layers/image_layer.ts
```diff
@@ -10,6 +10,9 @@ export type ImageLayerProps = {
   source: ImageChunkSource;
   region: Region;
   channelProps?: ChannelProps[];
+  transparent?: boolean;
+  opacity?: number;
+  blendMode?: blendMode;
```
shlomnissan: My previous comment apply here as well https://github.com/chanzuckerberg/imaging-active-learning/pull/228#discussion_r2058731550

I view configuration objects similarly to passing constructor parameters. If certain values aren't required to create a new instance, it's better to use mutators or access properties directly. Personally, I believe optionals are generally a bad idea, though that's just my opinion. In any case, this isn't a blocker — I just don't think these parameters are necessary.

packages/core/src/renderers/webgl_renderer.ts
```diff
@@ -41,9 +41,36 @@ export class WebGLRenderer extends Renderer {
...
-  protected renderObject(object: RenderableObject) {
+  protected beginTransparentPass(): void {
+    this.gl.enable(this.gl.BLEND);
+    this.gl.depthMask(false);
+  }
+
+  protected endTransparentPass(): void {
+    this.gl.disable(this.gl.BLEND);
+    this.gl.depthMask(true);
+  }
```
shlomnissan, line 52: A renderer will go through various types of context switching, and we probably don't want to create a separate virtual method for each one. I realize my earlier comment led us in that direction, but after thinking it over, I believe a better approach would be to make the `render()` function itself virtual and let each renderer handle the internal details. This would result in a more robust and easier-to-understand solution. If you feel it's out of scope for now, we can create an issue to track it — it's something I'd like to prioritize.

shlomnissan (approved):
Thanks, Phoenix! Great work on this PR. I left a few comments, but otherwise, it's good to go from my side 🚀

packages/core/src/layers/image_layer.ts
phoenixAja: done

packages/core/src/renderers/webgl_renderer.ts
phoenixAja: That makes a lot of sense. I agree that having the render() method itself be virtual would scale better and avoid bloating the base class with lots of pass-specific hooks. It’ll also make renderers like WebGLRenderer easier to understand in isolation. Here's an [issue](https://github.com/chanzuckerberg/imaging-active-learning/issues/236) to track

PR #229, opened 2025-04-17
fix(react): update css filename in exports

kyleawayan:
### Summary

Importing the React library using `@import "@idetik/react/style.css";` does not work anymore.

I think with the new Vite updates, there was a change of the output CSS filename. In the `dist` folder, it is now `index.css`, instead of `style.css`. I updated the `package.json` to update the filename, keeping `@import "@idetik/react/style.css";` valid, while also adding support for `@import "@idetik/react/index.css";`

### Related Issue
Closes #<issue-number>

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -29,14 +29,15 @@
   "main": "./dist/index.umd.cjs",
   "module": "./dist/index.js",
   "types": "./dist/index.d.ts",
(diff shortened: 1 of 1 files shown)


aganders3 (approved):
Thanks! I'm not sure what changed this, but I trust these changes make it work for you :)

PR #231, opened 2025-04-18
Create a simple logger class

### LINKED ISSUE #42: Create a simple Logger class
At this stage of development, we need to implement log tracing with predefined levels and labels for filtering. Let's start with a simple Logger class that outputs directly to the console.

shlomnissan:
### Summary

This PR introduces a centralized Logger class for consistent logging:

- Supports debug, info, warn, and error log levels
- Color-coded output for improved terminal readability
- Timestamps included on all logs
- Optional module/class tagging (auto-detected via stack trace)
- Log level filtering based on env

It also includes a usage example.

### Related Issue
Closes #42 

### Tests & Checks

Screenshot 2025-04-18 at 9 53 17 AM

### DIFF
--- a/packages/core/src/utilities/logger.ts
+++ b/packages/core/src/utilities/logger.ts
@@ -1,0 +1,108 @@
+type LogLevel = "info" | "warn" | "error" | "debug";
+
+const Levels = {
+  debug: 10,
+  info: 20,
+  warn: 30,
+  error: 40,
+} satisfies Record<LogLevel, number>;
+
+const Colors = {
+  debug: "\x1b[90m", // gray
+  info: "\x1b[36m", // cyan
+  warn: "\x1b[33m", // yellow
+  error: "\x1b[31m", // red
+} satisfies Record<LogLevel, string>;
+
+const regex = /at new\s+(.*)\s+\(/;
+
+export class Logger {
+  private static logLevel_: LogLevel =
+    import.meta.env.MODE === "production" ? "warn" : "debug";
+
+  public static setLogLevel(level: LogLevel) {
+    Logger.logLevel_ = level;
+  }
+
+  public static debug(
+    message: string,
+    moduleName?: string,
+    ...params: unknown[]
+  ) {
+    Logger.log("debug", message, moduleName, ...params);
+  }
+
+  public static info(
+    message: string,
+    moduleName?: string,
+    ...params: unknown[]
+  ) {
+    Logger.log("info", message, moduleName, ...params);
+  }
+
+  public static warn(
+    message: string,
+    moduleName?: string,
+    ...params: unknown[]
+  ) {
+    Logger.log("warn", message, moduleName, ...params);
+  }
+
+  public static error(
+    message: string,
+    moduleName?: string,
+    ...params: unknown[]
+  ) {
+    Logger.log("error", message, moduleName, ...params);
+  }
+
+  private static log(
+    level: LogLevel,
+    message: string,
+    moduleName?: string,
+    ...args: unknown[]
+  ) {
+    if (Levels[level] < Levels[Logger.logLevel_]) return;
+
+    const timestamp = new Date().toISOString();
+    const module = moduleName ?? Logger.detectModuleName();
+    const color = Colors[level];
+    const tag = `[${timestamp}][${level.toUpperCase()}][${module}]`;
+    const output = [`${color}${tag}`, message, ...args];
+
+    switch (level) {
+      case "debug":
+        console.debug(...output);
+        break;
+      case "info":
+        console.info(...output);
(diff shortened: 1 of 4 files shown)


packages/core/tsconfig.json
```diff
@@ -31,7 +31,7 @@
     "noUnusedParameters": true,
     "noFallthroughCasesInSwitch": true,
 
-    "types": ["vite-plugin-glsl/ext"]
+    "types": ["vite-plugin-glsl/ext", "vite/client"]
```
shlomnissan, line 34: Include the vite client type declarations, which include `import.meta.env`.

packages/core/src/utilities/logger.ts
```diff
@@ -0,0 +1,108 @@
...
+} satisfies Record<LogLevel, number>;
+
+const Colors = {
+  debug: "\x1b[90m", // gray
+  info: "\x1b[36m", // cyan
+  warn: "\x1b[33m", // yellow
+  error: "\x1b[31m", // red
+} satisfies Record<LogLevel, string>;
+
+const regex = /at new\s+(.*)\s+\(/;
```
aganders3: nit: I'd suggest changing this to a more descriptive name like `captureModuleNameRegex`

packages/core/src/utilities/logger.ts
```diff
@@ -0,0 +1,108 @@
...
+    moduleName?: string,
+    ...params: unknown[]
+  ) {
+    Logger.log("debug", message, moduleName, ...params);
+  }
+
+  public static info(
+    message: string,
+    moduleName?: string,
+    ...params: unknown[]
```
aganders3: I commonly write logs like `console.log("Initialized foo", foo)` so I can inspect the value of an object. At least in chrome this is pretty convenient and lets you drill into the object interactively.

I think with this type signature that would become `Logger.info("Initialized foo", undefined, foo)`, which I find awkward. How often do you expect setting the `moduleName` explicitly? Maybe in normal usage it should always be inferred - `log` could be public for where you need to be explicit.

Maybe my usage is not a good general logging pattern as I am still pretty new to JS.

packages/core/src/utilities/logger.ts
```diff
@@ -0,0 +1,108 @@
...
+        console.warn(...output);
+        break;
+      case "error":
+        console.error(...output);
+        break;
+    }
+  }
+
+  private static detectModuleName(): string {
+    // This logic is based on the format of the stack trace in the V8 engine
```
aganders3: Since this doesn't work outside of Chrome, perhaps we should skip adding the `module` to the tag in other contexts.

packages/core/src/utilities/logger.ts
```diff
@@ -0,0 +1,108 @@
...
+    message: string,
+    moduleName?: string,
+    ...args: unknown[]
+  ) {
+    if (Levels[level] < Levels[Logger.logLevel_]) return;
+
+    const timestamp = new Date().toISOString();
+    const module = moduleName ?? Logger.detectModuleName();
+    const color = Colors[level];
+    const tag = `[${timestamp}][${level.toUpperCase()}][${module}]`;
```
aganders3: It would be nice to have a file + line number if possible, but maybe we don't want to present this in production context (where we're likely minified anyway). Same goes for the module name itself, I suppose.

aganders3 (approved):
This generally LGTM and having some structure is better than what we've been doing. I had a few comments/questions but happy to merge as-is and improve this over time as-needed.

packages/core/src/utilities/logger.ts
shlomnissan: No longer relevant -- I removed the dynamic module name parser.

packages/core/src/utilities/logger.ts
shlomnissan: No longer relevant -- I made the module name a required parameter.

packages/core/src/utilities/logger.ts
shlomnissan: No longer relevant -- module name is a required parameter.

packages/core/src/utilities/logger.ts
shlomnissan: I don't think we can reliably extract line numbers :/

aganders3 (approved):
Looks good, thanks!

It's a huge nit, but I could imagine putting the `moduleName` arg before `message` now that it's required. This is a little more intuitive to me as it reflects the order these items appear in the log message. I don't feel strongly about this though.

shlomnissan:
> It's a huge nit, but I could imagine putting the moduleName arg before message now that it's required

@aganders3 Oh I agree, I'll make the change before merging

PR #233, opened 2025-04-24
refactor: OmeZarrImageViewer into composable hook-based architecture for layout flexibility

phoenixAja:
### Summary
This PR refactors the original `OmeZarrImageViewer` into a composable, layout-agnostic architecture that enables embedding the viewer into applications with custom UI designs — such as the new HITL tomogram viewer.

### Motivation:
The previous OmeZarrImageViewer tightly coupled internal viewer state with layout-specific rendering. This made it difficult to adapt the viewer to any specialized designs.

By moving viewer logic into useOmeZarrViewer hook, we can now:
* Compose the viewer anywhere in the app
* Style controls differently or place them in new containers
* Retain internal logic (slice management, layer loading) without duplication

### Key Changes
* Extracts all internal state and side effects (image loading, layer creation, Z index, channel controls, etc.) into a new custom hook: `useOmeZarrViewer`.
* Refactors the viewer component to use this hook and render only the presentational elements (Renderer, ChannelControlsList, control overlays).
* Preserves the original visual layout by default (Renderer background + channel and control overlays), but makes all parts individually movable.

### Tests & Checks
tested that current test application loads/works as expected


packages/react/package.json
```diff
@@ -4,7 +4,7 @@
   "version": "0.0.1",
   "type": "module",
   "scripts": {
-    "dev": "vite",
+    "dev": "vite --port 3000",
```
phoenixAja: i was running into cors issues with the example image pulled from biohub so i changed this.

packages/react/src/components/viewers/OmeZarrImageViewer/OmeZarrImageViewer.tsx
```diff
@@ -1,237 +1,63 @@
...
-      }
-      return false;
-    },
-    [setCamera]
-  );
-
-  // destructure props to narrow useEffect dependencies
+export function OmeZarrImageViewer(props: OmeZarrViewerContainerProps) {
   const {
+    sourceUrl,
```
aganders3, line 27: These were previously destructured to satisfy the linter when adding dependencies to `useEffect`, but that's no longer needed. I think in that case I'd prefer to _not_ destructure any of the props. I'm ambivalent about this though so do whatever feels right :).

packages/react/src/components/viewers/OmeZarrImageViewer/useOmeZarrImageViewer.ts
aganders3, line 1: I'm not sure the convention when using fractal directory structure, but I am used to hooks being in a `hooks` directory at the same level as `components`.

packages/react/src/components/viewers/OmeZarrImageViewer/OmeZarrImageViewer.tsx
```diff
@@ -292,24 +120,25 @@ export function OmeZarrImageViewer(
...
-                ? `Load 3D high-res (${props.allSlicesSizeEstimate})`
+              {allSlicesSizeEstimate
+                ? `Load 3D high-res (${allSlicesSizeEstimate})`
                 : "Load 3D high-res"}
             </Button>
           )}
-          <div className={cns("flex", "justify-end", "w-1/3")}>
-            {!loading && (
-              // TODO: "tag" is not very semantic but this looks okay
+          <div className="flex justify-end w-1/3">
```
aganders3, line 128: I think we should be consistent with using `cns` or have some convention to follow (i.e. use a single string if 3 or fewer classes?).

bchu1 (approved):

packages/react/package.json
aganders3: I don't do it very often, but this will make it a little more annoying to run two instances (for example separate branches) at the same time.

aganders3 (approved):
Looks good! I think this is more readable. Maybe in the future separate "features" of the viewer can even be split into separate hooks?

packages/react/src/components/viewers/OmeZarrImageViewer/useOmeZarrImageViewer.ts
phoenixAja: i'll move it to hooks

packages/react/package.json
phoenixAja:
```suggestion
    "dev": "vite",
```

PR #240, opened 2025-05-05
feat: Improve customizability and integrator facing API of React components

bchu1:
- Removes duplicate state in controls components, adds callbacks to ImageSeriesLayer.ts so that React can hook into its state directly using `useSyncExternalStore()`.
 - Caches the initial channels metadata response so that resetting controls doesn't require a refetch (in the future I think this the logic for fetching could probably benefit from moving outside of the viewer component as well).
 - Updates README to include instructions for integrators.
 - Removes some Organelle Box specific styles from the demo `<App>`.
 - Adds CSS customizability to the viewer component.
 - Moves `<ChannelControlsList>` out of the viewer. Removes all props from the component so it just works by itself. More will be done to this component later.
 - Creates a provider `<IdetikProvider>` to store global state. Integrators must now wrap their application in this component.
 - Creates a helper hook `useIdetik()` to access the provider's state.
 - Hooks `<ChannelControlsList>` into `<IdetikProvider>`.
 - Moves some state from `useOmeZarrImageViewer()` into the global provider.

Next PR: Remove more hardcoded styles, add more CSS customizability.

### DIFF
--- a/packages/react/src/components/hooks/useIdetik.ts
+++ b/packages/react/src/components/hooks/useIdetik.ts
@@ -1,0 +1,29 @@
+import { ImageSeriesLayer } from "@idetik/core";
+import { createContext, useContext } from "react";
+
+export interface IdetikContextValue {
+  imageSeriesLayer?: ImageSeriesLayer;
+  setImageSeriesLayer: (imageLayer: ImageSeriesLayer) => void;
+}
+
+export const IdetikContext = createContext<IdetikContextValue>({
+  setImageSeriesLayer: () => {
+    throw new Error(
+      "<OmeZarrImageViewer> was initialized but your application was not wrapped with <IdetikProvider>."
+    );
+  },
+});
+
+/**
+ * Gives you access to Idetik global state to write our own custom components.
+ *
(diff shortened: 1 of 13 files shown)


packages/react/src/components/hooks/useIdetik.ts
```diff
@@ -0,0 +1,39 @@
...
+  channels: ChannelProps[];
+  setChannels: (channels: ChannelProps[]) => void;
+  resetChannels: () => void;
+
+  channelControls: ChannelControl[]; // Same order as channels.
+  setChannelControls: React.Dispatch<
+    React.SetStateAction<ChannelControl[] | undefined>
+  >;
+
+  setImageSeriesLayer: React.Dispatch<
```
phoenixAja, line 21: why did you add these spaces?

phoenixAja (approved):
excited to use this! thanks

packages/react/src/components/hooks/useIdetik.ts
bchu1: trying to group them 😆 maybe it just made it uglier..

kyleawayan:
This looks awesome! Thanks!

Issue #59, opened 2024-09-27
Add a simple Color datatype


aganders3:
> Introducing a simple Color datatype would be beneficial.

_Originally posted by @shlomnissan in https://github.com/chanzuckerberg/imaging-active-learning/pull/46#discussion_r1775419144_

andy-sweet:
@aganders3 : was this closed by #242 ?

aganders3:
Good call - yes I think so! At least the basics are there.

PR #242, opened 2025-05-06
feat: add Points renderable object

aganders3:
### Summary
This PR adds support for displaying points with a new renderable object. The implementation is based on [this tech spec](https://docs.google.com/document/d/1EE2KP2WdeqSRcOrRXy3bNISeYiME8ts2ixPDeWecZ6I/edit?tab=t.0), which is still under review.

Points are drawn in a single pass with attributes:
* position
* color
* marker

where _marker_ is an integer that indexes into a sprite array on the renderable object. Then the fragment shader is able to sample from that sprite array to get the alpha value for the marker (symbol). Right now the library includes three basic binary symbols calculated on the CPU: circles, squares, and triangles.

The added example shows fetching [point annotations (picked particles) and images from the CryoET Data Portal](https://cryoetdataportal.czscience.com/runs/17682). This PR does not add a new layer, but the example includes a layer for demonstration purposes. This layer adjusts point color/size based on proximity to the current slice. As mentioned in the spec, these changes could also be implemented in a more advanced shader at some point (similarly for the `ProjectedLine` renderable).

If dynamically changing attributes (or even the number of points) is a common operation, we should consider using non-interleaved vertex attributes and reusable buffers to improve performance.

https://github.com/user-attachments/assets/58650c20-5f03-42d3-8a0f-52e2ff7ed6c9

Here is roughly the same view in Neuroglancer:

https://github.com/user-attachments/assets/4fe3d93e-0914-4398-bf8f-b6b786a925b8

The current motivating project for this work is the "Try CryLens" demo - they would like to use our library to:
* show CryoET images from the data portal
* allow users to click near a pre-picked particle to "find similar particles"
* show these similar particles overlaid on the image, perhaps with some indication of similarity (i.e. distance in embedding space)

### Related Issue
N/A

### Tests & Checks
Current suite of tests passes
New example appears to be correctly rendering point annotations from the CryoET data portal


packages/core/src/renderers/webgl_textures.ts
```diff
@@ -61,11 +62,6 @@ export class WebGLTextures {
     return texture;
   }
 
-  private configureTexture(texture: Texture) {
-    this.configureTextureParameters(texture);
-    this.allocateTextureStorage(texture);
-  }
-
```
aganders3, line 68: I was getting an error when updating the image slice with points also in the scene. I think it's due to the mixed configuration of the texture parameters (image texture vs. the sprite atlas for the points). So I split these functions to call `allocate` only once, but `configure` whenever updating the texture.

packages/core/src/core/geometry.ts
```diff
@@ -32,7 +35,7 @@ export class Geometry extends Node {
   }
 
   public get itemSize() {
-    return this.vertexData_.length / this.stride;
+    return this.vertexData_.byteLength / this.stride;
```
aganders3, line 38: This was causing an issue when drawing without an index buffer - basically it would only draw a fraction of the points because this number was wrong.

phoenixAja: interesting, maybe good to leave this as a comment?

packages/core/src/objects/geometry/points_geometry.ts
```diff
@@ -0,0 +1,68 @@
+import { vec3 } from "gl-matrix";
+
+import { Geometry } from "core/geometry";
+
+interface PointProperties {
+  position: vec3;
+  color: vec3;
```
phoenixAja: I think a more explicit type would improve readability and safety here. For example:

```
type PointPosition = { x: number; y: number; z: number };
type PointColor = { r: number; g: number; b: number };
```

This would let us destructure properties more clearly when serializing:

```
points.flatMap((p) => [
  p.position.x,
  p.position.y,
  p.position.z,
  p.color.r,
  p.color.g,
  p.color.b,
  p.size,
  p.marker,
]);
```

packages/core/src/objects/renderable/points.ts
```diff
@@ -0,0 +1,92 @@
+import { RenderableObject } from "core/renderable_object";
+import { PointsGeometry } from "objects/geometry/points_geometry";
+import { Texture2DArray } from "objects/textures/texture_2d_array";
```
phoenixAja: can you update these to relative imports? See [this](https://github.com/chanzuckerberg/imaging-active-learning/pull/243) for context!

packages/core/src/objects/geometry/points_geometry.ts
aganders3: Both good ideas! If it's okay I'd prefer to keep them out of this PR though. Color in particular we have an issue for (#59). Position is also a good idea, but it would benefit other classes (projected line, tracks) as well so I'd rather consider it in a broader context.

packages/core/src/core/geometry.ts
aganders3: Maybe the property name should change? This is passed in as the `count` to `drawArrays`. `stride` is calculated in bytes so it makes sense to me to use the same units in the numerator. To me this just felt like a bug we haven't seen because for other renderable objects we use an index buffer (so they use `drawElements` instead of `drawArrays`).

packages/core/src/objects/textures/texture.ts
```diff
@@ -11,6 +11,31 @@ export type TextureDataType = "unsigned_byte" | "unsigned_short" | "float";
...
+  | typeof Uint8Array
+  | typeof Uint16Array
+  | typeof Float32Array;
+
+export function concatenateTypedArrays(
+  data: DataTextureTypedArray[]
+): DataTextureTypedArray {
+  const eachLength = data[0].length;
+  const totalLength = data.length * eachLength;
+  const Constructor = data[0].constructor as DataTextureTypedArrayConstructor;
```
phoenixAja: Could we refine the DataTextureTypedArray type so that the constructor doesn't need to be cast manually here?

packages/core/src/objects/textures/texture_2d_array.ts
```diff
@@ -58,4 +59,16 @@ export class Texture2DArray extends Texture {
...
+
+  public static createWithArrays(
+    data: DataTextureTypedArray[],
+    width: number,
+    height: number
+  ) {
+    const array = concatenateTypedArrays(data);
+    const texture = new Texture2DArray(array, width, height);
+    texture.unpackRowLength = width;
+    texture.unpackAlignment = 1;
```
phoenixAja: is this always 1?

packages/core/src/renderers/webgl_buffers.ts
```diff
@@ -60,6 +60,9 @@ export class WebGLBuffers {
       if (attr.type == "next_position") idx = 4;
       if (attr.type == "direction") idx = 5;
       if (attr.type == "path_proportion") idx = 6;
+      if (attr.type == "color") idx = 7;
+      if (attr.type == "size") idx = 8;
+      if (attr.type == "marker") idx = 9;
```
phoenixAja, line 65: out of scope for this PR, but i feel like this could be improved with a mapping

```
const attributeIndexMap: Record<string, number> = {
  position: 0,
  normal: 1,
  uv: 2,
  previous_position: 3,
  next_position: 4,
  direction: 5,
  path_proportion: 6,
  color: 7,
  size: 8,
  marker: 9,
};

```

so then this looks a lot cleaner:
```
attributes.forEach((attr) => {
  const idx = attributeIndexMap[attr.type];
...
  }
});
```

packages/core/src/objects/textures/texture.ts
aganders3: What I want is for this to be like a generic `fn concatenate<T>(in: T[]) -> T`. This is the first thing that came to mind but I'm sure there's a better way.

packages/core/src/objects/textures/texture_2d_array.ts
aganders3: My understanding is that alignment is kind of a vestigial appendage from older GPUs that would perform better with 4-byte alignment. See the notes on [webgl2 fundamentals](https://webgl2fundamentals.org/webgl/lessons/webgl-data-textures.html) for a bit more info.

The other ways to set this would be to:
* use the largest value of 1, 2, 4, 8 (valid alignment) that evenly divides the row length.
* pad the rows to fit some chosen alignment

I think the performance penalty on modern systems will be negligible and 1 should always work, so I set it to that.

packages/core/src/renderers/webgl_buffers.ts
aganders3: I agree - I discussed briefly with @shlomnissan as well who is more familiar with how binding attributes works. I think there's a way to bind implicitly or perhaps by name instead of index, and we can skip using `layout` in the shaders.

edit: I think it would be similar to how we're handling uniforms now - I can make a PR for this

packages/core/src/core/geometry.ts
phoenixAja: property name change would work!

packages/core/src/objects/textures/texture.ts
phoenixAja: Yeah, I mean we could pass in the constructor explicitly like:

`ctor: { new(length: number): DataTextureTypedArray }` 

but that would require knowing the type at the call site, which runs into the same issue — just shifted earlier. So we’re trading one cast for another unless we restructure the API.

packages/core/src/renderers/webgl_buffers.ts
shlomnissan: IIRC what we discussed was binding by index or by name (using layout locations or [`glBindAttribLocation`](https://docs.gl/es3/glBindAttribLocation)). I looked into it at that time and this is what I remember:

The recommended approach is to use static bindings, which can be achieved either by specifying `layout(location = X)` directly in the vertex shader or by calling `glBindAttribLocation` in application code before linking the program. The alternative—querying attribute locations at runtime using `glGetAttribLocation`—is obviously less robust.

The key difference between `layout(location = X)` and `glBindAttribLocation` lies in who owns the mapping between attribute names and locations:
- With layout, the shader defines the mapping.
- With glBindAttribLocation, the application defines the mapping before linking.

The coupling between shader and application still exists either way—you must match names or locations—but I _personally_ prefer `glBindAttribLocation` because name-based coupling is easier to grep for and tends to make consistency issues more obvious. You can argue that layout locations are better because they eliminate the need to do any bindings before link time.

Either way, I agree with @phoenixAja using an enum will be better, and if we are using an enum, we might as well use `glBindAttribLocation` to centralize the mapping, but the latter is not a big deal either way -- especially when we implement shader includes, in which case all vertex attributes can be specified in a single file.

packages/core/src/objects/textures/texture_2d_array.ts
shlomnissan: AFAIK this is true for WebGL because of the high-level browser abstractions. In OpenGL, a performance penalty from unaligned data is possible depending on the driver or hardware, but in most cases it's negligible. I generally consider alignment because modern APIs like Vulkan require strict memory alignment. That said setting the unpack alignment to 1 in this case is probs the best choice.

packages/core/src/objects/textures/texture.ts
shlomnissan: TS type tricks are so annoying! I'm a little confused about why we need the `DataTextureTypedArrayConstructor` type to begin with? Wouldn't something like this work:

```ts
const Constructor = Object.getPrototypeOf(data[0]).constructor;
const result = new Constructor(totalLength);
```

packages/core/examples/points/main.ts
```diff
@@ -0,0 +1,166 @@
...
+} from "@";
+
+import { vec3 } from "gl-matrix";
+
+const imageUrl =
+  "https://files.cryoetdataportal.cziscience.com/10445/TS_100_3/Reconstructions/VoxelSpacing4.990/Tomograms/100/TS_100_3.zarr";
+const ribosomesUrl =
+  "https://files.cryoetdataportal.cziscience.com/10445/TS_100_3/Reconstructions/VoxelSpacing4.990/Annotations/104/cytosolic_ribosome-1.0_point.ndjson";
+const ferritinUrl =
+  "https://files.cryoetdataportal.cziscience.com/10445/TS_100_3/Reconstructions/VoxelSpacing4.990/Annotations/101/ferritin_complex-1.0_point.ndjson";
```
shlomnissan:
```suggestion
const baseUrl = "https://files.cryoetdataportal.cziscience.com/10445/TS_100_3/Reconstructions/VoxelSpacing4.990";
const imageUrl = `${baseUrl}/Tomograms/100/TS_100_3.zarr`;
const ribosomesUrl = `${baseUrl}/Annotations/104/cytosolic_ribosome-1.0_point.ndjson`;
const ferritinUrl = `${baseUrl}/Annotations/101/ferritin_complex-1.0_point.ndjson`;
```

packages/core/examples/points/main.ts
```diff
@@ -0,0 +1,166 @@
...
+// image scale for the highest res + points coords
+const IMAGE_SCALE_0 = 4.99; // nm/px
+// image scale for the lowest res (the image we show)
+const IMAGE_SCALE_2 = 19.96; // nm/px
```
shlomnissan: 1. How these scale values were determined? Are they derived from the image metadata?
2. Are we switching between low/high-res in this example? Why do we need both values?

packages/core/examples/points/main.ts
```diff
@@ -0,0 +1,166 @@
...
+const INITIAL_Z_POSITION = 918.16;
+
+const fetchNDJson = async (url: string) => {
```
shlomnissan: Nit: The name `fetchNDJson` might be a bit misleading—it suggests only fetching/parsing, but the function also returns an array of point locations. Consider renaming it.

shlomnissan:
@aganders3 I'm finding it a bit difficult to draw a clear distinction between layers and renderable objects. For example, I think of a 'point' as a renderable object, and a 'points layer' as something that renders multiple points. I hadn’t expected us to introduce a separate 'points geometry' type. Could you share more about how you're thinking about this? It would help me better understand the structure as I review the PR.

packages/core/examples/points/main.ts
```diff
@@ -0,0 +1,166 @@
...
+const ferritinLocations = await fetchNDJson(ferritinUrl);
+
+class Particles extends Layer {
+  private points_: vec3[] = [];
+  private color_: [number, number, number] = [0, 0, 0];
+  private marker_: number = 0;
+  private needsUpdate_: boolean = true;
+  public position: number = 0;
+
+  constructor(points: vec3[], color: [number, number, number], marker: number) {
```
shlomnissan: I think it makes sense to introduce a simple RGB color type now, and gradually extend and migrate it over time—but totally up to you.

packages/core/examples/points/main.ts
```diff
@@ -0,0 +1,166 @@
...
+  });
+};
+
+const ribosomeLocations = await fetchNDJson(ribosomesUrl);
+const ferritinLocations = await fetchNDJson(ferritinUrl);
+
+class Particles extends Layer {
+  private points_: vec3[] = [];
+  private color_: [number, number, number] = [0, 0, 0];
+  private marker_: number = 0;
```
shlomnissan: Would it make sense to declare these variables as `readonly`?

packages/core/examples/points/main.ts
```diff
@@ -0,0 +1,166 @@
...
+  private needsUpdate_: boolean = true;
+  public position: number = 0;
```
shlomnissan: Is public visibility necessary here? I didn’t see any external usage in this PR, but let me know if I missed something.

packages/core/examples/points/main.ts
```diff
@@ -0,0 +1,166 @@
...
+
+  constructor(points: vec3[], color: [number, number, number], marker: number) {
+    super();
+    this.setState("initialized");
+    this.points_ = points;
+    this.color_ = color;
+    this.marker_ = marker;
+    this.update();
```
shlomnissan: Calling `update()` directly from a layer's constructor seems problematic—it can lead to issues like incomplete initialization or redundant updates that result in inconsistent state. I understand this is example code, but if these examples are intended to stick around, we should keep them clean and accurate, since users may look to them as a reference for how to use the library.

packages/core/examples/points/main.ts
```diff
@@ -0,0 +1,166 @@
...
+    this.setState("ready");
+  }
+
+  public setPosition(z: number) {
```
shlomnissan: The name `setPosition` could be misleading here, since it's setting only the z component. How about `setDepth()`?

aganders3:
Good questions, and I still struggle with these definitions so this is a good exercise.

Here's how I think about it right now:
* _Renderable objects_ represent a combination of uniforms and shaders that are drawn with a single call. So I would not expect a CPU loop in `render(object)`.
* _Geometry_ contains the vertex attributes you can use to affect _how_ the object is drawn in a performant way.
* _Layers_ are collections of renderable objects that provide a higher-level API and may represent a more specific scientific visualization. They map the scientific data into their renderable objects by encoding it into attributes and uniforms.
  * For example a renderable object would likely have no concept of a time or slice number, but a layer would.
  * A "Tracks" layer might combine both points and lines, which need to be drawn separately.
  * As another example, a layer might map "time since last cell division" into "point color" for example. "Point color" has no inherent scientific meaning, but is a tool for encoding data.

Some of this comes down to anticipating use-cases to leverage GPU parallelism. I think it would be inefficient to draw points separately if the renderable object was a single "Point" and the shader would only draw one. This may not have a huge impact when drawing a few hundred points (particles on a CryoET image), but I think it will be a problem when drawing tens of thousands (point cloud of zebrafish embryo cells).

It's less clear to me _which_ attributes should be configurable per-point within a single call.  I think this then touches on your questions about geometry. Maybe if we want to draw a point cloud with separate marker types we should do each marker as separate renderable object - I'm not sure. Per-point size and color seem definitely valuable to me to encode additional information. I think this becomes less of a problem with non-interleaved attributes and shader snippets - we can then enable all sorts of performant features with little/no cost of not using them.

packages/core/examples/points/main.ts
```diff
@@ -0,0 +1,166 @@
...
+    this.position = z;
+    this.needsUpdate_ = true;
+  }
+
+  public update() {
+    /**
+     * TODO: re-creating the geometry is not efficient,
+     * but it's the simplest way to update all the
+     * vertex attributes for now.
```
shlomnissan: I think TODOs should be more actionable. In this case, something like: `Refactor this function once geometry objects support in-place data updates to avoid recreating objects and buffers on every update` would be clearer.

shlomnissan: I don't think our library is at the stage where we need to work around existing types. Instead, we should treat new work as an opportunity to extend our types where it makes sense. For example, if you're recreating the data because the `Geometry` type doesn't support something, we should consider extending `Geometry`. And if we need to support non-interleaved buffers, we should make sure that's properly sequenced and planned. That said, this approach is totally fine for this PR—just something to keep in mind. I might take a pass through our open issues soon so we can start prioritizing them.

packages/core/src/objects/textures/texture.ts
aganders3: Thanks! This indeed works. I had initially tried just using `data[0].constructor` which doesn't work. I don't really know what `getPrototypeOf` does, so I have some reading to do.

packages/core/examples/points/main.ts
```diff
@@ -0,0 +1,166 @@
...
+let debounce: number;
+zSlider.addEventListener("input", (event) => {
+  clearTimeout(debounce);
+  const value = (event.target as HTMLInputElement).valueAsNumber;
+  debounce = setTimeout(async () => {
+    imageLayer.setPosition(value * IMAGE_SCALE_2);
+    ribosomes.setPosition(value * IMAGE_SCALE_2);
+    ferritin.setPosition(value * IMAGE_SCALE_2);
+  }, 20);
+});
```
shlomnissan: "I’m fine with using Lodash for common utilities like debouncing—especially in example code—as long as we can import individual functions to avoid unnecessary bloat.

packages/core/examples/points/main.ts
```diff
@@ -0,0 +1,166 @@
...
+const imageLayer = new ImageSeriesLayer({
+  source: imageSource,
+  region,
+  seriesDimensionName: zDimName,
+  channelProps: [{ color: [1, 1, 1], contrastLimits: [-0.00001, 0.00001] }],
+});
+const setCameraFrame = (newState: LayerState) => {
+  if (newState === "ready" && imageLayer.extent !== undefined) {
+    camera.setFrame(0, imageLayer.extent.x, 0, imageLayer.extent.y);
+    renderer.setControls(new PanZoomControls(camera, camera.position));
```
shlomnissan, line 221: Is there a reason we need to create `PanZoomControls` twice? Could we just set it up here and skip the initialization-time one?

packages/core/src/renderers/webgl_buffers.ts
aganders3: > * With layout, the shader defines the mapping.
> * With glBindAttribLocation, the application defines the mapping before linking.

It feels like we kind of have both at the moment...the shader defines the mapping but then the application keeps its own mapping (in this series of `if` statements) as well, and we have to keep them in sync. I feel like it would just be easier to go by name.

In any case I think it's out of scope here, but I'd be happy to explore the refactor in another PR.

packages/core/examples/points/main.ts
```diff
@@ -0,0 +1,166 @@
...
+const imageSource = new OmeZarrImageSource(imageUrl);
+const loader = await imageSource.open();
+const attributes = await loader.loadAttributes();
+const zDimName = "z";
+const zAxisIndex = attributes.dimensionNames.findIndex(
+  (dim) => dim === zDimName
+);
+const zMin = 0;
+const zMax = attributes.shape[zAxisIndex];
+const zSlider = document.querySelector<HTMLInputElement>("#z-slider")!;
```
shlomnissan, line 203: It might be a good idea to start using something like [lil-gui](https://lil-gui.georgealways.com/) for these examples—especially if we plan to keep them around.

packages/core/src/objects/geometry/points_geometry.ts
shlomnissan: I’m a bit surprised that `gl-matrix` doesn’t support named accessors like `x`, `y`, and `z`. As for holding off on introducing a Color type, waiting for a broader refactor makes sense—but I also think it’s perfectly fine to introduce a new type now and update things incrementally. With a small team, it’s easy to share context around recent changes.

packages/core/src/objects/geometry/points_geometry.ts
```diff
@@ -0,0 +1,68 @@
...
+  size: number;
+  // marker is an index into a texture array
+  marker: number;
+}
+
+export class PointsGeometry extends Geometry {
```
shlomnissan: In this case, introducing a `PointsGeometry` feels a bit unnecessary. Similarly, passing a geometry into an `ImageRenderable` also seems redundant. To me, the term 'renderable object' implies that the geometry should be encapsulated within it. Why not simply instantiate a plain geometry object directly inside the renderable object classes? After all, a geometry is just a configurable buffer. I realize I may have contributed to this pattern by creating geometry instances when defining geometric primitives—but my intent there was to support factory-style creation, not to promote external geometry management in every case.

I read your [response](https://github.com/chanzuckerberg/imaging-active-learning/pull/242#issuecomment-2876924025) and I think your definition of "renderable object" is great, but I still don't understand why this abstraction is needed (at least for now when we don't need a point cloud with separate marker types.

packages/core/src/objects/geometry/points_geometry.ts
```diff
@@ -0,0 +1,68 @@
...
+    super();
+    this.addAttribute({
+      type: "position",
+      itemSize: 3,
+      offset: 0,
+    });
+    this.addAttribute({
+      type: "color",
+      itemSize: 3,
+      offset: 3 * Float32Array.BYTES_PER_ELEMENT,
```
shlomnissan: Can we calculate the offset dynamically based on existing attributes? I think we already have a built-in accessor for calculating the current stride.

packages/core/src/objects/geometry/points_geometry.ts
```diff
@@ -0,0 +1,68 @@
...
+        point.marker,
+      ])
+    );
+  }
+
+  public get type() {
+    return "PointsGeometry";
+  }
+
+  public getMarkerRange() {
```
shlomnissan: 1. This further reinforces my sense that the abstraction might not be strictly necessary, since the method is always called by the renderable object itself.
2. What does a min/max marker range means? Are these the min/max slices that represent each shape in a texture array?

packages/core/src/objects/renderable/points.ts
```diff
@@ -0,0 +1,92 @@
+import { RenderableObject } from "../../core/renderable_object";
+import { PointsGeometry } from "../geometry/points_geometry";
+import { Texture2DArray } from "../textures/texture_2d_array";
+
+type ProceduralMarker = "circle" | "square" | "triangle";
+type Marker = ProceduralMarker | Float32Array;
+const WIDTH = 256;
+const HEIGHT = 256;
```
shlomnissan: Are these the width and height of each individual sprite, or of the entire atlas? Maybe rename for clarity?

packages/core/src/objects/renderable/points.ts
```diff
@@ -0,0 +1,92 @@
...
+
+export class Points extends RenderableObject {
+  private atlas_: Texture2DArray;
+
+  constructor(geometry: PointsGeometry, markerAtlas?: Texture2DArray) {
```
shlomnissan: I think having the option to provide a custom atlas is a nice idea, but it feels a bit premature for the initial use case. If I did want to use a custom atlas, would passing it here be enough? Or would it also require additional changes—like making the sprite dimensions configurable or extending the current set of supported shapes (circle, square, triangle)?

packages/core/src/objects/renderable/points.ts
```diff
@@ -0,0 +1,92 @@
...
+    super();
+    const [minMarker, maxMarker] = geometry.getMarkerRange();
+    this.atlas_ =
+      markerAtlas ?? makeSpriteAtlas(["circle", "square", "triangle"]);
```
shlomnissan: It feels a bit odd that we need to generate new atlases when the dimensions are hard-coded and the output is always the same. Also, is dynamically drawing the atlas strictly necessary? Couldn’t we just use a high-resolution precomputed atlas and rely on mipmapping to support different sizes?

packages/core/src/objects/renderable/points.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    if (minMarker < 0 || maxMarker >= this.atlas_.depth) {
+      throw new Error(
+        `Markers must be in the range [0, ${this.atlas_.depth - 1}] (marker atlas depth)`
```
shlomnissan: I find the term 'atlas depth' a bit confusing—since markers are 2D images, it's not immediately clear what 'depth' refers to. Is it meant to indicate the number of sprites or slices in the atlas? If I ran into this error message without reading the code, I would fine this confusing.

packages/core/src/objects/renderable/points.ts
```diff
@@ -0,0 +1,92 @@
...
+  const data = new Float32Array(WIDTH * HEIGHT);
+  for (let i = 0; i < HEIGHT; i++) {
+    for (let j = 0; j < WIDTH; j++) {
+      if (j >= (WIDTH - i) / 2 && j <= (WIDTH + i) / 2) {
+        data[i * WIDTH + j] = 1.0;
+      }
+    }
+  }
+  return data;
+};
```
shlomnissan: The code for dynamically generating an atlas feels a bit out of place here. As mentioned earlier, I don’t think dynamic generation is strictly necessary at this stage—and even if we do need it, I’m not sure the renderable object should be responsible for creating textures. A cleaner approach might be to treat the atlas as a required input, along with metadata describing its configuration, and have the layer always pass it in. The atlas generation logic seems better suited as a separate utility or resource, rather than something embedded in the renderable object itself.

packages/core/src/objects/renderable/points.ts
```diff
@@ -0,0 +1,92 @@
...
+
+// TODO: can we do border colors with this method?
+// TODO: do we need to soften the edges?
+const markerDispatch: Map<ProceduralMarker, () => Float32Array> = new Map([
```
shlomnissan: Could you say more about why this needs to be so dynamic at this stage? Even if we're generating the atlas, why not just include all available sprites by default? Right now it supports:

- Providing a custom atlas
- Generating an atlas
- Selecting which markers to include

That level of flexibility feels a bit out of scope for this iteration. It might be better to start simpler and expand later with real requirements.

packages/core/src/objects/textures/texture_2d_array.ts
```diff
@@ -58,4 +60,16 @@ export class Texture2DArray extends Texture {
     texture.unpackAlignment = chunk.rowAlignmentBytes;
     return texture;
   }
+
+  public static createWithArrays(
```
shlomnissan: This looks like a utility that's not directly related to textures—should it live in a separate helper function? Also, do we really need to concatenate typed arrays in the first place? I assume this ties back to how the sprite sheet is structured, but it raises the question: why are we generating each sprite in its own array instead of building them into a single buffer from the start?

packages/core/src/objects/textures/texture.ts
```diff
@@ -11,6 +11,31 @@ export type TextureDataType = "unsigned_byte" | "unsigned_short" | "float";
 export type TextureUnpackRowAlignment = 1 | 2 | 4 | 8;
 
 export type DataTextureTypedArray = Uint8Array | Uint16Array | Float32Array;
+type DataTextureTypedArrayConstructor =
+  | typeof Uint8Array
+  | typeof Uint16Array
+  | typeof Float32Array;
+
+export function concatenateTypedArrays(
```
shlomnissan: This looks like a utility that's not directly related to textures—should it live in a separate helper function? Also, do we really need to concatenate typed arrays in the first place? I assume this ties back to how the sprite sheet is structured, but it raises the question: why are we generating each sprite in its own array instead of building them into a single buffer from the start?

packages/core/src/renderers/shaders/points_frag.glsl
```diff
@@ -0,0 +1,20 @@
...
+layout (location = 0) out vec4 fragColor;
+
+uniform mediump sampler2DArray markerAtlas;
+
+in vec4 color;
+flat in uint marker;
+
+
+void main() {
+    float alpha = texture(markerAtlas, vec3(gl_PointCoord, marker)).r;
```
shlomnissan, line 14: Should we multiply the final alpha by the atlas alpha to support soft edges or anti-aliasing? Right now, we're discarding transparent pixels but not using the alpha channel for blending.

packages/core/src/renderers/shaders/points_frag.glsl
```diff
@@ -0,0 +1,20 @@
...
+    if (alpha < 1e-3) {
```
shlomnissan: Would it make sense to define the alpha discard threshold as a named constant for readability?

packages/core/src/renderers/shaders/points_vert.glsl
```diff
@@ -0,0 +1,23 @@
+#version 300 es
+
+precision mediump float;
+
+layout (location = 0) in vec3 inPosition;
+layout (location = 7) in vec3 inColor;
+layout (location = 8) in float inSize;
+layout (location = 9) in float inMarker;
```
shlomnissan, line 8: As we begin refactoring shaders, I think it would be helpful to establish some naming guidelines we can all follow and share. I’m not sure if there’s a configurable linter available, but having some consistency would go a long way.

packages/core/src/renderers/webgl_renderer.ts
```diff
@@ -152,6 +151,17 @@ export class WebGLRenderer extends Renderer {
     return this.shaders_.get(type)!;
   }
 
+  private getShaderMode(type: Shader) {
+    switch (shaderCode[type].mode) {
```
shlomnissan: 1. I didn’t realize we were handling primitive mode in the shader—this feels a bit out of place. Shouldn’t that be defined by the renderable object instead? It seems more appropriate for the draw call level rather than something embedded in the shader. Curious to hear your thoughts.
2. Regardless, I think `getShaderMode` isn’t the most descriptive name. If possible, I’d suggest moving the primitive mode out of the shader in this PR.

shlomnissan (changes requested):
Great work—it's always exciting to see a new renderable type come together! I'm marking this as blocking for now because I'd like to better understand the decision to introduce a new geometry type. On the non-blocking side, I think it's worth revisiting the flexibility around dynamic shape atlas generation at this stage, and I’m also curious about a few smaller choices—like assigning primitive mode in the shader and the need for concatenating typed arrays. Let me know if you'd like to chat about any of this over zoom :) 

Apologies for the delay in reviewing this PR

packages/core/src/renderers/webgl_buffers.ts
shlomnissan: > I feel like it would just be easier to go by name

I agree. It’s a bit hard to quantify since there’s strong coupling either way, but the value becomes more apparent when the application already defines an enum mapping—it makes the integration feel more structured.

> In any case I think it's out of scope here

Can you add a new issue with some of the context from this thread ^ It's so easy for me to forget setup-related, "set and forget" decisions.

packages/core/src/renderers/shaders/points_frag.glsl
aganders3: I think we need to enable `OES_texture_float_linear` and linear filtering on the texture for this to make sense with binary markers. I'm not sure what platforms we need to support but that also suggests to me checking and falling back to nearest sampling if not available.

packages/core/src/renderers/webgl_buffers.ts
aganders3: I think we already have a relevant issue in https://github.com/chanzuckerberg/imaging-active-learning/issues/62 - this mention should tie it to this conversation.

packages/core/src/objects/geometry/points_geometry.ts
aganders3: > In this case, introducing a `PointsGeometry` feels a bit unnecessary. Similarly, passing a geometry into an `ImageRenderable` also seems redundant. To me, the term 'renderable object' implies that the geometry should be encapsulated within it.

I see your point on this now - I'd be happy to make this change! I think there is value in having these separate geometry types, but that's _IFF_  the we're passing geometry into the renderable objects. This was the pattern previously but I think we can get rid of that whole layer of abstraction for many cases.

> I realize I may have contributed to this pattern by creating geometry instances when defining geometric primitives—but my intent there was to support factory-style creation

I see this too, and I think my misinterpretation comes from "geometry" being a bit overloaded.

aganders3:
Thanks - I think I have a fairly clear idea of how to improve this now.

> I'm marking this as blocking for now because I'd like to better understand the decision to introduce a new geometry type.

I'm totally on board with removing the geometry type. As you mentioned I think this can also be done for our other renderable objects, so I'm happy to explore that in another PR.

> On the non-blocking side, I think it's worth revisiting the flexibility around dynamic shape atlas generation at this stage, and I’m also curious about a few smaller choices—like assigning primitive mode in the shader and the need for concatenating typed arrays.

I think all of these can be addressed and simplified as well. I still think it's fine to generate the default marker atlas dynamically, but I can refactor this to be simpler and just initialized the default once for example. Likewise I think you're right we can remove the need for concatenating the typed arrays. I find it convenient but perhaps not a big enough value-add for a kind of fiddly thing at this point.

packages/core/examples/points/main.ts
aganders3: I've never used lodash but I see it as a dependency in just about every project, so I'm cool with that too.

packages/core/examples/points/main.ts
aganders3: I kind of like keeping the examples super minimal, but in the end lil-gui menus look _a lot_ nicer than unstyled HTML and the code probably gets out of the way more compared to all the boilerplate you end up with. So I'm +1 on this.

packages/core/src/objects/geometry/points_geometry.ts
aganders3: I went ahead with adding a Color type and using it here plus a few places it crept into from there. If this closes #59 we should make sure to create a new issue to use `Color` everywhere it makes sense.

packages/core/src/objects/textures/texture.ts
aganders3: Removed

packages/core/src/objects/textures/texture_2d_array.ts
aganders3: Removed by refactor

packages/core/src/objects/textures/texture.ts
aganders3: Removed

packages/core/src/objects/renderable/points.ts
aganders3: I agree they don't need to be dynamically generated, and I updated that so they're only calculated on first-use. I think it's fine to have them procedurally generated though I did move these to the "example" code for what that's worth, but I could see having them in the library somewhere to provide more shapes at some point.

I'm not sure we need mipmapping because the practical size range is limited, and they're billboarded towards the camera.

packages/core/src/objects/renderable/points.ts
aganders3: >  I’m not sure the renderable object should be responsible for creating textures.

I don't see a problem with this - in fact I find it more intuitive to create textures within a renderable object rather than a Layer. I prefer the code within the layer to stay higher-level, but...

> A cleaner approach might be to treat the atlas as a required input, along with metadata describing its configuration, and have the layer always pass it in.

I do like the idea of making it a required input, so I moved this in any case for now.

> The atlas generation logic seems better suited as a separate utility or resource, rather than something embedded in the renderable object itself.

This seems reasonable to me as well. For now I think it's fine to keep it in the example.

packages/core/src/objects/renderable/points.ts
aganders3: Yeah I got a bit wrapped up here. I agree this is unnecessary and I removed it.

packages/core/src/objects/textures/texture_2d_array.ts
aganders3: Removed in refactor

packages/core/src/renderers/shaders/points_frag.glsl
aganders3: To provide proper contrast I think the renderable should also support a secondary color (i.e. a border color). I think this can be added at the same time in a future PR.

packages/core/src/renderers/webgl_renderer.ts
aganders3: I think shaders are generally pretty coupled to the primitive type, so it makes sense to me to define them in one spot. This could be a small problem for implementing another (WebGPU) renderer. In WebGPU this would not use the `'point-list'` primitive topology, because it's not very capable. Instead that would likely to use instancing and `'triangle-list'`.

I don't feel strongly about this though, so I moved this to the renderable object.

shlomnissan (approved):
Unblocking since I'm away. Merge when you feel it's ready 🚀

phoenixAja (approved):
nice!

Issue #236, opened 2025-04-29
Make Renderer.render() an abstract method


phoenixAja:
Right now, the Renderer base class defines a fixed render() method and uses small hooks like beginTransparentPass() and endTransparentPass() to manage GL state. As we add more rendering stages, this approach can become harder to maintain.

We should make render() an abstract method and let each concrete renderer (like WebGLRenderer) define its own render loop. This will make the code cleaner and easier to extend.

andy-sweet:
@phoenixAja : this was closed by #244 right?

If not, please reopen.

PR #244, opened 2025-05-08
refactor: make renderer method virtual, add WebGLState class

phoenixAja:
This PR addresses [#236](https://github.com/chanzuckerberg/imaging-active-learning/issues/236) by making the `Renderer.render()` method abstract and moving the full render loop into WebGLRenderer. This makes it easier to reason about and extend renderers in the future.

Changes:

* Removed beginTransparentPass() and endTransparentPass() hook methods.

* Introduced a new WebGLState utility to deduplicate common GL state changes (e.g., `gl.enable`, `gl.depthMask`, `gl.blendFunc`). It tracks previous values to avoid redundant driver calls, which should improve performance

* Added `syncActiveCamera()` helper in the base class to encapsulate camera switching logic.

* Added `renderLayer()` into WebGLRenderer for layer-level object rendering.


packages/core/src/core/renderer.ts
```diff
@@ -36,32 +33,15 @@ export abstract class Renderer {
     });
   }
 
-  public render(layerManager: LayerManager, camera: Camera) {
-    this.clear();
+  protected syncActiveCamera(camera: Camera) {
```
shlomnissan: 1. Consider renaming this method to `setActiveCamera` for consistency. Alternatively, we could define a mutator method like `protected set activeCamera(camera: Camera)` to make the intent clearer.
2. In `updateActiveCamera`, we currently infer the camera type at runtime to decide between `setAspectRatio` and `setViewportAspectRatio`. We could simplify this by introducing an abstract method on the base `Camera` class that all subclasses implement—this would remove the need for type checks. This change seems straightforward enough to include in this PR, though it’s optional.
3. Still thinking this one through, but I’m not convinced we need to store a reference to the active camera in the renderer. It might be cleaner to handle resizing within the base camera class and trigger update logic from the renderer when needed. Let me know if you have any thoughts about this.

packages/core/src/core/renderer.ts
```diff
@@ -36,32 +33,15 @@ export abstract class Renderer {
...
-    this.beginTransparentPass();
-    for (const layer of transparent) {
-      layer.update();
-      if (layer.state !== "ready") continue;
-      layer.objects.forEach((_, i) => this.renderObject(layer, i));
-    }
-    this.endTransparentPass();
   }
 
+  public abstract render(layerManager: LayerManager, camera: Camera): void;
```
shlomnissan, line 40: :+1:

packages/core/src/renderers/WebGLState.ts
```diff
@@ -0,0 +1,41 @@
...
+  private capabilities_ = new Map<GLenum, boolean>();
+  private depthMaskEnabled_: boolean | null = null;
+  private blendSrcFactor_: GLenum | null = null;
+  private blendDstFactor_: GLenum | null = null;
+
+  constructor(gl: WebGL2RenderingContext) {
+    this.gl_ = gl;
+  }
+
+  enable(cap: GLenum) {
```
shlomnissan: I prefer to explicitly declare access specifiers (`public`, `private`, `protected`) for clarity and consistency. In this case, I think it would be a cleaner design to make the low-level enable and disable methods private. This way, external consumers of the class interact only through higher-level, intention-revealing methods like `state.setDepthTesting(true)` instead of directly calling something like `state.enable(GL_DEPTH_TEST)`.

Encapsulating these OpenGL-specific operations behind well-named public methods not only improves readability but also decouples external code from implementation details, making the API easier to use and maintain.

packages/core/src/renderers/WebGLState.ts
```diff
@@ -0,0 +1,41 @@
+export class WebGLState {
+  private readonly gl_: WebGL2RenderingContext;
+
+  private capabilities_ = new Map<GLenum, boolean>();
```
shlomnissan: I’ve seen this referred to as “features” before, but I’ve always felt that “capabilities” is more accurate and better aligned with OpenGL terminology. If you’re looking for something even more explicit, `enabledCapabilities` would be a solid alternative—it clearly communicates that the map is tracking which capabilities are currently enabled.

packages/core/src/renderers/webgl_renderer.ts
```diff
@@ -26,6 +29,7 @@ export class WebGLRenderer extends Renderer {
   private readonly shaders_: Map<Shader, WebGLShaderProgram>;
   private readonly bindings_: WebGLBuffers;
   private readonly textures_: WebGLTextures;
+  private readonly webglState_: WebGLState;
```
shlomnissan: I suggest renaming `webglState_` to `state_` for consistency with other variables like `shaders_`, `textures_`, and `bindings_`.

packages/core/src/renderers/webgl_renderer.ts
```diff
@@ -62,17 +86,23 @@ export class WebGLRenderer extends Renderer {
     if (layer.transparent) {
       switch (layer.blendMode) {
         case "additive":
-          this.gl.blendFunc(this.gl.SRC_ALPHA, this.gl.ONE);
+          this.webglState_.setBlendFunc(this.gl.SRC_ALPHA, this.gl.ONE);
```
shlomnissan: The goal of this state class is to eventually encapsulate all WebGL-specific context switching calls (even things like `gl.viewport`, `gl.clear`, etc) and decouple them from the rest of the code. With that in mind, I think we can move the entire switch statement, along with the logic for enabling and disabling blending, into the WebGL state class and simplify the API (pass a blending mode instead of a source and destination).

Here's how it's implemented in Three.js: [WebGLState.js](/dev/src/renderers/webgl/WebGLState.js#L682)

And here's a simpler example that follows the same strategy:

```cpp
enum class Blending {
    None,
    Normal,
    Additive,
    Subtractive,
    Multiply
};

auto GLState::SetBlending(Blending blending) -> void {
    if (curr_blending_ != blending) {
        if (blending == Blending::None) {
            Disable(GL_BLEND);
        } else {
            Enable(GL_BLEND);
            switch (blending) {
            case Blending::Normal:
                glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA);
                break;
            case Blending::Additive:
                glBlendFunc(GL_SRC_ALPHA, GL_ONE);
                break;
            case Blending::Subtractive:
                glBlendFunc(GL_ZERO, GL_ONE_MINUS_SRC_COLOR);
                break;
            case Blending::Multiply:
                glBlendFunc(GL_ZERO, GL_SRC_COLOR);
                break;
            case Blending::None:
                break;
            }
        }
        curr_blending_ = blending;
    }
}
```

shlomnissan (changes requested):
This is a strong start. I'm requesting changes to encourage you to consider improving the WebGL state API by encapsulating the source and destination blend factors (see my last comment). If you see a better approach or have concerns, feel free to push back—otherwise, everything else looks great!

aganders3:
I think this also relates to https://github.com/chanzuckerberg/imaging-active-learning/issues/234, I'm just adding the link for posterity.

I agree this is a good direction, but it's not urgent so worth making some of the suggested improvements before merging. Thanks for picking this up!

packages/core/src/core/renderer.ts
phoenixAja: done, added a setter

packages/core/src/renderers/WebGLState.ts
phoenixAja: renamed to enabledCapabilities

packages/core/src/renderers/webgl_renderer.ts
phoenixAja: good idea, added

packages/core/src/renderers/WebGLState.ts
phoenixAja: done

shlomnissan (approved):
LGTM 🚀

PR #247, opened 2025-05-09
fix: Add "use client" to API components

bchu1:
I don't think this is 100% necessary, but it might be helpful for Next.js apps? Especially since I noticed the Organelle Box code was importing the viewer component dynamically?

### DIFF
--- a/packages/react/src/components/viewers/OmeZarrImageViewer/OmeZarrImageViewer.tsx
+++ b/packages/react/src/components/viewers/OmeZarrImageViewer/OmeZarrImageViewer.tsx
@@ -1,3 +1,5 @@
+"use client";
+
 import { Renderer } from "./components/Renderer";
 import { useOmeZarrViewer } from "./useOmeZarrImageViewer";
 import { Region } from "@idetik/core";
--- a/packages/react/src/components/viewers/OmeZarrImageViewer/components/ChannelControlsList/ChannelControlsList.tsx
+++ b/packages/react/src/components/viewers/OmeZarrImageViewer/components/ChannelControlsList/ChannelControlsList.tsx
@@ -1,3 +1,5 @@
+"use client";
+
 import {
   AccordionHeader,
   AccordionDetails,


kyleawayan:
I think it may not be? I think the dynamic import you saw (the `next/dynamic` one) is more for the rendering libraries.

To import this component into Next.js though, the component using it will always have to be inside a client component anyway (and to be able to use `next/dynamic` as well), so the implementation would always have to look like below in Next.js. So I think the `"use client"` in the actual library is not needed!

```tsx
"use client";

import dynamic from "next/dynamic";
// ...
const OmeZarrImageViewer = dynamic(
  () =>
    import("@idetik/react") // ...
```

bchu1:
> I think it may not be? I think the dynamic import you saw (the `next/dynamic` one) is more for the rendering libraries.
> 
> To import this component into Next.js though, the component using it will always have to be inside a client component anyway (and to be able to use `next/dynamic` as well), so the implementation would always have to look like below in Next.js. So I think the `"use client"` in the actual library is not needed!
> 
> ```tsx
> "use client";
> 
> import dynamic from "next/dynamic";
> // ...
> const OmeZarrImageViewer = dynamic(
>   () =>
>     import("@idetik/react") // ...
> ```

Oh I see, is `dynamic()` just an optimization?

edit: ah I just saw your parent component is also `"use client"`

kyleawayan:
I didn't put in the snippet above, but the `dynamic()` is there so we can [skip SSR](https://nextjs.org/docs/app/guides/lazy-loading#skipping-ssr), as rendering libraries such as three.js and like we have now, don't work when server side rendered. And yep to use `dynamic()` it has to be inside a client component, so that means idetik, at least for Next.js App Router, the parent component will always have to be client component to be able to disable SSR

bchu1:
> I didn't put in the snippet above, but the `dynamic()` is there so we can [skip SSR](https://nextjs.org/docs/app/guides/lazy-loading#skipping-ssr), as rendering libraries such as three.js and like we have now, don't work when server side rendered. And yep to use `dynamic()` it has to be inside a client component, so that means idetik, at least for Next.js App Router, the parent component will always have to be client component to be able to disable SSR

Interesting, what happens in your app if you get rid of `ssr: false`?

If there's something the Idetik package can do to prevent integrators from having to go down this route I'd like to look into it 🤔

kyleawayan:
I think the page just crashes lol, I think this is just something we have to address in the documentation, as I have to do this with other 3D libraries as well

bchu1:
> I think the page just crashes lol, I think this is just something we have to address in the documentation, as I have to do this with other 3D libraries as well

Super interesting, I'll take a look, thanks for explaining!

bchu1:
@phoenixAja @kyleawayan FWIW, I was able to load the viewer fine in the Embrella Next.js app without dynamic importing 🤷 (though I found some other bugs, which I will fix now 😆)

image

kyleawayan:
Oh interesting -- do you know if the page is being server side rendered though? (sorry I'm not familiar with Embrella)

aganders3:
I thought you couldn't create a WebGL context server-side, but I'm over my head here. I don't think SSR will save much since the data will need to be transferred to the client side for interactive rendering. Worst-case could this cause data to be fetched both server-side and client side?

Eliminating the dynamic import would be great though!

bchu1:
> Oh interesting -- do you know if the page is being server side rendered though? (sorry I'm not familiar with Embrella)

It's a Next.js app, if that's what you're asking?

bchu1:
> I thought you couldn't create a WebGL context server-side, but I'm over my head here. I don't think SSR will save much since the data will need to be transferred to the client side for interactive rendering. Worst-case could this cause data to be fetched both server-side and client side?
> 
> Eliminating the dynamic import would be great though!

Yeah I think the expectation is that SSR would just return nothing, it was just weird to me that that might cause an error

kyleawayan:
The error is on the Next.js side, this is what happens when `ssr: true`
```
Error: Switched to client rendering because the server rendering errored:

WebGL2RenderingContext is not defined
```
Then the viewer loads

bchu1:
> The error is on the Next.js side, this is what happens when `ssr: true`
> 
> ```
> Error: Switched to client rendering because the server rendering errored:
> 
> WebGL2RenderingContext is not defined
> ```
> 
> Then the viewer loads

Aha, I see it too.

PR #250, opened 2025-05-12
fix: Fix errors when integrating in SSR context

bchu1:
This should remove the need (for now...) of dynamically importing the library.

 - Removes immediate execution of code that references variables that only exist in the browser.
 - Updates more relative imports.

### DIFF
--- a/packages/core/src/renderers/webgl_shader_program.ts
+++ b/packages/core/src/renderers/webgl_shader_program.ts
@@ -157,34 +157,40 @@ export class WebGLShaderProgram {
   }
 }
 
-const SAMPLER_TYPES: ReadonlySet<GLenum> = new Set<GLenum>([
-  WebGL2RenderingContext.SAMPLER_2D,
-  WebGL2RenderingContext.SAMPLER_CUBE,
-  WebGL2RenderingContext.SAMPLER_3D,
-  WebGL2RenderingContext.SAMPLER_2D_ARRAY,
-  WebGL2RenderingContext.SAMPLER_2D_SHADOW,
-  WebGL2RenderingContext.SAMPLER_CUBE_SHADOW,
-  WebGL2RenderingContext.SAMPLER_2D_ARRAY_SHADOW,
-  WebGL2RenderingContext.INT_SAMPLER_2D,
-  WebGL2RenderingContext.INT_SAMPLER_3D,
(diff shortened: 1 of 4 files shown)


packages/core/src/renderers/webgl_shader_program.ts
```diff
@@ -157,34 +157,40 @@ export class WebGLShaderProgram {
...
-  WebGL2RenderingContext.INT_SAMPLER_2D_ARRAY,
-  WebGL2RenderingContext.UNSIGNED_INT_SAMPLER_2D,
-  WebGL2RenderingContext.UNSIGNED_INT_SAMPLER_3D,
-  WebGL2RenderingContext.UNSIGNED_INT_SAMPLER_CUBE,
-  WebGL2RenderingContext.UNSIGNED_INT_SAMPLER_2D_ARRAY,
-  WebGL2RenderingContext.MAX_SAMPLES,
-  WebGL2RenderingContext.SAMPLER_BINDING,
-]);
+const SAMPLER_TYPES: ReadonlySet<GLenum> =
+  typeof window !== "undefined" // Don't error in SSR contexts.
```
phoenixAja, line 161: can we throw an exception here so it doesn't error silently?

packages/react/src/components/Main.tsx
```diff
@@ -2,7 +2,7 @@ import React from "react";
 import { createRoot } from "react-dom/client";
 import cns from "classnames";
 
-import AppWithProviders from "./AppWithProviders.tsx";
+import AppWithProviders from "./AppWithProviders";
```
phoenixAja, line 5: huh thought i patched that with the submodule integration..

packages/react/src/components/viewers/OmeZarrImageViewer/components/ChannelControlsList/ChannelControlsList.tsx
```diff
@@ -7,7 +7,7 @@ import {
 import cns from "classnames";
 import { ChannelControl } from "./components/ChannelControl";
 import { ChannelProps } from "@idetik/core";
-import { useIdetik } from "components/hooks";
+import { useIdetik } from "../../../../hooks";
```
phoenixAja, line 10: i wish there was a better way...

phoenixAja (approved):
TY Brian!

PR #251, opened 2025-05-13
feat: Simplify <IdetikProvider>

bchu1:
In preparation for final refactoring PR, condense the context value by deleting properties derived from `imageSeriesLayer` and just exposing it directly.

Also add type safety to correlate `isInitialized` to `imageSeriesLayer` being defined.


phoenixAja (approved):
cool

PR #253, opened 2025-05-14
feat: Apply contrast slider changes to all tomogram channels in viewer

phoenixAja:
### Summary
Updated the contrast adjustment logic in `useOmeZarrImageViewer` to apply the selected contrastLimits to all channels.
Ensures that contrast changes from the UI are reflected across every channel, supporting both single-channel and multi-channel data.

### Tests & Checks
* tested examples

https://github.com/user-attachments/assets/717c060a-545e-4f8c-900b-d8f6b433229f

### DIFF
--- a/packages/react/src/components/viewers/OmeZarrImageViewer/useOmeZarrImageViewer.ts
+++ b/packages/react/src/components/viewers/OmeZarrImageViewer/useOmeZarrImageViewer.ts
@@ -57,7 +57,7 @@ export function useOmeZarrViewer({
   }, [imageLayer]);
 
   useEffect(() => {
-    const newSource = new OmeZarrImageSource(sourceUrl, 0);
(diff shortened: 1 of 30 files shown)


packages/react/src/components/viewers/OmeZarrImageViewer/useOmeZarrImageViewer.ts
```diff
@@ -140,6 +140,22 @@ export function useOmeZarrViewer({
     // eslint-disable-next-line react-hooks/exhaustive-deps -- Deps that trigger layer creation.
   }, [source, sourceUrl, region, seriesDimensionName]);
 
+  useEffect(() => {
```
bchu1, line 143: Add a comment that this is for when the fallback channels change?

bchu1 (approved):

PR #255, opened 2025-05-15
refactor: improve fallback handling for greyscale defaults in OmeZarr viewer

phoenixAja:
This PR addresses feedback on how greyscale defaults are handled when OMERO channels are unavailable:

* Renamed contrastLimits to fallbackContrastLimits to better convey its intended usage.
* Introduced defaultGreyscaleChannel() to standardize fallback channel creation with optional contrast limits.
* Updated omeroToChannelControls() to accept an optional default channel for cases where OMERO channel metadata is absent.
* Removed a redundant useEffect that applied contrastLimits directly to imageLayer channels; updates are now handled via the useIdetik context store.
* Downgraded console.error to console.warn for preload slice failures, reflecting that these are non-critical (ex: in organellebox).


bchu1 (approved):
Nice

PR #258, opened 2025-05-22
refactor: support middle-Z initialization and resolution selection

phoenixAja:
### Summary
- Adds support for initializing the viewer at the middle Z slice via a new `shouldLoadMiddleZ` flag in `useOmeZarrViewer` and `OmeZarrImageViewer`.
- Adds `resolutionLevel` support for controlling which resolution to load in `OmeZarrImageSource`.
- Refactors `loadOmeroDefaultZ` to include a fallback and console warning if the metadata field is missing.
- Cleans up excessive `console.debug` and `console.log` statements in `ImageSeriesLayer` and related hooks to reduce log noise.
- Consolidates state updates for `zIndex` using `useEffect`.

### Related Issue
Closes #<issue-number>

### Tests & Checks
checked embrella/ example app

### DIFF
--- a/packages/core/src/layers/image_series_layer.ts
+++ b/packages/core/src/layers/image_series_layer.ts
@@ -232,32 +232,16 @@ export class ImageSeriesLayer extends Layer {
     const chunk = await loader.loadChunk(pointRegion, this.scheduler_);
 
     this.dataChunks_[index] = chunk;
-    // console.debug(
-    //   `Loaded data for position ${position} (array index ${index})`
-    // );
     if (!token) {
-      // console.debug(
-      //   `Not setting data for position ${position} (array index ${index}) - loaded in background`
-      // );
       return;
     }
 
-    if (token.canceled) {
-      // console.debug(
-      //   `Not setting data for position ${position} (array index ${index}) - canceled by subsequent request`
-      // );
-    } else {
-      // console.debug(
-      //   `Setting data for position ${position} (array index ${index})`
-      // );
-      this.loadingToken_ = null;
-      this.setData(chunk);
-      this.setState("ready");
-    }
+    this.loadingToken_ = null;
+    this.setData(chunk);
+    this.setState("ready");
   }
 
   public async preloadSeries() {
-    console.debug(`Preloading series for dim ${this.seriesDimensionName_}`);
     const { length } = await this.loadSeriesAttributes();
     // Load remaining slices concurrently, exclude the token so they don't get set
     const loadPromises = [];
@@ -276,11 +260,6 @@ export class ImageSeriesLayer extends Layer {
           console.error(`Error loading slice: ${result.reason}`);
         }
       }
-    }
-    if (!results.some((result) => result.status === "rejected")) {
-      console.debug(
-        `Loaded all ${this.dataChunks_.length} slices for dim ${this.seriesDimensionName_}`
-      );
     }
   }
 
--- a/packages/core/src/layers/image_series_layer.ts
+++ b/packages/core/src/layers/image_series_layer.ts
@@ -235,10 +235,11 @@ export class ImageSeriesLayer extends Layer {
     if (!token) {
       return;
     }
-
-    this.loadingToken_ = null;
-    this.setData(chunk);
-    this.setState("ready");
+    if (!token.canceled) {
+      this.loadingToken_ = null;
+      this.setData(chunk);
+      this.setState("ready");
+    }
   }
 
   public async preloadSeries() {
--- a/packages/core/src/layers/image_series_layer.ts
+++ b/packages/core/src/layers/image_series_layer.ts
@@ -232,10 +232,8 @@ export class ImageSeriesLayer extends Layer {
     const chunk = await loader.loadChunk(pointRegion, this.scheduler_);
 
     this.dataChunks_[index] = chunk;
-    if (!token) {
-      return;
-    }
-    if (!token.canceled) {
+
+    if (token && !token.canceled) {
       this.loadingToken_ = null;
       this.setData(chunk);
       this.setState("ready");
--- a/packages/core/src/layers/image_series_layer.ts
+++ b/packages/core/src/layers/image_series_layer.ts
@@ -232,24 +232,24 @@ export class ImageSeriesLayer extends Layer {
     const chunk = await loader.loadChunk(pointRegion, this.scheduler_);
 
     this.dataChunks_[index] = chunk;
-    console.debug(
-      `Loaded data for position ${position} (array index ${index})`
-    );
+    // console.debug(
+    //   `Loaded data for position ${position} (array index ${index})`
+    // );
     if (!token) {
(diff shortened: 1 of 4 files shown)


packages/core/src/layers/image_series_layer.ts
```diff
@@ -232,32 +232,16 @@ export class ImageSeriesLayer extends Layer {
     const chunk = await loader.loadChunk(pointRegion, this.scheduler_);
 
     this.dataChunks_[index] = chunk;
-    console.debug(
-      `Loaded data for position ${position} (array index ${index})`
-    );
```
phoenixAja, line 237: this was bloating the console, figured we can just add console debugs as we need them

packages/core/src/layers/image_series_layer.ts
```diff
@@ -232,32 +232,16 @@ export class ImageSeriesLayer extends Layer {
...
-      console.debug(
-        `Setting data for position ${position} (array index ${index})`
-      );
-      this.loadingToken_ = null;
-      this.setData(chunk);
-      this.setState("ready");
-    }
+    this.loadingToken_ = null;
+    this.setData(chunk);
+    this.setState("ready");
```
aganders3: Did you remove the check for `token.canceled` on purpose?

packages/react/src/components/hooks/useOmeZarrImageViewer.ts
```diff
@@ -39,18 +42,28 @@ export function useOmeZarrViewer({
...
   const [loading, setLoading] = useState(true);
   const [allSlicesLoaded, setAllSlicesLoaded] = useState(false);
   const { setImageSeriesLayer, clearImageSeriesLayer, setChannelControls } =
     useIdetik();
 
+  useEffect(() => {
+    if (imageLayer) {
+      imageLayer.setIndex(zIndex);
+    }
+  }, [imageLayer, zIndex]);
```
aganders3, line 65: Is there a chance the `zIndex` gets set using a stale value (possibly out of range) if the `imageLayer` changes? I think this may have been why I was not using `zIndex` as state before.

Without `zIndex` state I think you can use `zValue = 0.5` to get the middle slice (you called it `zNormalized` which is a better name).

packages/react/src/components/hooks/useOmeZarrImageViewer.ts
```diff
@@ -157,18 +187,27 @@ export function useOmeZarrViewer({
...
-      const zNormalized = defaultZ / (max - min);
+      let initialZ: number;
+      if (shouldLoadMiddleZ) {
+        // Calculate middle z value from shape array
+        const zShape = attrs.shape[zIdx];
+        initialZ = Math.floor(zShape / 2);
+      } else {
+        // Use default Z from metadata
+        initialZ = await loadOmeroDefaultZ(sourceUrl);
+      }
```
aganders3: I think there's a comment somewhere about this, but I think these calculations (and the omero default Z) are only valid if the region is "full" for Z.

aganders3 (approved):
I think this is okay to merge if the issue in `image_series_layer` is resolved.

packages/core/src/layers/image_series_layer.ts
phoenixAja: oh whoops i did because i thought it was only giving a console error, i didn't see that it had a return as well. Thanks for catching!

phoenixAja: oh actually yes i did remove it because if the token was cancelled all it did was console.debug? Do you want me to add that back?

packages/react/src/components/hooks/useOmeZarrImageViewer.ts
phoenixAja: do you think we should add a check for that before proceeding with the calculations?

packages/react/src/components/hooks/useOmeZarrImageViewer.ts
phoenixAja: replaced zIndex with zValue

packages/react/src/components/hooks/useOmeZarrImageViewer.ts
phoenixAja: added check

packages/core/src/layers/image_series_layer.ts
aganders3: You don't need to add back the debug, but I think it needs `if (!token.canceled)` now around setting the data/state.

packages/react/src/components/hooks/useOmeZarrImageViewer.ts
aganders3: I'm okay with a check or a TODO for this - I think we need to better support getting info about the available data from the layer.

packages/react/src/components/hooks/useOmeZarrImageViewer.ts
```diff
@@ -194,16 +194,40 @@ export function useOmeZarrViewer({
...
+          const zShape = attrs.shape[zIdx];
+          initialZ = Math.floor(zShape / 2);
+        } else {
+          initialZ = await loadOmeroDefaultZ(sourceUrl);
+        }
+      } else if (zRegion) {
+        switch (zRegion.index?.type) {
+          case "point":
+            initialZ = zRegion.index.value;
+            break;
```
aganders3, line 213: I think there's already a check (somewhere...) that the `seriesDimension` is nor a point.

packages/core/src/layers/image_series_layer.ts
phoenixAja: oh right, thanks. lol

packages/react/src/components/hooks/useOmeZarrImageViewer.ts
phoenixAja: this was mostly to satisfy the types

packages/react/src/components/viewers/OmeZarrImageViewer/OmeZarrImageViewer.tsx
```diff
@@ -23,6 +23,9 @@ interface OmeZarrViewerContainerProps {
   onLoadAllSlicesClicked?: () => void;
   onAllSlicesLoaded?: () => void;
   onLoadAllSlicesAborted?: () => void;
+  resolutionLevel?: number;
+  shouldAutoLoadAllSlices?: boolean;
+  shouldLoadMiddleZ?: boolean;
```
bchu1, line 28: nit - might make more sense to be `initialSlice?: 'start' | 'middle' | 'end';`? not important to do right now

bchu1 (approved):

packages/react/src/components/viewers/OmeZarrImageViewer/OmeZarrImageViewer.tsx
phoenixAja: i was planning on refactoring this more, i'll add that to the batch

PR #260, opened 2025-05-23
Update OrthographicCamera to use transform for zoom

aganders3:
## Summary
This is a refactor in preparation for progressive chunk streaming and more immediately for the CryoLens project, which needs to get world position from a mouse click. This PR consolidates camera behavior and fixes several coordinate transformation issues that were causing incorrect zoom and positioning in orthographic cameras.

This also allows us to get the world position from a click, with the caveat that the depth of the object of interest in clip-space must be known a priori (object picking has not been implemented). In the CryoET particles/points example, the necessary code would look like this. I'm not adding this here, as I'm hoping the interaction can be integrated into the layer. Prior to this PR this was broken, as `clipToWorld` was wrong (zoom and pan work fine because it was still _relatively_ correct).

```
const canvas = document.getElementById("canvas") as HTMLCanvasElement;
canvas.addEventListener("click", (event) => {
  const x = event.clientX;
  const y = event.clientY;
  const clipPos = renderer.clientToClip([x, y]);
  const worldPos = camera.clipToWorld(clipPos);
  console.log("World position:", worldPos);
});
```

Out of scope here was refactoring the controls away from the renderer, but I'm happy to bring that in now that we have the Idetik runtime.

### Camera System Refactoring
- **Unified camera interface**: Both `OrthographicCamera` and `PerspectiveCamera` now implement consistent `zoom(factor)` and `setAspectRatio(aspectRatio)` methods (also added as abstract to the base class)
- **Fixed orthographic zoom**: Moved from projection matrix manipulation to proper transform scaling, ensuring zoom works correctly with world-to-clip coordinate transformations
- **Improved clipToWorld**: Fixed coordinate transformation to properly handle camera transforms

### Transform System Improvements  
- **Clearer naming**: ~~Since this transform is composed as-needed from components, I feel `SrtTransform` (srt == scale, rotate, translate) gives a better indication of what this is doing. `AffineTransform` is more generic and flexible in how it is composed. (I'm fine to be overruled on this. I think SRT is very common in graphics, but I was not familiar with it from a scientific computing perspective.)~~
_Edit: and apparently it's commonly called TRS, which is also counterintuitive to me. For these reasons I will accept I may not be the best person to name this. In any case I'm going to change it to `TrsTransform` for now. I still think Affine doesn't quite fit (even though of course the resulting matrix is affine)._
- **Cleaner, more consistent API**: Added `set` and `add` methods for rotation, scale, and translation. So these values can either be set or composed with their existing values.

### Updated Examples and Components
- Updated tracks example to use new unified camera APIs
- Fixed Ultrack prototype renderer to use new zoom method

### Related Issue
Related to #89

### Tests & Checks
Tested examples for accurate pan/zoom "feel", and logging the world position in the points example with
- All existing transform tests updated to use new API methods
- Camera zoom and coordinate transformations working correctly across examples
- No breaking changes to core rendering functionality


aganders3:
This video shows using an endpoint for the CryoLens backend to get the "nearest" 10 particles. This is hacked into the points example, but not committed here. The payload for their endpoint is just taking calculating `worldPos` from the renderer/camera, and adding `zSlider.valueAsNumber` for the Z-value. I think this is just using physical distance, not distance in embedding space, but that confirms the coordinate systems are compatible because these results look correct.

https://github.com/user-attachments/assets/8f71c90b-1ae0-45e4-86f1-60d63a193fea

packages/core/src/core/transforms.ts
```diff
@@ -1,19 +1,27 @@
...
-export class AffineTransform {
+export class TrsTransform {
   private dirty_ = true;
   private matrix_ = mat4.create();
   private rotation_ = quat.create();
   private translation_ = vec3.create();
   private scale_ = vec3.fromValues(1, 1, 1);
 
-  public rotate(q: quat) {
+  public addRotation(q: quat) {
```
andy-sweet, line 10: Just to double-check (as I understood from the PR description), there is an intentional change in semantics here in addition to the name change. This function now just modifies the rotation parameter of the `TrsTransform` instead of composing a rotation with the existing transform.

aganders3: Yes, that's correct. There is no longer an affordance on the transform itself for composing rotation with the _complete_ transform. All the methods on this class operate on the T, R, and S components separately, but `set*` methods will overwrite their components, while `add*` will compose with the existing component. Perhaps even this is unnecessary, since a caller can just get the current components separately as well (and then use them with `set*`).

I think composing rotation with the complete transform would be enabled by a `TransformChain` if/when we need it.

shlomnissan: Note on the idea of `TransformChain` (if I understand the intent): I built a similar abstraction in a past project, aiming to handle chained TRS transforms more explicitly. In practice, I ran into challenges scaling it across the entire rendering engine—especially when interacting with systems that expected matrix-based transforms.

Ultimately, I found that composing transforms via direct matrix multiplication was both simpler and sufficient for most use cases. When needed, a helper method to decompose a resulting matrix back into T/R/S components can bridge the gap if you want to rehydrate a `TRSTransform` structure. This approach kept the API surface smaller and easier to integrate across the engine.

packages/core/test/transform.test.ts
```diff
@@ -2,164 +2,433 @@ import { mat4, vec3, quat } from "gl-matrix";
...
+  t.setTranslation(vec3.fromValues(1, 2, 3));
+
+  const inverse1 = t.inverse;
+  const inverse2 = t.inverse;
+
+  expect(inverse1).not.toBe(inverse2);
+  expectMatrixEquals(inverse1, inverse2);
+});
+
+test("dirty flag invalidates cache on all operations", () => {
```
andy-sweet: Testing cache hits/misses is the kind of test I would typically avoid because it depends on implementation details and thus needs to be maintained as the implementation changes. Instead I prefer testing the correctness of the result after multiple calls, which should need no maintenance.

andy-sweet (approved):
Approving because I don't see any major issues and this is an improvement over `main` that also unblocks other applications.

In future work, I think it is worth reconsidering how the controls and camera are coupled. I would also want to reconsider `Camera` as a `RenderableObject` given that it really only uses the transform from that inheritance.

packages/core/test/transform.test.ts
aganders3: This is a good point. I guess I was concerned more about continual misses, but that should probably be caught by benchmarking/profiling and not tests.

PR #261, opened 2025-05-27
Update core examples to use Idetik runtime

shlomnissan:
### Summary

This PR updates all core examples in the `examples` directory to use the new `Idetik`
runtime. The examples themselves remain unchanged; only the runtime integration
has been updated.

### Tests & Checks
- All tests passes
- Manual verification of each individual example

### DIFF
--- a/packages/core/src/idetik.ts
+++ b/packages/core/src/idetik.ts
@@ -3,22 +3,23 @@ import { Layer } from "./core/layer";
 import { LayerManager } from "./core/layer_manager";
 import { WebGLRenderer } from "./renderers/webgl_renderer";
 import { PanZoomControls } from "./objects/cameras/controls";
+import { Logger } from "./utilities/logger";
 
 type IdetikParams = {
   canvasSelector: string;
-  camera: Camera;
+  camera?: Camera;
   controls?: PanZoomControls;
   layers?: Layer[];
 };
 
 export class Idetik {
   public layerManager: LayerManager;
-  public camera: Camera;
+  public camera?: Camera;
 
   private readonly renderer_: WebGLRenderer;
 
   constructor(params: IdetikParams) {
-    this.camera = params.camera;
+    this.camera = params.camera || undefined;
     this.layerManager = new LayerManager();
     this.renderer_ = new WebGLRenderer(params.canvasSelector);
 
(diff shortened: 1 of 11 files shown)


packages/core/src/idetik.ts
```diff
@@ -3,22 +3,23 @@ import { Layer } from "./core/layer";
 import { LayerManager } from "./core/layer_manager";
 import { WebGLRenderer } from "./renderers/webgl_renderer";
 import { PanZoomControls } from "./objects/cameras/controls";
+import { Logger } from "./utilities/logger";
 
 type IdetikParams = {
   canvasSelector: string;
-  camera: Camera;
+  camera?: Camera;
```
aganders3, line 10: Is there a way to make this non-optional (and therefore non-nullable)?

packages/core/src/idetik.ts
```diff
@@ -3,22 +3,23 @@ import { Layer } from "./core/layer";
...
   controls?: PanZoomControls;
```
aganders3, line 11: Not a new change in this PR, but should this be the base `CameraControls` class? Obviously we only have one controls class now, and it's likely to get refactored, so this isn't a blocker.

aganders3 (approved):
Looks good - thanks! I tested locally and the examples are working for me as well.

packages/core/src/idetik.ts
shlomnissan: I made this optional for now because one of the examples uses the renderer's width and height to create a camera, and I wanted to avoid changing the example code at this stage. Using the canvas's width and height isn’t ideal, especially on retina displays where the internal buffer can be larger—so we really need to get these dimensions from the renderer after it’s initialized.

Deferring camera initialization seemed like the simplest workaround for now, but I’m open to better ideas. In general, I think using optionals or default values when constructing objects is a bad practice, but I didn’t have a clear alternative at the time and wanted to get this finished. Do you have any suggestions?

packages/core/src/idetik.ts
shlomnissan: Yes, we can include this as part of the camera refactor.

packages/core/src/idetik.ts
aganders3: Okay I see. That example is using it to set the aspect ratio, but this should be quickly adjusted by the renderer (soon runtime) which calls `updateActiveCamera` or something. So in that case I think it can just be removed.

Beyond that I can't think of a reason the camera should need anything from the runtime to be set in advance. I guess perhaps getting bounds from the layer(s), which might not be updated without starting the runtime. In that case it makes more sense to me to update the camera after instantiation, but I guess this just pushes down the problem of optional/default values into the camera (this is already done in some examples).

shlomnissan: Ah yeah, I didn't even look into the "why" but it makes sense. I might push a quick follow up that removes the "set controls" or maybe just remove the param from the constructor and make the camera non-optional.

PR #263, opened 2025-05-29
Use Color utility class in a few more places

andy-sweet:
### Summary
After the introduction of the `Color` utility class in #242, I remembered a few other places where that would be preferable instead of using mutable arrays.

This change uses `Color` in those places. It also cleans up some related typing of `ChannelProps` to be mostly consistent across examples, reducing the need for repeated type casting expressions like `x as [number, number, number]`.

### Related Issue
Related to #59 

### Tests & Checks
I manually tested my changes by running the examples to ensure they continue to work as expected.


andy-sweet:
@aganders3 : I assigned you as a reviewer because you introduced this class. Other comments are welcome though.

packages/core/src/objects/renderable/projected_line.ts
```diff
@@ -70,7 +70,7 @@ export class ProjectedLine extends RenderableObject {
 
   public override getUniforms() {
     return {
-      LineColor: this.color,
+      LineColor: this.color.rgb,
```
andy-sweet, line 73: I believe `getUniforms` is called on every render pass, so this introduces a very small but frequent cost because `Color.rgb` calls clone. I think that making `Color` immutable and putting the onus on the client to clone is probably the way to go, but didn't make that change yet.

packages/core/src/renderers/webgl_renderer.ts
```diff
@@ -133,7 +133,12 @@ export class WebGLRenderer extends Renderer {
   }
 
   protected clear() {
-    this.gl.clearColor(...this.backgroundColor);
```
andy-sweet, line 136: At first I thought I'd be able to keep the spread operator with `this.backgroundColor.rgb`, but that doesn't work because `Color.rgb` is a `vec3` which is a union of a tuple and a typed array. I think I'd prefer that `Color.rgb` and `Color.rgba` are specifically typed to tuples of numbers, which would allow the spread here because the length of the tuple is known.

aganders3: This makes sense. We're not doing any math with this at the moment (and none would be complicated) so the `Color` class could store `[number, number, number, number]`. What do you think?

packages/core/src/objects/renderable/projected_line.ts
aganders3: I think it's okay to make this change here, especially if changing return types for `rgb` and `rgba` properties as described in your other comment.

packages/core/src/renderers/webgl_renderer.ts
andy-sweet: Sounds good to me. I'll make the update later today.

packages/core/examples/image2d_from_omezarr5d_u8/main.ts
```diff
@@ -1,5 +1,7 @@
 import { vec3 } from "gl-matrix";
 import {
+  ChannelProps,
```
aganders3, line 3: I seem to remember you creating `ChannelProps` with the goal of not having to import it.

I guess this is still true it doesn't _have_ to be imported, but when it's not set inline (in the constructor) TypeScript can't infer the type?

aganders3 (approved):
Looks good with or without the discussed changes - feel free to add them here or merge and open a separate PR.

packages/core/examples/image2d_from_omezarr5d_u8/main.ts
andy-sweet: Yes, I think this is generally the lesser of two evils (the other bigger one being the need for `x as [number, number]` ). You can also inline the usage of this object when constructing the layer, which negates the need for the import because the typing is bound to the constructor call.

packages/react/src/lib/color.ts
```diff
@@ -1,19 +0,0 @@
-// Convert hex string to RGB [0-1] values
-export const hexToRgb = (hex: string): [number, number, number] => {
```
andy-sweet, line 2: I think these are safe to remove unless they are exported elsewhere somehow?

packages/core/src/objects/renderable/image_renderable.ts
```diff
@@ -93,9 +92,8 @@ export class ImageRenderable extends RenderableObject {
 
       // All channels (including defaults) are already in this.channels_
       this.channels_.forEach((channel) => {
-        const channelColor = Color.from(channel.color);
         visible.push(channel.visible);
-        color.push(...channelColor.rgb);
+        color.push(...channel.color.rgb);
```
andy-sweet, line 96: I think this and the similar change above were safe before, but want to check I'm not missing something.

packages/core/src/objects/renderable/projected_line.ts
andy-sweet: `Color.rgb` still makes a (shallow) copy, so no big change here. Whereas `Color.rgba` returns the `readonly` array.

andy-sweet:
Make a couple of changes in `image_renderable` and also in the react component (removed the color utils there), so please take another look when you can.

packages/core/src/core/color.ts
```diff
@@ -19,32 +13,31 @@ export class Color {
...
     }
-    this.rgb_ = vec3.fromValues(r, g, b);
-    this.alpha_ = a ?? 1.0;
+    this.rgba_ = [r, g, b, a ?? 1.0];
   }
 
-  public get rgb(): vec3 {
-    return vec3.clone(this.rgb_);
+  public get rgb(): [number, number, number] {
+    return [this.rgba_[0], this.rgba_[1], this.rgba_[2]];
```
aganders3, line 20: Would it keep behavior more similar to `rgba()` to return `this.rgba_.slice(0, 3)` instead?

packages/core/src/objects/renderable/image_renderable.ts
aganders3: Yes, both look correct to me as they have already gone through `validateChannel`.

packages/core/src/core/color.ts
```diff
@@ -1,16 +1,10 @@
 import { vec3, vec4 } from "gl-matrix";
 
-export type ColorLike =
-  | Color
-  | vec3
-  | vec4
-  | [number, number, number]
-  | [number, number, number, number];
+export type ColorLike = Color | vec3 | vec4;
```
aganders3, line 3: I assume there's not different behavior with this change because the vec types also accept the `number[]` types?

packages/core/src/core/color.ts
aganders3: Googling suggests it does not, and this will still allocate.

`return this.rgba_ as readonly [number, number, number]` apparently may do what we want, but I think that would deserve double-checking and a comment.

aganders3: > apparently may do what we want

I don't think my suggestion there works, it passes static checks but dynamic use will show length = 4 for example. So maybe just ignore these ideas for now :/

andy-sweet: Yeah, I went through these same motions and went with what I have as a simplest, least offensive option for now :)

packages/core/src/core/color.ts
andy-sweet: That's my understanding too - the `vec3` definition already includes `[number, number, number]` so the explicit union is redundant (I think!). I considered tightening this definition to only use the tuples of numbers, but don't mind accepting `Float32Array` too right now.

PR #264, opened 2025-05-29
Use idetik name instead of imaging-active-learning

andy-sweet:
### Summary
This replaces a few instances of the old repo / top-level package name (imaging-active-learning) with the new name (idetik). Given that GitHub handles the rename pretty gracefully (i.e. by redirecting), most of these aren't strictly needed but clean things up for newcomers.


aganders3 (approved):
Everything works for me after removing `node_modules` and reinstalling.

PR #265, opened 2025-06-02
[fix] Pass through prop for `shouldAutoLoadAllSlices`

aganders3:
### Summary
This passes through the `shouldAutoLoadAllSlices` prop, which was added in #258 but ends up hard-coded to `true`.

@phoenixAja @bchu1 please take a look when updating, in case this affects the CZII HITL app.

### Related Issue
[See bug discussion in Slack](https://chanzuckerbergteam.slack.com/archives/C08JLLH6QKT/p1748551049926779)

### Tests & Checks
Tested manually in react example app


bchu1 (approved):
Thanks!

PR #266, opened 2025-06-03
Inject Idetik context to individual layers

shlomnissan:
### Summary

This PR introduces an `IdetikContext` type, which currently holds a reference
to an empty `ChunkManager` class. The context is created during runtime
initialization and passed to the `LayerManager`, which then injects it into each
layer added to the scene.

The intent is for layers to use their `onAttached` hook to register new sources with the `ChunkManager` and potentially access visible chunks from it. The exact mechanism for interacting with the chunk manager is still open for discussion. One option is for each layer to retain a reference to the chunk manager or the context itself and request chunks via a method like `chunkManager.getVisibleChunks(source)`. Alternatively, the chunk manager could return a dedicated store instance for each source, which layers can then use directly.

### Tests & Checks

- All tests and checks are passing.

### DIFF
--- a/packages/core/src/core/chunk_manager.ts
+++ b/packages/core/src/core/chunk_manager.ts
@@ -1,0 +1,11 @@
+import { Camera } from "../objects/cameras/camera";
+
+export class ChunkManager {
+  public getVisibleChunks() {
+    // TODO: implement
+  }
+
+  public update(_camera: Camera, _bufferWidth: number, _bufferHeight: number) {
+    // TODO: implement
+  }
+}
--- a/packages/core/src/core/layer.ts
+++ b/packages/core/src/core/layer.ts
@@ -1,3 +1,4 @@
+import { IdetikContext } from "../idetik";
 import { RenderableObject } from "./renderable_object";
 import { clamp } from "../utilities/clamp";
 
@@ -25,9 +26,9 @@ export abstract class Layer {
   public blendMode: blendMode;
 
   constructor({
-    transparent: transparent = false,
+    transparent = false,
     opacity = 1.0,
-    blendMode: blendMode = "normal",
+    blendMode = "normal",
   }: LayerOptions = {}) {
     if (opacity < 0 || opacity > 1) {
       console.warn(
@@ -51,6 +52,12 @@ export abstract class Layer {
   }
 
   public abstract update(): void;
+
+  // TODO: Consider making this an abstract method once chunk manager
+  // integration is finalized. Most layers will likely need access to the chunk
+  // manager, but for now, we allow optional overrides to avoid requiring
+  // placeholder implementations.
+  public onAttached(_context: IdetikContext) {}
 
   public get objects() {
     return this.objects_;
--- a/packages/core/src/core/layer_manager.ts
+++ b/packages/core/src/core/layer_manager.ts
@@ -39,7 +39,11 @@ export class LayerManager {
     this.layers_.splice(index, 1);
   }
 
-  public get layers() {
+  public removeAll() {
+    this.layers_.length = 0;
+  }
+
+  public get layers(): readonly Layer[] {
     return this.layers_;
   }
 }
--- a/packages/core/src/core/layer_manager.ts
+++ b/packages/core/src/core/layer_manager.ts
@@ -1,7 +1,15 @@
 import { Layer } from "./layer";
+import { IdetikContext } from "../idetik";
 
 export class LayerManager {
-  private layers_: Layer[] = [];
+  private readonly layers_: Layer[] = [];
+
+  // TODO: Make this non-optional when react components use the Idetik Runtime
+  private readonly context_?: IdetikContext;
+
+  constructor(context?: IdetikContext) {
+    this.context_ = context;
+  }
 
   public partitionLayers(): {
     opaque: Layer[];
@@ -23,6 +31,9 @@ export class LayerManager {
 
   public add(layer: Layer) {
     this.layers_.push(layer);
+    if (this.context_) {
+      layer.onAttached(this.context_);
+    }
   }
   public remove(index: number) {
     this.layers_.splice(index, 1);
--- a/packages/core/src/idetik.ts
+++ b/packages/core/src/idetik.ts
@@ -4,6 +4,7 @@ import { LayerManager } from "./core/layer_manager";
 import { WebGLRenderer } from "./renderers/webgl_renderer";
 import { PanZoomControls } from "./objects/cameras/controls";
 import { Logger } from "./utilities/logger";
+import { ChunkManager } from "./core/chunk_manager";
 
 type IdetikParams = {
   canvasSelector: string;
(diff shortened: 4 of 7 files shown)


packages/core/src/core/chunk_manager.ts
```diff
@@ -0,0 +1,11 @@
+import { Camera } from "../objects/cameras/camera";
+
+export class ChunkManager {
+  public getVisibleChunks() {
+    // TODO: implement
+  }
+
+  public update(_camera: Camera, _width: number, _height: number) {
```
aganders3: `_width` and `_height` here are the size of the canvas in pixels, right? Can we perhaps come up with more descriptive names or use an object to give structure to these params?

packages/core/src/core/layer.ts
```diff
@@ -1,3 +1,4 @@
+import { IdetikContext } from "idetik";
```
aganders3:
```suggestion
import { IdetikContext } from "../idetik";
```

packages/core/src/core/layer_manager.ts
```diff
@@ -1,7 +1,15 @@
 import { Layer } from "./layer";
+import { IdetikContext } from "idetik";
```
aganders3:
```suggestion
import { IdetikContext } from "../idetik";
```

packages/core/src/core/chunk_manager.ts
```diff
@@ -0,0 +1,11 @@
+import { Camera } from "../objects/cameras/camera";
+
+export class ChunkManager {
+  public getVisibleChunks() {
+    // TODO: implement
+  }
```
aganders3, line 6: Should the layer (or a chunk source) be an parameter here such that a layer only gets visible chunks relevant to itself?

packages/core/src/idetik.ts
```diff
@@ -12,16 +13,28 @@ type IdetikParams = {
...
+export type IdetikContext = {
+  chunkManager: ChunkManager;
+};
+
 export class Idetik {
   public layerManager: LayerManager;
   public camera: Camera;
 
+  private readonly context_: IdetikContext;
   private readonly renderer_: WebGLRenderer;
```
aganders3, line 25: We're a long way off from another rendering backend, but is it worth keeping this typed as a generic `Renderer`? I suppose that would require changing `IdetikParams` to take a renderer instead of the canvas selector.

(sorry, unrelated to this PR)

aganders3 (approved):
Looks good!

packages/core/src/core/chunk_manager.ts
shlomnissan: It's the canvas width and height scaled by the device pixel ratio—this gives the target size for the WebGL drawing buffer. We could rename it to `bufferWidth` and `bufferHeight` to make that purpose more explicit.

packages/core/src/core/layer_manager.ts
shlomnissan: I'm fine with making these changes for now, but in general, we should avoid using relative paths. Managing multiple packages adds some complexity, but it should still be fairly straightforward.

packages/core/src/core/chunk_manager.ts
shlomnissan: As noted in the PR overview, I'm considering two options:

1. Pass the chunk store directly to the method.
2. Define a new type (e.g., ChunkSource) that exposes an interface for accessing a specific data source.

With the second approach, the layer code might look something like this:

```ts
private chunkSource_: ChunkSource;

public onAttached(context: IdetikContext) {
    this.chunkSource_ = context.chunkManager.addSource(this.source_);
 }

public update() {
    if (this.state == "ready") {
        const visibleChunks = this.chunkSource_.getVisibleChunks();
        // generate image renderable objects
    }
}

File: packages/core/src/idetik.ts (hunk 5 above)
Comment (shlomnissan) [reply packages/core/src/idetik.ts:25]: Yep, we can extend the Idetik parameters object to include a renderer type. It’s probably not critical at this stage, but feel free to add a patch for it if you think it’s worthwhile.

File: packages/core/src/core/chunk_manager.ts (hunk 1 above)
Comment (shlomnissan) [reply packages/core/src/core/chunk_manager.ts]: Even `bufferWidth` and `bufferHeight` aren't as descriptive as I'd like. This is a good example of where adding documentation comments can really help clarify their purpose.

Comment (andy-sweet) [reply packages/core/src/core/chunk_manager.ts]: I prefer `bufferWidth` over `width` and think it's good enough for now. I think documenting `ChunkManager` to help clarify things is worth it at some point, because this will be an important call for Idetik.

File: packages/core/src/core/layer.ts [hunk 6]
```diff
@@ -52,6 +53,12 @@ export abstract class Layer {
 
   public abstract update(): void;
 
+  // TODO: Consider making this an abstract method once chunk manager
+  // integration is finalized. Most layers will likely need access to the chunk
+  // manager, but for now, we allow optional overrides to avoid requiring
```
andy-sweet, line 58: I could imagine some explicitly in-memory layers, especially utilities (e.g. `AxesLayer`), that do not need the chunk manager. There might be other things in the context they need though, so I don't mind making this abstract now or later.

packages/core/src/core/layer_manager.ts
```diff
@@ -23,6 +31,9 @@ export class LayerManager {
 
   public add(layer: Layer) {
     this.layers_.push(layer);
+    if (this.context_) {
```
andy-sweet, line 34: Now that `add` does more than `this.layers_.push`, we have to be a little careful to ensure we call `add` instead of `LayerManager.layers.push` - I wonder if it's worth defensively protecting that be returning a readonly array of layers for the property instead?

packages/core/src/layers/image_layer.ts
```diff
@@ -35,6 +36,10 @@ export class ImageLayer extends Layer {
     this.channelProps_ = channelProps;
   }
 
+  public onAttached(_context: IdetikContext): void {
```
andy-sweet, line 39: Will we need a similar `onDetached`? For the chunk manager, I could imagine maybe removing a source (if no layers need that source), which might free up some resources (e.g. cache space).

andy-sweet (approved):
Ideally I would like to see this do a little more before merging, but I may be missing some context on the overall implementation plan. If this unblocks that plan, feel free to merge. All of the comments are non-blocking.

packages/core/src/core/layer_manager.ts
aganders3: I would prefer absolute paths everywhere, but @phoenixAja had trouble integrating into other projects and the best current option we could figure out was to be consistent with using relative paths everywhere.

packages/core/src/core/chunk_manager.ts
aganders3: Thanks - this makes sense. Sorry I just misread the PR overview!

packages/core/src/core/chunk_manager.ts
aganders3: I'd suggest something like this, but I'm not fussed about it. It's not like this is going to be used in a ton of places so I don't see a lot of potential for confusion.
```
public update(_camera: Camera, scaledCanvasDims: { width: number, height: number })

File: packages/core/src/core/layer.ts (hunk 6 above)
Comment (shlomnissan) [reply packages/core/src/core/layer.ts:58]: Let's make it abstract later.

File: packages/core/src/core/layer_manager.ts (hunk 7 above)
Comment (shlomnissan) [reply packages/core/src/core/layer_manager.ts:34]: Good callout! I made it read-only.

```ts
public get layers(): readonly Layer[] {
    return this.layers_;
}
```

packages/core/src/layers/image_layer.ts
shlomnissan: Yes, we might. We have a follow-up task planned for Q3 currently titled “data transfer optimizations,” which will cover areas like caching, eviction strategies, and efficient CPU/GPU data transfers. That would be a good time to evaluate whether we need an onDetached, particularly if we introduce logic for explicitly removing sources that are no longer needed by any layers.

shlomnissan:
> Ideally I would like to see this do a little more before merging

@andy-sweet I understand. I see this as a good stopping point for now because introducing the empty `ChunkManager` class—while minimal—creates the necessary structure and access pattern for layers. With this in place, we can start dividing and parallelizing the remaining work within the chunk manager more effectively.

PR #267, opened 2025-06-03
feat: Refactor React components to use new Idetik object

bchu1:
- Fixes some styling on the demo page.
 - Completes refactoring of all redundant state into the global context.
 - Makes `<Renderer>` directly depend on the global context.
 - I'm not very happy with the typecasts required to deal with the `Layer[]`, but how we want to address that is a separate issue.

### DIFF
--- a/packages/react/src/components/hooks/useIdetik.ts
+++ b/packages/react/src/components/hooks/useIdetik.ts
@@ -1,4 +1,4 @@
-import { ChannelProps, ImageSeriesLayer } from "@idetik/core";
+import { ChannelProps, Idetik } from "@idetik/core";
 import { createContext, useContext } from "react";
 
 export interface ChannelControl {
@@ -6,33 +6,24 @@ export interface ChannelControl {
   contrastRange: [number, number];
 }
 
-export type IdetikContextValue =
+export type IdetikContextValue = (
   | {
       isInitialized: true;
-      imageSeriesLayer: ImageSeriesLayer; // Only defined when isInitialized.
-      channels: ChannelProps[];
-      channelControls: ChannelControl[]; // Same order as channels.
-      setImageSeriesLayer: React.Dispatch<
-        React.SetStateAction<ImageSeriesLayer | undefined>
-      >;
-      clearImageSeriesLayer: () => void;
-      setChannelControls: React.Dispatch<
-        React.SetStateAction<Array<ChannelControl>>
-      >;
+      idetik: Idetik;
     }
   | {
       isInitialized: false;
-      imageSeriesLayer: undefined;
-      channels: ChannelProps[];
-      channelControls: ChannelControl[];
-      setImageSeriesLayer: React.Dispatch<
-        React.SetStateAction<ImageSeriesLayer | undefined>
-      >;
-      clearImageSeriesLayer: () => void;
-      setChannelControls: React.Dispatch<
-        React.SetStateAction<Array<ChannelControl>>
-      >;
-    };
+      idetik: undefined;
+    }
+) & {
+  channels: ChannelProps[];
+  channelControls: ChannelControl[]; // Same order as channels.
+  setIdetik: React.Dispatch<React.SetStateAction<Idetik | undefined>>;
+  clear: () => void;
+  setChannelControls: React.Dispatch<
+    React.SetStateAction<Array<ChannelControl>>
+  >;
+};
 
 export const IdetikContext = createContext<IdetikContextValue | undefined>(
   undefined
--- a/packages/react/src/components/hooks/useIdetik.ts
+++ b/packages/react/src/components/hooks/useIdetik.ts
@@ -19,7 +19,6 @@ export type IdetikContextValue = (
   channels: ChannelProps[];
   channelControls: ChannelControl[]; // Same order as channels.
   setIdetik: React.Dispatch<React.SetStateAction<Idetik | undefined>>;
-  clear: () => void;
   setChannelControls: React.Dispatch<
     React.SetStateAction<Array<ChannelControl>>
   >;
--- a/packages/react/src/components/hooks/useIdetik.ts
+++ b/packages/react/src/components/hooks/useIdetik.ts
@@ -16,6 +16,7 @@ export type IdetikContextValue = (
       idetik: undefined;
     }
 ) & {
+  // TODO: Refactor into 1 array that lives in each layer.
   channels: ChannelProps[];
   channelControls: ChannelControl[]; // Same order as channels.
   setIdetik: React.Dispatch<React.SetStateAction<Idetik | undefined>>;
--- a/packages/react/src/components/hooks/useIdetik.ts
+++ b/packages/react/src/components/hooks/useIdetik.ts
@@ -19,7 +19,7 @@ export type IdetikContextValue = (
   // TODO: Refactor into 1 array that lives in each layer.
   channels: ChannelProps[];
   channelControls: ChannelControl[]; // Same order as channels.
-  setIdetik: React.Dispatch<React.SetStateAction<Idetik | undefined>>;
+  setIdetik: (idetik: Idetik) => void;
   setChannelControls: React.Dispatch<
     React.SetStateAction<Array<ChannelControl>>
   >;
--- a/packages/react/src/components/hooks/useIdetik.ts
+++ b/packages/react/src/components/hooks/useIdetik.ts
@@ -19,6 +19,7 @@ export type IdetikContextValue = (
   // TODO: Refactor into 1 array that lives in each layer.
   channels: ChannelProps[];
   channelControls: ChannelControl[]; // Same order as channels.
+  // TODO: Remove need to integrator to set Idetik.
   setIdetik: (idetik: Idetik) => void;
   setChannelControls: React.Dispatch<
     React.SetStateAction<Array<ChannelControl>>
--- a/packages/react/src/components/hooks/useOmeZarrImageViewer.ts
+++ b/packages/react/src/components/hooks/useOmeZarrImageViewer.ts
@@ -51,7 +51,7 @@ export function useOmeZarrViewer({
   const [zValue, setZValue] = useState(0.5);
   const [loading, setLoading] = useState(true);
   const [allSlicesLoaded, setAllSlicesLoaded] = useState(false);
-  const { idetik, setIdetik, clear, setChannelControls } = useIdetik();
+  const { idetik, setIdetik, setChannelControls } = useIdetik();
 
   useEffect(() => {
     const newSource = new OmeZarrImageSource(sourceUrl, resolutionLevel);
@@ -120,7 +120,6 @@ export function useOmeZarrViewer({
                 })
               );
             } else {
-              idetik.layerManager.remove(0);
               idetik.layerManager.add(layer);
             }
             setLoading(false);
@@ -162,7 +161,7 @@ export function useOmeZarrViewer({
 
     return () => {
       shouldSetLayer = false;
-      clear();
+      idetik?.layerManager.remove(0);
     };
     // eslint-disable-next-line react-hooks/exhaustive-deps -- Deps that trigger layer creation.
   }, [source, sourceUrl, region, seriesDimensionName, shouldAutoLoadAllSlices]);
--- a/packages/react/src/components/hooks/useOmeZarrImageViewer.ts
+++ b/packages/react/src/components/hooks/useOmeZarrImageViewer.ts
@@ -53,13 +53,6 @@ export function useOmeZarrViewer({
   const [allSlicesLoaded, setAllSlicesLoaded] = useState(false);
   const { idetik, setIdetik, clear, setChannelControls } = useIdetik();
 
-  const zIndex = Math.round(zValue * (zRange[1] - zRange[0]) + zRange[0]);
-  useEffect(() => {
-    if (idetik !== undefined) {
-      (idetik.layerManager.layers[0] as ImageSeriesLayer)?.setIndex(zIndex);
-    }
-  }, [idetik, zIndex]);
-
   useEffect(() => {
     const newSource = new OmeZarrImageSource(sourceUrl, resolutionLevel);
     setSource(newSource);
@@ -108,6 +101,9 @@ export function useOmeZarrViewer({
           seriesDimensionName,
           channelProps,
         });
+        layer.setIndex(
+          Math.round(zValue * (zRange[1] - zRange[0]) + zRange[0])
+        );
 
         onLayerCreated?.();
 
--- a/packages/react/src/components/hooks/useOmeZarrImageViewer.ts
+++ b/packages/react/src/components/hooks/useOmeZarrImageViewer.ts
@@ -1,12 +1,12 @@
 import { useState, useEffect, useCallback } from "react";
 import {
-  LayerManager,
   OmeZarrImageSource,
   OrthographicCamera,
   ImageSeriesLayer,
   Region,
   loadOmeroChannels,
   loadOmeroDefaultZ,
+  Idetik,
 } from "@idetik/core";
 
 import {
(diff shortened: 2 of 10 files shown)


aganders3:
Thanks! This looks like mostly an improvement already.

edit: though it feels like the pattern of `setIdetik(new Idetik(...))` here is pretty clunky and perhaps costly as it is re-initializing a `WebGLRenderer`.

I was thinking about this refactor this morning, partly trying to understand some of these react concepts (contexts and useSyncExternalStore).

Does it make sense to implement the `subscribe`/`getSnapshot` for the `Idetik` runtime object? Or maybe implement these individually on `Camera` and `LayerManager` instead? I don't want the main library getting a bunch of react-specific code, but this seems like it would provide a reasonable starting point for designing an events system we likely need anyway. We might be able to add such methods in wrapper classes within the react library.

Here's what I was thinking:
* Add interface for `useSyncExternalStore` to `Idetik`
* Add `removeLayer` to `LayerManager`
* `useIdetik` would output something like `{ layerManager, addLayer, removeLayer, camera }` or `{ idetikRuntime, addLayer, removeLayer }`

Now react apps would work directly with `@idetik/core` to create individual layers based on user interactions. Components may still need to keep local state of the layers they create in order to modify or remove them; that gets rid of the casting, but is it too much duplicate state? Would we then need `useXXLayer` hooks to set up each layer's syncable state if they are configurable (e.g. channel properties)?

bchu1:
> Thanks! This looks like mostly an improvement already.
> 
> edit: though it feels like the pattern of `setIdetik(new Idetik(...))` here is pretty clunky and perhaps costly as it is re-initializing a `WebGLRenderer`.

I'm not very happy with the whole sequence of how things are initialized either. `setIdetik()` should only be called on `onFirstLoad()` though?
> 
> I was thinking about this refactor this morning, partly trying to understand some of these react concepts (contexts and useSyncExternalStore).
> 
> Does it make sense to implement the `subscribe`/`getSnapshot` for the `Idetik` runtime object? Or maybe implement these individually on `Camera` and `LayerManager` instead? I don't want the main library getting a bunch of react-specific code, but this seems like it would provide a reasonable starting point for designing an events system we likely need anyway. We might be able to add such methods in wrapper classes within the react library.

If `Idetik` has those methods, it would have to either store the data (seems like a bad idea) or delegate to one of the composed classes anyways, at that point it would just be a thin proxy. I think `LayerManager` could make sense as the location for the channel callbacks, given its name? 😆 But I'm not seeing what that would give us.

I don't think having callbacks is necessarily React specific, any clients that need to listen to the viewer's state will need it, though there are other ways to do it. We also already have classes with callbacks?
> 
> Here's what I was thinking:
> 
> * Add interface for `useSyncExternalStore` to `Idetik`
> * Add `removeLayer` to `LayerManager`
> * `useIdetik` would output something like `{ layerManager, addLayer, removeLayer, camera }` or `{ idetikRuntime, addLayer, removeLayer }`
> 
> Now react apps would work directly with `@idetik/core` to create individual layers based on user interactions. Components may still need to keep local state of the layers they create in order to modify or remove them; that gets rid of the casting, but is it too much duplicate state? Would we then need `useXXLayer` hooks to set up each layer's syncable state if they are configurable (e.g. channel properties)?

I think once you introduce any duplicate state at all you're in a bad place. I do think we need more abstractions like `addLayer()` and `removeLayer()` though, but one of the new things with the `Idetik` object is it wants the `layers` and `camera` to be initialized upfront, which kind of requires the integrator to reach past its abstraction layer just to create it (though it's still a lot better than before!). Right now the React component has to do data fetching just to initialize them properly. Maybe `Idetik` should accept initialization functions, that way `Idetik` can be immediately created once and then clients could continue to interact with `Idetik` solely via the setters it exposes? cc @shlomnissan

bchu1:
> > Thanks! This looks like mostly an improvement already.
> > edit: though it feels like the pattern of `setIdetik(new Idetik(...))` here is pretty clunky and perhaps costly as it is re-initializing a `WebGLRenderer`.
> 
> I'm not very happy with the whole sequence of how things are initialized either. `setIdetik()` should only be called on `onFirstLoad()` though?
> 
> > I was thinking about this refactor this morning, partly trying to understand some of these react concepts (contexts and useSyncExternalStore).
> > Does it make sense to implement the `subscribe`/`getSnapshot` for the `Idetik` runtime object? Or maybe implement these individually on `Camera` and `LayerManager` instead? I don't want the main library getting a bunch of react-specific code, but this seems like it would provide a reasonable starting point for designing an events system we likely need anyway. We might be able to add such methods in wrapper classes within the react library.
> 
> If `Idetik` has those methods, it would have to either store the data (seems like a bad idea) or delegate to one of the composed classes anyways, at that point it would just be a thin proxy. I think `LayerManager` could make sense as the location for the channel callbacks, given its name? 😆 But I'm not seeing what that would give us.
> 
> I don't think having callbacks is necessarily React specific, any clients that need to listen to the viewer's state will need it, though there are other ways to do it. We also already have classes with callbacks?
> 
> > Here's what I was thinking:
> > 
> > * Add interface for `useSyncExternalStore` to `Idetik`
> > * Add `removeLayer` to `LayerManager`
> > * `useIdetik` would output something like `{ layerManager, addLayer, removeLayer, camera }` or `{ idetikRuntime, addLayer, removeLayer }`
> > 
> > Now react apps would work directly with `@idetik/core` to create individual layers based on user interactions. Components may still need to keep local state of the layers they create in order to modify or remove them; that gets rid of the casting, but is it too much duplicate state? Would we then need `useXXLayer` hooks to set up each layer's syncable state if they are configurable (e.g. channel properties)?
> 
> I think once you introduce any duplicate state at all you're in a bad place. I do think we need more abstractions like `addLayer()` and `removeLayer()` though, but one of the new things with the `Idetik` object is it wants the `layers` and `camera` to be initialized upfront, which kind of requires the integrator to reach past its abstraction layer just to create it (though it's still a lot better than before!). Right now the React component has to do data fetching just to initialize them properly. Maybe `Idetik` should accept initialization functions, that way `Idetik` can be immediately created once and then clients could continue to interact with `Idetik` solely via the setters it exposes? cc @shlomnissan

In either case, at least initializing `Idetik` with no layers and then adding/removing the single layer is a good idea, let me try.

packages/react/src/components/hooks/useIdetik.ts
```diff
@@ -1,38 +1,29 @@
...
-      clearImageSeriesLayer: () => void;
-      setChannelControls: React.Dispatch<
-        React.SetStateAction<Array<ChannelControl>>
-      >;
-    };
+      idetik: undefined;
+    }
+) & {
+  channels: ChannelProps[];
+  channelControls: ChannelControl[]; // Same order as channels.
```
shlomnissan, line 21: I'm a bit concerned about relying on the comment `// Same order as channels` to ensure that `channelControls` aligns with channels. This creates a soft contract that's easy to break and isn't enforced by the type system. Is there a way to enforce an explicit relationship between these two params that guarantees the order?

packages/react/src/components/hooks/useIdetik.ts
```diff
@@ -1,38 +1,29 @@
...
+  setIdetik: React.Dispatch<React.SetStateAction<Idetik | undefined>>;
```
shlomnissan: Why is the runtime allowed to be undefined? It's been a while since I last worked with React, so there might be a straightforward reason I'm missing. The only use case I can think of is resetting or clearing the viewer state — but should that really be the responsibility of the library?

packages/react/src/components/hooks/useOmeZarrImageViewer.ts
```diff
@@ -47,45 +47,32 @@ export function useOmeZarrViewer({
...
-  }, [imageLayer, zIndex]);
-
-  useEffect(() => {
-    if (imageLayer) {
-      const newManager = new LayerManager();
-      newManager.add(imageLayer);
-      setLayerManager(newManager);
-    }
-  }, [imageLayer]);
+  const { idetik, setIdetik, clear, setChannelControls } = useIdetik();
```
shlomnissan: We might be able to avoid exposing `setIdetik` directly by allowing `useIdetik` to accept initialization parameters. That way, the hook can handle creating the runtime internally, ensuring it's always set up correctly. This would help encapsulate state management and make the public API cleaner. We can also use this to guarantee that `idetik` is always initialized.

packages/react/src/components/hooks/useOmeZarrImageViewer.ts
```diff
@@ -47,45 +47,32 @@ export function useOmeZarrViewer({
...
-      return true;
-    }
-    return false;
-  }, []);
+  const zoomToFit = useCallback(
+    (layer: ImageSeriesLayer): OrthographicCamera | void => {
+      if (layer.extent) {
+        const cam = new OrthographicCamera(
+          0,
+          layer.extent.x,
```
shlomnissan, line 67: Can you remind me what the layer extent is and why it's used to initialize the camera?

packages/react/src/components/hooks/useOmeZarrImageViewer.ts
```diff
@@ -114,12 +101,16 @@ export function useOmeZarrViewer({
           seriesDimensionName,
           channelProps,
         });
+        layer.setIndex(
```
shlomnissan, line 104: I think there’s an opportunity to refactor this line for clarity and robustness. We could extract the mapping logic into a separate utility function and make it a bit more defensive—clamping the input range, for example. I haven’t tested it yet, but something along these lines might work:

```ts
function mapNormalizedToIndex(value: number, [min, max]: [number, number]): number {
  const clamped = Math.max(0, Math.min(1, value));
  return Math.round(clamped * (max - min) + min);
}
```

Generally, I feel like this type of logic is out-of-scope for a thin React API wrapper. We might want to implement utilities like this in the core library. @aganders3 thoughts?

packages/react/src/components/hooks/useOmeZarrImageViewer.ts
```diff
@@ -114,12 +101,16 @@ export function useOmeZarrViewer({
...
+          Math.round(zValue * (zRange[1] - zRange[0]) + zRange[0])
+        );
 
         onLayerCreated?.();
 
         const onFirstLoad = async () => {
           if (!shouldSetLayer || !layer) return;
-          if (zoomToFit(layer)) {
+          const camera = zoomToFit(layer);
+          if (camera !== undefined) {
```
shlomnissan, line 114: In what scenarios do we actually need to allow the camera to be undefined?

packages/react/src/components/hooks/useOmeZarrImageViewer.ts
```diff
@@ -256,7 +250,9 @@ export function useOmeZarrViewer({
       try {
         // Compute zIndex from zValue and zRange
         const zIndex = Math.round(zValue * (zRange[1] - zRange[0]) + zRange[0]);
-        await imageLayer.setIndex(zIndex);
+        await (idetik.layerManager.layers[0] as ImageSeriesLayer)?.setIndex(
```
shlomnissan, line 258: Why are we reading layers back from the layer manager and casting them to `ImageSeriesLayer`? Shouldn't the React component be responsible for tracking a reference to the specific layer instance it created? Also, since we need to support multiple layer types, how does this approach generalize across them? It would be helpful to clarify how this fits into the broader layer management strategy.

This comment applies to all other layer casts in this PR.

bchu1:
@aganders3 I added a `remove()` to the `LayerManager` and everything is working! I wonder if the existence of `LayerManager` is redundant, is it fair to say Idetik is really just a "layer manager" in of itself? 😅 If we moved the `LayerManager` code into `Idetik` that would start to look closer to what you were talking about, I think? 🤔

packages/react/src/components/providers/IdetikProvider.tsx
```diff
@@ -14,54 +14,55 @@ import {
   IdetikContextValue,
 } from "../hooks/useIdetik";
 
-// The return value of the getSnapshot() argument to useSyncExternalStore() must be memoized to
+// The return value of the getSnapshot argument to useSyncExternalStore() must be memoized to
```
shlomnissan: Could you clarify what “infinite re-renders” refers to here? My understanding is that memoization is typically used for caching expensive operations, not as a mechanism to prevent re-renders — which might suggest a design issue. That said, if the `getSnapshot` function is returning a new reference each time (e.g. a new object or array), I can see how React would interpret that as a change and re-render unnecessarily. If that's the case, maybe we want to write a better comment.

shlomnissan (approved):
Looks good to me. I left a couple of non-blocking questions, but I'm not too concerned about the React layer at this stage—if it works, I'm happy for it to be merged. When we start thinking about a public React API, we’ll probably want to spend more time on its design and clarify ownership (e.g., whether it sits with the visualization team or SDS).

shlomnissan:
> I wonder if the existence of LayerManager is redundant

@bchu1 That's a good question, but I don’t think the existence of `LayerManager` is redundant. While it’s true that Idetik contains a layer manager, it’s not accurate to say that Idetik is a layer manager.

Idetik represents the runtime—it orchestrates multiple subsystems including the renderer, camera, chunk manager, and layers. The layer manager is just one component within that system, responsible for managing layers in isolation (e.g., adding/removing layers, managing draw order, etc.).

By keeping the layer manager as a separate class, we maintain modularity and separation of concerns, which allows us to:
- Swap out or reuse `LayerManager` independently if needed
- Keep the Idetik class focused on runtime orchestration
- Simplify testing and maintenance of the layer management logic

So yeh, the `LayerManager` class could theoretically be folded into Idetik, but keeping it as its own abstraction is consistent with a clean, extensible architecture.

packages/react/src/components/hooks/useIdetik.ts
bchu1: Yes, I personally think this should all be folded into one array that lives in each layer? Added a TODO.

packages/react/src/components/hooks/useIdetik.ts
bchu1: Great callout, fixed.

packages/react/src/components/hooks/useOmeZarrImageViewer.ts
bchu1: Don't know, this was existing code.

packages/react/src/components/hooks/useOmeZarrImageViewer.ts
bchu1: Added a TODO. Do you think it makes sense to initialize `Idetik` with an `undefined` `camera` though? Guess this is related to your below comment about the camera depending on layer data to be initialized..

packages/react/src/components/providers/IdetikProvider.tsx
aganders3: I agree and my understanding of the intended interface is that `getSnapshot` will should the same object if it's unchanged. The ways I've seen this worked around are to give the class implementing the interface a `version_` or storing `cachedSnapshot_` that the object invalidates when state changes.

packages/react/src/components/hooks/useOmeZarrImageViewer.ts
aganders3: Working in the React components has revealed issues from implicit state on the `ImageSeriesLayer` in particular that's hard to get at. Specifically it derives some information from its inputs (source and region) that is not explicitly stored anywhere: the number of images in the series, the scale in that dimension, and the currently displayed (and/or requested) index. This makes it hard to sync with a slider/scrubber component. It's even worse in the context of multiscale images where the number of slices may depend on the scale being shown.

I guess what I'm saying is that yes, I agree this could be better handled in the library. As we do more integration I think having a more robust callback or event system will become increasingly important.

packages/react/src/components/hooks/useOmeZarrImageViewer.ts
aganders3: The layer extent is the bounding box of the layer in virtual coordinates. This allows the camera to fit its frame tightly to the data being displayed. The other option is to initialize the camera with some default frame and alter it once the data are loaded and the extent is available, but this can sometimes cause jarring visual effects.

packages/react/src/components/hooks/useOmeZarrImageViewer.ts
bchu1: This is more of the existing camera issue. `zoomToFit()` returns `undefined` if the layer data is not loaded, but the way we have it right now it should always be loaded, so we could add a non-null assertion to `zoomToFit()` but I'd rather we put some more effort into refactoring this, since it doesn't seem like it makes that much sense for `zoomToFit()` to be doing camera initialization as well. Added a TODO.

packages/react/src/components/hooks/useOmeZarrImageViewer.ts
bchu1: This will get resolved when we genericize the component to accept multiple layers. We'll iterate through all layers and perform the relevant index-updating operation based on their types, something to that effect.

packages/react/src/components/providers/IdetikProvider.tsx
bchu1: Updated the comment.

PR #268, opened 2025-06-04
feat: Finish migrating rest of <Renderer> to Idetik object

bchu1:
- Adds `stop()` functionality to the object (I think this is only relevant if the integrator uses the library in such a way that the `<Renderer>` unmounts).
 - Removes some changes that were accidentally merged in the last PR 😅
 - Removes redundant argument type in `useOmeZarrImageViewer()`.
 - Reset camera position when source changes (this was regressed by the last PR).
 - Lift up `cameraControls` prop out of `<Renderer>`.
 - Use `Idetik.start()` in `<Renderer>`.


packages/core/src/idetik.ts
```diff
@@ -82,4 +87,8 @@ export class Idetik {
     render();
     return this;
   }
+
+  public stop() {
+    this.stopped_ = true;
```
andy-sweet: As `Idetik.start` starts the animation loop with `requestAnimationFrame`, I would expect `Idetik.stop` to stop it with `cancelAnimationFrame`.  I understand that `Idetik.stopped_` will cause any previous loops (e.g. from a react dev-mode mount) to not actually update/render Idetik, but looping with the early return seems untidy and wasteful.

I still need to catch up on the Idetik runtime and the related hooks/providers, so maybe this issue is handled elsewhere?

bchu1: Done. Good call

PR #269, opened 2025-06-05
Add ligfortran5 for macos error

JoOkuma:
### Summary

I recently changed my laptop and had trouble running the demo because of a missing fortran library.
Some colleagues faced a similar issue, I fixed this by adding an extra conda dependency.

Additional reference:
https://stackoverflow.com/questions/79565969/importerror-when-importing-numpy-missing-libgfortran-5-dylib-on-macos-vs-code


shlomnissan (approved):
:+1:

PR #270, opened 2025-06-05
Add chunk streaming example sandbox

shlomnissan:
### Summary

This PR introduces a new example demonstrating image loading via chunk
streaming. It also includes a sandboxed copy of `ImageLayer`, which we’ll
use to prototype chunk streaming support—such as integrating with the chunk
manager and generating per-tile image renderables. This isolated setup allows
us to iterate on chunk streaming features safely without impacting existing
examples or functionality.

### Tests & Checks
- All tests are passing


phoenixAja (approved):
ty!

PR #271, opened 2025-06-09
Remove "scale index" from OmeZarrImageLoader

shlomnissan:
### Summary

This PR updates the image loader implementation by removing the internal
`scaleIndex` member, enabling support for loading chunks across multiple LODs
independently. This is change is needed for the upcoming chunk streaming
functionality, where the chunk manager may request chunks from multiple LODs
within the same store. Additionally, the term "scale" has been renamed to "lod"
where applicable, to reduce ambiguity with image scale factors.

**The driving change here is passing a required LOD index to** `ImageChunkLoader::loadChunk`.

**Note**: I added an lod parameter to individual layers as a temporary solution for selecting the level of detail. This will be removed once chunk streaming is fully implemented. At that point, we may introduce an lod clamping parameter at the top level, likely within OmeZarrImageSource, to centralize LOD selection logic.

### Tests & Checks
- Tests are passing.
- Verified by ensuring that all are working.

### DIFF
--- a/packages/core/examples/image2d_from_omezarr4d_hcs/main.ts
+++ b/packages/core/examples/image2d_from_omezarr4d_hcs/main.ts
@@ -56,7 +56,7 @@ const onImageChange = async () => {
   app.layerManager.removeAll();
   const imageUrl =
     plateUrl + "/" + wellSelector.value + "/" + imageSelector.value;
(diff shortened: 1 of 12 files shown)


phoenixAja (approved):
looks solid! This is a nice refactor

PR #272, opened 2025-06-09
feat: implement LOD selection and scale-aware image loading

phoenixAja:
### Summary

This PR adds support for resolution-dependent Level of Detail (LOD) selection and dynamic image reloading in the image layer system. The viewer now computes the optimal scale index for the current zoom level and screen resolution, and reloads the image from the appropriate scale level.

### Key Changes

- Added `computeLOD()` to `ChunkManager`, which calculates the best scale index based on pixels per world unit vs texels per world unit, using a soft scoring function that penalizes undersampling.
- Implemented `loadAvailableScales()` in `ImageLayerXYZ` to extract scale metadata from NGFF multiscales or fallback loaders.
- Connected LOD computation to runtime context via `getCamera()` and `getViewport()`.
- Introduced `updateLOD()` and `reloadWithScale()` to selectively reload chunks at a new LOD only when needed.
- Ensured initial and runtime LOD transitions are efficient and avoid unnecessary reloading.
- Added error handling for loader fallbacks and scale index validation.

### Tests & Checks
checked that LOD was properly returned in chunk_streaming example

### DIFF
--- a/packages/core/src/core/chunk_manager.ts
+++ b/packages/core/src/core/chunk_manager.ts
@@ -89,7 +89,7 @@ export class ChunkManagerSource {
   ): number {
     // Get the actual visible bounds to determine virtual width
     const viewExtent = this.calculateVisibleBounds(camera);
-    const virtualWidth = viewExtent.worldWidth;
+    const virtualWidth = viewExtent.virtualWidth;
     const virtualUnitsPerScreenPixel = virtualWidth / bufferWidth;
 
     const numLods = availableScales.length;
@@ -104,8 +104,8 @@ export class ChunkManagerSource {
 
   // Calculate the visible bounds in virtual space.
   private calculateVisibleBounds(camera: Camera): {
-    worldWidth: number;
-    worldHeight: number;
+    virtualWidth: number;
+    virtualHeight: number;
   } {
     // Screen space corners (normalized device coordinates: -1 to +1)
     const topLeftNDC = vec4.fromValues(-1, 1, 0, 1);
@@ -119,7 +119,7 @@ export class ChunkManagerSource {
     );
     const invViewProjection = mat4.invert(mat4.create(), viewProjection);
 
-    // Transform to world space
+    // Transform to virtual space
     const topLeftWorld = vec4.transformMat4(
       vec4.create(),
       topLeftNDC,
@@ -131,10 +131,10 @@ export class ChunkManagerSource {
       invViewProjection
     );
 
-    const worldWidth = Math.abs(bottomRightWorld[0] - topLeftWorld[0]);
-    const worldHeight = Math.abs(topLeftWorld[1] - bottomRightWorld[1]);
+    const virtualWidth = Math.abs(bottomRightWorld[0] - topLeftWorld[0]);
+    const virtualHeight = Math.abs(topLeftWorld[1] - bottomRightWorld[1]);
 
-    return { worldWidth, worldHeight };
+    return { virtualWidth, virtualHeight };
   }
 }
 
--- a/packages/core/src/core/chunk_manager.ts
+++ b/packages/core/src/core/chunk_manager.ts
@@ -79,61 +79,20 @@ export class ChunkManagerSource {
   public computeLOD(
     camera: Camera,
     bufferWidth: number, // screen/canvas width in pixels
-    bufferHeight: number, // screen/canvas height in pixels
     availableScales: number[][] // scale factors per LOD, where each scale is [c, z, y, x]
   ): number {
-
-    // Calculate world-space dimensions of the current visible view
+    // Get the actual visible bounds to determine virtual width
     const viewExtent = this.calculateVisibleBounds(camera);
+    const virtualWidth = viewExtent.worldWidth;
 
-    // Calculate desired screen resolution: pixels per world unit
-    // i.e., how many screen pixels span one unit of virtual space (zoom-dependent)
-    const desiredResolutionX = bufferWidth / viewExtent.worldWidth;
-    const desiredResolutionY = bufferHeight / viewExtent.worldHeight;
+    const virtualUnitsPerScreenPixel = virtualWidth / bufferWidth;
 
-    // Choose the higher resolution between X and Y (higher pixels per unit) to avoid aliasing
-    const desiredResolution = Math.max(desiredResolutionX, desiredResolutionY);
+    const numLods = availableScales.length;
+    const lodShift = numLods - 1;
+    const lodF = lodShift + Math.log2(1 / virtualUnitsPerScreenPixel);
 
-    // Select the LOD with resolution closest to what's needed - prefer higher resolution over lower resolution
-    let bestScaleIndex = 0;
-    let bestResolutionMatch = Infinity;
-
-    for (let i = 0; i < availableScales.length; i++) {
-      const scale = availableScales[i];
-
-      // Assume last two dimensions are spatial (y, x) — scale = world units per texel
-      const scaleX = scale[scale.length - 1];
-      const scaleY = scale[scale.length - 2];
-
-      // Convert resolution to texels per world unit
-      // Higher = more detail; lower = coarser
-      // Less than 1 texel per world unit = undersampling → aliasing risk
-      const resolutionX = 1.0 / scaleX;
-      const resolutionY = 1.0 / scaleY;
-      const resolution = Math.min(resolutionX, resolutionY); // Conservative estimate
-
-      const resolutionRatio = resolution / desiredResolution;
-
-      // Scoring:
-      // - If ratio >= 1.0 → oversampling → mildly penalize
-      // - If ratio < 1.0 → undersampling → penalize more strongly
-      let score: number;
-      if (resolutionRatio >= 1.0) {
-        // Resolution is higher than desired - score based on how much excess detail
-        score = resolutionRatio;
-      } else {
-        // Resolution is lower than desired - penalize but not too harshly
-        // Use 1/ratio so that the higher resolution ( that's closer to desired) gets better score
-        score = 1.0 / resolutionRatio;
-      }
-
-      if (score < bestResolutionMatch) {
-        bestResolutionMatch = score;
-        bestScaleIndex = i;
-      }
-    }
-
-    return bestScaleIndex;
+    const maxLod = numLods - 1;
+    return Math.max(0, Math.min(maxLod, Math.floor(lodF)));
   }
 
   // Calculate the visible bounds in virtual space.
--- a/packages/core/src/core/chunk_manager.ts
+++ b/packages/core/src/core/chunk_manager.ts
@@ -33,15 +33,10 @@ export class ChunkManagerSource {
       return undefined;
     }
 
-    console.log('Loading chunk with LOD:', this.currentLOD_, 'Region:', this.region_);
-    console.log('Available attributes:', this.attributes_);
-
     try {
       return await this.loader_.loadChunk(this.region_, this.currentLOD_);
     } catch (error) {
       console.warn("Failed to reload with new scale:", error);
-      console.warn("Current LOD:", this.currentLOD_);
-      console.warn("Attributes length:", this.attributes_.length);
       return undefined;
     }
   }
@@ -51,7 +46,6 @@ export class ChunkManagerSource {
     let lodResult: number;
 
     if (availableScales.length === 0) {
-      console.warn("No scales available, using default LOD 0");
       lodResult = 0; // Use first LOD level when no scales available
       if (firstPass) {
         this.currentLOD_ = lodResult;
@@ -92,14 +86,12 @@ export class ChunkManagerSource {
 
     // Check for invalid values
     if (!isFinite(virtualWidth) || virtualWidth <= 0 || !isFinite(bufferWidth) || bufferWidth <= 0) {
-      console.warn('Invalid dimensions for LOD calculation:', { virtualWidth, bufferWidth });
       return 0; // Default to highest resolution
     }
 
     const virtualUnitsPerScreenPixel = virtualWidth / bufferWidth;
 
     if (!isFinite(virtualUnitsPerScreenPixel) || virtualUnitsPerScreenPixel <= 0) {
-      console.warn('Invalid virtualUnitsPerScreenPixel:', virtualUnitsPerScreenPixel);
       return 0;
     }
 
@@ -167,12 +159,10 @@ export class ChunkManager {
   public async addSource(source: ImageChunkSource) {
     let existing = this.sources_.get(source);
     if (!existing) {
-      console.log("adding sourceq", source);
       const loader = await source.open();
       const attrs = await loader.loadAttributes();
       existing = new ChunkManagerSource(loader, attrs);
       this.sources_.set(source, existing);
-      console.log("added source", source);
     }
     return existing;
   }
--- a/packages/core/src/core/chunk_manager.ts
+++ b/packages/core/src/core/chunk_manager.ts
@@ -33,27 +33,32 @@ export class ChunkManagerSource {
       return undefined;
     }
 
+    console.log('Loading chunk with LOD:', this.currentLOD_, 'Region:', this.region_);
+    console.log('Available attributes:', this.attributes_);
+
     try {
       return await this.loader_.loadChunk(this.region_, this.currentLOD_);
     } catch (error) {
       console.warn("Failed to reload with new scale:", error);
+      console.warn("Current LOD:", this.currentLOD_);
+      console.warn("Attributes length:", this.attributes_.length);
       return undefined;
     }
   }
 
-  public async updateLOD(camera: Camera, bufferWidth: number, bufferHeight: number, firstPass: boolean = false): Promise<ImageChunk | undefined> {
+  public async updateLOD(camera: Camera, bufferWidth: number, firstPass: boolean = false): Promise<ImageChunk | undefined> {
     const availableScales = this.attributes_.map(attr => attr.scale);
     let lodResult: number;
+
     if (availableScales.length === 0) {
-      console.warn("No scales available");
-      lodResult = this.attributes_.length - 1;
+      console.warn("No scales available, using default LOD 0");
+      lodResult = 0; // Use first LOD level when no scales available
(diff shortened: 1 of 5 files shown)


packages/core/src/core/chunk_manager.ts
```diff
@@ -8,4 +17,107 @@ export class ChunkManager {
...
+      // Convert resolution to texels per world unit
+      // Higher = more detail; lower = coarser
+      // Less than 1 texel per world unit = undersampling → aliasing risk
+      const resolutionX = 1.0 / scaleX;
+      const resolutionY = 1.0 / scaleY;
+      const resolution = Math.min(resolutionX, resolutionY); // Conservative estimate
+
+      const resolutionRatio = resolution / desiredResolution;
+
+      // Scoring:
```
phoenixAja: I changed the implementation of LOD selection from the original C++ prototype. Instead of directly choosing the closest high-res scale, this version uses a scoring function to evaluate all LODs based on how well their scale matches the desired resolution.

Downsampling causes blur and potential aliasing, so we assign it a higher penalty.

Upsampling has the potential to waste memory/I/O, but is visually accurate, so we assign it a lower penalty.

We compute the score for each LOD and select the one with the lowest penalty, balancing image quality vs performance.

Happy to tweak the penalty weights or revisit the logic if people have thoughts

packages/core/src/core/chunk_manager.ts
```diff
@@ -1,5 +1,14 @@
 import { Camera } from "../objects/cameras/camera";
 
+export interface LODResult {
+  // Optimal scale index to use (0 = highest resolution)
+  scaleIndex: number;
+  // Effective resolution in pixels per world unit
+  resolution: number;
+  // Scale factor for this level (world units per pixel)
+  scaleFactor: number;
+}
```
shlomnissan: I’m not sure it’s worth defining a separate type solely for the return value—especially if we’re not instantiating it manually. We should generally prefer type inference in these cases. Also, as we discussed yesterday, I don’t think layers need awareness of the current LOD, so this might be unnecessary altogether.

packages/core/src/core/chunk_manager.ts
```diff
@@ -8,4 +17,107 @@ export class ChunkManager {
   public update(_camera: Camera, _bufferWidth: number, _bufferHeight: number) {
     // TODO: implement
   }
+
+  public computeLOD(
+    camera: Camera,
+    bufferWidth: number, // screen/canvas width in pixels
+    bufferHeight: number, // screen/canvas height in pixels
+    availableScales: number[][] // scale factors per LOD, where each scale is [c, z, y, x]
+  ): LODResult {
```
shlomnissan: - I think this should be a private method. The update method in the chunk manager is meant to internally invoke `computeLOD`, and since we’re already passing in the camera and buffer dimensions, there’s no need to extend the context object. See `idetik.ts`, line 74, for reference.
- Another point from our Slack conversation yesterday—the chunk manager will eventually need to support multiple sources. Since we should be able to retrieve scales or LODs directly from each source’s loader, there’s no need to invoke this function externally.
- Do we actually need to know the scales per LOD? I was under the impression that we could determine the appropriate LOD based on pixel density, calculated from the camera and window dimensions. My intuition is that LOD selection should be image-independent. For example, in the prototype, we could simplify the `computeLOD` method to something like this, since the decision relied purely on pixel density rather than image-specific scales.

```cpp
auto ChunkManager::ComputeLod(const OrthographicCamera& camera) const -> int {
    const auto virtual_width = camera.Width() / glm::length(glm::vec3{camera.transform[0]});
    const auto virtual_units_per_screen_pixel =  virtual_width / window_dims_.width;
    return std::log2(1 / virtual_units_per_screen_pixel);
}
```
The calculation above is entirely independent of the image itself. That said, I might be missing something—happy to step through this function together over Zoom to clarify what’s actually needed here.

packages/core/src/core/chunk_manager.ts
```diff
@@ -8,4 +17,107 @@ export class ChunkManager {
...
+    if (availableScales.length === 0) {
+      throw new Error("No scales available");
```
shlomnissan: I know scale values aren’t strictly guaranteed, but should we throw an error if they’re missing, or fall back to a default instead? If we do decide to throw, it probably makes more sense to do so at the store level. My intuition is that if the image can be rendered as a chunked source, the `ChunkManager` should be able to support it as well.

packages/core/src/core/chunk_manager.ts
```diff
@@ -8,4 +17,107 @@ export class ChunkManager {
...
+    }
+
+    // Calculate world-space dimensions of the current visible view
+    const viewExtent = this.calculateViewExtent(
+      camera,
+      bufferWidth,
+      bufferHeight
+    );
```
shlomnissan: - Instead of `viewExtent`, which implies size rather than range or position, consider `visibleBounds`. The same thinking should be applied to the function name.
- Since the function is intended to return the visible bounds in virtual space, I think the comment should be updated to reflect that. I also think we need to decide if we use the term world space or virtual space. I feel like the latter is more accurate, but I don't think it matters that much as long as we are consistent.

packages/core/src/core/chunk_manager.ts
```diff
@@ -8,4 +17,107 @@ export class ChunkManager {
...
+      resolution: effectiveResolution,
+      scaleFactor: effectiveScaleFactor,
+    };
+  }
+
+  // Calculate the world space extent (width/height) currently visible in the camera view.
+  private calculateViewExtent(
+    camera: Camera,
+    _bufferWidth: number, // screen width
+    _bufferHeight: number // screen height
```
shlomnissan: Is there a reason we include the buffer's width and height even though they’re not used?

packages/core/src/core/chunk_manager.ts
```diff
@@ -8,4 +17,107 @@ export class ChunkManager {
...
+  ): { worldWidth: number; worldHeight: number } {
+    // Screen space corners (normalized device coordinates: -1 to +1)
+    const topLeft = camera.clipToWorld([-1, 1, 0]); // camera space to world space
+    const topRight = camera.clipToWorld([1, 1, 0]);
+    const bottomLeft = camera.clipToWorld([-1, -1, 0]);
+    const bottomRight = camera.clipToWorld([1, -1, 0]);
```
shlomnissan: I don’t think we should use `clipToWorld` here. It performs a matrix inverse and two matrix transformations, which is an expensive operation in a tight loop. In this case, it’s called four times in succession. Instead, I’d suggest computing the inverse of the projection-view matrix once and using it to transform two vectors, similar to what’s done in the prototype.

packages/core/src/idetik.ts
```diff
@@ -15,6 +15,8 @@ type IdetikParams = {
 
 export type IdetikContext = {
   chunkManager: ChunkManager;
+  getCamera: () => Camera;
+  getViewport: () => { width: number; height: number };
```
shlomnissan: As I mentioned earlier, I don’t believe we need to extend the context object—these values are already accessible to the chunk manager through its `update` function.

packages/core/src/layers/image_layer_xyz.ts
```diff
@@ -1,11 +1,17 @@
 import { Layer, LayerOptions } from "../core/layer";
 import { IdetikContext } from "../idetik";
 import { Region } from "../data/region";
-import { ImageChunk, ImageChunkSource } from "../data/image_chunk";
+import {
```
shlomnissan: As we discussed yesterday, I don’t believe the layer should be aware of LOD. The current changes break the encapsulation that the chunk manager is meant to enforce.

For completeness, I’ll reiterate the point I made in our Slack conversation:

My thinking is all layers will share the same chunk manager, but I don’t think layers should be responsible for any logic related to LOD, chunk visibility, or similar concerns—those should remain internal to the chunk manager. A clean pattern might look like this:

In the onAttached hook:
```ts
this.chunkManagerSource_ = context.chunkManager->addSource(ImageChunkSource);
```

Then, in the layer’s update method:
```
const chunks = this.chunkManagerSource_.getVisibleChunks();
// Iterate through chunks and create image renderables as needed
```
This keeps the layer’s responsibilities focused on rendering, while encapsulating chunk management details.

The `addSource` method in the chunk manager will create a new `ChunkManagerSource` type, unless a source already exists for this particular `ImageChunkSource`, in which case it will return the cached instance.

This approach allows the chunk manager to support multiple sources with a clean API—no need to keep passing the source as a key when retrieving visible chunks. It also keeps caching minimal and ensures that layers remain decoupled from concerns like LOD, chunk intersections, and related logic.

shlomnissan (changes requested):
I think we should revise this PR to better align with the goal of encapsulating LOD calculations within the `ChunkManager`. I’ve left a few other comments with questions and suggested changes as well.

Even though I don’t think we’re quite there yet, you did a great job implementing the LOD logic. Your approach differs from what I originally envisioned. In my prototype, I computed a continuous estimate of pixel density based on camera zoom, assuming that all LODs follow a consistent downsampling factor (e.g., 2× per level). That assumption might not always hold—though I believe @aganders3 mentioned it’s mostly true?

Your approach, by contrast, is more image-driven. It evaluates per-LOD scale values directly and is robust to irregularities like non-power-of-two downsampling or misaligned pyramids. That makes it more flexible and data-aware.

I haven't verified the correctness of your calculations but I'll dig into it next. I just wanted to post comments quickly so you are not blocked.

Both approaches have trade-offs. If we expect arbitrary or irregular scaling, your method is safer. But if scale levels are regular, my approach is faster, source-agnostic, and more generalized.

My hunch is that your solution is better suited for our use case, but I’ll defer to @aganders3 and @andy-sweet, who have more domain knowledge here.

packages/core/src/core/chunk_manager.ts
phoenixAja: i like virtual space better than world space personally, but i felt like in the code base world space was used more frequently.

packages/core/src/core/chunk_manager.ts
andy-sweet: For what it's worth, I like the separation introduced by Shlomi here. I'd maybe take that a step further and remove the log2, which makes an assumption about the levels of detail in the image. At that point, I'd also associate the virtual/camera units per screen pixel as something being associated more with the renderer that could be passed through to the chunkmanager rather than the chunkmanager itself.

packages/core/src/core/chunk_manager.ts
andy-sweet: I think of virtual space being associated with a single multi-resolution image. So if the intent is for the chunkmanager to handle multiple multi-resolution image sources with different virtual spaces, then I prefer world space.

packages/core/src/core/chunk_manager.ts
```diff
@@ -8,4 +17,107 @@ export class ChunkManager {
...
+      if (score < bestResolutionMatch) {
+        bestResolutionMatch = score;
+        bestScaleIndex = i;
+      }
+    }
+
+    const selectedScale = availableScales[bestScaleIndex];
+    const scaleX = selectedScale[selectedScale.length - 1];
+    const scaleY = selectedScale[selectedScale.length - 2];
+    const effectiveResolution = Math.min(1.0 / scaleX, 1.0 / scaleY);
```
andy-sweet: Nit: this is not currently used and can be derived from `scaleFactor`. Unless there is an anticipated need to reuse this value multiple times, could we just remove it to save a little computation and simplify the return type?

packages/core/src/core/chunk_manager.ts
phoenixAja: no i can remove, this was leftover from something i was doing earlier

packages/core/src/core/chunk_manager.ts
phoenixAja: removed

packages/core/src/core/chunk_manager.ts
phoenixAja: this was mostly used for logging, i removed it thanks!

packages/core/src/layers/image_layer_xyz.ts
phoenixAja: done

packages/core/src/core/chunk_manager.ts
phoenixAja: done, copied the C++ implementation

packages/core/src/core/chunk_manager.ts
phoenixAja: changed logic to match prototype

packages/core/src/idetik.ts
phoenixAja: removed

packages/core/src/index.ts
```diff
@@ -9,6 +9,7 @@ export type { CameraControls } from "./objects/cameras/controls";
 export { Layer } from "./core/layer";
 export type { LayerState } from "./core/layer";
 export { LayerManager } from "./core/layer_manager";
+export { ChunkManager } from "./core/chunk_manager";
```
andy-sweet, line 12: Is this the main index of things we export to the public API? If so, I would not expect `ChunkManager` to be something we export as part Idetik's public API yet. My worry is that there will be a cascade of exported types that we might want to change in the future (e.g. `ImageChunk`).

Doesn't need to change here, but something to keep in mind as we make progress.

packages/core/src/layers/image_layer_xyz.ts
```diff
@@ -22,45 +22,35 @@ export class ImageLayerXYZ extends Layer {
...
 
   public async onAttached(context: IdetikContext) {
     this.chunkManagerSource_ = await context.chunkManager.addSource(
       this.source_
     );
+    this.chunkManagerSource_.setRegion(this.region_);
   }
 
-  public update() {
+  public async update() {
```
andy-sweet, line 45: Doesn't necessarily need to addressed here, but the abstract `Layer.update` is not `async`, so I was a little surprised to see the annotation here.

Most implementations make `async` calls, but as `update` is intended to be called in a rendering/polling loop, I'm assuming normal usage would not want/need to `await` this.

As `update` doesn't return anything, I don't think this is too important. But I haven't thought through this type of override before in javascript.

packages/core/src/core/chunk_manager.ts
```diff
@@ -4,30 +4,137 @@ import {
...
+    }
+
+    try {
+      const chunk = await this.loader_.loadChunk(
+        this.region_,
+        this.currentLOD_
+      );
+      this.chunks_[this.currentLOD_] = [chunk];
+    } catch (error) {
+      // Throttle error logging to prevent console spam in render loop
```
andy-sweet, line 76: What errors are you seeing? They may be related to errors we saw with the `ImageSeriesLayer`, which we solved by throttling the number of chunks being simultaneously fetched with zarrita (using https://github.com/chanzuckerberg/idetik/blob/07e58fe527743c85e1b37106f34c94cbf3c5e971/packages/core/src/layers/image_series_layer.ts#L44).

That was only defined within `ImageSeriesLayer` because we didn't want it too leak too much. But maybe it or something like it should live in the `ChunkManager`?

packages/core/src/core/chunk_manager.ts
```diff
@@ -4,30 +4,137 @@ import {
...
+    availableScales: number[][] // scale factors per LOD, where each scale is [c, z, y, x]
+  ): number {
+    // Get the actual visible bounds to determine virtual width
+    const viewExtent = this.calculateVisibleBounds(camera);
+    const virtualWidth = viewExtent.virtualWidth;
+    const virtualUnitsPerScreenPixel = virtualWidth / bufferWidth;
+
+    const numLods = availableScales.length;
+    const lodShift = numLods - 1;
+    const lodF = lodShift - Math.log2(1 / virtualUnitsPerScreenPixel);
```
andy-sweet, line 97: I think this is baking in the assumption of scales being separated by factors of 2. That should be asserted somewhere (i.e. we should at least log an error somewhere) to avoid being confused in the future if we ever see a dataset where that's not the case.

packages/core/src/layers/image_layer_xyz.ts
```diff
@@ -69,51 +59,60 @@ export class ImageLayerXYZ extends Layer {
     }
   }
 
-  public get channelProps(): ChannelProps[] | undefined {
-    return this.channelProps_;
-  }
+  private async updateChunks() {
```
andy-sweet, line 62: This is labelled as async, but I don't see any `await` or other `async` function in its call stack. Maybe I just missed it? Or maybe this will change in the future? Regardless, better to be consistent and correct within a PR.

packages/core/src/layers/image_layer_xyz.ts
andy-sweet: Also, maybe this function doesn't actually need to be async yet since `updateChunks` is not? Or maybe I'm wrong about that?

packages/core/src/layers/image_layer_xyz.ts
```diff
@@ -69,51 +59,60 @@ export class ImageLayerXYZ extends Layer {
...
+      this.setState("loading");
+    }
+
+    const chunks = this.chunkManagerSource_.getVisibleChunks();
 
-  private async load(region: Region) {
-    if (this.state !== "initialized") {
-      throw new Error(`Trying to load chunks more than once.`);
+    if (chunks.length > 0) {
+      this.processChunk(chunks[0]);
```
andy-sweet, line 72: If I understand this PR correctly, this will create a new texture and image renderable object on every tick of Idetik's render loop. Is that right? That's fine as a very temporary intermediate state, but I would expect this to be changed in the next one or two related PRs.

andy-sweet (commented):
Generally not too many surprises here and it's nice to see this start to come in. This helped me understand some design choices that I had not fully grasped yet either (e.g. the need for `ChunkManagerSource.setRegion`).

I think some/most of my feedback may be addressed by https://github.com/chanzuckerberg/idetik/pull/278 but I haven't looked through that yet.

There seem to be some intermediate states in this PR, some of which I'm not comfortable approving on their own. If they are fixed by https://github.com/chanzuckerberg/idetik/pull/278 , you should feel comfortable merging this and help getting that in soon after.

Also, it looks like the PR description needs to be updated to match the new changes.

PR #273, opened 2025-06-10
Add ChunkManagerSource and refactor

shlomnissan:
### Summary

This PR introduces a `ChunkManagerSource` interface that allows layers to
interact with the chunk manager indirectly. It also adds placeholder
structures and comments to outline the intended chunk manager architecture
without fully implementing it, helping to unblock downstream development.

### Tests & Checks

- All checks have passed
- Verified the chunk streaming example is not broken

### DIFF
--- a/packages/core/src/core/chunk_manager.ts
+++ b/packages/core/src/core/chunk_manager.ts
@@ -1,11 +1,77 @@
-import { Camera } from "../objects/cameras/camera";
+import {
+  ImageChunk,
+  ImageChunkLoader,
+  ImageChunkSource,
+  LoaderAttributes,
+} from "@/data/image_chunk";
 
-export class ChunkManager {
-  public getVisibleChunks() {
-    // TODO: implement
+import { Camera } from "../objects/cameras/camera";
+import { vec2, vec4, mat4 } from "gl-matrix";
+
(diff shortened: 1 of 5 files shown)


phoenixAja (approved):
lg!

packages/core/src/core/chunk_manager.ts
```diff
@@ -1,11 +1,77 @@
...
+} from "@/data/image_chunk";
+
 import { Camera } from "../objects/cameras/camera";
+import { vec2, vec4, mat4 } from "gl-matrix";
+
+type Bounds = { min: vec2; max: vec2 };
+
+export class ChunkManagerSource {
+  // @ts-expect-error unused for now
+  private readonly chunks_: ImageChunk[][] = [];
```
andy-sweet, line 15: In neuroglancer, chunk has a general meaning across data types (images, meshes, etc.), so chunk manager does too: https://github.com/google/neuroglancer/tree/master/src/chunk_manager#readme

In Idetik, we only have image chunks so far, so I might expect this class to be called `ImageChunkManager` and `ImageChunkManagerSource` to help clarify that.

In the future, I expect we'll at least want Idetik to be able to handle different chunk types. Do we have a technical design for a more general purpose chunk manager?

PR #274, opened 2025-06-10
feat: Changes for Organelle Box time use case

bchu1:
Next PR: Rename `z` &#8594; `slice`.

Adds text customizability.

New demo is at http://localhost:5173/src/examples/organelle-box-time/


packages/react/src/components/hooks/useOmeZarrImageViewer.ts
```diff
@@ -219,11 +222,19 @@ export function useOmeZarrViewer({
...
+            initialZ = 0;
+            break;
+          case "middle":
+            initialZ = Math.floor(zShape / 2);
+            break;
+          case "end":
+            initialZ = zShape;
+            break;
+          default:
+            initialZ = await loadOmeroDefaultZ(sourceUrl);
```
aganders3: nit: I'd consider using `0` as the default instead, and making this explicit with `initialIndex === 'omeroDefaultZ'` or something, because `defaultZ` only has the correct meaning if the series dimension is Z. The omero metadata also supports `defaultT` but I think in general it's less useful and just setting the initial timeframe to 0 is good.

packages/react/src/components/hooks/useOmeZarrImageViewer.ts
```diff
@@ -37,10 +36,14 @@ export interface OmeZarrImageViewerProps {
...
   onLoadAllSlicesAborted?: () => void;
+  indexIndicatorText?:
+    | string
+    | ((currentIndex: number, totalIndexes: number) => string);
+  loadAllButtonText?: string | (() => string);
   fallbackContrastLimits?: [number, number];
   lod?: number;
   shouldAutoLoadAllSlices?: boolean;
-  shouldLoadMiddleZ?: boolean;
+  initialIndex?: "start" | "middle" | "end";
```
aganders3: As an idea - this could be `"start" | "middle" | "end" | number;`, but everything is clunky since we don't know the number of slices available without loading some metadata first. Likewise `defaultZ` is based on the number of slices in the high res image, which is awkward when the region is not "full" in that dimension.

I think we're likely to keep tweaking this, so no need to change it here.

packages/react/src/examples/organelle-box-time/App.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { useState } from "react";
+import { ChannelControlsList } from "../../components/viewers/OmeZarrImageViewer/components/ChannelControlsList";
+
+const sourceUrl =
+  "https://public.czbiohub.org/organelle_box/datasets/A549/2024_11_07_A549_SEC61_DENV_cropped.zarr";
+const wellPath = "B/3";
+const region: Region = [
+  { dimension: "T", index: { type: "full" } },
+  { dimension: "C", index: { type: "full" } },
+  { dimension: "Z", index: { type: "point", value: 0 } },
```
aganders3: Again it's clunky, but we might want to set this to the omero `defaultZ` (at least when this is done on the portal - for the demo it may not matter much).

packages/react/src/examples/organelle-box-time/index.html
aganders3, line 1: This file structure works for now, but I'm wondering if we can use a similar structure to the `core` repo where the examples don't get put in `src`.

Again doesn't need to be in this PR.

packages/react/src/examples/organelle-box-time/App.tsx
```diff
@@ -0,0 +1,52 @@
...
+        sourceUrl={imageUrl}
+        region={region}
+        seriesDimensionName="T"
+        initialIndex="start"
+        loadAllButtonText="Load entire duration (230MB)"
+        classNames={{
+          root: "bg-dark-sds-color-primitive-gray-100 flex-auto min-h-0",
+        }}
+        indexIndicatorText={(currentIndex, totalIndexes) =>
+          `${Math.floor(currentIndex / 60)}h${currentIndex % 60}m of ${Math.floor(totalIndexes / 60)}h${totalIndexes % 60}m`
```
aganders3: I like that this can take a function to format! I think the data we have is in seconds though (based on the Zarr metadata).

aganders3 (approved):
Looks great - thanks!

packages/react/src/examples/organelle-box-time/App.tsx
bchu1: FYI @kyleawayan

packages/react/src/examples/organelle-box-time/index.html
bchu1: Done.

PR #275, opened 2025-06-11
fix: Fix only half of slices in demo loading

bchu1:
Z slider was only going halfway.

LMK if I should just ignore this for now, I know there's a bunch of things in progress.


aganders3 (approved):
Thanks!

andy-sweet (approved):
I also took a quick look at this. I didn't follow all the related changes that caused the bug. But this fix looks right to me.

PR #279, opened 2025-06-16
fix: add backdrop-blur-md to the slice/time indicator, and slider

kyleawayan:
### Summary

Forgot to add the same background blur styles to the slice/time indicator and slider -- it now matches the background blur of the `ChannelControlsList` (`backdrop-blur-md`) -- basically needed wherever `bg-black/75` is
https://github.com/chanzuckerberg/idetik/blob/8a2d75941ad5a4304630dad9e7baaa11f7381e48/packages/react/src/components/viewers/OmeZarrImageViewer/components/ChannelControlsList/ChannelControlsList.tsx#L47-L51

### Related Issue
Closes #<issue-number>


bchu1:
You can use the new `classNames` prop to customize CSS! They should be `sliceIndicator` and `sliceSliderContainer`

kyleawayan:
Oh ok that sounds cool! But I think this is actually a bug (but purely cosmetic) for the default styles even.

bchu1:
> Oh ok that sounds cool! But I think this is actually a bug (but purely cosmetic) for the default styles even.

Oh I see

bchu1 (approved):

PR #280, opened 2025-06-17
Image loading using chunks

shlomnissan:
### Summary

This PR introduces image loading via the chunk manager, currently limited to
the single image layer. To maintain backward compatibility and accommodate the
evolving API design, the chunk manager is only used when the region passed to
the image layer specifies "full" for both the `x` and `y` axes.

At this stage, LOD handling and tile intersection logic are not yet implemented
—these will be addressed in a follow-up PR. For now, all chunks are treated
as visible, and a fixed LOD is used.

**Note**: As discussed in our last engineering sync, this is an interim implementation. Once chunk-based loading is in place for both image layers, we’ll revisit the public API and refactor our examples to rely exclusively on the chunk manager for image data loading.

### Tests & Checks
- All tests/checks have passed.
- Manual verification that existing examples are working as expected.

https://github.com/user-attachments/assets/5cb8d17b-9aa8-4515-b65b-d0e624e5f132


packages/core/src/core/chunk_manager.ts
```diff
@@ -4,42 +4,99 @@ import {
...
-
-    // TODO: create chunks for each LOD without loading data
+    this.region_ = region;
+
+    // generate chunks for each LOD without loading data
+    this.chunks_ = Array(attrs.length)
+      .fill(null)
+      .map(() => []);
+    for (let lod = 0; lod < attrs.length; ++lod) {
+      const xIdx = region.findIndex((entry) => entry.dimension === "x");
```
andy-sweet: The way I interpret the [OME-NGFF v0.4 specification](https://ngff.openmicroscopy.org/0.4/#multiscale-md) the names of axes in multi-scale images do not have to be "c", "t", "z", "y", "x".

However, the examples use those specifically and that's what I've mostly seen in practice.

One exception is that some OME-Zarrs use upper-case instead. And that seems to be the case for the [OrganelleBox data](https://public.czbiohub.org/organelle_box/datasets/A549/organelle_box_crop_v1.zarr/ATG101/MeOH/000000/.zattrs).

A quick, maybe good-enough, workaround for now is to check for lower and upper case or convert to lower. 

Alternatively, you could pass through the spatial dimensions as done for the z/t dimension name in the image series layer.

Another idea is to always use the last two dimension names from `attrs` since the loader makes a similar assumption, which should be consistent with the OME-NGFF spec.

packages/core/src/data/ome_zarr_image_loader.ts
```diff
@@ -169,6 +206,43 @@ export class OmeZarrImageLoader {
...
+  ) {
+    const indices = this.regionToIndices(region, attrs);
+    const array = await zarr.open.v2(this.root_.resolve(attrs.datasetPath), {
+      kind: "array",
+      attrs: false,
+    });
+
+    const output = new Map();
+    region.forEach((entry, dimIdx) => {
+      if (entry.dimension === "x") {
```
andy-sweet, line 225: See the similar comment in `chunk_manager` about assuming `x` and `y` names.

packages/core/examples/chunk_streaming/main.ts
```diff
@@ -21,11 +21,11 @@ const region: Region = [
...
   { dimension: "c", index: { type: "point", value: 0 } },
   { dimension: "z", index: { type: "point", value: 300 } },
-  { dimension: "y", index: { type: "interval", start: top, stop: bottom } },
-  { dimension: "x", index: { type: "interval", start: left, stop: right } },
+  { dimension: "y", index: { type: "full" } },
+  { dimension: "x", index: { type: "full" } },
 ];
 const channelProps = [{ contrastLimits: [0, 255] as [number, number] }];
-const layer = new ImageLayerXYZ({ source, region, channelProps });
+const layer = new ImageLayer({ source, region, channelProps });
```
andy-sweet: I found it useful to add an `AxesLayer` to help verify that the chunks are being offset/scaled as intended with something like this

```typescript
const axes = new AxesLayer({ length: 2000, width: 0.01 });
```

and then appending that to the Idetik layer list.

Would be better with ticks and labels, or a scale-bar in addition, but might be a useful addition?

packages/core/src/layers/image_series_layer.ts
```diff
@@ -171,7 +170,7 @@ export class ImageSeriesLayer extends Layer {
         x: chunk.shape.x * chunk.scale.x,
         y: chunk.shape.y * chunk.scale.y,
       };
-    } else {
+    } else if (chunk.data) {
       this.texture_.data = chunk.data;
```
andy-sweet, line 174: Is there any use in handling the case when `chunk.data` is undefined with a log warning/error here? I don't think we expect to reach that case here right now. But that might also go away soon, so maybe not worth it.

packages/core/src/core/chunk_manager.ts
andy-sweet: This causes some of the examples to log an error (e.g. image2d_from_omezarr4d_hcs), but since they don't actually need the chunk manager, I think they continue to work as expected.

packages/core/src/core/chunk_manager.ts
```diff
@@ -4,42 +4,99 @@ import {
...
+      const chunkHeight = attrs[lod].chunks[yIdx];
+      const chunksX = Math.ceil(attrs[lod].shape[xIdx] / chunkWidth);
+      const chunksY = Math.ceil(attrs[lod].shape[yIdx] / chunkHeight);
+      const channels = attrs[lod].shape.length === 3 ? attrs[lod].shape[0] : 1;
+      for (let x = 0; x < chunksX; ++x) {
+        for (let y = 0; y < chunksY; ++y) {
+          this.chunks_[lod].push({
+            state: "unloaded",
+            lod,
+            visible: true, // TODO:(shlomnissan) should be set to false
```
andy-sweet, line 54: Should this be set to `false` here and then updated to `true` in `updateChunks`? That seems a bit redundant with `state === "loaded"`, so this is probably intended for another purpose?

packages/core/src/core/chunk_manager.ts
```diff
@@ -4,42 +4,99 @@ import {
...
   public getVisibleChunks(): ImageChunk[] {
-    return [];
+    return this.chunks_[curr_lod].filter((e) => e.visible);
   }
 
-  public updateChunks(_: Bounds) {
-    // TODO: intersection test, set visibility and load data
+  public async updateChunks(_: Bounds) {
+    // TODO: map the LOD factor from the chunk manager to an available LOD in image space
+    // TODO: replace the following block with loading based on intersection tests
```
andy-sweet, line 83: Is this also where you would expect to some chunk eviction? Maybe based on some kind of distance to the current bounds?

andy-sweet (approved):
Generally looks great and the example works well for me. No blocking feedback, but the hard-coding of "xyz" dimension names will quickly become too fragile. I mentioned some quick workarounds, but this is where a more explicit mapping between Idetik's visualized dimensions and a source's dimensions might be needed.

If that's not provided by the user/developer, we could likely write some simple-ish logic to automatically infer some sensible correspondences (e.g. by order), but that's work for another PR.

phoenixAja (approved):
all makes sense and looks good to me

packages/core/src/layers/image_layer.ts
```diff
@@ -39,19 +43,44 @@ export class ImageLayer extends Layer {
...
-        throw new Error(`Unhandled LayerState case: ${exhaustiveCheck}`);
+    if (this.useChunkManager_) {
+      if (!this.chunkManagerSource_) return;
+      const chunks = this.chunkManagerSource_.getVisibleChunks();
+      chunks.forEach((chunk) => {
+        if (chunk.state === "loaded" && !this.visibleChunks_.includes(chunk)) {
+          this.visibleChunks_.push(chunk);
+          this.addObject(this.createImage(chunk, this.channelProps));
+        }
+        this.setState("ready");
```
aganders3: I think this will fire a `ready` event frequently. Is there somewhere else it can go?

packages/core/src/layers/image_layer.ts
```diff
@@ -39,19 +43,44 @@ export class ImageLayer extends Layer {
...
-        break;
-      default: {
-        const exhaustiveCheck: never = this.state;
-        throw new Error(`Unhandled LayerState case: ${exhaustiveCheck}`);
+    if (this.useChunkManager_) {
+      if (!this.chunkManagerSource_) return;
+      const chunks = this.chunkManagerSource_.getVisibleChunks();
+      chunks.forEach((chunk) => {
+        if (chunk.state === "loaded" && !this.visibleChunks_.includes(chunk)) {
+          this.visibleChunks_.push(chunk);
```
aganders3: Chunks are never removed from `this.visibleChunks`. Maybe this is intentional until implementing visibility/LOD selection, so you could just add a TODO. Though I think the implementation could be started without visibility/LOD selection.

Also, should this instead be a `Map<ImageChunk, ImageRenderable>`? I think that was how it was described in the spec, and will help to remove renderable objects.

packages/core/src/data/ome_zarr_image_loader.ts
```diff
@@ -114,13 +146,17 @@ export class OmeZarrImageLoader {
...
+      state: "loaded",
+      lod: lod,
+      visible: true,
       data: subarray.data,
       shape: {
         x: subarray.shape[subarray.shape.length - 1],
         y: subarray.shape[subarray.shape.length - 2],
         c: subarray.shape.length === 3 ? subarray.shape[0] : 1,
       },
+      chunkIndex: { x: 0, y: 0 },
```
aganders3, line 162: This is a little odd, but I think this is just an artifact of re-using the `ImageChunk` type, so I think it's fine for now.

I still think `chunkIndex` may need to be expanded to n-d, but that can also come later.

packages/core/examples/chunk_streaming/main.ts
shlomnissan: Sure, I’ll add an axes layer. FWIW I see this example as temporary—it’s primarily for development purposes. I think we should eventually curate a more meaningful set of examples, but we can address that when we start building the Idetik website.

packages/core/src/data/ome_zarr_image_loader.ts
```diff
@@ -59,7 +59,39 @@ export class OmeZarrImageLoader {
...
 
-  async loadChunk(
+  public async loadChunkDataFromRegion(chunk: ImageChunk, region: Region) {
+    const attrs = this.getImageAttributes()[chunk.lod];
+    const array = await zarr.open.v2(this.root_.resolve(attrs.datasetPath), {
+      kind: "array",
+      attrs: false,
+    });
+
+    const dimInfo = await this.getDimInfoMap(region, chunk, attrs);
```
aganders3, line 69: It's fairly fast, but this causes a read of the same `.zarray` for every chunk in the current LOD. Maybe this can be cached somehow?

I'm a little confused on the the role of `dimInfoMap` in general. I can give it a closer read, but this seems like something that could change or just go away depending on the final public API.

packages/core/src/data/ome_zarr_image_loader.ts
```diff
@@ -169,6 +206,43 @@ export class OmeZarrImageLoader {
...
+      if (entry.dimension === "y") {
+        const value = 0; // not used for y dimension
+        output.set("y", { dimIdx, chunkIdx: chunk.chunkIndex.y, value });
+        return;
+      }
+
+      const value = indices[dimIdx];
+      if (typeof value !== "number") {
+        throw new Error(`Expected numeric index for ${entry.dimension}`);
+      }
```
aganders3, line 240: This restricts the current implementation to single-channel, right? This is another case where using logical chunks and letting zarrita's `zarr.get` handle the bookkeeping makes more sense to me than relying on source data chunks and dimension ordering.

packages/core/src/core/chunk_manager.ts
shlomnissan: Yes, visibility will be determined in `updateChunks`.

The chunk's state refers to the status of its underlying data. Some chunks may be loaded but not currently visible, while others might be potentially visible, but their data won’t be loaded until the LOD they belong to becomes active.

packages/core/src/core/chunk_manager.ts
shlomnissan: Yes, I believe chunk eviction will be triggered from here as well. I’ve only added comments for the immediate functionality we plan to implement in the first iteration. We’ll be addressing caching and eviction strategies soon—initially, this might be something simple like a fixed-size LRU cache, but we may explore more robust approaches based on distance. I haven’t given it too much thought yet; the main goal is to establish a baseline we can use to benchmark different strategies, which is part of Q3 roadmap.

packages/core/src/layers/image_series_layer.ts
shlomnissan: > might also go away soon, so maybe not worth it

That was my thinking as well. Once we integrate chunking into the image series layer, I expect the first step will be factoring out the current "load region" logic. At that point, this will likely be removed—possibly along with the existing layer state management in general—unless we identify a need to keep it.

packages/core/src/layers/image_layer.ts
```diff
@@ -39,19 +43,44 @@ export class ImageLayer extends Layer {
     this.region_ = region;
     this.lod_ = lod;
     this.channelProps_ = channelProps;
+
+    const x = region.find((r) => r.dimension === "x");
+    const y = region.find((r) => r.dimension === "y");
+    this.useChunkManager_ =
+      x?.index.type === "full" && y?.index.type === "full";
```
aganders3: We might want this to also check for point dimensions in `t`, `c`, and `z` to ensure existing multi-channel implementations continue to work (though most things use `ImageSeriesLayer`).

packages/core/src/layers/image_layer.ts
```diff
@@ -39,19 +43,44 @@ export class ImageLayer extends Layer {
...
+    console.log(this.useChunkManager_);
```
aganders3: Let's remove this log or use the logging system with a more informative message.

packages/core/src/core/chunk_manager.ts
shlomnissan: Ah, interesting — I assumed that uppercase axis names like X/Y are not valid in OME-NGFF, since the spec consistently uses lowercase. I’ll convert axis names to lowercase before doing any comparisons to be safe.

I also couldn’t find any guarantee in the spec that the last two dimensions are always y and x. So, it seems safer to rely on inspecting the axes array and comparing axis names directly to determine the order. Am I missing something?

aganders3 (approved):
This looks great!

I have the same concerns as @andy-sweet about the dimension names and ordering, but nothing blocking.

It also looks like #275 is either back here or was not actually fixed. That's an `ImageSeriesLayer` so it's unlikely to be the fault of this PR and we can address it separately.

Reloading shows how clogged the main thread gets when decoding from the browser cache. That should definitely be part of our Q3 roadmap.

packages/core/src/core/chunk_manager.ts
aganders3: You're right - v0.4 (and I believe v0.5) says:
> In addition, the entries MUST be ordered by "type" where the "time" axis must come first (if present), followed by the "channel" or custom axis (if present) and the axes of type "space". If there are three spatial axes where two correspond to the image plane ("yx") and images are stacked along the other (anisotropic) axis ("z"), the spatial axes SHOULD be ordered as "zyx".

I think [RFC-3](https://ngff.openmicroscopy.org/rfc/3/index.html) is very likely to be accepted though:
> This document proposes removing any restrictions on the number of dimensions stored in OME-Zarr arrays. Additionally, it removes restrictions on the names and types of included dimensions.

shlomnissan: > This causes some of the examples to log an error (e.g. image2d_from_omezarr4d_hcs)

This error was caused by a bug that I’ve now fixed. This example should not have used the chunk manager in the first place, since it includes an interval along the channel (C) axis. That said, the chunk manager probably should support this case. Next, I’ll work on handling channels properly. The goal is to have all examples eventually use the chunk manager.

> This document proposes removing any restrictions on the number of dimensions stored in OME-Zarr arrays. Additionally, it removes restrictions on the names and types of included dimensions.

Just to confirm, NGFF are the maintainer of the OME-Zarr spec, correct? @aganders3 either way, I ended up comparing axis name explicitly.

packages/core/src/layers/image_layer.ts
shlomnissan: Good catch! I’ll set it conditionally for now until we decide how to handle layer states more broadly.

packages/core/src/layers/image_layer.ts
shlomnissan: This was a debug log I forgot to remove. I agree, this is a good log to have.

packages/core/src/data/ome_zarr_image_loader.ts
shlomnissan: Yes, it’s currently limited to a single channel. Supporting multiple channels is my next task. I’m intentionally avoiding handling arbitrary shapes for now (or even fixed shapes with varying strides), but if `zarr.get()` simplifies channel access, I’ll reconsider.

packages/core/src/data/ome_zarr_image_loader.ts
shlomnissan: Yes, that was intentional. I assumed Zarrita handles caching and didn’t verify it, relying on HTTP caching as a fallback. At some point, we should consider implementing our own Zarr loader—especially if the chunk manager will support chunk ranges—as it would give us better control over caching, OME-Zarr metadata validation, and more. Not a priority for now, though.

> I'm a little confused on the the role of dimInfoMap in general

It’s just remapping a region into a shape that’s easier to iterate over when building the chunk indices array. The easiest way to understand what it’s doing is to log the output. I could probably remove it, but I think it improves readability for now. Ultimately, we’ll likely need to refactor the OME-Zarr loader more broadly once the chunk manager becomes the default for loading data—at that point, we can revisit how we organize things.

packages/core/src/data/ome_zarr_image_loader.ts
shlomnissan: > this is just an artifact of re-using the ImageChunk

Yep. Once the chunk manager becomes the default for loading data, this function will be removed altogether.

packages/core/src/layers/image_layer.ts
shlomnissan: > Maybe this is intentional until implementing visibility/LOD selection

Yes. It is intentional. I will add a TODO and use a map to store chunk <-> images.

packages/core/src/core/chunk_manager.ts
aganders3: OME is "Open Microscopy Environment" and is the name of a group. NGFF stands for "Next Generation File Format", so OME-NGFF and OME-Zarr are interchangeable in my mind (though technically I think NGFF refers to a _process_ and OME-Zarr is a specific _spec_...much like Frankenstein is actually the _doctor_, not the monster). So I think it's more OME who are the maintainers.

All this to say comparing with literal axis names is fine for now, but the comping changes suggests to me that in the future we will want to more explicitly map "data space" to screen space without imposing constraints on the source data in the library. I think individual applications we make can be more restrictive with the format they expect (axes names/order).

packages/core/src/layers/image_layer.ts
shlomnissan: @aganders3 
```ts
private updateChunks() {
  if (!this.chunkManagerSource_) return;

  // Temporary until we decide how to approach state management
  if (this.state !== "ready") {
    this.setState("ready");
  }

  // This loop removes image renderables for chunks that are no longer visible.
  // While this approach works for now, it may be more efficient in the future
  // to reuse renderables by updating their underlying data instead of repeatedly
  // creating new texture objects. Note: GPU resources are not currently being
  // released, so this will also need to be addressed soon.
  this.visibleChunks_.forEach((image, chunk) => {
    if (!chunk.visible) {
      this.removeObject(image);
      this.visibleChunks_.delete(chunk); // safe
    }
  });

  this.chunkManagerSource_.getVisibleChunks().forEach((chunk) => {
    if (chunk.state === "loaded" && !this.visibleChunks_.has(chunk)) {
      const image = this.createImage(chunk, this.channelProps);
      this.visibleChunks_.set(chunk, image);
      this.addObject(image);
    }
  });
}

public update() {
  if (this.useChunkManager_) {
    this.updateChunks();
  } else {
    // switch (this.state)
  }
}
```

PR #281, opened 2025-06-18
feat: Dynamic LOD Computation for ChunkManager

phoenixAja:
This PR replaces the previously hardcoded curr_lod constant with logic that dynamically computes the appropriate level of detail (LOD) based on the current view bounds and screen resolution.

The ChunkManagerSource now uses the visible virtual width and the buffer (canvas) width to calculate a screen-space resolution ratio, from which it derives the most suitable LOD. This allows the system to adaptively select a coarser or finer LOD as the user zooms in and out, setting the stage for progressive loading and improved performance at scale.

LOD computation happens during ChunkManager.update, prior to triggering chunk updates.

### DIFF
--- a/packages/core/src/core/chunk_manager.ts
+++ b/packages/core/src/core/chunk_manager.ts
@@ -16,6 +16,8 @@ export class ChunkManagerSource {
   private readonly region_;
   private readonly attrs_: LoaderAttributes[];
   public currentLOD_: number = 0;
+  private readonly xIdx_: number;
+  private readonly yIdx_: number;
 
   constructor(
     loader: ImageChunkLoader,
@@ -26,6 +28,16 @@ export class ChunkManagerSource {
     this.region_ = region;
     this.attrs_ = attrs;
     this.currentLOD_ = 1;
+
+    this.xIdx_ = region.findIndex(
+      (entry) => entry.dimension.toLocaleLowerCase() === "x"
+    );
+    this.yIdx_ = region.findIndex(
+      (entry) => entry.dimension.toLocaleLowerCase() === "y"
+    );
+    if (this.xIdx_ === -1 || this.yIdx_ === -1) {
+      throw new Error("Missing required spatial axis x/y");
+    }
 
     // generate chunks for each LOD without loading data
     this.chunks_ = Array(this.attrs_.length)
@@ -91,8 +103,9 @@ export class ChunkManagerSource {
     for (let i = 1; i < availableScales.length; i++) {
       const prev = availableScales[i - 1];
       const curr = availableScales[i];
-      const rx = prev[0] / curr[0];
-      const ry = prev[1] / curr[1];
+      const rx = curr[this.xIdx_] / prev[this.xIdx_];
+      const ry = curr[this.yIdx_] / prev[this.yIdx_];
+
       if (Math.abs(rx - 2) > EPS || Math.abs(ry - 2) > EPS) {
         console.error(
           `Scale ratio between levels ${i - 1} and ${i} is (${rx}, ${ry}), expected (2.0, 2.0)`
--- a/packages/core/src/core/chunk_manager.ts
+++ b/packages/core/src/core/chunk_manager.ts
@@ -12,7 +12,7 @@ import { almostEqual } from "@/utilities/almost_equal";
 type Bounds = { min: vec2; max: vec2 };
 
 export class ChunkManagerSource {
-  private readonly chunks_: ImageChunk[][] = [];
+  private readonly chunks_: ImageChunk[][];
   private readonly loader_;
   private readonly region_;
   private readonly attrs_: LoaderAttributes[];
@@ -102,9 +102,6 @@ export class ChunkManagerSource {
       const ry = curr[yIdx] / prev[yIdx];
 
       if (!almostEqual(rx, 2) || !almostEqual(ry, 2)) {
-        console.error(
-          `Scale ratio between levels ${i - 1} and ${i} is (${rx}, ${ry}), expected (2.0, 2.0)`
-        );
         throw new Error(
           `Scales must be separated by factors of 2. Got ratio (${rx}, ${ry}) between scales ${prev} and ${curr}`
         );
--- a/packages/core/src/core/chunk_manager.ts
+++ b/packages/core/src/core/chunk_manager.ts
@@ -114,12 +114,9 @@ export class ChunkManagerSource {
     const virtualWidth = Math.abs(visibleBounds.max[0] - visibleBounds.min[0]);
     const virtualUnitsPerScreenPixel = virtualWidth / bufferWidth;
 
-    const numLods = availableScales.length;
-    const lodShift = numLods - 1;
-    const lodF = lodShift - Math.log2(1 / virtualUnitsPerScreenPixel);
-
-    const maxLod = numLods - 1;
-    const newLOD = Math.max(0, Math.min(maxLod, Math.floor(lodF)));
+    const maxLod = availableScales.length - 1;
+    const lodF = maxLod - Math.log2(1 / virtualUnitsPerScreenPixel);
+    const newLod = Math.max(0, Math.min(maxLod, Math.floor(lodF)));
 
     if (newLOD !== this.currentLOD_) {
       console.log(`LOD changed from ${this.currentLOD_} to ${newLOD}`);
--- a/packages/core/src/core/chunk_manager.ts
+++ b/packages/core/src/core/chunk_manager.ts
@@ -114,10 +114,8 @@ export class ChunkManagerSource {
   }
 
   public setLOD(lodFactor: number): void {
-    const numLods = this.attrs_.length;
-    const lodShift = numLods - 1;
-    const lodF = lodShift - lodFactor;
-    const newLOD = Math.max(0, Math.min(numLods - 1, Math.floor(lodF)));
+    const maxLOD = this.attrs_.length - 1;
+    const targetLOD = Math.max(0, Math.min(maxLOD, Math.floor(maxLOD - lodFactor)));
 
     if (newLOD !== this.currentLOD_) {
       console.debug(`LOD changed from ${this.currentLOD_} to ${newLOD}`);
--- a/packages/core/src/core/chunk_manager.ts
+++ b/packages/core/src/core/chunk_manager.ts
@@ -7,7 +7,7 @@ import {
 import { Region } from "../data/region";
 import { Camera } from "../objects/cameras/camera";
 import { vec2, vec4, mat4 } from "gl-matrix";
-import { almostEqual } from "@/utilities/almost_equal";
+import { almostEqual } from "../utilities/almost_equal";
 
 type Bounds = { min: vec2; max: vec2 };
 
--- a/packages/core/src/core/chunk_manager.ts
+++ b/packages/core/src/core/chunk_manager.ts
@@ -7,6 +7,7 @@ import {
 import { Region } from "../data/region";
 import { Camera } from "../objects/cameras/camera";
 import { vec2, vec4, mat4 } from "gl-matrix";
+import { almostEqual } from "@/utilities/almost_equal";
 
 type Bounds = { min: vec2; max: vec2 };
 
@@ -15,9 +16,10 @@ export class ChunkManagerSource {
   private readonly loader_;
   private readonly region_;
   private readonly attrs_: LoaderAttributes[];
-  public currentLOD_: number = 0;
+  private currentLOD_: number = 0;
   private readonly xIdx_: number;
   private readonly yIdx_: number;
+  private readonly channelIdx_: number;
 
   constructor(
     loader: ImageChunkLoader,
@@ -27,7 +29,7 @@ export class ChunkManagerSource {
     this.loader_ = loader;
     this.region_ = region;
     this.attrs_ = attrs;
-    this.currentLOD_ = 1;
+    this.currentLOD_ = 0;
 
     this.xIdx_ = region.findIndex(
       (entry) => entry.dimension.toLocaleLowerCase() === "x"
@@ -38,7 +40,14 @@ export class ChunkManagerSource {
     if (this.xIdx_ === -1 || this.yIdx_ === -1) {
       throw new Error("Missing required spatial axis x/y");
     }
-
+    let channelIdx = region.findIndex(
+      (entry) => entry.dimension.toLowerCase() === "c"
+    );
+    if (channelIdx === -1) {
+      channelIdx = 0;
+    }
+    this.validateScaleRatios(this.xIdx_, this.yIdx_);
+    this.channelIdx_ = channelIdx;
     // generate chunks for each LOD without loading data
     this.chunks_ = Array(this.attrs_.length)
       .fill(null)
@@ -53,7 +62,9 @@ export class ChunkManagerSource {
         this.attrs_[lod].shape[this.yIdx_] / chunkHeight
       );
       const channels =
-        this.attrs_[lod].shape.length === 3 ? this.attrs_[lod].shape[0] : 1;
+        this.attrs_[lod].shape.length === 3
+          ? this.attrs_[lod].shape[this.channelIdx_]
+          : 1;
       for (let x = 0; x < chunksX; ++x) {
         for (let y = 0; y < chunksY; ++y) {
           this.chunks_[lod].push({
@@ -82,25 +93,15 @@ export class ChunkManagerSource {
     }
   }
 
-  public getVisibleChunks(): ImageChunk[] {
-    return this.chunks_[this.currentLOD_].filter((e) => e.visible);
-  }
-
-  public computeLOD(
-    visibleBounds: Bounds,
-    bufferWidth: number // screen/canvas width in pixels
-  ): void {
+  private validateScaleRatios(xIdx: number, yIdx: number): void {
     const availableScales = this.attrs_.map((attr) => attr.scale);
-
-    // Assert that scales are separated by factors of 2
-    const EPS = 1e-6;
     for (let i = 1; i < availableScales.length; i++) {
       const prev = availableScales[i - 1];
       const curr = availableScales[i];
-      const rx = curr[this.xIdx_] / prev[this.xIdx_];
-      const ry = curr[this.yIdx_] / prev[this.yIdx_];
+      const rx = curr[xIdx] / prev[xIdx];
+      const ry = curr[yIdx] / prev[yIdx];
 
-      if (Math.abs(rx - 2) > EPS || Math.abs(ry - 2) > EPS) {
+      if (!almostEqual(rx, 2) || !almostEqual(ry, 2)) {
         console.error(
           `Scale ratio between levels ${i - 1} and ${i} is (${rx}, ${ry}), expected (2.0, 2.0)`
         );
@@ -109,20 +110,20 @@ export class ChunkManagerSource {
         );
       }
     }
+  }
 
-    // Calculate virtual width from visible bounds
-    const virtualWidth = Math.abs(visibleBounds.max[0] - visibleBounds.min[0]);
-    const virtualUnitsPerScreenPixel = virtualWidth / bufferWidth;
+  public getVisibleChunks(): ImageChunk[] {
+    return this.chunks_[this.currentLOD_].filter((e) => e.visible);
+  }
 
-    const numLods = availableScales.length;
+  public setLOD(lodFactor: number): void {
+    const numLods = this.attrs_.length;
     const lodShift = numLods - 1;
-    const lodF = lodShift - Math.log2(1 / virtualUnitsPerScreenPixel);
-
-    const maxLod = numLods - 1;
-    const newLOD = Math.max(0, Math.min(maxLod, Math.floor(lodF)));
+    const lodF = lodShift - lodFactor;
+    const newLOD = Math.max(0, Math.min(numLods - 1, Math.floor(lodF)));
 
     if (newLOD !== this.currentLOD_) {
-      console.log(`LOD changed from ${this.currentLOD_} to ${newLOD}`);
+      console.debug(`LOD changed from ${this.currentLOD_} to ${newLOD}`);
       this.currentLOD_ = newLOD;
     }
   }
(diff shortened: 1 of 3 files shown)


packages/core/src/core/chunk_manager.ts
```diff
@@ -25,27 +26,34 @@ export class ChunkManagerSource {
...
-        throw new Error("Missing required spatial axis x/y");
-      }
-
-      const chunkWidth = attrs[lod].chunks[xIdx];
-      const chunkHeight = attrs[lod].chunks[yIdx];
-      const chunksX = Math.ceil(attrs[lod].shape[xIdx] / chunkWidth);
-      const chunksY = Math.ceil(attrs[lod].shape[yIdx] / chunkHeight);
-      const channels = attrs[lod].shape.length === 3 ? attrs[lod].shape[0] : 1;
+      const channels =
+        this.attrs_[lod].shape.length === 3 ? this.attrs_[lod].shape[0] : 1;
```
aganders3: The current recommended axis order is tczyx so axis 0 is often time. We may need to start specifying the channel axis name or something. For now I propose we use a similar pattern as above:

```
this.channelIdx_ = region.findIndex(
      (entry) => entry.dimension.toLocaleLowerCase() === "c"
    );
```

packages/core/src/core/chunk_manager.ts
```diff
@@ -25,27 +26,34 @@ export class ChunkManagerSource {
...
       for (let x = 0; x < chunksX; ++x) {
         for (let y = 0; y < chunksY; ++y) {
           this.chunks_[lod].push({
```
aganders3, line 70: I guess I didn't catch this before. Do we need this to be a list-of-lists? The chunks themselves already have a `lod` property. I'm sure it's somewhat more efficient but feels slightly redundant.

packages/core/src/core/chunk_manager.ts
```diff
@@ -75,13 +83,54 @@ export class ChunkManagerSource {
...
+
+      if (Math.abs(rx - 2) > EPS || Math.abs(ry - 2) > EPS) {
+        console.error(
+          `Scale ratio between levels ${i - 1} and ${i} is (${rx}, ${ry}), expected (2.0, 2.0)`
+        );
+        throw new Error(
+          `Scales must be separated by factors of 2. Got ratio (${rx}, ${ry}) between scales ${prev} and ${curr}`
+        );
+      }
+    }
```
aganders3: I recommend factoring this out into a function, and doing the validation once in the constructor.

packages/core/src/core/chunk_manager.ts
```diff
@@ -75,13 +83,54 @@ export class ChunkManagerSource {
...
+
+    const numLods = availableScales.length;
+    const lodShift = numLods - 1;
+    const lodF = lodShift - Math.log2(1 / virtualUnitsPerScreenPixel);
+
+    const maxLod = numLods - 1;
+    const newLOD = Math.max(0, Math.min(maxLod, Math.floor(lodF)));
+
+    if (newLOD !== this.currentLOD_) {
+      console.log(`LOD changed from ${this.currentLOD_} to ${newLOD}`);
```
aganders3: Keep the log, but use the logging system and make it debug-level.

packages/core/src/core/chunk_manager.ts
```diff
@@ -75,13 +83,54 @@ export class ChunkManagerSource {
...
+    // Calculate virtual width from visible bounds
+    const virtualWidth = Math.abs(visibleBounds.max[0] - visibleBounds.min[0]);
+    const virtualUnitsPerScreenPixel = virtualWidth / bufferWidth;
+
+    const numLods = availableScales.length;
+    const lodShift = numLods - 1;
+    const lodF = lodShift - Math.log2(1 / virtualUnitsPerScreenPixel);
+
+    const maxLod = numLods - 1;
+    const newLOD = Math.max(0, Math.min(maxLod, Math.floor(lodF)));
```
aganders3:
```suggestion
    const maxLod = availableScales.length - 1;
    const lodF = maxLod - Math.log2(1 / virtualUnitsPerScreenPixel);
    const newLod = Math.max(0, Math.min(maxLod, Math.floor(lodF)));
```

nit: I don't feel like these extra variables increase readability. I'd simplify to this (or something like it) or use more informative names.

Also we should stick to a convention on `LOD` vs `Lod`.

packages/core/src/core/chunk_manager.ts
```diff
@@ -75,13 +83,54 @@ export class ChunkManagerSource {
...
+      console.log(`LOD changed from ${this.currentLOD_} to ${newLOD}`);
+      this.currentLOD_ = newLOD;
+    }
   }
 
   public async updateChunks(_: Bounds) {
     // TODO: map the LOD factor from the chunk manager to an available LOD in image space
     // TODO: replace the following block with loading based on intersection tests
-    for (const chunk of this.chunks_[curr_lod]) {
+    for (const chunk of this.chunks_[this.currentLOD_]) {
```
aganders3, line 132: Again it doesn't come from this PR but I think we're going to want to iterate _all_ chunks here anyway (to mark not visible, etc.), so I think `this.chunks_` could be a flat array. Then this becomes something like:

```
        const needsToBeLoaded = chunk.visible && chunk.state !== "loaded";
        if (needsToBeLoaded) {
           ...
```

packages/core/src/core/chunk_manager.ts
```diff
@@ -112,6 +161,7 @@ export class ChunkManager {
     // TODO: compute the LOD factor
 
     for (const [_, chunkManagerSource] of this.sources_) {
+      chunkManagerSource.computeLOD(visibleBounds, _bufferWidth);
```
aganders3: There was some discussion that LOD could be calculated by the `ChunkManager` itself, without requiring data about the individual source. Where did we land on that?

I guess here it at least needs to be per-source since different sources may have a different number of available scales.

cc @shlomnissan

packages/core/src/core/chunk_manager.ts
```diff
@@ -10,13 +10,14 @@ import { vec2, vec4, mat4 } from "gl-matrix";
 
 type Bounds = { min: vec2; max: vec2 };
 
-// temporary value. LOD will be computed dynamically
-const curr_lod = 1;
-
 export class ChunkManagerSource {
```
aganders3, line 14: We had some discussion about keeping a each file to a single class. I'm personally fine with closely related classes like this being in one file, but I wanted to call it out in case others feel more strongly about it.

packages/core/src/core/chunk_manager.ts
```diff
@@ -10,13 +10,14 @@ import { vec2, vec4, mat4 } from "gl-matrix";
...
   private readonly chunks_: ImageChunk[][] = [];
```
aganders3: I don't think this needs a default value (anymore?)

packages/core/src/core/chunk_manager.ts
```diff
@@ -25,27 +26,34 @@ export class ChunkManagerSource {
...
+    this.yIdx_ = region.findIndex(
+      (entry) => entry.dimension.toLocaleLowerCase() === "y"
+    );
+    if (this.xIdx_ === -1 || this.yIdx_ === -1) {
+      throw new Error("Missing required spatial axis x/y");
+    }
 
     // generate chunks for each LOD without loading data
-    this.chunks_ = Array(attrs.length)
+    this.chunks_ = Array(this.attrs_.length)
```
aganders3, line 52: nit: Consider moving this to a private method like `this.chunks_ = this.initializeEmptyChunks()` or something.

aganders3 (approved):
This is great and seems to be working well for me!

I am approving to keep things moving since I think this is well on the right path, but I left some optional feedback to address now or take note.

More generally I see a few artifacts at the edges of the renderables that flicker when zooming. I'm guessing these can be fixed by tweaking some rendering options? I don't think this is introduced by your PR, but I did not notice them before.
Screenshot 2025-06-20 at 11 31 44 AM

packages/core/src/core/chunk_manager.ts
shlomnissan: I typically follow the one-class-per-file principle, but in this case, I see the chunk manager source as the interface through which layers interact with the chunk manager—or perhaps more accurately, the chunk manager acts as a coordinator of these sources. Given that relationship, I think colocating the classes makes sense.

packages/core/src/core/chunk_manager.ts
shlomnissan: It’s a simple one-liner, so the extra indirection might be overkill but I don't feel strongly about it either way.

packages/core/src/core/chunk_manager.ts
```diff
@@ -25,27 +26,34 @@ export class ChunkManagerSource {
   ) {
     this.loader_ = loader;
     this.region_ = region;
+    this.attrs_ = attrs;
+    this.currentLOD_ = 1;
```
shlomnissan: This choice feels somewhat arbitrary to me—I initially used 1 as a placeholder without much reasoning. Now that we're computing the value dynamically, it might make more sense to start with the lowest resolution, or use 0, assuming the LOD is determined before we load any data.

packages/core/src/core/chunk_manager.ts
shlomnissan: Good point! I agree with your suggestion.

One potentially more robust alternative could be to convert `Region` into a class with a method or adding a global helper method for retrieving the index of a given axis from a region. This would reduce boilerplate and ensure consistent application of safety checks like `toLowerCase()` during comparisons.

That said, I think long term we might also want to reconsider the `Region` abstraction altogether. In the context of chunking, we likely don’t need to support intervals or points separately, which simplifies the model.

Something @andy-sweet mentioned that really resonated with me was the idea of thinking about what we currently call a region as a virtual camera view. Building on that, I’ve been considering redefining `Region` as a virtual camera object—one that exists in the same dimensional space as the source, allowing us to position the "camera" along each axis (essentially selecting a point in that space). We could then transform this virtual camera into a presentation-specific one, like an orthographic camera.

As I understand Andy’s point, the key insight is that every axis should be treated uniformly—instead of assigning special semantics to x/y, we should aim for a consistent, axis-agnostic approach.

packages/core/src/core/chunk_manager.ts
shlomnissan: Totally agree. This is borderline blocking for me—unnecessary computations shouldn’t be executed in a tight loop.

shlomnissan: Other than factoring this out to a function and only calling it once, I would consider a utility function too

```ts
function almostEqual(a: number, b: number, epsilon = 1e-6): boolean {
  return Math.abs(a - b) <= epsilon;
}
```

packages/core/src/core/chunk_manager.ts
```diff
@@ -75,13 +83,54 @@ export class ChunkManagerSource {
...
+      }
+    }
+
+    // Calculate virtual width from visible bounds
+    const virtualWidth = Math.abs(visibleBounds.max[0] - visibleBounds.min[0]);
+    const virtualUnitsPerScreenPixel = virtualWidth / bufferWidth;
+
+    const numLods = availableScales.length;
+    const lodShift = numLods - 1;
+    const lodF = lodShift - Math.log2(1 / virtualUnitsPerScreenPixel);
```
shlomnissan: This comment is related to @aganders3 previous comment https://github.com/chanzuckerberg/idetik/pull/281/files#r2159245981

I think this calculation can be computed once at the chunk manager level. Then, you can pass `lodFactor` down to the sources to determine the appropriate level based on the number of scales.

```ts
const virtualWidth = Math.abs(visibleBounds.max[0] - visibleBounds.min[0]);
const virtualUnitsPerScreenPixel = virtualWidth / bufferWidth;
const lodFactor = Math.log2(1 / virtualUnitsPerScreenPixel);
```

So maybe we have a `computeLODFactor` in the chunk manager, and a `setLOD(const lodFactor: number)` in the chunk manager source that applies offset and clamps.

I realize this might seem like a minor optimization, but in a tight loop, it’s worth avoiding redundant calculations wherever possible.

packages/core/src/core/chunk_manager.ts
```diff
@@ -75,13 +83,54 @@ export class ChunkManagerSource {
...
+    // Assert that scales are separated by factors of 2
+    const EPS = 1e-6;
+    for (let i = 1; i < availableScales.length; i++) {
+      const prev = availableScales[i - 1];
+      const curr = availableScales[i];
+      const rx = curr[this.xIdx_] / prev[this.xIdx_];
+      const ry = curr[this.yIdx_] / prev[this.yIdx_];
+
+      if (Math.abs(rx - 2) > EPS || Math.abs(ry - 2) > EPS) {
+        console.error(
```
shlomnissan: Use `logger.error` instead of an explicit console log if possible.

packages/core/src/core/chunk_manager.ts
shlomnissan: You're right—that's on me. Since we're already iterating over all chunks to determine visibility, there's no real need for nested arrays at this stage. As it stands, I think we don’t have a use case that requires iterating over chunks by LOD, so the added structure may be unnecessary for now.

packages/core/src/core/chunk_manager.ts
shlomnissan: Yes, the calculation can—and should—be split: compute the factor at the chunk manager level, and handle the offset and clamping at the source level, if possible. I’ve added a comment near the relevant code to clarify this.

packages/core/src/layers/image_layer.ts
```diff
@@ -29,20 +29,17 @@ export class ImageLayer extends Layer {
   private visibleChunks_: Map<ImageChunk, ImageRenderable> = new Map();
 
   // TODO:(shlomnissan) Remove this parameter when chunk manager is used by default
-  private readonly lod_?: number;
```
shlomnissan, line 32: I believe this member variable is still necessary until the chunk manager becomes the sole entry point for loading data. At the moment, we sometimes need to load an interval along a specific axis with a fixed LOD, rather than defaulting to the lowest resolution—which is the original reason I introduced this variable. I'm pretty sure it's a dependency to one of the examples, but I might be wrong.

packages/core/src/layers/image_layer.ts
```diff
@@ -132,7 +129,7 @@ export class ImageLayer extends Layer {
     this.setState("loading");
     const loader = await this.source_.open();
     const attributes = await loader.loadAttributes();
-    const lod = this.lod_ ?? attributes.length - 1;
+    const lod = this.chunkManagerSource_?.currentLOD_ ?? attributes.length - 1;
```
shlomnissan: I think this should be reverted because this function is only called when we are not using the chunk manager is not used.

packages/core/src/core/chunk_manager.ts
```diff
@@ -10,13 +10,14 @@ import { vec2, vec4, mat4 } from "gl-matrix";
...
 
-// temporary value. LOD will be computed dynamically
-const curr_lod = 1;
-
 export class ChunkManagerSource {
   private readonly chunks_: ImageChunk[][] = [];
   private readonly loader_;
   private readonly region_;
+  private readonly attrs_: LoaderAttributes[];
+  public currentLOD_: number = 0;
```
shlomnissan: I don't think this member variable should be accessed publicly.

shlomnissan (changes requested):
This change looks great overall—thanks for sticking with it! I'm sorry you had to rewrite it a couple of times; I’ll make sure to plan things more clearly on my end next time.

I'm blocking because I think there may have been a misunderstanding regarding the image layer changes—specifically around the removal of the `lod_` parameter. We may still need to support a fixed LOD option in scenarios where the chunk manager isn't used. Additionally, I think some of the concerns Ashley raised are valid. In particular, I'd like to see the unnecessary calculations addressed before this PR is merged.

shlomnissan:
> More generally I see a few artifacts at the edges of the renderables that flicker when zooming. I'm guessing these can be fixed by tweaking some rendering options? I don't think this is introduced by your PR, but I did not notice them before.

@aganders3 I think we can reduce some of the artifacts by disabling depth testing, which I believe is currently enabled by default? And ensure that we use "nearest filtering" for now.

But yeah, it is unrelated to this PR.

packages/core/src/core/chunk_manager.ts
aganders3: Sorry I meant not just this line creating the empty array, but initializing all the chunks without data.

shlomnissan: Ahh that makes more sense

packages/core/src/core/chunk_manager.ts
andy-sweet: Given the throw right after this has similar information, I would prefer just removing this log.

packages/core/src/core/chunk_manager.ts
```diff
@@ -75,13 +83,54 @@ export class ChunkManagerSource {
...
+  }
+
+  public computeLOD(
+    visibleBounds: Bounds,
+    bufferWidth: number // screen/canvas width in pixels
+  ): void {
+    const availableScales = this.attrs_.map((attr) => attr.scale);
+
+    // Assert that scales are separated by factors of 2
+    const EPS = 1e-6;
```
andy-sweet: Is there a strong reason to define our own (much looser) tolerance rather than `Number.EPSILON`?

packages/core/src/core/chunk_manager.ts
```diff
@@ -112,6 +161,7 @@ export class ChunkManager {
     // TODO: compute the LOD factor
```
andy-sweet:
```suggestion
```

andy-sweet (commented):
Love to see this coming in. I don't have too much to add on top of the existing comments, some of which need addressing before I would approved. Otherwise, I added some minor comments.

One of the examples is broken (image2d_from_omezarr5d_u8), but it's broken on main too in a slightly different way, so need to fix here. I can follow up with a fix after this goes in, or maybe even before.

andy-sweet:
> One of the examples is broken (image2d_from_omezarr5d_u8), but it's broken on main too in a slightly different way, so need to fix here. I can follow up with a fix after this goes in, or maybe even before.

Actually, the fix is easier on this branch than on `main`, though I don't entirely understand why yet: https://github.com/chanzuckerberg/idetik/compare/phoenix-lod-take2...andy-sweet/fix-dims-lod?expand=1

Probably best to avoid polluting this PR with those changes, but feel free to bring in if desired.

packages/core/src/core/chunk_manager.ts
phoenixAja: good idea, moved to constructor and created almostEqual util function

packages/core/src/core/chunk_manager.ts
phoenixAja: changed to zero

packages/core/src/core/chunk_manager.ts
phoenixAja: done

packages/core/src/core/chunk_manager.ts
phoenixAja: done!

packages/core/src/layers/image_layer.ts
phoenixAja: reverted

packages/core/src/core/chunk_manager.ts
```diff
@@ -7,16 +7,19 @@ import {
...
+import { almostEqual } from "../utilities/almost_equal";
 
 type Bounds = { min: vec2; max: vec2 };
 
-// temporary value. LOD will be computed dynamically
-const curr_lod = 1;
-
 export class ChunkManagerSource {
-  private readonly chunks_: ImageChunk[][] = [];
+  private readonly chunks_: ImageChunk[][];
```
shlomnissan, line 15: @aganders3 made a good point about flattening this list, especially since we consistently iterate across LODs.

https://github.com/chanzuckerberg/idetik/pull/281#discussion_r2159209738

packages/core/src/core/chunk_manager.ts
```diff
@@ -25,27 +28,43 @@ export class ChunkManagerSource {
...
+    if (this.xIdx_ === -1 || this.yIdx_ === -1) {
+      throw new Error("Missing required spatial axis x/y");
+    }
+    let channelIdx = region.findIndex(
+      (entry) => entry.dimension.toLowerCase() === "c"
+    );
+    if (channelIdx === -1) {
+      channelIdx = 0;
+    }
+    this.validateScaleRatios(this.xIdx_, this.yIdx_);
```
shlomnissan: This line should likely be moved to line 43, before setting the channel index.

packages/core/src/core/chunk_manager.ts
```diff
@@ -61,27 +80,55 @@ export class ChunkManagerSource {
...
   public getVisibleChunks(): ImageChunk[] {
-    return this.chunks_[curr_lod].filter((e) => e.visible);
+    return this.chunks_[this.currentLOD_].filter((e) => e.visible);
+  }
+
+  public setLOD(lodFactor: number): void {
+    const numLods = this.attrs_.length;
+    const lodShift = numLods - 1;
+    const lodF = lodShift - lodFactor;
+    const newLOD = Math.max(0, Math.min(numLods - 1, Math.floor(lodF)));
```
shlomnissan:
```suggestion
    const maxLOD = this.attrs_.length - 1;
    const targetLOD = Math.max(0, Math.min(maxLOD, Math.floor(maxLOD - lodFactor)));
```

packages/core/src/core/chunk_manager.ts
```diff
@@ -61,27 +80,55 @@ export class ChunkManagerSource {
...
+      }
+    }
+  }
+
   public getVisibleChunks(): ImageChunk[] {
-    return this.chunks_[curr_lod].filter((e) => e.visible);
+    return this.chunks_[this.currentLOD_].filter((e) => e.visible);
+  }
+
+  public setLOD(lodFactor: number): void {
```
shlomnissan, line 116: This should probably be a private method. You can pass `lodFactor` to the public `update` function and invoke `setLOD` internally. That way, all source updates in the chunk manager go through a single entry point.

shlomnissan (approved):
Looks great! I left a couple of comments for you to consider.

PR #282, opened 2025-06-20
refactor: core changes for react refactoring

aganders3:
### Summary
This PR refactors the core `Idetik` class to accept an `HTMLCanvasElement` directly instead of a canvas selector string, while maintaining backward compatibility during a transition period. This works better for React integration because the canvas element may not be available at the time of component initialization, so when using the context we don't have to rely as much on mount effects, forward refs, or prop drilling. I also think it's a more direct/intuitive interface in general.

This provides a cleaner, more direct API while maintaining backward compatibility and simplifies some of the work on React integration slated for a future PR.

- *Core API Change*: Modified `IdetikParams` to accept both `canvas?: HTMLCanvasElement` and
  `canvasSelector?: string` with validation to ensure exactly one is provided
  - **Note**: This change is temporarily backward compatible, because the current react context expects it and is hard to detangle.
- *Idetik Runtime*: Expose `canvas` as public readonly property on the runtime object so applications (CryoLens) can directly register event listeners on it. Also provide access to `clientToClip` through the runtime (implemented on the `Renderer` at the moment because it holds the canvas ref). This is also probably temporary and could be replaced by Layers registering interactions or adding "resources" that can register interactions.
- *WebGL Renderer*: Updated to consistently use `HTMLCanvasElement` internally
  - **Note - breaking change**: users should be using the `Idetik` runtime object anyway, but if not they will need to update to pass a canvas element instead of a selector to the `WebGLRenderer`
- *Examples Updated*: All core examples now use `document.querySelector<HTMLCanvasElement>()` pattern
- *React Integration*: Minimal changes to maintain compatibility - fixed circular dependency in hooks
- *Ultrack Updates*: Updated to use canvas element instead of selector string
- *Package Dependencies*: Updated React package to properly declare `@idetik/core` as peer dependency

**LayerManager Changes**
New Layer Management Methods:
- Added `remove(layer: Layer)` - removes a specific layer instance (safer than index-based removal, and more convenient when you have a layer reference in React)
  - **Note - breaking change**: existing usage of `remove` with the layer index need to be updated to pass the layer, or use `removeByIndex(index: number)` if you need the old behavior

React Integration Support:
- Added `addLayersChangeCallback`, `removeLayersChangeCallback` for subscribing to layer changes with `useSyncExternalStore`
- This follows the pattern established in `ImageSeriesLayer` for synchronizing react state with Idetik state

**Layer Type Property**

Added Abstract readonly `type` property. This is a small change, but it helps in a few ways and
follows the pattern we've already established in the codebase:
- Better debugging - you can easily identify layer types in dev tools
- Enables type-safe layer filtering and management in React components
- Supports future features like layer-specific UI controls
- Makes layer serialization/deserialization possible

### Testing
I tested the examples, the demo apps in `@idetik/react`, and the ultrack prototype
The refactor is tested by the CryoLens example as well, which is using my [WIP refactor branch](https://github.com/chanzuckerberg/idetik/tree/aganders3/react-refactor-ideas), which this code is extracted from.

### DIFF
--- a/packages/core/examples/image2d_from_omezarr4d_hcs/main.ts
+++ b/packages/core/examples/image2d_from_omezarr4d_hcs/main.ts
@@ -81,7 +81,7 @@ const onImageChange = async () => {
   });
   app.layerManager.add(newLayer);
   newLayer.addStateChangeCallback((state) => {
-    if (state === "ready" && newLayer?.extent) {
+    if (state === "ready" && newLayer.extent) {
       camera.setFrame(0, newLayer.extent.x, 0, newLayer.extent.y);
       camera.update();
       controls.panTarget = camera.position;
--- a/packages/core/examples/image2d_from_omezarr4d_hcs/main.ts
+++ b/packages/core/examples/image2d_from_omezarr4d_hcs/main.ts
@@ -41,8 +41,6 @@ const app = new Idetik({
   controls,
 }).start();
 
-let currentLayer: ImageLayer | undefined;
-
 const region: Region = [
   { dimension: "T", index: { type: "point", value: 0 } },
   { dimension: "C", index: { type: "interval", start: 1, stop: 3 } },
@@ -55,9 +53,7 @@ const imageSelector = document.querySelector("#image") as HTMLSelectElement;
 
 const onImageChange = async () => {
   console.debug("onImageChange: ", imageSelector.value);
-  if (currentLayer) {
-    app.layerManager.remove(currentLayer);
-  }
+  app.layerManager.removeAll();
   const imageUrl =
     plateUrl + "/" + wellSelector.value + "/" + imageSelector.value;
   const source = new OmeZarrImageSource(imageUrl);
@@ -74,7 +70,7 @@ const onImageChange = async () => {
     [omeroChannels[1].window.start, omeroChannels[1].window.end],
     [omeroChannels[2].window.start, omeroChannels[2].window.end],
   ];
-  currentLayer = new ImageLayer({
+  const newLayer = new ImageLayer({
     source,
     region,
     channelProps: [
@@ -83,10 +79,10 @@ const onImageChange = async () => {
     ],
     lod: 0,
   });
-  app.layerManager.add(currentLayer);
-  currentLayer.addStateChangeCallback((state) => {
-    if (state === "ready" && currentLayer?.extent) {
-      camera.setFrame(0, currentLayer.extent.x, 0, currentLayer.extent.y);
+  app.layerManager.add(newLayer);
+  newLayer.addStateChangeCallback((state) => {
+    if (state === "ready" && newLayer?.extent) {
+      camera.setFrame(0, newLayer.extent.x, 0, newLayer.extent.y);
       camera.update();
       controls.panTarget = camera.position;
     }
@@ -95,9 +91,7 @@ const onImageChange = async () => {
 
 const onWellChange = async () => {
   console.debug("onWellChange: ", wellSelector.value);
-  if (currentLayer) {
-    app.layerManager.remove(currentLayer);
-  }
+  app.layerManager.removeAll();
   const path = wellSelector.value;
   const well = await loadOmeZarrWell(plateUrl, path);
   console.debug("well", well);
--- a/packages/core/examples/image2d_from_omezarr4d_hcs/main.ts
+++ b/packages/core/examples/image2d_from_omezarr4d_hcs/main.ts
@@ -36,10 +36,12 @@ wellPaths.forEach((path) => {
 const camera = new OrthographicCamera(0, 840, 0, 360);
 const controls = new PanZoomControls(camera, camera.position);
 const app = new Idetik({
-  canvasSelector: "canvas",
+  canvas: document.querySelector<HTMLCanvasElement>("canvas")!,
   camera,
   controls,
 }).start();
+
+let currentLayer: ImageLayer | undefined;
 
 const region: Region = [
   { dimension: "T", index: { type: "point", value: 0 } },
@@ -53,7 +55,9 @@ const imageSelector = document.querySelector("#image") as HTMLSelectElement;
 
 const onImageChange = async () => {
   console.debug("onImageChange: ", imageSelector.value);
-  app.layerManager.removeAll();
+  if (currentLayer) {
+    app.layerManager.remove(currentLayer);
+  }
   const imageUrl =
     plateUrl + "/" + wellSelector.value + "/" + imageSelector.value;
   const source = new OmeZarrImageSource(imageUrl);
@@ -70,7 +74,7 @@ const onImageChange = async () => {
     [omeroChannels[1].window.start, omeroChannels[1].window.end],
     [omeroChannels[2].window.start, omeroChannels[2].window.end],
   ];
-  const layer = new ImageLayer({
+  currentLayer = new ImageLayer({
     source,
     region,
     channelProps: [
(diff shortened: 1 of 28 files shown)


packages/core/src/objects/cameras/camera.ts
```diff
@@ -5,7 +5,6 @@ export abstract class Camera extends RenderableObject {
   protected projectionMatrix_ = mat4.create();
   protected near_ = 0;
   protected far_ = 0;
-  protected zoom_ = 1;
```
aganders3, line 8: This was leftover and unused from previous camera refactoring

packages/core/src/core/layer_manager.ts
```diff
@@ -3,6 +3,8 @@ import { IdetikContext } from "../idetik";
 
 export class LayerManager {
   private readonly layers_: Layer[] = [];
+  private layersSnapshot_: Layer[] | null = null;
```
bchu1: could we remove this and just have the mutation methods create new arrays? could also make `private layers_: ReadonlyArray<Layer> = [];` to prevent accidental mutation

packages/core/src/idetik.ts
```diff
@@ -20,15 +22,30 @@ export type IdetikContext = {
...
+      throw new Error("Either canvas or canvasSelector must be provided");
+    }
+    if (params.canvas && params.canvasSelector) {
+      throw new Error("Cannot provide both canvas and canvasSelector");
+    }
+
     this.camera = params.camera;
-    this.renderer_ = new WebGLRenderer(params.canvasSelector);
+    const canvas =
+      params.canvas ||
```
bchu1: nit - `params.canvas ??`

packages/core/src/idetik.ts
```diff
@@ -62,6 +79,13 @@ export class Idetik {
     this.renderer_.setControls(controls);
   }
 
+  // TODO: this can be moved directly to this calss, but will need access to (and possibly expose)
```
bchu1: typo

packages/react/package.json
```diff
@@ -55,6 +55,7 @@
   },
   "peerDependencies": {
     "@czi-sds/components": "^22.3.0",
+    "@idetik/core": "^0.0.1",
```
bchu1, line 58: intentional?

packages/core/examples/image2d_from_omezarr4d_hcs/main.ts
```diff
@@ -36,11 +36,13 @@ wellPaths.forEach((path) => {
 const camera = new OrthographicCamera(0, 840, 0, 360);
 const controls = new PanZoomControls(camera, camera.position);
 const app = new Idetik({
-  canvasSelector: "canvas",
+  canvas: document.querySelector<HTMLCanvasElement>("canvas")!,
   camera,
   controls,
 }).start();
 
+let currentLayer: ImageLayer | undefined;
```
andy-sweet: Nit: using `currentLayer` with `LayerManager.remove` is slightly more complicated than before, and doesn't add any obvious benefit, so revert?

packages/core/examples/points/main.ts
```diff
@@ -214,8 +215,9 @@ const imageLayer = new ImageSeriesLayer({
 });
 
 const camera = new OrthographicCamera(0, 1024, 0, 1024, -10000, 10000);
+const canvas = document.querySelector<HTMLCanvasElement>("canvas")!;
```
andy-sweet: Nit: inline this as before and as with the other examples?

packages/core/examples/points/main.ts
```diff
@@ -57,6 +57,7 @@ const virusLikeLocations = await fetchPositionsFromNDJson(virusLikeUrl);
 type Marker = "circle" | "square" | "triangle";
 
 class Particles extends Layer {
+  public type = "Particles";
```
andy-sweet: Nit: I understand this is allowed by the typescript compiler (in the spirit of sub-classes being allowed to add more capability), but there's no reason to add capability here. 

```suggestion
  public readonly type = "Particles";
```

packages/core/examples/tracks/main.ts
```diff
@@ -116,8 +116,9 @@ const { xMin: left, xMax: right, yMin: top, yMax: bottom } = lineLayer.extent;
 const camera = new OrthographicCamera(left, right, top, bottom);
 camera.zoom(0.5);
 
+const canvas = document.querySelector<HTMLCanvasElement>("canvas")!;
```
andy-sweet: Nit: inline as before and in other examples?

packages/core/test/webgl_renderer.test.ts
```diff
@@ -4,18 +4,14 @@ import { WebGLRenderer } from "@";
 
 test("Instantiate WebGLRenderer", () => {
   document.body.innerHTML = '<canvas id="canvas"></canvas>';
-  const renderer = new WebGLRenderer("#canvas");
+  const canvas = document.querySelector<HTMLCanvasElement>("#canvas")!;
+  const renderer = new WebGLRenderer(canvas);
   expect(renderer).toBeDefined();
 });
 
-test("Instantiate WebGLRenderer with invalid selector", () => {
```
andy-sweet, line 11: From the PR description, I understood that we are maintaining backwards compatibility with the selector for now. Therefore, this test still has some purpose and I would only remove it when removing compatibility.

packages/core/test/layer.test.ts
```diff
@@ -3,6 +3,8 @@ import { expect, test, vi } from "vitest";
 import { Layer } from "@";
 
 class TestLayer extends Layer {
+  public type = "TestLayer";
```
andy-sweet: Nit

```suggestion
  public readonly type = "TestLayer";
```

packages/core/src/idetik.ts
```suggestion
  // TODO: this can be moved directly to this class, but will need access to (and possibly expose)
```

packages/core/src/core/layer_manager.ts
```diff
@@ -34,16 +36,58 @@ export class LayerManager {
     if (this.context_) {
       layer.onAttached(this.context_);
     }
+    this.notifyLayersChanged();
   }
-  public remove(index: number) {
+
+  public remove(layer: Layer) {
```
andy-sweet, line 41: This is a public API break. I'm fine making that break, but I would reword your mention of it in the PR description highlighting that it is a break and you should use the newly added `removeByIndex` instead of `remove` if you need the old behavior.

packages/core/src/core/layer_manager.ts
andy-sweet: In favor of this too, especially if nothing depends on the mutable behavior right now.

andy-sweet (approved):
No blocking feedback, though some updates to the PR description would be appreciated.

packages/core/src/core/layer_manager.ts
aganders3: Thanks! I was unaware of `ReadonlyArray` but I like this.

packages/core/test/webgl_renderer.test.ts
aganders3: Hm, it's a good point and makes me realize this is actually a breaking change. With this PR the renderer expects a canvas element (only) and it's the runtime that can accept either. Unfortunately there are no tests for the runtime object yet, but I can add some here and include a test for this.

packages/react/package.json
aganders3: Yes, this was causing some problems with using the `Color` class in the react library, because internally it uses `instanceof` but I think we were getting two copies of the class.

packages/core/src/core/layer_manager.ts
aganders3: Even better - we can remove `getLayersSnapshot` with this because the property getter will be returning a "snapshot".

bchu1 (approved):

packages/core/examples/image2d_from_omezarr4d_hcs/main.ts
```diff
@@ -79,10 +79,10 @@ const onImageChange = async () => {
     ],
     lod: 0,
   });
-  app.layerManager.add(layer);
-  layer.addStateChangeCallback((state) => {
-    if (state === "ready" && layer.extent) {
-      camera.setFrame(0, layer.extent.x, 0, layer.extent.y);
+  app.layerManager.add(newLayer);
+  newLayer.addStateChangeCallback((state) => {
+    if (state === "ready" && newLayer?.extent) {
```
andy-sweet: How can `newLayer` be undefined here?

```suggestion
    if (state === "ready" && newLayer.extent) {
```

andy-sweet (approved):
Still looks good to me other than a small nit where i requested changes

aganders3:
Thanks for the good reviews!

Issue #222, opened 2025-04-07
Enhance panzoom controls for mobile and trackpad use


aganders3:
Yes, I think it should either be configurable, or we should look into some convention. I just felt like the zooming was sluggish so I increased it a bit.

It also is a fixed number per wheel event, but I think we might be able to use the actual `deltaY` value to do something more natural feeling.

_Originally posted by @aganders3 in https://github.com/chanzuckerberg/imaging-active-learning/pull/220#discussion_r2031578275_

aganders3:
Also note zoom does not work on mobile - we need to map the panzoom controls to touch events as well.

PR #285, opened 2025-06-24
Fix pan-zoom controls when leaving the canvas

### LINKED ISSUE #284: Pan-zoom controls behave unexpectedly when leaving the canvas
Steps to reproduce on macOS 15.5, Chrome (37.0.7151.120 (Official Build) (arm64))

1. Run an example with pan-zoom controls (e.g. `image2d_from_omezarr5d_u16`).
2. Click and hold the left mouse button to start panning.
3. Move the mouse cursor outside of the canvas (e.g. by moving it out of the window).
4. Release the left mouse button.
5. Move the mouse cursor back onto the canvas.

After (5) you should notice that the controls continue to pan, whereas I would expect them not too because the button has been released.

A variant of this issue can be found in the react component, where there are elements that overlap with the canvas that absorb the mouse events in a way that is similar to leaving the canvas.

andy-sweet:
### Summary
In `PanZoomControls` this change now uses pointer events instead of mouse events and treats the `pointercancel` event in the same way as a `pointerup` event, which prevents the surprising and undesirable behavior described in #284. Another benefit of using pointer events is that we can capture and release the pointer event ID, which allows us to continue responding to pointer events when the pointer is down but strictly outside the target.

### Related Issue
Closes #284

### Tests & Checks
I manually tested an example with pan-zoom controls and the main React component.

### DIFF
--- a/packages/core/src/objects/cameras/controls.ts
+++ b/packages/core/src/objects/cameras/controls.ts
@@ -43,9 +43,9 @@ export class PanZoomControls implements CameraControls {
         (event: Event) => this.wheel(event as WheelEvent, clientToWorld),
       ],
       [
-        "mousedown",
+        "pointerdown",
         (event: Event) =>
-          this.mousedown(event as MouseEvent, target, clientToWorld),
+          this.pointerdown(event as PointerEvent, target, clientToWorld),
       ],
     ];
   }
@@ -66,17 +66,23 @@ export class PanZoomControls implements CameraControls {
     this.pan(deltaWorld);
   }
 
-  private mousedown(
-    event: MouseEvent,
+  private pointerdown(
+    event: PointerEvent,
     target: EventTarget,
     clientToWorld: ClientToWorld
   ) {
     const clientStart = vec2.fromValues(event.clientX, event.clientY);
     const worldStart = clientToWorld(clientStart, this.clipDepth);
+    const pointerId = event.pointerId;
 
-    const onMouseMove = (event: Event) => {
-      if (!(event instanceof MouseEvent)) {
-        throw new Error("Expected MouseEvent");
+    // capture the pointer to ensure we receive events even when outside the element
+    if (target instanceof Element) {
+      target.setPointerCapture(pointerId);
+    }
+
+    const onPointerMove = (event: Event) => {
+      if (!(event instanceof PointerEvent) || event.pointerId !== pointerId) {
+        return;
       }
       const clientPos = vec2.fromValues(event.clientX, event.clientY);
       const worldPos = clientToWorld(clientPos, this.clipDepth);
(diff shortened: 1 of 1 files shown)


andy-sweet:
@aganders3 : I assigned you as a reviewer because you originally wrote these controls. Other comments and reviews are welcome.

aganders3:
Thanks for calling this out and starting a fix! This change does feel like an improvement, but I am concerned about the issue with overlapping canvas elements.

As an alternative I found [`setPointerCapture`](https://developer.mozilla.org/en-US/docs/Web/API/Element/setPointerCapture), which I think may be more appropriate here. This allows the canvas to continue responding to events after mousing out of the canvas. So for example you can keep panning the image when your mouse moves out as long as you keep holding down the button.

I'm not sure I fully understand how to use the API but the behavior with these changes feels natural to me. I pushed to a separate branch based on this one: https://github.com/chanzuckerberg/idetik/compare/andy-sweet/fix-controls...aganders3/fix-controls

What do you think?

andy-sweet:
> I pushed to a separate branch based on this one: [andy-sweet/fix-controls...aganders3/fix-controls](https://github.com/chanzuckerberg/idetik/compare/andy-sweet/fix-controls...aganders3/fix-controls)

It's generous to say that's based on this branch, but I'll take it!

> What do you think?

I like the behavior. It implements a similar idea I had in the PR description, but did not investigate further:

> To avoid this side-effect, we could consider capturing some information on leaving that we could use on entering, but this feels fragile, so I avoided it for now.

I also think using pointer events instead of mouse events might have some advantages when it comes to supporting other input devices, so let's go with your approach.

Do you want to create a new PR or merge your branch in here?

aganders3:
I'm happy to just push it here! It's based on your branch in that I pulled your branch to test, and edited it. 😅

> I also think using pointer events instead of mouse events might have some advantages when it comes to supporting other input devices

I hope so - it would be great to add "pinch to zoom" soon! Of course we have an issue for this in #222 but I've done basically zero research on it.

aganders3 (approved):
I guess now I'm not sure how valuable a review from me is. This seems like only an improvement to me though.

packages/core/src/objects/cameras/controls.ts
```diff
@@ -66,31 +66,46 @@ export class PanZoomControls implements CameraControls {
...
     const worldStart = clientToWorld(clientStart, this.clipDepth);
+    const pointerId = event.pointerId;
 
-    const onMouseMove = (event: Event) => {
-      if (!(event instanceof MouseEvent)) {
-        throw new Error("Expected MouseEvent");
+    // capture the pointer to ensure we receive events even when outside the element
+    if (target instanceof Element) {
+      target.setPointerCapture(pointerId);
+    }
```
aganders3, line 81: Maybe some of these guards can be removed with stricter typing in the interfaces?

andy-sweet: TIL that things other than elements can be event targets though I guess I have used the document/window in this way before. I think this guard is fine for now, though I might add a warning given this is unexpected.

packages/core/src/objects/cameras/controls.ts
```diff
@@ -66,31 +66,46 @@ export class PanZoomControls implements CameraControls {
...
+
+    const onPointerMove = (event: Event) => {
+      if (!(event instanceof PointerEvent) || event.pointerId !== pointerId) {
```
andy-sweet, line 84: Do you know a way to make the second part of the condition true in this usage (i.e. where we remove this callback on pointerup/cancel)? Maybe this could happen with two different input devices?

andy-sweet:
> I guess now I'm not sure how valuable a review from me is. This seems like only an improvement to me though.

Well, you're basically the author now, and I'm the reviewer. So I think it's fine. I'll consider adding some logs and afterwards will merge this.

packages/core/src/objects/cameras/controls.ts
aganders3: Yes or I was thinking of multi-touch, where I think each finger gets an ID as it appears in a `pointerdown` event.

andy-sweet: Ahh, another TIL. Thanks for clarifying.

PR #286, opened 2025-06-24
refactor: flatten LOD chunk structure

phoenixAja:
## Refactor: Flatten LOD chunk structure

This PR simplifies the internal `chunks_` data structure in `ChunkManagerSource` by flattening the list-of-lists into a single `ImageChunk[]`. Previously, chunks were stored as a nested `ImageChunk[][]` keyed by LOD index, but this is redundant since each `ImageChunk` already contains a `lod` property.

No behavioral changes; this is a pure internal cleanup.


shlomnissan (approved):
:+1:

PR #287, opened 2025-06-24
feat: Load only visible chunks based on view bounds

phoenixAja:
## Load only visible chunks based on view bounds

This PR introduces visibility culling to the chunk loading system. Instead of loading all chunks at the current LOD, we now compute which chunks intersect the visible view bounds and only load those.

###  Features

- **New `getVisibleChunks()` API**: Returns only chunks that are both `visible` and `loaded`, ready for rendering.
- **View-aware chunk loading**: `updateChunks()` now computes visibility using screen-projected bounds and loads only chunks that are in-view and currently `"unloaded"`.
- **Visibility invalidation optimization**: Adds `hasVisibleBoundsChanged()` to avoid recomputing visibility if the camera hasn't moved.
- **Error handling**: Gracefully catches chunk loading errors and resets chunk state to `"unloaded"`.

### Benefits

- Reduces memory and network load by only loading chunks the user can actually see.
- Avoids repeated visibility computation when the view is unchanged.

### Tests: 
verified chunk_streaming example is only loading the chunks within the current viewport

### DIFF
--- a/packages/core/src/core/chunk_manager.ts
+++ b/packages/core/src/core/chunk_manager.ts
@@ -131,8 +131,6 @@ export class ChunkManagerSource {
   }
 
   public async updateChunks(visibleBounds: Bounds) {
-    // TODO: map the LOD factor from the chunk manager to an available LOD in image space
-    // TODO: replace the following block with loading based on intersection tests
     if (this.hasVisibleBoundsChanged(visibleBounds)) {
       this.computeVisibleChunks(visibleBounds);
       this.lastVisibleBounds_ = {
--- a/packages/core/src/core/chunk_manager.ts
+++ b/packages/core/src/core/chunk_manager.ts
@@ -69,7 +69,7 @@ export class ChunkManagerSource {
           this.chunks_.push({
             state: "unloaded",
             lod,
-            visible: true, // TODO:(shlomnissan) should be set to false
+            visible: false,
             shape: {
               x: chunkWidth,
               y: chunkHeight,
--- a/packages/core/src/core/chunk_manager.ts
+++ b/packages/core/src/core/chunk_manager.ts
@@ -110,7 +110,9 @@ export class ChunkManagerSource {
   }
 
   public getVisibleChunks(): ImageChunk[] {
-    return this.chunks_[this.currentLOD_].filter((e) => e.visible);
+    const visibleChunks = this.chunks_[this.currentLOD_].filter((e) => e.visible);
+    const visibleAndReady = visibleChunks.filter((e) => e.state === "loaded");
+    return visibleAndReady;
   }
 
   public setLOD(lodFactor: number): void {
@@ -126,15 +128,59 @@ export class ChunkManagerSource {
     }
   }
 
-  public async updateChunks(_: Bounds) {
-    // TODO: map the LOD factor from the chunk manager to an available LOD in image space
-    // TODO: replace the following block with loading based on intersection tests
+  public async updateChunks(visibleBounds: Bounds) {
+    this.computeVisibleChunks(visibleBounds);
+
     for (const chunk of this.chunks_[this.currentLOD_]) {
-      if (chunk.state === "unloaded") {
+      if (chunk.visible && chunk.state === "unloaded") {
         chunk.state = "loading";
         this.loader_.loadChunkDataFromRegion(chunk, this.region_).then(() => {
           chunk.state = "loaded";
+        }).catch((error) => {
+          console.error(`Error loading chunk (${chunk.chunkIndex?.x},${chunk.chunkIndex?.y}): ${error}`);
+          chunk.state = "unloaded";
         });
+      }
+    }
+  }
+
+  private computeVisibleChunks(visibleBounds: Bounds): void {
+    const currentChunks = this.chunks_[this.currentLOD_];
+    if (currentChunks.length === 0) return;
+
+    // Get chunk size in virtual coordinates for current LOD
+    const firstChunk = currentChunks[0];
+    const chunkVirtualWidth = firstChunk.shape.x * firstChunk.scale.x;
+    const chunkVirtualHeight = firstChunk.shape.y * firstChunk.scale.y;
+
+    // Compute chunk index range using analytical approach
+    const chunkIndexX1 = Math.floor(visibleBounds.min[0] / chunkVirtualWidth);
+    const chunkIndexX2 = Math.floor(visibleBounds.max[0] / chunkVirtualWidth);
+    const chunkIndexY1 = Math.floor(visibleBounds.min[1] / chunkVirtualHeight);
+    const chunkIndexY2 = Math.floor(visibleBounds.max[1] / chunkVirtualHeight);
+
+    // Ensure min/max are in correct order
+    const minChunkIndexX = Math.min(chunkIndexX1, chunkIndexX2);
+    const maxChunkIndexX = Math.max(chunkIndexX1, chunkIndexX2);
+    const minChunkIndexY = Math.min(chunkIndexY1, chunkIndexY2);
+    const maxChunkIndexY = Math.max(chunkIndexY1, chunkIndexY2);
+
+    // Reset all chunks to not visible first
+    for (const chunk of currentChunks) {
+      chunk.visible = false;
+    }
+
+    // Set visible chunks based on index range
+    for (const chunk of currentChunks) {
+      if (!chunk.chunkIndex) continue;
+      const { x, y } = chunk.chunkIndex;
+      if (
+        x >= minChunkIndexX &&
+        x <= maxChunkIndexX &&
+        y >= minChunkIndexY &&
+        y <= maxChunkIndexY
+      ) {
+        chunk.visible = true;
       }
     }
   }
--- a/packages/core/src/core/chunk_manager.ts
+++ b/packages/core/src/core/chunk_manager.ts
@@ -157,41 +157,35 @@ export class ChunkManagerSource {
 
   private updateChunkVisibility(visibleBounds: Bounds): void {
     if (this.chunks_.length === 0) return;
-    const firstChunkAtLOD = this.chunks_.find(
-      (chunk) => chunk.lod === this.currentLOD_
-    );
-    if (!firstChunkAtLOD) return;
-    const chunkVirtualWidth = firstChunkAtLOD.shape.x * firstChunkAtLOD.scale.x;
-    const chunkVirtualHeight =
-      firstChunkAtLOD.shape.y * firstChunkAtLOD.scale.y;
 
-    const minChunkIndexX = Math.floor(visibleBounds.min[0] / chunkVirtualWidth);
-    const maxChunkIndexX = Math.ceil(visibleBounds.max[0] / chunkVirtualWidth);
-    const minChunkIndexY = Math.floor(
-      visibleBounds.min[1] / chunkVirtualHeight
-    );
-    const maxChunkIndexY = Math.ceil(visibleBounds.max[1] / chunkVirtualHeight);
-
-    // Reset visibility and set visible chunks based on index range in a single pass
     for (const chunk of this.chunks_) {
-      // exit early if chunks are not at the current LOD
-      if (chunk.lod !== this.currentLOD_) {
+      if (!chunk.chunkIndex) {
         chunk.visible = false;
         continue;
       }
 
-      chunk.visible = false;
-      if (chunk.chunkIndex) {
-        const { x, y } = chunk.chunkIndex;
-        if (
-          x >= minChunkIndexX &&
-          x <= maxChunkIndexX &&
-          y >= minChunkIndexY &&
-          y <= maxChunkIndexY
-        ) {
-          chunk.visible = true;
-        }
-      }
+      const chunkVirtualWidth = chunk.shape.x * chunk.scale.x;
+      const chunkVirtualHeight = chunk.shape.y * chunk.scale.y;
+
+      const minChunkIndexX = Math.floor(
+        visibleBounds.min[0] / chunkVirtualWidth
+      );
+      const maxChunkIndexX = Math.ceil(
+        visibleBounds.max[0] / chunkVirtualWidth
+      );
+      const minChunkIndexY = Math.floor(
+        visibleBounds.min[1] / chunkVirtualHeight
+      );
+      const maxChunkIndexY = Math.ceil(
+        visibleBounds.max[1] / chunkVirtualHeight
+      );
+
+      const { x, y } = chunk.chunkIndex;
+      chunk.visible =
+        x >= minChunkIndexX &&
+        x <= maxChunkIndexX &&
+        y >= minChunkIndexY &&
+        y <= maxChunkIndexY;
     }
   }
 
--- a/packages/core/src/core/chunk_manager.ts
+++ b/packages/core/src/core/chunk_manager.ts
@@ -173,22 +173,19 @@ export class ChunkManagerSource {
     const minChunkIndexY = Math.min(chunkIndexY1, chunkIndexY2);
     const maxChunkIndexY = Math.max(chunkIndexY1, chunkIndexY2);
 
-    // Reset all chunks to not visible first
+    // Reset visibility and set visible chunks based on index range in a single pass
     for (const chunk of this.chunks_) {
       chunk.visible = false;
-    }
-
-    // Set visible chunks based on index range
-    for (const chunk of this.chunks_) {
-      if (!chunk.chunkIndex) continue;
-      const { x, y } = chunk.chunkIndex;
-      if (
-        x >= minChunkIndexX &&
-        x <= maxChunkIndexX &&
-        y >= minChunkIndexY &&
-        y <= maxChunkIndexY
-      ) {
-        chunk.visible = true;
+      if (chunk.chunkIndex) {
+        const { x, y } = chunk.chunkIndex;
+        if (
+          x >= minChunkIndexX &&
+          x <= maxChunkIndexX &&
+          y >= minChunkIndexY &&
+          y <= maxChunkIndexY
+        ) {
+          chunk.visible = true;
+        }
       }
     }
   }
--- a/packages/core/src/core/chunk_manager.ts
+++ b/packages/core/src/core/chunk_manager.ts
@@ -8,6 +8,7 @@ import { Region } from "../data/region";
 import { Camera } from "../objects/cameras/camera";
 import { vec2, vec4, mat4 } from "gl-matrix";
 import { almostEqual } from "../utilities/almost_equal";
+import { Logger } from "../utilities/logger";
 
 type Bounds = { min: vec2; max: vec2 };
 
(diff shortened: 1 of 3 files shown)


packages/core/src/core/chunk_manager.ts
```diff
@@ -126,18 +130,85 @@ export class ChunkManagerSource {
...
+
+    // Set visible chunks based on index range
+    for (const chunk of this.chunks_) {
+      if (!chunk.chunkIndex) continue;
+      const { x, y } = chunk.chunkIndex;
+      if (
+        x >= minChunkIndexX &&
+        x <= maxChunkIndexX &&
+        y >= minChunkIndexY &&
+        y <= maxChunkIndexY
```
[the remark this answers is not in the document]
phoenixAja: i think this is safe to assume

packages/core/src/core/chunk_manager.ts
```diff
@@ -126,18 +130,82 @@ export class ChunkManagerSource {
...
+      }
+    }
+  }
+
+  private hasVisibleBoundsChanged(visibleBounds: Bounds): boolean {
+    if (!this.lastVisibleBounds_) {
+      return true;
+    }
+
+    const EPSILON = 1e-6;
```
andy-sweet: Given that `almostEquals` was defined in the last PR and is already imported and used in this module, use it here too?

packages/core/src/core/chunk_manager.ts
```diff
@@ -126,18 +130,82 @@ export class ChunkManagerSource {
...
+
+  private computeVisibleChunks(visibleBounds: Bounds): void {
+    if (this.chunks_.length === 0) return;
+
+    const firstChunk = this.chunks_[0];
+    const chunkVirtualWidth = firstChunk.shape.x * firstChunk.scale.x;
+    const chunkVirtualHeight = firstChunk.shape.y * firstChunk.scale.y;
+
+    const chunkIndexX1 = Math.floor(visibleBounds.min[0] / chunkVirtualWidth);
+    const chunkIndexX2 = Math.floor(visibleBounds.max[0] / chunkVirtualWidth);
```
andy-sweet: I would expect to see `ceil` instead of `floor` with the max values / upper bounds. I did a little digging and found that `visibleBounds.min[1]` was greater than `visibleBounds.max[1]` which was unexpected. I think you may need to change the normalized coordinates defined for topLeft and bottomRight in `computeVisibleBounds` to be (-1, -1) and (1, 1) respectively?

If that expectation is correct, then I think you could also remove the calls to min/max below since we know which ones are bigger/smaller.

packages/core/src/core/chunk_manager.ts
```diff
@@ -126,18 +130,82 @@ export class ChunkManagerSource {
...
+            chunk.state = "unloaded";
+          });
       }
     }
   }
+
+  private computeVisibleChunks(visibleBounds: Bounds): void {
+    if (this.chunks_.length === 0) return;
+
+    const firstChunk = this.chunks_[0];
```
andy-sweet: Don't you need the first chunk of the current LOD? While the shapes may often be the same, I assume the scales are different.

andy-sweet (changes requested):
Mostly looks as expected, but I think I spotted a few issues that may be causing the chunks to not be marked as visible when they should be.

https://github.com/user-attachments/assets/555b3100-284b-4b94-bf86-2886e431efd6

I made an attempt to address some of my comments here which might fix that issue, but I'm not certain: https://github.com/chanzuckerberg/idetik/compare/phoenix/visible-bounds...andy-sweet/fix-bounds?expand=1

packages/core/src/core/chunk_manager.ts
phoenixAja: you're right, i forgot i did that lol

packages/core/src/core/chunk_manager.ts
andy-sweet: Or maybe you need to define the virtual width and height per LOD and move some of this into the loop below in an efficient way? I'm not sure if the visibility of chunks not in the current LOD needs to be reflected (or if they could just be false).

packages/core/src/core/chunk_manager.ts
```diff
@@ -126,18 +130,82 @@ export class ChunkManagerSource {
...
+    const chunkVirtualWidth = firstChunk.shape.x * firstChunk.scale.x;
+    const chunkVirtualHeight = firstChunk.shape.y * firstChunk.scale.y;
+
+    const chunkIndexX1 = Math.floor(visibleBounds.min[0] / chunkVirtualWidth);
```
andy-sweet: This calculation may also need to take into account the `offset` of each chunk because there may be a `translation` in the metadata. That is not properly accounted for when initializing the chunks, so it does not need to happen in this PR though.

packages/core/src/core/chunk_manager.ts
```diff
@@ -126,18 +130,82 @@ export class ChunkManagerSource {
...
+            console.error(
+              `Error loading chunk (${chunk.chunkIndex?.x},${chunk.chunkIndex?.y}): ${error}`
+            );
+            chunk.state = "unloaded";
+          });
       }
     }
   }
+
+  private computeVisibleChunks(visibleBounds: Bounds): void {
```
andy-sweet: As this has the side-effect of updating the visibility of the chunks, it would be better to give it a name that indicates that (e.g. `updateVisibleChunks` or `updateChunkVisibility`).

packages/core/src/core/chunk_manager.ts
phoenixAja: can you elaborate? Not sure i'm following

andy-sweet: The OME-NGFF allows a translation transform for each multi-scale level (and maybe an overall one too?), which among other things is sometimes used to maintain a meaningful origin across processed images (e.g. after cropping).

We currently capture it [as an image attribute](https://github.com/chanzuckerberg/idetik/blob/2459e46ddb9b3868b01a7624ee0b19ef3cadd431/packages/core/src/data/ome_zarr_image_loader.ts#L21) and use it in the old style of loading, but the chunks in the chunk manager currently ignore it.

For that reason and because it's not common, this is definitely one of those times to make an issue rather than fix in this PR.

packages/core/src/core/chunk_manager.ts
```diff
@@ -126,20 +130,82 @@ export class ChunkManagerSource {
...
+    if (this.hasVisibleBoundsChanged(visibleBounds)) {
+      this.computeVisibleChunks(visibleBounds);
+      this.lastVisibleBounds_ = {
+        min: vec2.clone(visibleBounds.min),
+        max: vec2.clone(visibleBounds.max),
+      };
+    }
     for (const chunk of this.chunks_) {
-      if (chunk.state === "unloaded") {
+      if (chunk.visible && chunk.state === "unloaded") {
```
andy-sweet: I can't comment on the actual line because GitHub, but should the TODO on line 72 be fixed by initializing visible to `false`?

phoenixAja: removed

packages/core/src/core/chunk_manager.ts
```diff
@@ -109,7 +110,10 @@ export class ChunkManagerSource {
 
   public getVisibleChunks(): ImageChunk[] {
```
andy-sweet: I think that visible is a bit overloaded in this module. This method's definition of visible (from its name getVisibleChunks) is "should this chunk be rendered" whereas `chunk.visible` seems to be "is this chunk in the visible bounds".

If we need to distinguish between those, they should be distinguished by name - here are not great naming ideas: `inVisibleBounds`, `toBeRendered`, `ready`.

packages/core/src/core/chunk_manager.ts
```diff
@@ -126,18 +130,79 @@ export class ChunkManagerSource {
...
+    const minChunkIndexY = Math.floor(
+      visibleBounds.min[1] / chunkVirtualHeight
+    );
+    const maxChunkIndexY = Math.ceil(visibleBounds.max[1] / chunkVirtualHeight);
+
+    // Reset visibility and set visible chunks based on index range in a single pass
+    for (const chunk of this.chunks_) {
+      chunk.visible = false;
+      if (chunk.chunkIndex) {
+        const { x, y } = chunk.chunkIndex;
```
andy-sweet: I think this will set `chunk.visible` incorrectly for chunks not at the current LOD because this `chunkIndex` is defined with the chunks of each LOD (i.e. they are not comparable across LODs, even though the virtual width/height are).

To fix this you could set `chunk.visible` to be false for chunks not at the current LOD, though I'm not sure if this is the intended behavior (i.e. i'm not sure exactly what visible is intended to mean here).

If visible is intended to mean inVisibleBounds (regardless of the current LOD), you could find the min/max chunk indices for each LOD and use that to check those bounds.

andy-sweet: Another related factor to consider here is that `setLOD` is currently public. If the LOD should determine visibility of chunks, that I think that public call may need call this function to update chunk visibility.

Currently, `update` always calls `setLOD` then `updateChunks`, so I think an LOD manually set outside of the chunk manager will effectively be overwritten. Therefore you should probably just make `setLOD` private?

andy-sweet (commented):
This is working for me now, but I think you should wait for @shlomnissan to take a look and approve this before merging in case some of my thinking is not quite right here. I have also have a few remaining comments and questions that should probably be resolved.

andy-sweet:
Another random comment that's not specific to this PR, but I think that the chunk manager is worthy of some decent test coverage. It's a critical part of Idetik and its design + implementation is non-trivial enough where tests become useful, especially as we improve and optimize it.

packages/core/src/core/chunk_manager.ts
```diff
@@ -109,7 +110,10 @@ export class ChunkManagerSource {
 
   public getVisibleChunks(): ImageChunk[] {
     return this.chunks_.filter(
-      (chunk) => chunk.lod === this.currentLOD_ && chunk.visible
+      (chunk) =>
+        chunk.lod === this.currentLOD_ &&
+        chunk.visible &&
+        chunk.state === "loaded"
```
shlomnissan, line 101: I think this is a great fit for this PR. As a quick follow-up, we should add logic to ensure there are no visual gaps in the image. One approach would be to always return the lowest LOD chunks, and avoid hiding the previous LOD until the current one is fully loaded. I think that's what the POC is doing.

packages/core/src/core/chunk_manager.ts
shlomnissan: I agree—the name “visibleChunks” doesn’t clearly reflect what the method actually returns. I’m fine with keeping it simple, like `getChunks()`. While it doesn’t add much clarity, it also avoids making assumptions about the nature of the chunks, which helps hide implementation details and gives us more flexibility going forward.

packages/core/src/core/chunk_manager.ts
phoenixAja: thanks! i made a new method that encapsulates both and made setLOD private.

packages/core/src/core/chunk_manager.ts
phoenixAja: i renamed some of the methods, let me know your thoughts!

phoenixAja: i wont resolve until you agree if it's clearer now

packages/core/src/core/chunk_manager.ts
```diff
@@ -91,29 +92,65 @@ export class ChunkManagerSource {
...
-          `Scales must be separated by factors of 2. Got ratio (${rx}, ${ry}) between scales ${prev} and ${curr}`
-        );
+  /**
+   * Loads any currently visible chunks that are not yet loaded.
+   * Recomputes visibility if the visible bounds have changed since the last update.
+   *
+   * @param visibleBounds - The virtual-space bounding box of the current camera view.
+   */
+  public loadVisibleChunks(visibleBounds: Bounds) {
+    if (this.hasVisibleBoundsChanged(visibleBounds)) {
```
shlomnissan: 1. Consider moving the visible bounds change check (`hasVisibleBoundsChanged`) to the call site. This would make `loadVisibleChunks` solely responsible for loading chunks, while leaving visibility update decisions to the caller.
2. Rather than calling `hasVisibleBoundsChanged`, could we just compare the bounds directly? You already create a deep copy, so I think the following should work:

```ts
if (visibleBounds !== lastVisibleBounds_) {
  // ...
}
```

It's not clear to me that there's benefit in using `almostEqual` on each axes.

phoenixAja:
> is worthy of some decent test coverage. It's a critical part of Idetik and its design + implementation is non-trivial enough where tests become useful, especially as we improve and optimize it.

made an issue to capture this [here](https://github.com/chanzuckerberg/idetik/issues/291)

packages/core/src/core/chunk_manager.ts
```diff
@@ -91,29 +92,65 @@ export class ChunkManagerSource {
...
+      this.updateChunkVisibility(visibleBounds);
+      this.lastVisibleBounds_ = {
+        min: vec2.clone(visibleBounds.min),
+        max: vec2.clone(visibleBounds.max),
+      };
+    }
+    for (const chunk of this.chunks_) {
```
shlomnissan: I’d suggest extracting this loop into a separate private method (e.g., `processChunkData`). This would make the code easier to extend in the future—for example, if we want to add support for unloading data or handle additional chunk states.

packages/core/src/core/chunk_manager.ts
```diff
@@ -91,29 +92,65 @@ export class ChunkManagerSource {
...
+        chunk.state = "loading";
+        this.loader_
+          .loadChunkDataFromRegion(chunk, this.region_)
+          .then(() => {
+            chunk.state = "loaded";
+          })
+          .catch((error) => {
+            console.error(
+              `Error loading chunk (${chunk.chunkIndex?.x},${chunk.chunkIndex?.y}): ${error}`
+            );
```
shlomnissan: Use `Logger.error()` instead.

packages/core/src/core/chunk_manager.ts
```diff
@@ -91,29 +92,65 @@ export class ChunkManagerSource {
...
+            chunk.state = "unloaded";
```
shlomnissan: I think this is fine for now, but we may want to introduce a "failed to load" state in the near future.

packages/core/src/core/chunk_manager.ts
phoenixAja: @shlomnissan what do you think?

packages/core/src/core/chunk_manager.ts
```diff
@@ -91,29 +92,65 @@ export class ChunkManagerSource {
...
-      (chunk) => chunk.lod === this.currentLOD_ && chunk.visible
-    );
+  /**
+   * Updates the current level-of-detail (LOD) based on a continuous LOD factor,
+   * and recalculates chunk visibility for the given view bounds.
+   *
+   * @param lodFactor - A continuous value used to select the LOD level (higher = coarser).
+   * @param visibleBounds - The virtual-space bounds used to determine chunk visibility.
+   */
+  public refreshViewState(lodFactor: number, visibleBounds: Bounds) {
```
shlomnissan: 1. I’m open to standardizing comments, but I think it’s important that they be consistent and thoughtfully applied. I’m generally not a fan of verbose documentation comments that describe every step the function performs—this tends to introduce a form of coupling where changes to the code often lead to outdated comments, which can be more harmful than having no comments at all. In this case, the function is already well-structured and readable through its method calls, so I’d suggest removing all the documentation comments from this file for now.

2. I’d also recommend renaming this function to `update()` for consistency. The current name, `refreshViewState`, feels a bit unclear and potentially misleading. `update()` aligns better with common naming conventions and is more intuitive (at least to me) in this context.

packages/core/src/core/chunk_manager.ts
```diff
@@ -91,29 +92,65 @@ export class ChunkManagerSource {
...
+    this.setLOD(lodFactor);
+    this.updateChunkVisibility(visibleBounds);
+    return this.loadVisibleChunks(visibleBounds);
```
shlomnissan: Why are we returning `this.loadVisibleChunks(visibleBounds);` here? I think this function returns no value.

packages/core/src/core/chunk_manager.ts
```diff
@@ -126,18 +163,74 @@ export class ChunkManagerSource {
     }
   }
 
-  public async updateChunks(_: Bounds) {
-    // TODO: map the LOD factor from the chunk manager to an available LOD in image space
-    // TODO: replace the following block with loading based on intersection tests
+  private updateChunkVisibility(visibleBounds: Bounds): void {
+    if (this.chunks_.length === 0) return;
```
shlomnissan: This might be unnecessary? I don’t really consider “no chunks” to be an error state (though maybe that’s up for debate). That said, I don’t expect it to be a common case either. If there are no chunks, `firstChunkAtLOD` will be falsy, so the function will exit early regardless. I would probably add a `Logger.warn()` if that's the case.

packages/core/src/core/chunk_manager.ts
```diff
@@ -126,18 +163,74 @@ export class ChunkManagerSource {
...
+    const firstChunkAtLOD = this.chunks_.find(
```
shlomnissan: I guess there was value in having a list of lists to store chunks?

packages/core/src/core/chunk_manager.ts
```diff
@@ -126,18 +163,74 @@ export class ChunkManagerSource {
...
+      (chunk) => chunk.lod === this.currentLOD_
```
shlomnissan: I believe we should be computing visibility for all chunks, not just those at the current LOD. This is important for supporting features like data unloading and for continuing to rely on chunks from previous LODs while new ones are still loading.

packages/core/src/core/chunk_manager.ts
```diff
@@ -126,18 +163,74 @@ export class ChunkManagerSource {
...
+    );
+    if (!firstChunkAtLOD) return;
+    const chunkVirtualWidth = firstChunkAtLOD.shape.x * firstChunkAtLOD.scale.x;
+    const chunkVirtualHeight =
+      firstChunkAtLOD.shape.y * firstChunkAtLOD.scale.y;
```
shlomnissan: Are these actually the virtual dimensions of the chunk? I think shape * scale represents the physical dimensions? Also, do we really need the first chunk here? It seems like any chunk at the current LOD would do. If that’s the case, it might be cleaner and more efficient to maintain a mapping of physical chunk dimensions per LOD, so we don’t need to recompute this every time.

packages/core/src/core/chunk_manager.ts
```diff
@@ -126,18 +163,74 @@ export class ChunkManagerSource {
...
+
+    const minChunkIndexX = Math.floor(visibleBounds.min[0] / chunkVirtualWidth);
+    const maxChunkIndexX = Math.ceil(visibleBounds.max[0] / chunkVirtualWidth);
+    const minChunkIndexY = Math.floor(
+      visibleBounds.min[1] / chunkVirtualHeight
+    );
+    const maxChunkIndexY = Math.ceil(visibleBounds.max[1] / chunkVirtualHeight);
```
shlomnissan: I think this is more or less how the prototype handled spatial filtering, but @aganders3 suggested a slightly different approach that I documented in the tech spec. It’s not a blocker, but I’m curious if you had a chance to review that before settling on this implementation.

packages/core/src/core/chunk_manager.ts
```diff
@@ -126,18 +163,74 @@ export class ChunkManagerSource {
...
-      if (chunk.state === "unloaded") {
-        chunk.state = "loading";
-        this.loader_.loadChunkDataFromRegion(chunk, this.region_).then(() => {
-          chunk.state = "loaded";
-        });
+      // exit early if chunks are not at the current LOD
+      if (chunk.lod !== this.currentLOD_) {
+        chunk.visible = false;
+        continue;
+      }
```
shlomnissan: Visibility shouldn’t be tied directly to the LOD level. Even if we only intend to update chunks for a specific LOD (which I don't think is the case), we shouldn’t assume that all other chunks are always hidden.

packages/core/src/core/chunk_manager.ts
```diff
@@ -126,18 +163,74 @@ export class ChunkManagerSource {
...
+
+      if (!almostEqual(rx, 2) || !almostEqual(ry, 2)) {
+        throw new Error(
+          `Scales must be separated by factors of 2. Got ratio (${rx}, ${ry}) between scales ${prev} and ${curr}`
+        );
+      }
+    }
+  }
+
+  private hasVisibleBoundsChanged(visibleBounds: Bounds): boolean {
```
shlomnissan: As I mentioned in previous comments, I'm not sure this is needed, or that it adds value, but feel free to push back. More generally, I feel like we it would be better to track if "visible bounds have changed" by marking the camera, but that's not a blocker for this PR.

shlomnissan (changes requested):
Overall, this looks great! I haven’t had a chance to fully verify correctness yet, though. How did you test that it works as expected? It’s always helpful to include that kind of detail in the PR summary.

I'm marking this as blocking for now because I think there may be a misunderstanding around chunk visibility. Specifically, I believe we need to track visible chunks across all LODs, not just the current one. Let me know if you see it differently, happy to discuss.

I also would have liked to see the simpler intersection approach that was outlined in the spec and mentioned by Ashley. Not sure if it's viable, but I think it’s worth trying out (not blocking). That said, I think there are still some opportunities to simplify parts of the code overall.

We're very close... You did an EXCELLENT job on this so far!

andy-sweet (approved):
My feedback has been addressed, so approving.

packages/core/src/core/chunk_manager.ts
phoenixAja: yeah! i thought i was doing ashley's approach.. i'll have to look at the spec again

packages/core/src/core/chunk_manager.ts
phoenixAja: i was only adding the documentation to the public methods, but i can take it out. i can rename refreshViewState to update for consistency

phoenixAja:
>  How did you test that it works as expected? It’s always helpful to include that kind of detail in the PR summary.

i did add that to the PR summary, i said: verified chunk_streaming example is only loading the chunks within the current viewport

i logged the visible chunks to see how they changed on viewport change

packages/core/src/core/chunk_manager.ts
phoenixAja: yeah that makes sense

packages/core/src/core/chunk_manager.ts
phoenixAja: mistake

packages/core/src/core/chunk_manager.ts
phoenixAja: we're not using firstChunkAtLOD since now we're loading all the chunks within the viewport regardless of LOD

packages/core/src/core/chunk_manager.ts
phoenixAja: We no longer need to reference the first chunk anymore, since we're computing visibility across all chunks (not just the current LOD).

Also — I was under the impression that shape * scale gives us the chunk dimensions in virtual space, since both shape and scale come directly from attrs. That should mean we're working in world/virtual units, not physical ones right?

packages/core/src/core/chunk_manager.ts
phoenixAja: Yep, agreed — happy to keep it as "unloaded" for now, and i can follow up with a "failed to load" state in a new PR

packages/core/src/core/chunk_manager.ts
phoenixAja: Yep, that makes sense — I like the idea of using the lowest available LOD as a fallback while higher-res chunks are still loading. I’ll follow up in a future PR to add logic for stitching across LODs and preventing gaps!

packages/core/src/core/chunk_manager.ts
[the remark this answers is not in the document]
phoenixAja: cc @shlomnissan

packages/core/src/core/chunk_manager.ts
shlomnissan: @andy-sweet, we're using the `regionToIndices` function when loading a chunk in `OmeZarrImageLoader::loadChunkDataFromRegion` (via the `getDimInfoMap` function) and I think it does take translation into account?

Is the offset returned by `regionToIndices` in data space sufficient for our needs? If not, could you provide some context as to why? Or better yet, is there a data source with a different origin we can use to test?

packages/core/src/core/chunk_manager.ts
shlomnissan: Chunk shape * scale transform generally represents the physical dimensions of the chunk in the coordinate space. In other words, it maps the chunk from pixel space into the physical coordinate system (e.g., microns, nanometers, etc.).

shlomnissan (approved):
Looks good to me :+1:

PR #288, opened 2025-06-24
Handle upper case dimension names in OME-Zarr chunk loader

andy-sweet:
### Summary
In the OME-Zarr chunk loading parts of the code, we assume that the last two spatial dimensions that we present will be named "x" or "y". We mostly do this in a case-insensitive way, but `getDimInfoMap` currently assumes lower-case only. See the image2d_from_omezarr5d_u8 example for a case where this currently fails.

This change is a short term fix to that method to be case-insensitive in a similar way to related parts of the code.

For a long term fix, we'll want a more robust way of handling dimension names and correspondences, but that is a much larger problem.

### Tests & Checks
I manually tested all the examples after this change including the currently broken image2d_from_omezarr5d_u8.

### DIFF
--- a/packages/core/src/data/ome_zarr_image_loader.ts
+++ b/packages/core/src/data/ome_zarr_image_loader.ts
@@ -222,15 +222,23 @@ export class OmeZarrImageLoader {
 
     const output = new Map();
     region.forEach((entry, dimIdx) => {
-      if (entry.dimension === "x") {
+      if (entry.dimension.toLocaleLowerCase() === "x") {
         const value = 0; // not used for x dimension
-        output.set("x", { dimIdx, chunkIdx: chunk.chunkIndex.x, value });
+        output.set(entry.dimension, {
+          dimIdx,
+          chunkIdx: chunk.chunkIndex.x,
+          value,
+        });
         return;
       }
 
-      if (entry.dimension === "y") {
+      if (entry.dimension.toLocaleLowerCase() === "y") {
         const value = 0; // not used for y dimension
-        output.set("y", { dimIdx, chunkIdx: chunk.chunkIndex.y, value });
+        output.set(entry.dimension, {
+          dimIdx,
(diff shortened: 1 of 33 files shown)


andy-sweet:
@shlomnissan : I assigned you to review this because I think you introduced some of these changes. Feel free to reassign to @phoenixAja if I got confused about the order of changes here.

packages/core/src/data/ome_zarr_image_loader.ts
```diff
@@ -222,15 +222,23 @@ export class OmeZarrImageLoader {
 
     const output = new Map();
     region.forEach((entry, dimIdx) => {
-      if (entry.dimension === "x") {
+      if (entry.dimension.toLowerCase() === "x") {
         const value = 0; // not used for x dimension
-        output.set("x", { dimIdx, chunkIdx: chunk.chunkIndex.x, value });
+        output.set(entry.dimension, {
```
shlomnissan: Since we already handle both lowercase and uppercase here, we can standardize the dimension to lowercase at this point and eliminate redundant case checks in downstream code.

Instead of saying `entry.dimension` we can revert this to `x`, and then we can remove the following checks

```ts
const dimInfo = await this.getDimInfoMap(region, chunk, attrs);
if (
  !(dimInfo.has("x") || dimInfo.has("X")) ||
  !(dimInfo.has("y") || dimInfo.has("Y"))
) {
  throw new Error("Missing required spatial axis x/y");
}
```
Not a big deal. As mentioned, this is a short-term fix. I think we should prioritize a more robust solution, especially once we have a single path for loading data from OMEZarr files.

shlomnissan (approved):

packages/core/src/data/ome_zarr_image_loader.ts
andy-sweet: Good idea. Done! And agreed on the longer term needs.

PR #289, opened 2025-06-24
Add resource disposal and refactor buffers/textures

### LINKED ISSUE #41: Add dispose functionality for geometries and textures
Currently, we allocate names and buffers for geometries and textures using the WebGL renderer but never deallocate them. This is especially crucial for textures, as we might need to free up GPU memory that's no longer in use before uploading new texture data.

shlomnissan:
### Summary

This PR refactors both `WebGLBuffers` and `WebGLTextures` to support upcoming work
on GPU resource lifetime management. It introduces `dispose()` and `disposeAll()`
methods to both classes for explicit resource cleanup.

**WebGLBuffers**
The primary change involves structuring buffer-related GPU handles (vao, vbo, ebo)
into a unified object. This simplifies resource tracking and disposal beyond basic
map cleanup. It will also help when adding wireframe index buffers.

**WebGLTextures**
The main improvements include eliminating repeated calls to common utility methods
(e.g., texture format/type lookups) and reorganizing logic to improve readability
and reduce duplication.

### ⚠️ Note
It may be easier to review by reading through the full files instead of diffing individual changes.

### Related Issue
Closes #41 

### Tests & Checks
- Manual verification
- All checks have passed

### DIFF
--- a/packages/core/src/renderers/webgl_buffers.ts
+++ b/packages/core/src/renderers/webgl_buffers.ts
@@ -1,69 +1,81 @@
 import { RenderableObject } from "../core/renderable_object";
+import { GeometryAttributeIndex } from "../core/geometry";
+
+type BufferHandles = {
+  vao: WebGLVertexArrayObject;
+  vbo: WebGLBuffer;
+  ebo?: WebGLBuffer;
+};
 
 export class WebGLBuffers {
   private readonly gl_: WebGL2RenderingContext;
-  private VAOs_: Map<string, WebGLVertexArrayObject> = new Map();
-  private currentVAO_: WebGLVertexArrayObject = 0;
+  private readonly buffers_: Map<RenderableObject, BufferHandles> = new Map();
+  private currentObject_: RenderableObject | null = null;
 
   constructor(gl: WebGL2RenderingContext) {
     this.gl_ = gl;
   }
 
-  public bind(object: RenderableObject) {
-    const uuid = object.geometry.uuid;
+  public bindObject(object: RenderableObject) {
+    if (this.alreadyActive(object)) return;
 
-    if (this.alreadyActive(uuid)) return;
-
-    let objectVAO = this.VAOs_.get(uuid) || null;
-    if (!objectVAO) {
-      objectVAO = this.createVAO();
+    if (!this.buffers_.has(object)) {
+      this.generateBuffers(object);
     }
 
-    this.gl_.bindVertexArray(objectVAO);
-    if (!this.VAOs_.has(uuid)) {
-      this.createBuffers(object);
-      this.VAOs_.set(uuid, objectVAO);
+    const buffers = this.buffers_.get(object);
+    if (!buffers) {
+      throw new Error("Failed to retrieve buffer handles for object");
     }
 
-    this.currentVAO_ = objectVAO!;
+    this.gl_.bindVertexArray(buffers.vao);
+    this.currentObject_ = object;
   }
 
-  private alreadyActive(uuid: string) {
-    if (this.currentVAO_ !== 0) {
-      return this.VAOs_.get(uuid) === this.currentVAO_;
+  public disposeObject(object: RenderableObject) {
+    const buffers = this.buffers_.get(object);
+    if (!buffers) return;
+
+    this.gl_.deleteVertexArray(buffers.vao);
+    this.gl_.deleteBuffer(buffers.vbo);
+
+    if (buffers.ebo) this.gl_.deleteBuffer(buffers.ebo);
+
+    this.buffers_.delete(object);
+    if (this.currentObject_ === object) {
+      this.currentObject_ = null;
     }
-    return false;
   }
 
-  private createVAO() {
+  public disposeAll() {
+    for (const object of this.buffers_.keys()) {
+      this.disposeObject(object);
+    }
+  }
+
+  private alreadyActive(object: RenderableObject) {
+    return this.currentObject_ === object;
+  }
+
+  private generateBuffers(object: RenderableObject) {
     const vao = this.gl_.createVertexArray();
     if (!vao) {
-      throw new Error(`Unable to generate a vertex array object name`);
+      throw new Error("Failed to create vertex array object (VAO)");
     }
-    return vao;
-  }
 
-  private createBuffers(object: RenderableObject) {
-    const buffer = this.gl_.createBuffer();
-    const { vertexData, indexData, attributes, stride } = object.geometry;
+    this.gl_.bindVertexArray(vao);
 
-    const bufferType = this.gl_.ARRAY_BUFFER;
-    this.gl_.bindBuffer(bufferType, buffer);
-    this.gl_.bufferData(bufferType, vertexData, this.gl_.STATIC_DRAW);
+    const { vertexData } = object.geometry;
+    const vboType = this.gl_.ARRAY_BUFFER;
+    const vbo = this.gl_.createBuffer();
+    if (!vbo) throw new Error("Failed to create vertex buffer (VBO)");
 
+    this.gl_.bindBuffer(vboType, vbo);
+    this.gl_.bufferData(vboType, vertexData, this.gl_.STATIC_DRAW);
+
+    const { attributes, stride } = object.geometry;
     attributes.forEach((attr) => {
-      let idx = -1;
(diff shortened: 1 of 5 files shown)


aganders3:
Thanks - in general this is a nice improvement, and a good way to see what cleanup may be required for each object.

I'm curious how you are thinking about the control flow for cleanup. Currently a Layer is responsible for creating a RenderableObject, but it has no mechanism for communicating directly to the WebGL* objects that the object's buffers, etc. should be disposed. I see the disposal methods are public, and I'm guessing it will be the Renderer calling them since it already owns the references to WebGLBuffers and WebGLTextures.

Alternatively I could imagine adding disposal callbacks to the object itself when generating buffers for it. Then a layer could simply remove it and call `renderableObject.dispose()` and it would clean itself up. Maybe this puts too much responsibility on the layer.

I'm also not quite sure how this will work for Textures (or even buffers) in use by multiple objects (if this is something we expect).

Sorry if this was discussed elsewhere while I was out! I'm just trying to put a little more context around the changes here. I'll take a closer look at exactly what's happening but this looks good to me.

shlomnissan:
@aganders3 These are all great questions. The main thought I had regarding object disposal is that it needs to be explicit for now. Meaning, some other part of the system has to specifically indicate that the underlying resources of a given renderable object should be released.

In the past, I’ve handled disposal implicitly, which works in languages that support deterministic object lifetimes. Similar to your suggestion, I’d attach a disposal callback to the object that gets triggered right before the object is destroyed, typically from the destructor.

I believe implicit disposal is also technically possible in JavaScript using the `FinalizationRegistry`, but I’m not very familiar with how object lifetimes work in that environment. I do know that Neuroglancer uses some form of reference counting, including across thread boundaries (e.g., worker threads can increment the count), but I’m not entirely sure what it’s used for or how the mechanism works.

To go back to your original question—we could define a `dispose()` method on the base renderable object class that marks the object for disposal using a flag. Then the renderer could clean up associated resources like buffers and textures. Alternatively, we could use a callback mechanism. Either way the layer would need to coordinate some of this, which isn’t ideal, but I’m okay with that for now?

Where I’m currently getting stuck is not so much how we dispose objects, but _when_ we should do it. For instance, if we have a renderable object backed by a vertex buffer and a texture, should we destroy it immediately when it’s removed from a layer? Probably not—especially if we might reuse it. In that case, it would make more sense to keep a pool of image renderables for different LODs, where we simply update the buffers as needed.

If you or @andy-sweet have thoughts on this, I’d love to hear them.

packages/core/src/renderers/webgl_buffers.ts
```diff
@@ -1,69 +1,81 @@
...
 
-  private alreadyActive(uuid: string) {
-    if (this.currentVAO_ !== 0) {
-      return this.VAOs_.get(uuid) === this.currentVAO_;
+  public disposeObject(object: RenderableObject) {
+    const buffers = this.buffers_.get(object);
+    if (!buffers) return;
+
+    this.gl_.deleteVertexArray(buffers.vao);
+    this.gl_.deleteBuffer(buffers.vbo);
```
andy-sweet, line 40: From the [MDN docs of deleteBuffer](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/deleteBuffer)

> Normally you don't need to call this method yourself, when the buffer object is dereferenced it will be marked as free.

So maybe we don't need to explicitly delete these? 

Of course, we can still explicitly delete at this point if we want to clean up resources ASAP.

packages/core/src/renderers/webgl_textures.ts
```diff
@@ -166,93 +168,92 @@ export class WebGLTextures {
...
-
-  private getGLWrapMode(mode: TextureWrapMode) {
-    switch (mode) {
-      case "repeat":
-        return this.gl_.REPEAT;
-      case "clamp_to_edge":
-        return this.gl_.CLAMP_TO_EDGE;
+      case "nearest": return this.gl_.NEAREST;
+      case "linear": return this.gl_.LINEAR;
+      default: throw new Error(`Unsupported texture filter: ${filter}`);
```
andy-sweet, line 188: Nit: I like the compile-time exhaustive check used elsewhere in idetik (e.g. in [image_layer](https://github.com/chanzuckerberg/idetik/blob/013f39ff48c744b6d17630c50c4224e2f10c83fd/packages/core/src/layers/image_layer.ts#L111).

Maybe we could roll that into a utility function to make it a one-liner to keep these compact. Doesn't have to happen in this PR, but you made a few changes like this.

andy-sweet (approved):
Approving, because this looks good to me and is internal only. I'll follow up with questions/thoughts about when to call these in a separate comment.

aganders3:
> The main thought I had regarding object disposal is that it needs to be explicit for now. ... I believe implicit disposal is also technically possible in JavaScript using the `FinalizationRegistry`...

100%, I think explicit is the way to go. I looked at `FinalizationRegistry` when thinking about this and writing my above comment, but it seems useless here as it's apparently not guaranteed to run in a timely manner or even at all. Since these GPU resources are valuable, I think it's better to just handle it all explicitly.

I think neuroglancer's reference counting is doing something different, but I think it could work for us if we expect a single `Texture` (or possibly, but perhaps less likely, `Geometry`) to be used by different renderable objects.

I'll give this all a bit more thought, but we can take the conversation elsewhere.

andy-sweet:
A few thoughts, but maybe not many clear answers :)

> Where I’m currently getting stuck is not so much how we dispose objects, but when we should do it. For instance, if we have a renderable object backed by a vertex buffer and a texture, should we destroy it immediately when it’s removed from a layer? Probably not—especially if we might reuse it. In that case, it would make more sense to keep a pool of image renderables for different LODs, where we simply update the buffers as needed.

Sorry if I'm being a bit simple here, but as things stand now, I think we should do it immediately because we are not reusing any textures, objects, or buffers from a pool. Layers explicitly create things like textures, so they should explicitly delete them too.

So I'd expect an Idetik developer to explicitly call `Layer.dispose` which calls `RenderableObject.dispose` which calls `Texture.dispose` which calls `gl.deleteTexture`. I'd expect a developer to do that when they don't have any need to render that layer any more.

For example, we should probably do that in the ultrack prototype right now [where we create a layer for each new task](https://github.com/chanzuckerberg/idetik/blob/b6cc19686d2f86ee4f3d290db34859cdaefcbba9/packages/ultrack-prototype/frontend/components/Renderer.tsx#L62). There is a TODO to reuse layers, but that seems like an application-specific optimization that might not always hold.

Another thought is that we expose layer instances directly in the API (i.e. the objects that effectively control the lifespan of textures, etc.), which means that developers can intentionally choose not to call dispose of their textures/buffers after removing them from the layer manager or after stopping idetik. Originally, I thought we might automatically call dispose when a layer is removed from the layer manager or when calling idetik.stop, but I don't think that's desirable.

andy-sweet:
Also, from an initial reading, it looks like dispose is safe, at least for textures. I.e. I could explicitly call `Layer.dispose`, but the layer could still be safely used for rendering, there would just be a performance hit to create a new webgltexture. Is that right? And is that intended?

packages/core/src/renderers/webgl_buffers.ts
shlomnissan: Ahhh I had no idea. I guess it's because the garbage collector takes care of it already? I'm pretty sure the OpenGL driver won’t automatically clean up resources unless the context is destroyed.

Anyway, I might leave it to avoid unpredictable delays and aligns with expectations from APIs like OpenGL. Also helps avoid confusion for those (like me) who assumed this was required, but I will add a comment explaining that.

From the [MDN docs of deleteBuffer](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/deleteBuffer)
> Note: This feature is available in Web Workers

I had no idea the WebGL context is available in Web Workers either.

shlomnissan:
> As things stand now, I think we should do it immediately because we are not reusing any textures, objects, or buffers from a pool. Layers explicitly create things like textures, so they should explicitly delete them too.

Oh yeah, I do think we need to start disposing resources across layers and renderable objects until we have a better plan for certain paths like the chunk manager, which currently just creates new "image renderables." I can add a `dipose()` virtual function to layers and renderable objects and hook it up in a follow-up.

PR #292, opened 2025-06-27
refactor: Simplified React context

aganders3:
## Summary
This PR refactors the `@idetik/react` package to use a simplified runtime-based architecture that decouples the React context from specific layer types and encourages more direct access to the Idetik runtime.

This work enabled the [Try CryoLens demo](https://github.com/chanzuckerberg/try-cryolens) which implements its own `Particles` layer as well as instantiates layers dynamically based on user actions.

### React Context Simplification

The `useIdetik` hook is updated to only provide direct runtime access instead of being tightly coupled to `ImageSeriesLayer`.

**Before**: Context was tightly coupled to `ImageSeriesLayer` with automatic channel state management:
```typescript
// Old: ImageSeriesLayer-specific context
channels: ChannelProps[];
channelControls: ChannelControl[];
setChannelControls: React.Dispatch<...>;
```

**After**: Generic runtime access that works with any layer types:
```typescript
// New: Runtime-focused context
export type IdetikContextValue = 
  | { isReady: true; runtime: Idetik; }
  | { isReady: false; runtime: null; initializeWithCanvas: (canvas: HTMLCanvasElement) => void; };
```

A new `IdetikCanvas` component adds a `<canvas>` to the DOM and initializes the Idetik runtime/context for it.

### "Viewer"
The combination of these these things captures what I consider a "Viewer": a component that
* contains an `IdetikCanvas` to initialize the context
* manages creation and lifecycle of some layers

Our `OmeZarrImageViewer` then becomes the first example of this, and what I would consider a "reference" implementation. That is, it may have enough functionality for 80% of use-cases, but also serves as a guide for how to use the other provided components to build for the other 20% of use-cases. The logic from the `useOmeZarrImageViewer` hook has been moved _back_ into the component here, though that may not be a strictly necessary change. I think we can continue to iterate on this reference implementation (and create others) to represent best practices and clean patterns as they evolve.

This all means `Viewer` components will typically be _self-contained_ - one side-effect here is that the `ChannelControlsList` is no longer decoupled from the `OmeZarrImageViewer` (from a DOM perspective). Instead the `ChanelControlsList` takes its relevant layer as a prop, and manages state synchronization itself. **This is a breaking change** for downstream users of the `OmeZarrImageViewer`, who will no longer need to (or be able to) add/place a `ChannelControlsList` component themselves. There are a number of ways we could decouple this again (e.g. passing in a `onNewLayer` callback to lift the `ImageSeriesLayer` reference out of the viewer, or instantiating a `ChannelControlsList` hard-coded with the first layer from the runtime (effectively how it works now)).

## Future Work

A more declarative React-style interface (e.g., `<Layer>` components, or using layer-construction objects) was intentionally out of scope for this PR. However the changes proposed here establish a foundation that could enable such patterns in the future.

Some other components I developed but left off this PR, I'd like to include in a follow-on:
* `HCSImagePanel` to select a well/fov from an HCS OME-Zarr
* `IdetikLayerList` to list active layers and allow removal

One more item I have noted is kind of awkward: it's hard to maintain state that corresponds to library state, but perhaps shouldn't _be_ library state. A good example here is extra data for the channel controls. The library does not need to know a channel's name or contrast _range_ (limits of a contrast slider), but we need them to show the _control_. It's clunky and error-prone to need to keep these in sync and in the same order. Another example might be a layer having a display name. @bchu1 had a good idea to create an affordance for encoding these things as additional (arbitrary) properties on library objects at runtime.

### Related Issue
N/A

### Tests & Checks
Tested the demo apps for Z and T slicing with `npm run components` and manually testing both http://localhost:5173 and http://localhost:5173/examples/organelle-box-time/


packages/react/README.md
```diff
@@ -61,6 +55,104 @@ To track metrics, the `OmeZarrImageViewer` component accepts optional callbacks
...
+import { IdetikProvider, IdetikCanvas, useIdetik } from "@idetik/react";
+import { ImageSeriesLayer, OmeZarrImageSource } from "@idetik/core";
+import { useState, useEffect, useSyncExternalStore } from "react";
+
+function CustomViewer({ sourceUrl }: { sourceUrl: string }) {
+  const contextValue = useIdetik();
+  const [layer, setLayer] = useState<ImageSeriesLayer | null>(null);
+
+  // Get active layers using useSyncExternalStore
+  const activeLayers = useSyncExternalStore(
```
bchu1, line 71: why not put `useSyncExternalStore()` in the `<IdetikProvider>` and have users directly use `contextValue.runtime.layerManager.layers`?

packages/react/README.md
```diff
@@ -61,6 +55,104 @@ To track metrics, the `OmeZarrImageViewer` component accepts optional callbacks
...
+    () => {
+      if (!contextValue.isReady) return [];
+      return contextValue.runtime.layerManager.layers;
+    }
+  );
+
+  useEffect(() => {
+    if (!contextValue.isReady) return;
+
+    const runtime = contextValue.runtime;
```
bchu1: possible to just destructure `{ isReady, runtime }`?

packages/react/README.md
```diff
@@ -61,6 +55,104 @@ To track metrics, the `OmeZarrImageViewer` component accepts optional callbacks
...
+
+    const createLayer = async () => {
```
bchu1: nit - don't need this to be a separate function?

packages/react/README.md
```diff
@@ -61,6 +55,104 @@ To track metrics, the `OmeZarrImageViewer` component accepts optional callbacks
...
+- When `isReady: false`: Provides `initializeWithCanvas(canvas)` function
+- When `isReady: true`: Provides `runtime` (Idetik instance) and `canvas` (HTMLCanvasElement)
+
+**Layer Management**: Use `runtime.layerManager` directly:
+- `runtime.layerManager.add(layer)` - Add a layer
+- `runtime.layerManager.remove(layer)` - Remove a layer
+- `runtime.layerManager.layers` - Get current layers array
+- `runtime.layerManager.addLayersChangeCallback(callback)` - Subscribe to layer changes
+
+**Reactive Updates**: Use `useSyncExternalStore` to subscribe to runtime state changes. Prefer this
```
bchu1, line 138: I think we should just automatically subscribe to everything in core that exposes a callback in the provider

packages/react/README.md
aganders3: I tried this before by adding an `activeLayers` attribute on `IdetikContextValue`. It worked, but I removed it because most components just don't need it. If you make it a top-level property the context is constantly changing, which made it a little annoying to include in `useEffect` dependencies for example.

Another option would be to provide a second custom hook like `useSyncedActiveLayers` or that would just wrap this call to `useExternalSyncStore`.

If it's a common need though I'd be happy to put it in.

packages/react/README.md
aganders3: Anything exposed directly by the runtime you mean (at the moment just the `layers`)? I'd be fine with that. This was more a comment about things like the channel controls.

packages/react/src/components/providers/IdetikProvider.tsx
```diff
@@ -1,61 +1,44 @@
...
+      newIdetik.start();
+      setIdetik(newIdetik);
+    }
+  }, [canvasRef, idetik]);
+
+  const idetikContext: IdetikContextValue = idetik
+    ? {
+        isReady: true,
+        runtime: idetik,
+        canvas: canvasRef!,
```
bchu1: `canvas` is already exposed in the runtime right?

packages/react/src/components/providers/IdetikProvider.tsx
```diff
@@ -1,61 +1,44 @@
...
-  const channels = useSyncExternalStore(
-    imageSeriesLayer?.addChannelChangeCallback ?? (() => () => {}),
-    () => imageSeriesLayer?.channelProps ?? EMPTY_ARRAY,
-    () => EMPTY_ARRAY // Doesn't render anything on SSR
-  );
-  const [channelControls, setChannelControls] = useState<Array<ChannelControl>>(
-    []
-  );
+  const [idetik, setIdetik] = useState<Idetik | null>(null);
+  const [canvasRef, setCanvasRef] = useState<HTMLCanvasElement | null>(null);
```
bchu1: could get rid of this and just call the code in the `useEffect` in a callback that `initializeWithCanvas` points to?

aganders3: Mmm - yeah that's simpler. Something like this (also removing `canvas` as you noted)?

```typescript
export const IdetikProvider = ({ children }: PropsWithChildren) => {
  const [idetikContext, setIdetikContext] = useState<IdetikContextValue>(
    {
      isReady: false,
      runtime: null,
      initializeWithCanvas: (canvas: HTMLCanvasElement) => {
        const camera = new OrthographicCamera(0, 128, 0, 128, -1000, 1000);
        const newIdetik = new Idetik({
          canvas,
          camera,
          controls,
        });
        newIdetik.start();
        setIdetikContext({
          isReady: true,
          runtime: newIdetik,
        });
      }
    }
  );


  return (
    <IdetikContext.Provider value={idetikContext}>
      {children}
    </IdetikContext.Provider>
  );
};
```

packages/react/README.md
aganders3: Would this mean any component using `useIdetik` would re-render if any of the subscribed values changed? Maybe that's desirable?

packages/react/src/components/viewers/OmeZarrImageViewer/OmeZarrImageViewer.tsx
```diff
@@ -1,28 +1,345 @@
...
+      } catch (err) {
+        console.error("[Viewer] Failed to load OMERO metadata:", err);
+      }
+    };
+
+    createLayer();
+
+    return () => {
+      if (
+        imageLayerRef.current &&
```
bchu1, line 172: since we're only supporting 1 layer right now, what if we just got rid of `imageLayerRef` and did `removeAll()`?

packages/react/src/components/viewers/OmeZarrImageViewer/components/ChannelControlsList/ChannelControlsList.tsx
```diff
@@ -108,10 +116,10 @@ export function ChannelControlsList({ classNames }: ChannelControlsListProps) {
                 <ChannelControl
                   key={index}
                   channelIndex={index}
-                  label={channelControls[index]?.label ?? `Channel ${index}`}
+                  label={extraControlProps[index].label}
                   color={props.color}
                   contrastLimits={props.contrastLimits}
-                  contrastRange={contrastRange}
+                  contrastRange={contrastRange as [number, number]}
```
bchu1: nit - is the cast needed?

packages/react/vite.config.js
```diff
@@ -47,15 +47,18 @@ export default defineConfig(() => {
           'react-dom/client',
           '@czi-sds/components',
           '@mui/material',
+          '@idetik/core',
         ],
         output: {
+          banner: '"use client";\n',  // library is intended for client-side use for now
```
bchu1: can prolly get rid of the for now 😄

packages/react/src/components/viewers/OmeZarrImageViewer/OmeZarrImageViewer.tsx
aganders3: The code might become much simpler 😆

I have to think about it a little. There are a couple places it checks if `thisLayer === imageLayerRef.current` for example to short-circuit.

We need some reference to the layer to pass it to the `ChannelControlsList`, but we could get it from the runtime and cast it. I think this is how it works now?

I could also see needing to add this back when we add a segmentation layer, but the argument would be "just add it back then".

packages/react/src/components/viewers/OmeZarrImageViewer/components/ChannelControlsList/ChannelControlsList.tsx
aganders3: Apparently not!

packages/react/README.md
bchu1: we can also `useMemo()` the object that is passed into the context to only change when necessary (https://react.dev/reference/react/useContext#optimizing-re-renders-when-passing-objects-and-functions)

packages/react/vite.config.js
bchu1: sry I meant the "for now" in the comment, I'm totally good with adding the banner

packages/react/README.md
bchu1: pretty much, I think it should be ok? it's the same as if the state was all defined in the component calling `useIdetik()` anyways

packages/react/src/components/providers/IdetikProvider.tsx
bchu1: yeah, maybe also check if `idetik` already exists? I remember we were talking about caching the state in case of unmount/remounts

packages/react/README.md
aganders3: Sorry, maybe we can chat about it on Monday (almost EOD for me). Do you mean to have like a single call to `useIdetik` somewhere and anything extracted from it passed as props to children?

I was thinking any component inside the provider would call `useIdetik` itself in order to either update or query the state from the runtime.

So for example you might have an `HCSImagePanel` that shows `well` and `image` menus - when you select them it creates a layer and adds it to the runtime. Another component might let you select camera parameters or controls. Neither would need to be aware of other layer changes, nor should they re-render when the layers change - so I don't think they would need to be subscribed.

Then a `LayerList` component would call `useIdetik` and subscribe for layer changes (only).

packages/react/src/components/viewers/OmeZarrImageViewer/components/ChannelControlsList/ChannelControlsList.tsx
```diff
@@ -7,16 +7,34 @@ import {
...
-import { useIdetik } from "../../../../hooks";
+import { useSyncExternalStore } from "react";
+import { ExtraControlProps } from "../../utils";
 
 export interface ChannelControlsListProps {
+  // TODO: make this work with ImageLayer as well - need to refactor to add
+  // useSyncExternalStore-compatible methods to ImageLayer
+  layer: ImageSeriesLayer;
+  // TODO: it's awkward to have labels and contrastRanges as separate props
+  // but they're not needed for *rendering* so they don't belong in the library
```
andy-sweet, line 18: In the specific case of contrast ranges (and maybe labels too?), even though the library does not need them for rendering at the moment, it might be able to easily populate them while doing something else.

Another related thought I've been having is around source metadata. I think the library has to load some of that to render, and so if it's already fetched it, it may as well populate it. This could also simplify the OmeZarrImageViewer.

packages/react/src/index.ts
```diff
@@ -1,20 +1,5 @@
...
-export {
-  Renderer,
-  OmeZarrImageViewer,
-  type OmeZarrImageViewerProps,
-  ChannelControlsList,
-  useIdetik,
-  useOmeZarrViewer,
-  IdetikProvider,
-};
+export * from "./components";
```
andy-sweet, line 4: I'd expect some way to define non-exported components that might help make some of the bigger components here more readable, but aren't intended for public reuse. I was considering doing exactly that for a `ScaleBar` component where we might want to restrict external use of that to prevent inaccurate depictions.

aganders3: With the fractal structure/barrel files I think that's okay - if you don't export them somewhere along the chain they will stop bubbling up.

packages/react/README.md
```diff
@@ -61,6 +55,98 @@ To track metrics, the `OmeZarrImageViewer` component accepts optional callbacks
...
+```tsx
+import { IdetikProvider, IdetikCanvas, useIdetik } from "@idetik/react";
+import { ImageSeriesLayer, OmeZarrImageSource } from "@idetik/core";
+import { useState, useEffect, useSyncExternalStore } from "react";
+
+function CustomViewer({ sourceUrl }: { sourceUrl: string }) {
+  const [layer, setLayer] = useState<ImageSeriesLayer | null>(null);
+  const { isReady, runtime } = useIdetik();
+
+  // Get active layers using useSyncExternalStore
```
andy-sweet: Nit: for someone reading advanced usage, this comment is no more descriptive than the code beneath. I would either improve it or remove it.

```suggestion
```

packages/react/README.md
```diff
@@ -61,6 +55,98 @@ To track metrics, the `OmeZarrImageViewer` component accepts optional callbacks
...
+  useEffect(() => {
+    if (!isReady) return;
+
+    const source = new OmeZarrImageSource(sourceUrl);
+    const newLayer = new ImageSeriesLayer({
+      source,
+      region: yourRegion,
+      seriesDimensionName: "Z"
+    });
+    // Add layer to the runtime
```
andy-sweet: Nit: for someone reading advanced usage, this comment is no more descriptive than the code beneath. I would either improve it or remove it.

```suggestion
```

andy-sweet (commented):
I still need to digest this, especially as I hadn't fully caught up with the Idetik React components before these changes. But my overall impression is positive because I found this generally easier to follow that what's on `main`, maybe partly because it feels like there is less redirection, but also because it looks like there are some real simplifications too.

I'll take another look at this next week.

packages/react/README.md
bchu1: Sg. I meant just let everything rerender. I think what we currently have wouldn't trigger too much. If we do have to add some sort of state eventually that triggers a lot, we could look into some options... off the top of my head:

1. Split into multiple contexts and have `useIdetik()` take in an enum or something to determine which context to subscribe to
2. Stop using context and instead make `useIdetik()` take a selector function (like what state management libraries do) that gets passed into the `snapshot` argument of `useSyncExternalStore()`, make either core expose a single store that publishes all state changes or create a proxy store in react/ that subscribes to all of core's more fine grained callback stores.

Trying to avoid having users use `useSyncExternalStore()` directly since it's supposed to be an obscure hook that only libraries call. No one really knows about it 🫠 and it's kind of hard to use

packages/react/src/components/viewers/OmeZarrImageViewer/OmeZarrImageViewer.tsx
bchu1: oh interesting. I'm fine with leaving it. I wonder how controls would work once we have multiple layers, like how would they specify which layer gets the controls?

aganders3: I imagine something like a `LayerList` component that would let you select an "active" layer.

One idea we had was that layers themselves could define interactions (i.e. mouse/key listeners), and to do that we may need a concept of the "active" layer in the library itself.

packages/react/src/components/IdetikCanvas.tsx
```diff
@@ -0,0 +1,34 @@
+"use client";
+
+import { useCallback } from "react";
+import { useIdetik } from "../hooks/useIdetik";
+
+interface IdetikCanvasProps {
+  canvasId?: string;
+  style?: React.CSSProperties;
```
andy-sweet: Is passing explicit style needed? I'm trying to think if it could cause any issues (I can't think of too many important ones). It's also somewhat inconsistent with the other styling in the components which uses tailwind shorthand.

packages/react/vite.config.js
```diff
@@ -56,6 +57,7 @@ export default defineConfig(() => {
             'react-dom/client': 'ReactDOM.client',
             '@czi-sds/components': 'SDS',
             '@mui/material': 'MaterialUI',
+            '@idetik/core': 'IDETIKCore',
```
andy-sweet: Nit: unless IDETIK is an acronym, don't make it all upper-case (we're not NVIDIA are we?!)

```suggestion
            '@idetik/core': 'IdetikCore',
```

packages/react/src/components/viewers/OmeZarrImageViewer/components/ChannelControlsList/ChannelControlsList.tsx
```diff
@@ -7,16 +7,34 @@ import {
...
+  layer,
+  extraControlProps,
+  classNames,
+}: ChannelControlsListProps) {
+  const channels = useSyncExternalStore(
+    (callback) => {
+      layer.addChannelChangeCallback(callback);
+      return () => layer.removeChannelChangeCallback(callback);
+    },
+    () => layer.channelProps ?? []
```
andy-sweet: This used to have a comment regarding SSR. Do we still need that?

Or the assumption that all Idetik components are effectively client-side? Does that mean we need a "use-client" at the top of this file, or is the one in OmeZarrImageViewer.tsx enough?

andy-sweet (approved):
I only have minor non-blocking comments left, so I'm approving. You should probably wait to merge until @bchu1 approves too though.

packages/react/src/components/IdetikCanvas.tsx
bchu1: we could switch it to `className` to make it consistent?

packages/react/README.md
aganders3: > Trying to avoid having users use useSyncExternalStore() directly since it's supposed to be an obscure hook that only libraries call. No one really knows about it 🫠 and it's kind of hard to use

I think I agree with this. My thought is that we should basically be providing components for all uses (that we can think of) that would require it. The comment in the README about using it is more for developers or encouraging us to stick with this pattern vs. finding other ways to sync state as it comes up.

packages/react/src/components/viewers/OmeZarrImageViewer/components/ChannelControlsList/ChannelControlsList.tsx
aganders3: Yeah I'm ambivalent about what I wrote in that comment I guess. I could see certain things having a reasonable home in the library, but I feel like this is a challenge we may run into regardless. I'll keep it as a background thought for a while.

packages/react/src/components/IdetikCanvas.tsx
aganders3: I think the intent was to move all styling out of the components to generally be handled by downstream libraries (via these `style` props), but I'll be honest I don't quite understand the best practices here.

I do think perhaps `classNames` would be more accurate?

packages/react/src/components/viewers/OmeZarrImageViewer/components/ChannelControlsList/ChannelControlsList.tsx
aganders3: Hm, good question. I'm not quite sure when we need the extra SSR callback in this, and I've struggled a little trying to force the components to be client-side only. I guess it might be worth just adding it here (including the comment)?

packages/react/src/components/viewers/OmeZarrImageViewer/components/ChannelControlsList/ChannelControlsList.tsx
```diff
@@ -7,16 +7,36 @@ import {
...
-  const { idetik, isInitialized, channels, channelControls } = useIdetik();
+export function ChannelControlsList({
+  layer,
+  extraControlProps,
+  classNames,
+}: ChannelControlsListProps) {
+  const channels = useSyncExternalStore(
+    (callback) => {
+      layer.addChannelChangeCallback(callback);
+      return () => layer.removeChannelChangeCallback(callback);
```
bchu1, line 37: nice, I totally forgot about the cleanup

packages/react/src/components/viewers/OmeZarrImageViewer/components/ChannelControlsList/ChannelControlsList.tsx
```diff
@@ -7,16 +7,36 @@ import {
...
+    },
+    () => layer.channelProps ?? [],
```
bchu1: are you getting any infinite rerender errors from not making the empty array stable? Might be helpful to just make it stable anyways, in case something changes and people start running into this

bchu1 (approved):
Thanks Ashley! I'm good to go!

Issue #191, opened 2025-02-25
Add a scale bar


aganders3:
I'm not sure the best way to implement this right now. Drawing the line might be relatively easy, but adding text will be a bit more effort. We also need to make sure it makes sense in the context of the layers being shown. Thus, it's related to #83.

It's also possible we could do this (or just the text?) outside of the GL context in the component library.

edyoshikun:
It would be cool to dynamically generate this scalebar like in 
https://opencell.sf.czbiohub.org/target/1050

andy-sweet:
Requirements and technical design: https://docs.google.com/document/d/1yr9ul99CXgodVjxJhHMi60EUKx5-4gYnetxRhbMwCj0/

PR #293, opened 2025-06-27
Scale bar prototype

andy-sweet:
### Summary
This is a prototype for a simple approach (i.e. no new abstractions in Idetik other than a small addition to the public API) to drawing a scale bar in Idetik that allows style configuration. I used the React `OmeZarrImageViewer` component to drive this approach and ensure we have something useful there, given that will be the first real-world usage. But I also show how this can be used without React in one of the existing examples.

There a few things I don't like about this approach, but I think it might be the shortest path to something useful that make minimal additions to the Idetik public API.

This is a draft PR because this is prototype intended to be used a reference and should not be merged directly.

https://github.com/user-attachments/assets/efac476d-1161-4ab7-aeec-9298f30a2a84

### Related Issue
Related to #191


andy-sweet:
Closing this in favor of #297

PR #294, opened 2025-06-30
fix: Resize renderer on all <canvas> size changes

bchu1:


packages/core/test/webgl_renderer.test.ts
```diff
@@ -1,7 +1,17 @@
-import { expect, test } from "vitest";
+import { afterAll, beforeAll, expect, test, vi } from "vitest";
 
 import { WebGLRenderer } from "@";
 
+beforeAll(() => {
+  vi.stubGlobal(
```
andy-sweet: Which test fails and why without this stub? Some comment in the code to explain the need here would be ideal.

andy-sweet (changes requested):
Good catch on the need for ResizeObserver, but this causes significant flickering for me when resizing, which is not the case on main.

This is how the HCS example looks for me on this branch - warning if you are photosensitive!

https://github.com/user-attachments/assets/23d0cf38-b040-4aa1-aed2-2d7a1a651e00

Similar problem for all examples and the main React demo.

bchu1:
> Good catch on the need for ResizeObserver, but this causes significant flickering for me when resizing, which is not the case on main.
> 
> This is how the HCS example looks for me on this branch - warning if you are photosensitive!
> 
>  fix-resize.mp4 
> Similar problem for all examples and the main React demo.

This has been a pretty interesting rabbit hole 🐰 it works now (tested it in Chrome Firefox and Safari) though it's a bit sad that renderer.ts is incapable of "resizing itself". Basically when we set `canvas.width` and `canvas.height` it clears the `canvas`, so the only sure-fire way to prevent flickering is to render immediately after resizing.

I think this was only working sort of by accident before due to how the browser happens to order events and then RAF, but `ResizeObserver` callbacks are executed after RAF. I tried out wrapping the `ResizeObserver` callback in `setTimeout()` which actually worked to restore the previous order, but still produced some flickering in Firefox for whatever reason.

andy-sweet:
Sorry about the delay, I'll take another look at this later today.

packages/core/test/webgl_renderer.test.ts
```diff
@@ -1,7 +1,20 @@
-import { expect, test } from "vitest";
+import { afterAll, beforeAll, expect, test, vi } from "vitest";
 
 import { WebGLRenderer } from "@";
 
+beforeAll(() => {
+  // "Instantiate WebGLRenderer with no WebGL context" throws an unhandled error when Renderer's
```
andy-sweet: If I remove this and run all the tests, I do not get this failure. Maybe that's related to the recent changes? If so, maybe we can remove it here too.

Otherwise, I'd prefer to not have it be applied globally if it only applies to one test. Though point taken that we only have two tests here right now...

andy-sweet:
> This has been a pretty interesting rabbit hole 🐰 it works now (tested it in Chrome Firefox and Safari) though it's a bit sad that renderer.ts is incapable of "resizing itself".

Thanks for jumping down said hole and finding out!

I'd also prefer if renderer was capable of handling the resizes itself, but it's not a deal-breaker.

One somewhat simple idea I had was to push `Idetik.needsResize` into `Renderer`, in which case I think you could handle everything there. You might need to define the abstract `Renderer.render` to handle that resize (when needed) and then call the super method in any sub-classes, which I don't love so much.

If that doesn't work or seems unreasonable, I'm happy to approve this to fix those cases when we have canvas resizes without window resizes.

packages/core/test/webgl_renderer.test.ts
bchu1: Ah, yeah I don't need this anymore!

bchu1:
> > This has been a pretty interesting rabbit hole 🐰 it works now (tested it in Chrome Firefox and Safari) though it's a bit sad that renderer.ts is incapable of "resizing itself".
> 
> Thanks for jumping down said hole and finding out!
> 
> I'd also prefer if renderer was capable of handling the resizes itself, but it's not a deal-breaker.
> 
> One somewhat simple idea I had was to push `Idetik.needsResize` into `Renderer`, in which case I think you could handle everything there. You might need to define the abstract `Renderer.render` to handle that resize (when needed) and then call the super method in any sub-classes, which I don't love so much.
> 
> If that doesn't work or seems unreasonable, I'm happy to approve this to fix those cases when we have canvas resizes without window resizes.

I thought about this as well and felt that extra code complexity is better than laying a trap w.r.t requiring subclasses to call `super.render()` at a specific point

andy-sweet (approved):
Good enough reasoning for me!

Issue #47, opened 2024-09-19
Add debug rendering


aganders3:
> Rendering points and/or line strips with the triangles is nice for debugging purposes, as in the PR description. It was easy enough to render points or line strips instead of triangles by changing this accordingly, but I guess the PR description requires another call to draw?
> 
> How hard would it be to make this easy to enable for debugging geometry in general? Would be super useful for (especially when we start to think about tiling textures and other geometries). Definitely doesn't need to be done in this PR though.

_Originally posted by @andy-sweet in https://github.com/chanzuckerberg/imaging-active-learning/pull/46#discussion_r1766770899_

PR #295, opened 2025-07-01
(Stacked) Render wireframes on top of meshes [1/2]

shlomnissan:
### Summary

This PR introduces initial support for rendering wireframes over shaded
geometry. It adds on-demand generation of wireframe index data when a
renderable object is set to wireframe-on-shaded mode, and refactors the WebGL
renderer by extracting the draw call logic into a separate method. This lays
the groundwork for rendering the same object twice — once for the shaded mesh
and once for the wireframe overlay with different rendering parameters.

**Next steps (in a follow-up PR)**
- Add wireframe-specific VAO and EBO support in `WebGLBuffers`
- Introduce a flat-colored shader for wireframes
- Complete the integration in the `WebGLRenderer` logic

### Related Issue
Related issue: https://github.com/chanzuckerberg/idetik/issues/47 (close in the next PR)

### Tests & Checks
- All checks have passed
- Verified manually that all examples are working as expected

### DIFF
--- a/packages/core/src/core/geometry.ts
+++ b/packages/core/src/core/geometry.ts
@@ -1,4 +1,5 @@
 import { Node } from "./node";
+import { Logger } from "../utilities/logger";
 
 type GeometryAttributeType =
   | "position"
@@ -32,14 +33,16 @@ type GeometryAttribute = {
 };
 
 export class Geometry extends Node {
+  private readonly attributes_: GeometryAttribute[];
   protected vertexData_: Float32Array;
   protected indexData_: Uint32Array;
-  private readonly attributes_: GeometryAttribute[];
+  private wireframeIndexData_: Uint32Array;
 
   constructor(vertexData: number[] = [], indexData: number[] = []) {
     super();
     this.vertexData_ = new Float32Array(vertexData);
     this.indexData_ = new Uint32Array(indexData);
+    this.wireframeIndexData_ = new Uint32Array();
     this.attributes_ = [];
   }
 
@@ -54,7 +57,7 @@ export class Geometry extends Node {
   public get stride() {
     return (
       this.attributes_.reduce((acc, curr) => {
-        return (acc += curr.itemSize);
+        return acc + curr.itemSize;
       }, 0) * Float32Array.BYTES_PER_ELEMENT
     );
   }
@@ -67,11 +70,59 @@ export class Geometry extends Node {
     return this.indexData_;
   }
 
+  public get wireframeIndexData() {
+    return this.wireframeIndexData_;
+  }
+
   public get attributes() {
     return this.attributes_;
   }
 
   public get type() {
     return "Geometry";
+  }
+
+  public generateWireframeIndexData() {
+    const indexData = this.indexData_;
+
+    if (indexData.length === 0) {
+      Logger.warn(
+        "Geometry",
+        "Wireframe generation error: only indexed geometries are supported"
+      );
+      return;
+    }
+
+    if (indexData.length % 3 !== 0) {
+      Logger.warn("Geometry", "Wireframe generation error: non-triangle data");
+      return;
+    }
+
+    const edgeSet = new Set<string>();
+    const wireframeIndices: number[] = [];
+    const addEdge = (a: number, b: number) => {
+      // Normalize edge order and use a set to deduplicate,
+      // since shared edges between triangles would otherwise
+      // be added multiple times.
+      const i0 = Math.min(a, b);
+      const i1 = Math.max(a, b);
+      const key = `${i0},${i1}`;
+      if (!edgeSet.has(key)) {
+        edgeSet.add(key);
+        wireframeIndices.push(i0, i1);
+      }
+    };
+
+    for (let i = 0; i < indexData.length; i += 3) {
+      const i0 = indexData[i];
+      const i1 = indexData[i + 1];
+      const i2 = indexData[i + 2];
+
+      addEdge(i0, i1);
+      addEdge(i1, i2);
+      addEdge(i2, i0);
+    }
+
+    this.wireframeIndexData_ = new Uint32Array(wireframeIndices);
   }
 }
--- a/packages/core/src/core/geometry.ts
+++ b/packages/core/src/core/geometry.ts
@@ -98,7 +98,7 @@ export class Geometry extends Node {
       return;
     }
 
-    const edgeSet = new Set<string>();
+    const edgeSet = new Set<{ i0: number; i1: number }>();
     const wireframeIndices: number[] = [];
     const addEdge = (a: number, b: number) => {
       // Normalize edge order and use a set to deduplicate,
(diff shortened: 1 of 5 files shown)


packages/core/src/core/geometry.ts
```diff
@@ -67,11 +70,56 @@ export class Geometry extends Node {
...
   }
 
   public get type() {
     return "Geometry";
   }
+
+  public generateWireframeIndexData() {
+    const indexData = this.indexData_;
+
+    if (indexData.length === 0) {
```
shlomnissan, line 88: This is not strictly necessary. We can generate wireframe index data from vertex positions but it's less accurate.. Index buffers preserve the actual triangle connectivity of the mesh. I might add support for both but I think this should be fine for now.

packages/core/src/core/renderable_object.ts
```diff
@@ -8,11 +8,12 @@ import { Shader } from "../renderers/shaders";
...
 
 export abstract class RenderableObject extends Node {
+  private readonly textures_: Texture[] = [];
+  private readonly transform_ = new TrsTransform();
   private geometry_ = new Geometry();
-  private textures_: Texture[] = [];
-  private transform_ = new TrsTransform();
   private programName_: Shader | null = null;
   private primitive_: Primitive = "triangles";
+  private wireframeOnShaded_ = false;
```
shlomnissan: The "on shaded" part of this variable name is technically a misnomer since we’re not actually shading triangles but it's a widely understood term, so I’m okay with keeping it for clarity

andy-sweet:
Related issue: https://github.com/chanzuckerberg/idetik/issues/47

packages/core/src/renderers/webgl_renderer.ts
```diff
@@ -85,10 +75,54 @@ export class WebGLRenderer extends Renderer {
...
+          this.gl,
+          shaderCode[type].vertex,
+          shaderCode[type].fragment
+        )
+      );
+    }
+    return this.shaders_.get(type)!;
+  }
+
+  private drawObject(
```
andy-sweet, line 96: Nit: I typically find that single use private helper methods being defined directly under the method that uses them improves readability. It would also make the diff in this file much easier to read too!

aganders3:
I support adding wireframe rendering for debugging (and other) purposes, but I'm not totally sold on the implementation here. It feels like a pretty specific feature to be added to all renderable objects/geometry. Instead I would have expected a new `Wireframe` renderable object, perhaps with a factory function to take geometry from an existing renderable (or with `generateWireframeIndexData` as a pure free function). This would avoid needing to modify the renderer to render certain objects twice, but I guess may add unnecessary debug-object bookkeeping to layers.

Another option off the top of my head is to make this a more generic "debug mode" for renderables, which they could override. I say this because we might want other debug features besides wireframes. For example it may be useful to map texcoords -> colors, or render a test pattern instead of an image texture.

I don't feel too strongly about this as it's not really slated to be a public feature, so happy to see it go in as-is if people are happy with it. As far as code goes I think this looks good!

packages/core/src/core/geometry.ts
```diff
@@ -67,11 +70,59 @@ export class Geometry extends Node {
...
+
+    const edgeSet = new Set<string>();
+    const wireframeIndices: number[] = [];
+    const addEdge = (a: number, b: number) => {
+      // Normalize edge order and use a set to deduplicate,
+      // since shared edges between triangles would otherwise
+      // be added multiple times.
+      const i0 = Math.min(a, b);
+      const i1 = Math.max(a, b);
+      const key = `${i0},${i1}`;
```
andy-sweet: Why not use a pair of numbers as the key type to avoid conversion to a string here?

packages/core/src/core/renderable_object.ts
```diff
@@ -32,6 +33,16 @@ export abstract class RenderableObject extends Node {
 
   public set geometry(geometry: Geometry) {
     this.geometry_ = geometry;
+    if (this.wireframeOnShaded_) {
+      this.generateWireframeIndicesIfNeeded(true);
```
andy-sweet: Why do we need to force regeneration of the wireframe indices when setting the geometry here? As I understand it, the wireframe indices are stored on the geometry and are either empty (in which case they will be generated by `generateWireframeIndicesIfNeeded`) or should be consistent with that geometry.

packages/core/src/core/renderable_object.ts
andy-sweet: Optional: personally I find the name hard to read, so if it's a misnomer too, I wonder if we can just use something simpler like `hasWireframe` or `wireframeEnabled`.

andy-sweet (approved):
No blocking feedback, so approved. Just some questions and optional comments. Looking forward to seeing the wireframes in the follow-up!

shlomnissan:
@aganders3 I think that wireframe rendering is something that should apply to all geometries, and since every renderable is fundamentally tied to a geometry (as implied by the base class), it felt appropriate to add the logic there.

That said, I agree that having a dedicated wireframe geometry object that can be instantiated from an existing geometry is cleaner, especially in terms of managing vertex/index buffers. However, I decided to avoid it because renderable objects may be composed of multiple geometries (e.g., a segmented model), and we don’t want to manually manage wireframe versions of each geometry in such cases. This is likely why Three.js supports a "wireframe" property on standard materials (along with a wireframe geometry object.)

Regarding the idea of a more generic debug mode—I’m definitely aligned with that direction, but I’d prefer to keep things simple until we have a second concrete debug feature to support. At that point, we can refactor around a common mechanism with real use cases in mind.

Like you, I don’t feel too strongly about this implementation. It is simple enough in my view to change later on if it's causing issues, however, if there’s strong consensus around an alternative approach, I’m more happy to rewrite this :)

### Three.js references

If the material has the wireframe property set to true, the wireframe index buffer is used, and if it doesn't exist, it will be generated on the fly. This is the approach I'm exploring here

https://github.com/mrdoob/three.js/blob/dev/src/renderers/WebGLRenderer.js#L1111

Three.js also supports wireframe geometries, And while they do support groups, you do need to configure your segmented meshes accordingly

https://github.com/mrdoob/three.js/blob/dev/src/geometries/WireframeGeometry.js

packages/core/src/renderers/webgl_renderer.ts
shlomnissan: That makes sense. My background is in C/C++, where there's no hoisting and access specifiers like public, protected, and private are organized into blocks. But I'm happy to make the change...

packages/core/src/core/geometry.ts
shlomnissan: Good call. I’m not used to having a generic dictionary that accepts any key type without needing custom hash functions. In my own engine, I went with cantor pairing since I’m trying to avoid dynamic string generation whenever possible. I don't think it's a big deal here, but yeah, in this case, using a pair of numbers works well.

packages/core/src/core/renderable_object.ts
shlomnissan: Good call! Not sure how I missed this :/

PR #296, opened 2025-07-01
feat: add LOD fallback and cleanup logic to ChunkManager and ImageLayer

phoenixAja:
### Summary

This PR implements a fallback mechanism for tile-based image loading in `ChunkManagerSource` and integrates proper cleanup in `ImageLayer`. When the current LOD is not yet fully loaded, visible chunks from the lowest-resolution LOD will be rendered underneath to reduce perceptual gaps.

### Key Changes

- **`ChunkManagerSource.getChunks()`**:
  - Returns both current LOD and lowest-res fallback chunks, unless the current LOD *is* the fallback.
  - Ensures fallback tiles are visible until replaced.
  
- **`loadVisibleChunks()`**:
  - Loads both current and background (lowest-res) LOD chunks in parallel.
  - Selectively processes only those two LOD levels.
  
- **`getFallbackChunks()`**:
  - Returns only *visible + loaded* tiles at the lowest LOD (no overlap filtering).
  
- **`loadBackgroundLOD()`**:
  - Loads only visible, unloaded chunks at lowest LOD.
  - Leaves others untouched to reduce unnecessary IO.
  
- **`unloadUnneededLODs()`**:
  - Removes all loaded chunks from intermediate LODs when switching LOD level.
  - Preserves only current and fallback levels.

- **`ImageLayer` integration**:
  - Drops renderables for chunks no longer in `getChunks()` (e.g., LOD switch or fallback removal).

### Rationale

This fallback strategy:
- Improves perceived rendering continuity during navigation and zoom.
- Prevents rendering gaps during high-LOD load delays.
- Avoids excessive memory use by cleaning up intermediate-resolution tiles.

### Testing

- tested chunk_streaming example, verified that correct chunks were being returned (in getChunks) and that there were no blank spots when zooming in and out


packages/core/src/core/chunk_manager.ts
```diff
@@ -94,17 +96,75 @@ export class ChunkManagerSource {
...
+    const backgroundChunks = this.chunks_.filter(
+      (chunk) => chunk.lod === this.lowestResLOD_
+    );
+    const allVisibleBackgroundLoaded = backgroundChunks.every(
+      (chunk) => chunk.state === "loaded"
+    );
+
+    if (!allVisibleBackgroundLoaded) {
+      this.loadBackgroundLOD();
+    }
```
andy-sweet: Based on the current implement of `loadBackgroundLOD`, I think this could be simplified to

```suggestion
    this.loadBackgroundLOD();
```

without any change in behavior.

I'm not sure if `loadBackgroundLOD` will change in the future, but simplifying things could help keep the chunk manager easier to understand.

packages/core/src/core/chunk_manager.ts
```diff
@@ -94,17 +96,75 @@ export class ChunkManagerSource {
...
+          .then(() => {
+            chunk.state = "loaded";
+          })
+          .catch((error) => {
+            Logger.error(
+              "ChunkManager",
+              `Error loading background chunk (${chunk.chunkIndex?.x},${chunk.chunkIndex?.y}): ${error}`
+            );
+            chunk.state = "unloaded";
+          });
```
andy-sweet: Other than the string `background` in the log message, this is just the same as

```suggestion
        this.processChunkData(chunk);
```

Given that chunk manager is already quite complicated and may only get more complicated in the future, I'd lean towards reusing private helper methods and keeping things concise when possible.

packages/core/src/core/chunk_manager.ts
```diff
@@ -94,17 +96,75 @@ export class ChunkManagerSource {
...
     for (const chunk of this.chunks_) {
-      this.processChunkData(chunk);
+      // Only load chunks for current LOD
+      if (chunk.lod === this.currentLOD_ && chunk.state === "unloaded") {
+        this.processChunkData(chunk);
+      }
+    }
+  }
+
+  private getFallbackChunks(): ImageChunk[] {
```
andy-sweet: Unless you intend to reuse this very soon, I would probably inline it where it's used in `getChunks` too avoid to much indirection.

andy-sweet (commented):
I think the basic behavior of falling back to the lowest LOD is fine as a first pass.

But I also think the code here is more complicated than it needs to be. I suggested some simplifications, but maybe hold off on those before @shlomnissan takes a look.

packages/core/src/core/chunk_manager.ts
```diff
@@ -94,17 +96,63 @@ export class ChunkManagerSource {
...
+    if (this.currentLOD_ === this.lowestResLOD_) {
+      return currentLODChunks;
+    }
+
+    const fallbackChunks = this.getFallbackChunks();
+
+    return [...fallbackChunks, ...currentLODChunks];
   }
 
   public loadVisibleChunks() {
```
shlomnissan: I think we may want to mark this function private.

packages/core/src/core/chunk_manager.ts
shlomnissan: I agree with Andy's suggestion. It seems like we don't need the checks leading up to `this.loadBackgroundLOD`.

I'm not convinced we need to modify `loadVisibleChunks` at all. It already iterates through `this.chunks_`. We can modify `getChunks` to always include the highest LOD level (lowest resolution) to minimize changes.

packages/core/src/core/chunk_manager.ts
shlomnissan: ^^ 100%

packages/core/src/core/chunk_manager.ts
```diff
@@ -148,8 +194,24 @@ export class ChunkManagerSource {
     );
 
     if (targetLOD !== this.currentLOD_) {
-      console.debug(`LOD changed from ${this.currentLOD_} to ${targetLOD}`);
       this.currentLOD_ = targetLOD;
+
+      // Unload chunks from all LODs except current and background
+      this.unloadUnneededLODs();
```
shlomnissan: I'm a bit concerned about the concept of "unloading" data in this context. Given the current setup, is there any compelling reason to discard data once it's been uploaded to the GPU? From what I can tell, there isn’t.

That said, proper data unloading requires a more deliberate approach. Specifically, we need to think through:
- When is it appropriate to unload chunk data from CPU memory?
- When should GPU resources be released?
- How can we effectively reuse existing renderable objects and texture buffers?

Unless we're currently facing performance bottlenecks or running into memory constraints, I would suggest removing all unloading logic for now. We can treat it as a dedicated follow-up project when we have clearer requirements and use cases.

packages/core/src/layers/image_layer.ts
```diff
@@ -77,13 +77,15 @@ export class ImageLayer extends Layer {
...
+    // This loop removes image renderables for chunks that are no longer visible
+    // or no longer returned by getChunks() (e.g., due to LOD changes).
     // While this approach works for now, it may be more efficient in the future
     // to reuse renderables by updating their underlying data instead of repeatedly
     // creating new texture objects. Note: GPU resources are not currently being
     // released, so this will also need to be addressed soon.
+    const currentChunks = new Set(this.chunkManagerSource_.getChunks());
     this.visibleChunks_.forEach((image, chunk) => {
-      if (!chunk.visible) {
+      if (!chunk.visible || !currentChunks.has(chunk)) {
```
shlomnissan: - Is there a reason to check for visibility AND `!currentChunks.has(chunk)`?
- If we are keeping `currentChunks` I suggest we avoid calling `this.chunkManagerSource_.getChunks()` twice.

shlomnissan (changes requested):
I'm blocking because I'm not fully convinced that unloading image data from memory is the right direction at this stage—unless I'm missing some key context. I also see an opportunity to simplify the implementation overall, as @andy-sweet suggested.

From my perspective, the only piece that really needs to change is the logic inside `getChunks()`. I don’t yet see a reason to introduce the concepts of fallback chunks or background loading. If `getChunks()` collects all the chunks that should be loaded, then `loadVisibleChunks()` should simply load those as needed, including the low resolution chunks.

That said, you’ve spent more time in this part of the code, so please feel free to push back if there’s something I’m overlooking.

For reference, I think the C++ POC had a fallback strategy that was more robust. It prioritized low-res, but it also kept the previous LOD chunks that are still visible (assuming it's not the low-res ones) until the current LOD chunks are loaded. We should consider this approach as first pass for Idetik too:

```cpp
auto ChunkManager::GetChunks() -> std::vector<Chunk*> {
    std::vector<Chunk*> chunks;

    // 1. always include low-res tiles
    for (auto& chunk : chunks_[max_lod_]) {
        if (chunk.visible && chunk.State() == ChunkState::Loaded) {
            chunks.push_back(&chunk);
        }
    }

    // 2. if currently viewing low-res tiles, no need to include anything else
    if (curr_lod == max_lod_) return chunks;

    // 3. collect all visible chunks from the current LOD
    bool all_loaded = true;
    for (auto& chunk : chunks_[curr_lod]) {
        if (chunk.visible) {
            chunks.push_back(&chunk);
            if (chunk.State() != ChunkState::Loaded) {
                all_loaded = false;
            }
        }
    }

    // 4. if chunks from current LOD are not fully loaded, add visible chunks from previous LOD
    if (!all_loaded && curr_lod != max_lod_) {
        for (auto& chunk : chunks_[prev_lod]) {
            if (chunk.visible && chunk.State() == ChunkState::Loaded) {
                chunks.push_back(&chunk);
            }
        }
    }

    return chunks;
}
```

packages/core/src/core/chunk_manager.ts
phoenixAja: That makes sense, but getChunks() only returns chunks—it doesn't trigger any loading. If we remove the call to loadBackgroundLOD() from loadVisibleChunks(), we risk the fallback chunks never getting loaded unless we restructure how loading is triggered. I'm okay simplifying the logic, but I think it's safer to keep background loading explicit for now.

packages/core/src/core/chunk_manager.ts
phoenixAja: my thinking was that if you weren't actively looking at a chunk  you should unload it, but maybe this isn't as much of a big deal as i initially thought it would be. I'll remove the unload logic

packages/core/src/core/chunk_manager.ts
shlomnissan: Ah, that makes sense now. I got confused because I was reviewing the POC implementation, which relies on an implicit loading order — it always begins with the lowest LOD, so we can safely assume loading was triggered. Given that, your approach seems reasonable. Apologies for the misunderstanding :)

packages/core/src/core/chunk_manager.ts
```diff
@@ -123,8 +156,6 @@ export class ChunkManagerSource {
   }
 
   private processChunkData(chunk: ImageChunk): void {
-    if (!chunk.visible || chunk.state !== "unloaded") return;
```
shlomnissan, line 126: I think that removing this check means that we load chunks that are not visible as well? For what it's worth, I do think the visibility check need to move to `loadVisibleChunks`.

packages/core/src/layers/image_layer.ts
phoenixAja: yes since we are already filtering for visible chunks in getChunks i can remove that check and just keep !currentChunks.has(chunk)

packages/core/src/core/chunk_manager.ts
phoenixAja: moved to loadVisibleChunks

packages/core/src/core/chunk_manager.ts
```diff
@@ -94,17 +96,52 @@ export class ChunkManagerSource {
...
         chunk.visible &&
         chunk.state === "loaded"
     );
+
+    // If we're at the lowest resolution LOD, only return current LOD chunks
+    if (this.currentLOD_ === this.lowestResLOD_) {
+      return currentLODChunks;
+    }
+
+    const fallbackChunks = this.chunks_.filter(
```
shlomnissan: I think that `lowResChunks` might be a more descriptive name? Also, we're using the term "fallback chunks" and "background LODs." Whatever name we go for, maybe we pick one for consistency?

packages/core/src/core/chunk_manager.ts
```diff
@@ -94,17 +96,52 @@ export class ChunkManagerSource {
...
+        chunk.lod === this.lowestResLOD_ &&
+        chunk.visible &&
+        chunk.state === "loaded"
+    );
+    return [...fallbackChunks, ...currentLODChunks];
   }
 
-  public loadVisibleChunks() {
+  private loadVisibleChunks() {
+    // Load background LOD (lowest resolution) first if not already started
```
shlomnissan: If you rename this method to `this.loadLowResChunks()` or something along these lines, you could get rid of this comment.

packages/core/src/core/chunk_manager.ts
```diff
@@ -94,17 +96,52 @@ export class ChunkManagerSource {
...
+
+  private loadBackgroundLOD(): void {
+    const backgroundChunks = this.chunks_.filter(
+      (chunk) => chunk.lod === this.lowestResLOD_
+    );
+
+    for (const chunk of backgroundChunks) {
+      if (chunk.state === "unloaded") {
+        this.processChunkData(chunk);
+      }
```
shlomnissan: Combine these two loops?
```suggestion
  private loadBackgroundLOD(): void {
    for (const chunk of this.chunks_) {
      if (chunk.lod === this.lowestResLOD_ && chunk.state === "unloaded") {
        this.processChunkData(chunk);
      }
    }
  }
```

shlomnissan (approved):
Thanks for taking the time to iterate on this. I feel like the code is much easier to follow now.

I left a couple of comments that would be nice to address before merging, but I don't consider them blocking.

PR #297, opened 2025-07-02
Add general overlays to Idetik with scalebar and FPS overlay examples

andy-sweet:
### Summary
This adds a general concept of overlays to Idetik. Overlays are updated at the end of an Idetik render pass. This allows applications that use Idetik to update content they are responsible for in-sync with the Idetik render loop.

One example of an overlay that we need to support soon is a scale-bar. These changes introduce a simple scale-bar in the OME-Zarr u16 example to show one way how that can be done. [Another branch that builds on this](https://github.com/chanzuckerberg/idetik/compare/andy-sweet/scalebar-overlay...andy-sweet/scalebar-overlay-react?expand=1) shows how Idetik's React component can do that.

These changes also introduce a crude frames-per-second (FPS) overlay as a different example that uses the render call's newly added timestamp.

The main downside of this approach is that it allows users of the Idetik runtime to run arbitrary code in the Idetik render loop, which could have significant performance implications.

### Related Issue
Related to #191

### Tests & Checks
I manually tested the examples where I added overlays to check for expected behavior (e.g. the overall scale of the image).

### DIFF
--- a/packages/core/examples/chunk_streaming/fps_overlay.ts
+++ b/packages/core/examples/chunk_streaming/fps_overlay.ts
@@ -1,0 +1,31 @@
+type FpsOverlayProps = {
+  textDiv: HTMLDivElement;
+};
+
+export class FpsOverlay {
+  textDiv_: HTMLDivElement;
+  msSinceLastUpdate_: number = 0;
+  lastTimestamp_?: DOMHighResTimeStamp;
+
+  constructor(props: FpsOverlayProps) {
+    this.textDiv_ = props.textDiv;
+  }
+
+  public update(_idetik: unknown, timestamp?: DOMHighResTimeStamp): void {
+    if (timestamp === undefined) {
+      return;
+    }
+    if (this.lastTimestamp_ === undefined) {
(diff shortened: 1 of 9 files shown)


andy-sweet:
Reviewers: I chose you because you've already given feedback on this work. Other reviews and comments are welcome though!

packages/core/examples/chunk_streaming/fps_overlay.ts
```diff
@@ -0,0 +1,31 @@
+type FpsOverlayProps = {
+  textDiv: HTMLDivElement;
+};
+
+export class FpsOverlay {
+  textDiv_: HTMLDivElement;
+  msSinceLastUpdate_: number = 0;
```
bchu1: nit - don't need `: number`

bchu1 (approved):

packages/core/examples/chunk_streaming/fps_overlay.ts
andy-sweet: Removed.

PR #299, opened 2025-07-03
Support other integer valued images

### LINKED ISSUE #298: Support other common integer pixel types
Currently, we only support loading and rendering of three image pixel types (uint8, uint16, float32).

Ideally, we should support any data type that is easily supportable by JS and WebGL. At a minimum, we should add int8 and uint32 as somewhat common scientific image pixel types, which are pixel types sometimes used for dense segmentation or label images (see #177).

The goal of this issue is to at least support int8 and uint32 image pixel types.

andy-sweet:
### Summary
This adds support for all integer valued images that can be supported as WebGL textures.

Instead of adding a pair of new image fragment shaders to handle signed ints, I defined template shaders to handle signed int, unsigned int, and float valued textures that depend on `#define` directives being injected into the source at runtime.

### Related Issue
Closes #298

### Tests & Checks
- I modified an existing example to add an int8 mask overlay and tested that.
- I manually tested the existing examples.


andy-sweet:
Still a work in progress because I'm unsure if a shader template + search/replace the defines is a good way to go here. I also need to test the newly added integer types other than int8.

andy-sweet:
@shlomnissan : I requested you as a reviewer because you've worked on more sophisticated versions of this and already made some related comments. Other reviews and comments are welcome.

packages/core/src/renderers/webgl_shader_program.ts
```diff
@@ -27,11 +28,27 @@ export class WebGLShaderProgram {
...
     this.link();
   }
 
+  private replaceSourceDefines(
+    source: string,
+    defines?: Map<string, string>
+  ): string {
+    if (!defines) return source;
+    for (const [key, value] of defines) {
+      source = source.replace(key, value);
```
andy-sweet: This is pretty fragile. I would prefer to just insert the `#define X Y` strings into the beginning of the shader source, but I didn't want to deal with weird corner cases related to the requirement of the version statement needing to be the first statement (e.g. shaders defined one line). Maybe we can make stronger assumptions about our shader sources?

andy-sweet:
@aganders3 : you might want to take a look at this you because I remember you had some thoughts about reusing shader code.

packages/core/src/renderers/shaders/scalar_image_frag.glsl
```diff
@@ -4,7 +4,9 @@ precision mediump float;
 
 layout (location = 0) out vec4 fragColor;
 
-uniform mediump usampler2D texture0;
+// SAMPLER_TYPE must be defined by the application using this shader.
+uniform mediump SAMPLER_TYPE texture0;
```
andy-sweet: When the underlying texture data is int32 or uint32, this might need to be highp instead of mediump to ensure that we're retaining the precision of those values. I'm not quite sure what WebGL2 does when you pass through a texture of uniform with higher-precision than the shader supports though and I haven't seen this cause an issue with 32-bit integers yet (but maybe I haven't run into a high enough value?).

packages/core/src/renderers/webgl_shader_program.ts
aganders3: I think I'd also prefer using `#define X Y`, but I think I understand your concern here. All our shaders currently use `#version 300 es` so perhaps we just include that along with any `#define` lines at the beginning?

aganders3 (commented):
In general this looks good, but I'd like to see what using the built-in preprocessor (#defines) looks like.

We could also use maps or something to clean up some of the long if/else and switch cases, but I don't mind them this way either.

packages/core/src/renderers/shaders/scalar_image_frag.glsl
aganders3: I don't _think_ the precision matters for integer textures unless we're doing linear filtering, but we might want `highp` for 32-bit floats. If not we should consider rescaling/recasting on the CPU to save GPU memory.

packages/core/examples/image_mask_overlay/main.ts
```diff
@@ -0,0 +1,95 @@
...
+  OrthographicCamera,
+  Region,
+  Color,
+  ImageSeriesLayer,
+} from "@";
+
+const imageUrl =
+  "https://files.cryoetdataportal.cziscience.com/10000/TS_041/Reconstructions/VoxelSpacing13.480/Tomograms/100/TS_041.zarr";
+const labelsUrl =
+  "https://files.cryoetdataportal.cziscience.com/10000/TS_041/Reconstructions/VoxelSpacing13.480/Annotations/114/membrane-1.0_segmentationmask.zarr";
```
shlomnissan: Nit
```suggestion
const baseUrl =
  "https://files.cryoetdataportal.cziscience.com/10000/TS_041/Reconstructions/VoxelSpacing13.480";
const imageUrl = `${baseUrl}/Tomograms/100/TS_041.zarr`;
const labelsUrl = `${baseUrl}/Annotations/114/membrane-1.0_segmentationmask.zarr`;
```

packages/core/examples/image_mask_overlay/main.ts
```diff
@@ -0,0 +1,95 @@
...
+
+const imageSource = new OmeZarrImageSource(imageUrl);
+const labelsSource = new OmeZarrImageSource(labelsUrl);
+
+const loader = await imageSource.open();
+const attributes = await loader.loadAttributes();
+const lods = attributes.length;
+const attributesForLastLod = attributes[lods - 1];
```
shlomnissan: We might want to add a computed property to retrieve the LOD, so instead of having `loadAttributes` return an array, it could return an object that provides access to the LOD number and perhaps even something like `attributesForLastLOD`. Also, the repeated use of `lods - 1` feels a bit redundant. I personally find it clear, but I’d also be fine assigning it to a variable for clarity.

packages/core/examples/image_mask_overlay/main.ts
```diff
@@ -0,0 +1,95 @@
...
+
+const zDimName = "z";
+const zAxisIndex = attributesForLastLod.dimensionNames.findIndex(
+  (dim) => dim === zDimName
+);
+const zScale = attributesForLastLod.scale[zAxisIndex];
+const zMin = 50 * zScale;
+const zMax = 75 * zScale;
```
shlomnissan: Where do these numbers come from? Maybe add a comment?

packages/core/examples/image_mask_overlay/main.ts
```diff
@@ -0,0 +1,95 @@
...
+    layer.setPosition(value);
+  }
+});
+
+for (const layer of layers) {
+  layer.preloadSeries();
+  layer.setPosition(slider.valueAsNumber);
+}
+
+const camera = new OrthographicCamera(0, 240 * zScale, 0, 232 * zScale);
```
shlomnissan: Same question about these numbers. I know this is just an example, but I have no idea how we pick these numbers by reading this file. Comments will help clarify.

packages/core/src/objects/renderable/image_renderable.ts
```diff
@@ -114,11 +114,37 @@ export class ImageRenderable extends RenderableObject {
...
       throw new Error("un-textured image not implemented");
     } else if (texture.type == "Texture2D") {
-      this.programName =
-        texture.dataType == "float" ? "floatImage" : "uintImage";
+      if (
+        texture.dataType == "byte" ||
+        texture.dataType == "int" ||
+        texture.dataType == "short"
+      ) {
+        this.programName = "intImage";
```
shlomnissan: I see this a bit differently. Since we're always using the same program and just passing different attributes to it, it makes more sense to define a `fragmentDefines` map here, rather than selecting different program names to represent the various fragment defines.

packages/core/src/objects/textures/channel.ts
```diff
@@ -69,10 +69,18 @@ function contrastLimitsFromTexture(texture: Texture): [number, number] {
     return [0, 1];
   }
   switch (texture.dataType) {
+    case "byte":
+      return [-128, 127];
+    case "short":
+      return [-32768, 32767];
```
shlomnissan: These ranges feel a bit out of place. I’d consider moving them closer to the data format definition, since they’re closely related.

packages/core/src/renderers/shaders/index.ts
```diff
@@ -13,34 +11,58 @@ export type Shader =
...
+  projectedLine: {
+    vertex: projectedLineVertexShader,
+    fragment: projectedLineFragmentShader,
+  },
+  points: {
+    vertex: pointsVertexShader,
+    fragment: pointsFragmentShader,
+  },
+  // TODO: consolidate image shaders
+  floatImage: {
```
shlomnissan: As mentioned above, all of these entries use the same vertex and fragment shaders, so they correspond to the same program. I think the "fragment defines" should be passed to wherever we resolve them, since each entry here represents a distinct program (i.e., a unique vertex/fragment combination).

packages/core/src/renderers/shaders/scalar_image_array_frag.glsl
```diff
@@ -4,7 +4,9 @@ precision mediump float;
 
 layout (location = 0) out vec4 fragColor;
 
-uniform mediump usampler2DArray texture0;
+// SAMPLER_TYPE must be defined by the application using this shader.
+uniform mediump SAMPLER_TYPE texture0;
```
shlomnissan: I think we should inject preprocessor `#defines` at the top of the shader code to maintain a valid structure, especially when using `#ifdef`/`#if` blocks. (I'm currently using a GLSL linter that flags the program as invalid without them.) This also makes it clearer that these values are defined externally. Otherwise, we’d need to add comments next to each unknown string to explain what they represent.

packages/core/src/renderers/shaders/scalar_image_frag.glsl
andy-sweet: I found this buried in the [GLES 3.0 spec](https://registry.khronos.org/OpenGL/specs/es/3.0/GLSL_ES_Specification_3.00.pdf):

> For the texture sampling functions, the precision of the return type matches the precision of the
sampler type.
```
uniform lowp sampler2D sampler;
highp vec2 coord;
...
lowp vec4 col = texture (sampler, coord); // texture() returns lowp
```

A case where I thought this would matter is if we're using the return value from `texture` (or `texelFetch`?) from the primary image data to look up a color value somewhere else (e.g. in a colormap stored in a uniform or a texture).

packages/core/src/renderers/webgl_shader_program.ts
shlomnissan: That’s a fair point. To avoid breaking the requirement that the `#version` directive must be the first line, what I’ve done in the past is insert a custom pragma like:

```glsl
#pragma inject_attributes
```

immediately after the version directive, and then replace that line with the needed `#define` statements. It keeps the GLSL structure valid and makes the injection point explicit and easy to reason about.

shlomnissan (changes requested):
I think this is a solid starting point, but I’d like to see a couple of things addressed:
- The implementation should produce valid GLSL code using #defines
- The concept of a shader program as a vertex/fragment shader pair should be preserved

I’m requesting changes for now so we can tackle these before merging. That said, I’m happy to move forward with it as long as we commit to a quick follow-up.

packages/core/src/renderers/webgl_shader_program.ts
andy-sweet: I like the `#pragma` idea. I'll implement that.

packages/core/src/renderers/shaders/scalar_image_array_frag.glsl
andy-sweet: Can you share the GLSL linter? I want to at least run it manually anyway to check my changes are good.

packages/core/src/renderers/shaders/index.ts
andy-sweet: > As mentioned above, all of these entries use the same vertex and fragment shaders, so they correspond to the same program.

They are using the same fragment shader template, but they actually need to provide the defines to specify a complete/valid shader, so technically they are not the same program (actually they are not even programs yet, just sources).

I think this is more than just nit-picking because the way `WebGLRenderer.getShaderProgram` is defined (i.e. it creates programs as needed), I think each record the `shaderCode` object must provide everything needed to compile a program rather the program itself (i.e. `WebGLRenderer.getShaderProgram` is the resolution time).

I might be a bit confused here, so I might reach out directly to clarify this. But I also want to avoid falling into a shader-composition rabbit hole in this PR.

packages/core/src/objects/renderable/image_renderable.ts
andy-sweet: Maybe linked to my confusion in [my other comment](https://github.com/chanzuckerberg/idetik/pull/299/files#r2211314622), but if they're defined here, how would `WebGLRenderer` or `WebGLShaderProgram` be aware of them? Would we change the interface `RenderableObject` to also include defines for the shader programs?

andy-sweet:
> I think this is a solid starting point, but I’d like to see a couple of things addressed:
> 
> * The implementation should produce valid GLSL code using #defines
> * The concept of a shader program as a vertex/fragment shader pair should be preserved
> 
> I’m requesting changes for now so we can tackle these before merging. That said, I’m happy to move forward with it as long as we commit to a quick follow-up.

I think I addressed your main concerns with the latest commits, so let me know if I got the general gist. But still need to address some of the smaller ones, which I should be able to do tomorrow morning.

packages/core/src/renderers/shaders/scalar_image_frag.glsl
aganders3: Yeah I think you're right - I was thinking of the precision but it would also affect the range of values.

packages/core/src/renderers/webgl_renderer.ts
```diff
@@ -181,18 +183,20 @@ export class WebGLRenderer extends Renderer {
     this.gl.depthFunc(this.gl.LEQUAL);
   }
 
-  private getShaderProgram(type: Shader) {
-    if (!this.shaders_.has(type)) {
+  private getShaderProgram(props: ProgramProps) {
+    if (!this.shaders_.has(props)) {
```
andy-sweet: I think this will create a WebGLShaderProgram per RenderableObject instance, which is not what we want. Instead we need the props/key to be something that has the same hash when it has the same values (e.g. like a string composited from the props).

packages/core/src/renderers/shaders/index.ts
```diff
@@ -14,10 +12,14 @@ export type Shader =
...
   | "points"
   | "wireframe"
-  | "floatImage"
-  | "floatImageArray"
-  | "uintImage"
-  | "uintImageArray";
+  | "scalarImage"
+  | "scalarImageArray";
+
+export type ProgramProps = {
```
andy-sweet: The terminology here is bit confusing.

`Shader` defines a name that identifies a program's shaders' source code.

Whereas `ProgramProps` defines the properties needed to define a program (i.e. the shaders' source code and any defines).

If I'm confused about this, then maybe I should rename `ProgramProps` to `ShaderProps`?

Otherwise, maybe we should consider renaming `Shader` to `Program` and `shaderCode` to `programCode`?

packages/core/examples/image_mask_overlay/main.ts
andy-sweet: Agreed. I'm probably going to add a mask layer to one of the existing examples rather than add an entirely new one so that I can reuse some the example's non-trivial logic, which is better explained.

packages/core/src/objects/renderable/image_renderable.ts
shlomnissan: Yes, I will respond to the other comment.

packages/core/src/renderers/shaders/index.ts
shlomnissan: We already discussed this over Slack but I'm adding information here for future reference.

shlomnissan: Copied directly from Slack:

That’s a reasonable response and I think I might be missing your intent behind this setup. Also my original comment leaned a bit too much into semantics without clearly expressing the actual concern (which isn’t a major one in the short term): Technically, each unique set of `#defines` constitutes a different shader program—what I’d consider a program permutation. While the template remains the same, the defines themselves shouldn't be limited to just data types in the long run. For example, we might introduce defines like `USE_INSTANCING`, `USE_FLATSHADED`, or even ones with specific values like `SCALE` or `SIZE`. Each of these combinations would result in a distinct program, but their configuration will come externally—from the needs of the layer or renderable object—not from within the `shaderCode` object itself.

Another way to put it is that we can treat entries in the `shaderCode` object as program templates, where the configuration of each template should be determined by external properties — the sampler type should depend on the texture's underlying data type.

This is how Three.js injects defines
https://github.com/mrdoob/three.js/blob/dev/src/renderers/webgl/WebGLProgram.js#L534

Before an object is rendered, a "parameters" object is created with different configuration that depends on a variety of things: material (to choose the template), individual configuration for the material, lighting, cameras, unique features, etc. Then, this parameters object is passed to the program which reads these values and injects the right defines accordingly.

packages/core/src/renderers/shaders/scalar_image_array_frag.glsl
shlomnissan: I'm using the [GLSL Lint VSCode plugin](https://marketplace.visualstudio.com/items?itemName=dtoplak.vscode-glsllint). It depends on [glslang](https://github.com/KhronosGroup/glslang) which provides the GLSL code validation so it needs to be installed first. You can either build it from source or download a binary and then point the VSCode plugin to the executable. It's not the best linter, but it's good enough.

packages/core/examples/image_mask_overlay/main.ts
```diff
@@ -41,17 +43,33 @@ const channelProps: ChannelProps[] = [
...
+const imageLayer = new ImageSeriesLayer({
+  source: imageSource,
   region,
   seriesDimensionName: zDimName,
   channelProps,
 });
 
-layer.addStateChangeCallback((newState: LayerState) => {
+imageLayer.addStateChangeCallback((newState: LayerState) => {
   stateEl!.textContent = newState;
```
shlomnissan: I think this this assertion (`!`) is unnecessary.

packages/core/src/core/renderable_object.ts
```diff
@@ -3,16 +3,15 @@ import { Geometry } from "../core/geometry";
...
+import { ProgramProps } from "../renderers/shaders";
 
 export abstract class RenderableObject extends Node {
   public wireframeEnabled = false;
   private readonly textures_: Texture[] = [];
   private readonly transform_ = new TrsTransform();
   private geometry_ = new Geometry();
   private wireframeGeometry_: WireframeGeometry | null = null;
-  private programName_: Shader | null = null;
+  private programProps_: ProgramProps | null = null;
```
shlomnissan: I think `ProgramProps` should live outside the renderable object, since its configuration may depend on other entities in the future. For instance, while we currently apply transparency at the layer level (which may not be ideal), I could see us supporting more advanced blending techniques that require conditional logic in shaders mapped through defines. To support that flexibility, a more common structure would be:

- `ProgramProps` becomes its own class.
- Its constructor accepts a renderable object (and eventually could take in a layer, camera, or other context).
- The constructor is responsible for setting up all the properties that define the shader program permutation.
- During the render loop, the renderer creates a `ProgramProps` instance using the relevant input object.
- This instance is then passed into `WebGLShaderPrograms::get`, as you're doing now.

Also, in a future PR, we could hash the `ProgramProps` values to cache and reuse shader permutations more efficiently.

This isn’t a blocking concern, in the short-term we only need the renderable object to configure program props, but if you’re aligned with this direction, it may be worth making this change now rather than refactoring later.

packages/core/src/objects/renderable/image_renderable.ts
```diff
@@ -49,7 +61,7 @@ export class ImageRenderable extends RenderableObject {
 
   public addTexture(texture: Texture) {
     super.addTexture(texture);
-    this.setProgramName();
+    this.updateProgramProps();
```
shlomnissan: If you construct the program properties in the render loop, you won't need to keep track of updates, and since `ProgramProps` is just a POD, I don't think we need to worry about perf.

packages/core/src/renderers/shaders/index.ts
```diff
@@ -14,10 +12,14 @@ export type Shader =
...
+  name: Shader;
+  vertexDefines?: ReadonlyArray<[string, string]>;
+  fragmentDefines?: ReadonlyArray<[string, string]>;
```
shlomnissan: - I think explicitly defining the possible properties makes the code much easier to understand and maintain. It also leads to more consistent and predictable hashing for program permutations.
- I'm not sure we need to separate `vertexDefines` and `fragmentDefines`
- Also, I think `webgl_shader_program.ts` might be a better place for this type?

With that in mind, I’d suggest refactoring to something like this:

```ts
class ProgramProps {
  public name: string;
  public textureDataType?: ...;

  constructor(name: string, object: RenderableObject) {
    this.name = name;

    if (object.textures.length) {
      // Set textureDataType or other defines
    }
  }
}
```

packages/core/src/renderers/webgl_shader_programs.ts
```diff
@@ -0,0 +1,54 @@
...
+
+function programKey(props: ProgramProps): string {
+  const vertexDefines = props.vertexDefines
+    ? props.vertexDefines.join(",")
+    : "";
+  const fragmentDefines = props.fragmentDefines
+    ? props.fragmentDefines.join(",")
+    : "";
+  return `${props.name}:${vertexDefines}:${fragmentDefines}`;
+}
```
shlomnissan: Ahh yes! You already thought about this :)

The reason I think that hard-coding the properties (instead of using arrays) is better, is because _I think_ the current program key generation is order-dependent?

shlomnissan (commented):
I think we are very close. I left a couple of suggestions to consider/discuss before approving.

packages/core/src/renderers/webgl_shader_programs.ts
andy-sweet: Yes, that's right. I could sort them, but I don't like that as the defines could get pretty large. I like your idea of hard-coding the defines or the properties that effectively specify them, so let me go that route instead.

packages/core/examples/image_mask_overlay/main.ts
andy-sweet: Removed.

packages/core/src/core/renderable_object.ts
andy-sweet: I'm going to hold off on defining a constructor that accepts a `RenderableObject` because that inverts the current design of a `RenderableObject` choosing what program it uses, and is worthy of a more well thought-out change than this PR.

Ultimately, I want it to be easy to reason about how a renderable object is rendered (e.g. I want to be able to a read its shader code without too much mental gymnastics). I think this PR has already drifted away from that, and I think the suggestion above might make things even worse without careful consideration.

I'm going to the current form of this PR as far as it go and then also consider rolling some of these changes back a bit in a less dramatic PR.

andy-sweet:
After experimenting with doing something a bit more complex (see my [shader-programs](https://github.com/chanzuckerberg/idetik/compare/andy-sweet/shader-programs?expand=1) branch), I decided to roll this back to something closer its original crude form.

Compared to that original form, I use `#define` directives instead of search/replace, which should be more robust. The `shaderCode` record remains verbose, but is easy enough to reason about.

If we need something more complex (i.e. as we add more defines and variants of shaders), I think that effort warrants more design thought than can be captured in this one PR.

I appreciate the comments here, which helped me understand the current state of things a little better. Agreed with @shlomnissan that `RenderableObject.programName` doesn't really belong in `RenderableObject`, I think because it couples that to some kind of GL renderer (through the GLSL shader sources). But again, that feels out of scope for the goal here. I think a better time to address that is when we get more serious about introducing a second renderer (e.g. WebGPU).

shlomnissan (approved):

andy-sweet:
@aganders3 : as you wanted to see how this looks with `#define` directives and there have been some other updates + discussion, I'll wait for you to take another look before merging. I also modified an example that I think you originally wrote instead of introducing a new one - let me know if you'd like to change that.

packages/core/src/renderers/webgl_shader_programs.ts
```diff
@@ -0,0 +1,51 @@
...
+    return program;
+  }
+}
+
+function replaceSourceDefines(
+  source: string,
+  defines?: ReadonlyArray<[string, string]>
+): string {
+  if (defines === undefined) return source;
+  const definesSource = defines
```
aganders3: I'm not totally sure this would work nor how valuable it may be, but there's also a `#line` [GLSL preprocessor directive](https://www.khronos.org/opengl/wiki/Core_Language_(GLSL)#Preprocessor_directives) you might be able to add here with something like:

`#line __LINE__ - {defines.length}`

packages/core/src/renderers/shaders/index.ts
```diff
@@ -14,40 +12,59 @@ export type Shader =
...
-    uintImageArray: {
-      vertex: meshVertexShader,
-      fragment: uintImageArrayFragmentShader,
-    },
-  };
+type ShaderCode = {
+  vertex: string;
+  vertexDefines?: ReadonlyArray<[string, string]>;
+  fragment: string;
+  fragmentDefines?: ReadonlyArray<[string, string]>;
```
aganders3: nit: Would it be possible/preferable to use `ReadonlyMap` instead?

aganders3 (approved):
Looks good to me!

packages/core/src/renderers/shaders/scalar_image_frag.glsl
aganders3: I think this would be possible now with the infrastructure you added. You don't need to add it here but maybe create an issue in case we run into it?

packages/core/src/renderers/webgl_shader_programs.ts
andy-sweet: I assume the potential benefit is accurate line numbers in error messages and similar? I'll take a quick look into it.

packages/core/src/renderers/shaders/index.ts
andy-sweet: Sure, it was a ReadonlyMap at some point, so happy to bring it back.

andy-sweet: Oh wait, I was thinking of something else. What's the motivation behind a Map here? In our version of JS, it's slightly more annoying to iterate over a map than an array when this actually gets used.

aganders3: The motivation is purely that it feels more semantic/intuitive, but if it's more annoying in some way no worries.

packages/core/src/renderers/webgl_shader_programs.ts
andy-sweet: I added this, but note there are other things that can cause the line number to be misreported (e.g. the combination of our current import and vite configs seem to remove comments, which can remove lines).

packages/core/src/renderers/shaders/index.ts
andy-sweet: It requires an extra array when generating the string, but changed this to a `ReadonlyMap` to make it easier to think about here.

PR #300, opened 2025-07-07
(Stacked) Render wireframes on top of meshes [2/2]

### LINKED ISSUE #47: Add debug rendering

shlomnissan:
### Summary

This PR finalizes the wireframe-over-shaded-geometry feature. It introduces
dedicated VAO and EBO support for wireframes within WebGL buffers, adds a basic
passthrough shader for rendering flat colors, and completes the necessary
integration in `WebGLRenderer`.

**Next Steps**
I'm calling it done because the functionality is there, but I want to extend it in a follow-up:
- Implement layer-specific logic for displaying debug wireframes.
- Add color option for wireframes - for example, different wireframe colors may be desirable per LOD level.
- Apply polygon offset to ensure wireframes consistently render on top of shaded geometry.

### Related Issue
Closes #47 

### Tests & Checks
- Manually verified — see attached video.
- To enable wireframes, set `wireframeEnabled = true` on renderable objects.
- All existing examples render correctly with no regressions.
- All checks have passed.

https://github.com/user-attachments/assets/bad8dc00-4261-4d5f-bd3f-73b363eeaee4

### DIFF
--- a/packages/core/src/core/renderable_object.ts
+++ b/packages/core/src/core/renderable_object.ts
@@ -5,15 +5,15 @@ import { TrsTransform } from "../core/transforms";
 
 import { Shader } from "../renderers/shaders";
 
-export type Primitive = "triangles" | "points";
+export type Primitive = "triangles" | "points" | "lines";
 
 export abstract class RenderableObject extends Node {
+  public wireframeEnabled = false;
   private readonly textures_: Texture[] = [];
   private readonly transform_ = new TrsTransform();
   private geometry_ = new Geometry();
   private programName_: Shader | null = null;
   private primitive_: Primitive = "triangles";
-  private wireframeEnabled_ = false;
 
   public addTexture(texture: Texture) {
     this.textures_.push(texture);
@@ -33,16 +33,6 @@ export abstract class RenderableObject extends Node {
 
   public set geometry(geometry: Geometry) {
     this.geometry_ = geometry;
-    if (this.wireframeEnabled_) {
-      this.generateWireframeIndicesIfNeeded();
-    }
-  }
-
-  public set wireframeOnShaded(enabled: boolean) {
-    this.wireframeEnabled_ = enabled;
-    if (this.wireframeEnabled_) {
-      this.generateWireframeIndicesIfNeeded();
-    }
   }
 
   public get programName(): Shader {
@@ -62,14 +52,6 @@ export abstract class RenderableObject extends Node {
 
   protected set primitive(primitive: Primitive) {
     this.primitive_ = primitive;
-  }
-
-  private generateWireframeIndicesIfNeeded() {
-    const hasIndexData = this.geometry_.indexData.length > 0;
-    const hasWireframeIndexData = this.geometry_.wireframeIndexData.length > 0;
-    if (hasIndexData && !hasWireframeIndexData) {
-      this.geometry_.generateWireframeIndexData();
-    }
   }
 
   /**
--- a/packages/core/src/renderers/shaders/wireframe_frag.glsl
+++ b/packages/core/src/renderers/shaders/wireframe_frag.glsl
@@ -1,0 +1,11 @@
+#version 300 es
+
+precision mediump float;
+
+layout (location = 0) out vec4 fragColor;
+
+uniform float u_opacity;
+
+void main() {
+    fragColor = vec4(1.0, 1.0, 1.0, u_opacity);
+}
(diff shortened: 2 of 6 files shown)


packages/core/src/core/renderable_object.ts
```diff
@@ -64,14 +54,6 @@ export abstract class RenderableObject extends Node {
     this.primitive_ = primitive;
   }
 
-  private generateWireframeIndicesIfNeeded() {
```
shlomnissan, line 67: I moved the deferred generation of wireframe index buffers to the `GLBuffer` class.

packages/core/src/renderers/shaders/wireframe_frag.glsl
```diff
@@ -0,0 +1,11 @@
+#version 300 es
+
+precision mediump float;
+
+layout (location = 0) out vec4 fragColor;
+
+uniform float u_opacity;
+
+void main() {
+    fragColor = vec4(1.0, 1.0, 1.0, u_opacity);
```
shlomnissan, line 10: Just pass-through shaders for now, but I intend to add a color uniform for customized debugging.

packages/core/src/renderers/webgl_renderer.ts
```diff
@@ -132,8 +137,13 @@ export class WebGLRenderer extends Renderer {
       }
     }
 
-    const primitive = this.getGLPrimitve(object.primitive);
-    const index = object.geometry.indexData;
+    let index = object.geometry.indexData;
+    let primitive = this.getGLPrimitve(object.primitive);
```
andy-sweet, line 141: I just noticed the typo in `getGLPrimitve` (it's missing the last i). This isn't directly related to this PR, so need to fix it here. But if you don't, I'll open a separate PR to do so.

packages/core/src/renderers/webgl_renderer.ts
```diff
@@ -88,14 +88,19 @@ export class WebGLRenderer extends Renderer {
...
+      this.bindings_.bindObject(object, "lines");
+      const wireframeProgram = this.getShaderProgram("wireframe").use();
+      this.drawObject(layer, object, true, wireframeProgram);
+    }
   }
 
   private drawObject(
     layer: Layer,
     object: RenderableObject,
+    wireframe: boolean,
```
andy-sweet, line 103: I don't love the boolean parameter here because it effectively adds branching to this function. One thought is to refactor `drawObject` into `setProgram` (most of `drawObject` now) and then inline the actual draw calls into `renderObject` for the main object and its wireframe.

After describing it, I don't like that much more, so this is non-blocking and just something for your consideration.

packages/core/src/renderers/webgl_buffers.ts
```diff
@@ -5,31 +5,47 @@ type BufferHandles = {
...
-
-    if (!this.buffers_.has(object)) {
-      this.generateBuffers(object);
+  public bindObject(
+    object: RenderableObject,
+    indexType: IndexType = "triangles"
+  ) {
+    let buffers = this.buffers_.get(object);
+    if (buffers) {
+      const vao = indexType === "lines" ? buffers.wireframeVao : buffers.vao;
```
andy-sweet, line 29: The comparison of `indexType === "lines"` is used to branch this function twice. I'd pass through a boolean here instead to indicate the real intent of this branch, which is to bind the wireframe buffers instead of main object buffers. This is also protective if we later did want to support a renderable object that uses lines for rendering its main geometry (though that probably would have/allow wireframes).

In this case, I feel more strongly that defining two functions is the way to go here (i.e. `bindObject` and `bindObjectWireframes`) because so much of the function is dependent on that condition.

andy-sweet (changes requested):
Pretty close to approving, especially as this is a debug behavior that is not enabled. But there are a few places in the code that are now hard to follow in the non-debug case. I had some questions and suggestions. I'll approve if you accept some form of those or can explain them away.

andy-sweet:
Also, I noticed some slightly strange behavior in the chunk streaming where the triangles for the finest-resolution level are not shown. I'm not exactly sure how to reproduce. And I'm also not sure if that's a bug here or in chunk-streaming. Will try to follow-up with a video tomorrow.

aganders3:
> I noticed some slightly strange behavior in the chunk streaming where the triangles for the finest-resolution level are not shown.

I was seeing the same thing, but it goes away when I rebase this branch (cleanly) on `main`. Maybe it's a chunk streaming bug that was fixed by #296? I'm not sure why that doesn't show up in the changes here.

PR #301, opened 2025-07-08
Add a ScaleBar React component to OmeZarrImageViewer

### LINKED ISSUE #191: Add a scale bar

andy-sweet:
### Summary
This defines an internal ScaleBar component in the Idetik React package and uses it in the OmeZarrImageViewer component. The scale bar's visibility and alignment can be configured using the viewer's props. By default it is visible (because an OME-Zarr often has a useful units) and left-aligned to match the general layout.

I needed to reconfigure the div hierarchy in the OmeZarrImageViewer component to handle resizing somewhat sensibly, especially at small widths (e.g. mobile devices). I decided to place the scale bar in the bottom-left corner for now, because it seemed like the least disruptive option that was also good enough for the OrganelleBox portal launch.

I also allow units to be optionally returned when loading an OME-Zarr's attributes. In the React component, these units are fetched and updated when fetch the z-axis metadata.

https://github.com/user-attachments/assets/b359a5ce-c540-4e34-bdd9-9b4d5737f19e

### Related Issue
Closes #191

### Tests & Checks
I manually tested this with the main component and asserted that the described length roughly matches the pixel sizing/spacing described in the corresponding OME-Zarr file. I also manually created a more exact test using a Python script to create and then serve an OME-Zarr with integer pixel spacing.

### DIFF
--- a/packages/react/src/components/viewers/OmeZarrImageViewer/OmeZarrImageViewer.tsx
+++ b/packages/react/src/components/viewers/OmeZarrImageViewer/OmeZarrImageViewer.tsx
@@ -366,18 +366,12 @@ export function OmeZarrImageViewer(props: OmeZarrImageViewerProps) {
             "flex",
             "flex-col",
             "m-sds-l",
-            "p-sds-xs",
-            // TODO: decide if style should match others.
-            // "bg-black/75",
-            // "backdrop-blur-md",
-            // "rounded-sds-m",
-            // "shadow-sds-m",
             "w-1/5",
             "select-none",
             props.classNames?.scaleBarContainer
           )}
         >
-          <ScaleBar idetik={runtime} unit={unit} />
+          <ScaleBar unit={unit} align={"start"} />
         </div>
         <div
           className={cns(
--- a/packages/react/src/components/viewers/OmeZarrImageViewer/OmeZarrImageViewer.tsx
+++ b/packages/react/src/components/viewers/OmeZarrImageViewer/OmeZarrImageViewer.tsx
@@ -373,7 +373,7 @@ export function OmeZarrImageViewer(props: OmeZarrImageViewerProps) {
             "shadow-sds-m",
             "w-1/5",
             "select-none",
-            props.classNames?.scaleBarContainer,
+            props.classNames?.scaleBarContainer
           )}
         >
           <ScaleBar idetik={runtime} unit={unit} />
--- a/packages/react/src/components/viewers/OmeZarrImageViewer/OmeZarrImageViewer.tsx
+++ b/packages/react/src/components/viewers/OmeZarrImageViewer/OmeZarrImageViewer.tsx
@@ -373,7 +373,6 @@ export function OmeZarrImageViewer(props: OmeZarrImageViewerProps) {
             // "rounded-sds-m",
             // "shadow-sds-m",
             "w-1/5",
-            "box-border",
             "select-none",
             props.classNames?.scaleBarContainer
           )}
--- a/packages/react/src/components/viewers/OmeZarrImageViewer/OmeZarrImageViewer.tsx
+++ b/packages/react/src/components/viewers/OmeZarrImageViewer/OmeZarrImageViewer.tsx
@@ -43,7 +43,10 @@ interface OmeZarrImageViewerProps {
     sliceIndicator?: string;
     load3dButton?: string;
     sliceSliderContainer?: string;
-    scaleBarContainer?: string;
+  };
+  scaleBar?: {
+    visible?: boolean;
+    align?: "start" | "end" | "center";
   };
   onLayerCreated?: () => void;
   onFirstSliceLoaded?: () => void;
@@ -70,6 +73,10 @@ export function OmeZarrImageViewer(props: OmeZarrImageViewerProps) {
     initialIndex = "omeroDefaultZ",
     loadAllButtonText,
     indexIndicatorText,
+    scaleBar = {
+      visible: true,
+      align: "start",
+    }
   } = props;
 
   const { isReady: runtimeIsReady, runtime } = useIdetik();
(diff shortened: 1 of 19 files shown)


andy-sweet:
@bchu1 : I assigned you as a reviewer because you reviewed the previous related PR and for your React, CSS, and SDS expertise. Other comments and reviews are welcome.

packages/react/src/components/viewers/OmeZarrImageViewer/components/ScaleBar/ScaleBar.tsx
```diff
@@ -0,0 +1,208 @@
...
+    >
+      <div
+        ref={textDivRef}
+        className={cns(
+          "text-white",
+          "text-base",
+          "font-sds-code",
+          `text-align-${align}`
+        )}
+        style={{
```
andy-sweet: I'm unsure how to define the equivalent tailwind string here that actually takes an effect, but I would like to do that for consistency.

packages/react/src/components/viewers/OmeZarrImageViewer/components/ScaleBar/ScaleBar.tsx
```diff
@@ -0,0 +1,208 @@
...
+          textShadow:
+            "black 1px 1px 1px, black -1px -1px 1px, black 1px -1px 1px, black -1px 1px 1px",
+        }}
+      ></div>
+      <div
+        ref={lineDivRef}
+        className={cns("bg-white", "h-sds-m")}
+        style={{
```
andy-sweet: I'm unsure how to define the equivalent tailwind string here that actually takes an effect, but I would like to do that for consistency.

packages/react/src/components/viewers/OmeZarrImageViewer/OmeZarrImageViewer.tsx
```diff
@@ -342,82 +358,105 @@ export function OmeZarrImageViewer(props: OmeZarrImageViewerProps) {
       )}
       <div
         className={cns(
+          "flex",
+          "flex-row",
```
bchu1: nit - don't need, `row` is default

packages/react/src/components/viewers/OmeZarrImageViewer/components/ScaleBar/ScaleBar.tsx
```diff
@@ -0,0 +1,208 @@
...
+import { useEffect, useRef } from "react";
+import { Idetik, OrthographicCamera } from "@idetik/core";
+import cns from "classnames";
+import { useIdetik } from "hooks";
+
+function getUnitAbbreviation(unit: string | undefined): string {
+  if (unit === undefined) return "";
+  // From OME-NGFF v0.4 specification:
+  // https://ngff.openmicroscopy.org/0.4/#axes-md
+  switch (unit.toLowerCase()) {
```
bchu1: nit - could be a map, but performance prolly doesn't matter here?

packages/react/src/components/viewers/OmeZarrImageViewer/components/ScaleBar/ScaleBar.tsx
```diff
@@ -0,0 +1,208 @@
...
+  }
+}
+
+type ScaleBarProps = {
+  unit?: string;
+  align?: "start" | "center" | "end";
+};
+
+export function ScaleBar(props: ScaleBarProps) {
+  const idetikContext = useIdetik();
```
bchu1: nit - could destructure just `{ runtime }`

packages/react/src/components/viewers/OmeZarrImageViewer/components/ScaleBar/ScaleBar.tsx
bchu1: I think it would be `text-shadow-[black_1px_1px_1px,_black...]` bit ugly though

packages/react/src/components/viewers/OmeZarrImageViewer/components/ScaleBar/ScaleBar.tsx
bchu1: `border-[thin] border-solid border-black`

bchu1 (approved):

packages/react/src/components/viewers/OmeZarrImageViewer/components/ScaleBar/ScaleBar.tsx
andy-sweet: Thanks, done!

packages/react/src/components/viewers/OmeZarrImageViewer/components/ScaleBar/ScaleBar.tsx
andy-sweet: I actually needed to put the whole string in brackets to just pass through custom CSS `[text-shadow:black_1px_1px_1px,black_...]`. I think that's because we're on tailwind 3.4.17, but text-shadow was only added in 4.1. We could upgrade, but I'll leave that outside of this PR.

packages/react/src/components/viewers/OmeZarrImageViewer/OmeZarrImageViewer.tsx
andy-sweet: Removed.

packages/react/src/components/viewers/OmeZarrImageViewer/components/ScaleBar/ScaleBar.tsx
andy-sweet: It makes things slightly more compact, so done.

packages/react/src/components/viewers/OmeZarrImageViewer/components/ScaleBar/ScaleBar.tsx
andy-sweet: Done.

PR #302, opened 2025-07-08
(ONE LINE) Disable antialiasing to eliminate artifacts

shlomnissan:
### Summary

This PR disables antialiasing to avoid artifacts along tile edges. WebGL
enables multisample antialiasing (MSAA) by default, which is typically used
to smooth jagged edges. However, in tile-based rendering, MSAA can cause
visible seams or blending artifacts between tiles, as seen in our application.

Disabling MSAA is a common practice in precise scientific or technical
visualizations where pixel accuracy is more important than edge smoothing.
In our case, turning off MSAA resolves the artifacts observed in the chunked
data example.

One important note: MSAA cannot be toggled at runtime. If we ever need to
support selective antialiasing, we may need to introduce framebuffer-based
rendering with explicit control over sampling.

This change does not fix all visual issues. If linear filtering is enabled,
we still expect artifacts around tile edges due to texture interpolation.

https://en.wikipedia.org/wiki/Multisample_anti-aliasing

### Related Issue
N/A

### Tests & Checks
- Visual verification (see screenshots in full size)
- All checks have passed.

| Antialiasing enabled | Antialiasing disabled |
|----------|----------|
| Screenshot 2025-07-08 at 2 12 58 PM | Screenshot 2025-07-08 at 2 12 40 PM  |

### DIFF
--- a/packages/core/src/renderers/webgl_renderer.ts
+++ b/packages/core/src/renderers/webgl_renderer.ts
@@ -36,7 +36,10 @@ export class WebGLRenderer extends Renderer {
   constructor(canvas: HTMLCanvasElement) {
     super(canvas);
 
-    this.gl_ = this.canvas.getContext("webgl2", { depth: true });
+    this.gl_ = this.canvas.getContext("webgl2", {
+      depth: true,
+      antialias: false,
+    });
     if (!this.gl_) {
       throw new Error(`Failed to initialize WebGL2 context`);
     }


phoenixAja (approved):

andy-sweet (approved):
Approving because I think the edge artifacts are maybe a blocking bug in chunk streaming and this fixes it.

The only negative side-effects in the examples is the projected lines have some jaggies now, which is entirely expected.

Screenshot 2025-07-09 at 09 56 46

I don't view this as a blocking issue because we don't use the projected lines too much (biggest use is in the ultrack prototype) and the artifacts are more cosmetic than the tile edge artifacts (which are disruptive/misleading). As we add more layer types that may benefit from anti-aliasing (e.g. non-trivial meshes), we may need to think of a more sophisticated solution.

shlomnissan:
@andy-sweet Yep, I agree it’s not a blocker. There are a few shader-based methods for line antialiasing, and some of them are pretty straightforward to implement. When it comes to general antialiasing for specific layers, we’ve got a few options, but they’re likely to be a bit more involved.

I’m hoping we get a chance to experiment with virtual textures in Q3 or Q4 — that might be the cleanest way to avoid these artifacts altogether, even if it adds some localized complexity on the texturing side.

aganders3:
Now that we have layer blending/transparency, we can definitely do some AA in the line shader.

I also did some experimentation and these artifacts are avoided by using `clamp_to_edge` instead of `wrap` on the textures for the tiles, which makes sense to me and agrees with how the artifacts appear as I look closer. I think with MSAA some samples will be ~1/2 pixel outside the texture if the texcoords are [0, 1] inclusive. So we might be able to leave AA on (assuming we want it) and use that method.

PR #303, opened 2025-07-09
feat:  Add chunk info overlay to chunk_streaming example for progressive loading demos

phoenixAja:
###  Summary

  - Added a new ChunkInfoOverlay to the chunk_streaming example to visualize progressive loading in real-time
  - Provides detailed per-LOD breakdown of chunk loading status for better demo visibility

###  Changes

  - New file: chunk_info_overlay.ts - Overlay class that displays chunk loading metrics
  - Modified: index.html - Added styling and DOM element for chunk info display
  - Modified: main.ts - Integrated the chunk info overlay alongside existing FPS overlay

###  Features

  The overlay displays:
  - Total chunk count and loading status
  - Per-LOD breakdown showing visible vs rendered chunks
  - Current LOD level (updates with zoom)
  - Sample chunk details (size, scale, position)

https://github.com/user-attachments/assets/e34e43d7-9109-42c2-aa84-959b55887802


andy-sweet:
This is one of the uses I had in mind with overlays, so it's nice to see it so soon! I'll review this tomorrow. I'm slightly concerned by exposing more of Idetik with the public API additions here, but I generally prefer to do that and move quickly until we have most of the core functionality behavior we want in the library - and then we can clean things up later.

phoenixAja:
> This is one of the uses I had in mind with overlays, so it's nice to see it so soon! I'll review this tomorrow. I'm slightly concerned by exposing more of Idetik with the public API additions here, but I generally prefer to do that and move quickly until we have most of the core functionality behavior we want in the library - and then we can clean things up later.

No worries, there isn't a rush to merge this in. Also i was thinking about exposing more of the public API. I'm personally not too worried, especially because exposing the internals a bit would help any downstream users with their own debugging/logging. But happy to reconsider. Plus i don't think exposing these will mislead users on how to interact with the chunkManager

packages/core/examples/chunk_streaming/index.html
```diff
@@ -28,11 +28,25 @@
         background-color: rgba(0, 0, 0, 0.5);
         user-select: none;
       }
+      #chunk-info {
+        position: absolute;
+        top: 1em;
+        right: 1em;
+        width: 20%;
```
andy-sweet: Nit: making this wider prevents line wrap for me at reasonable window sizes.

```suggestion
        width: 40%;
```

packages/core/examples/chunk_streaming/index.html
```diff
@@ -28,11 +28,25 @@
...
+        margin: 1em;
+        padding: 1em;
+        color: white;
+        background-color: rgba(0, 0, 0, 0.5);
+        user-select: none;
+        font-family: monospace;
```
andy-sweet: Nit: move font-family and font-size to the body so that the overlays with text content share the same font.

andy-sweet (approved):
Approving because I'm OK exposing things like the chunks and chunk manager source right now, because we're still learning what we need from those. Just a few nits, which are optional.

PR #305, opened 2025-07-14
(Small) Move the primitive attribute to geometry

shlomnissan:
### Summary

This PR moves the primitive property from the renderable object base  
class to the geometry class. The primitive type is tightly coupled with
the geometry, so relocating it improves clarity and structure.  

This change also simplifies downstream work, such as wireframe  
rendering and supporting multiple geometries per renderable object.

### Related Issue
N/A

### Tests & Checks
- All checks have passed
- Manual verification of existing examples

### DIFF
--- a/packages/core/src/objects/renderable/points.ts
+++ b/packages/core/src/objects/renderable/points.ts
@@ -19,7 +19,6 @@ export class Points extends RenderableObject {
   constructor(points: PointProperties[], markerAtlas: Texture2DArray) {
     super();
     this.programName = "points";
-    this.primitive = "points";
     this.atlas_ = markerAtlas;
 
     points.forEach((point) => {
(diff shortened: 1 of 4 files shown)


packages/core/src/objects/renderable/points.ts
```diff
@@ -64,6 +63,8 @@ export class Points extends RenderableObject {
       offset: geometry.stride,
     });
 
+    geometry.primitive = "points";
```
andy-sweet: Thinking out aloud: this PR moves primitive into geometry because it is tightly coupled with the vertex data in the geometry, so should it be passed into the constructor and be less publicly mutable as with vertex data? Or is the intent to allow mutation of primitive to allow the same geometry to be rendered differently (in which case does it really belong in geometry)?

andy-sweet (approved):
I'm fine with this move. Just had a stream of consciousness though in one comment, that you can probably ignore.

packages/core/src/objects/renderable/points.ts
shlomnissan: > should it be passed into the constructor and be less publicly mutable as with vertex data?

Yes, I think it should. I might make this change before merging.

PR #307, opened 2025-07-15
Use canvas.clientWidth instead of idetik.width for scale-bar updates

### LINKED ISSUE #306: ScaleBar does not update correctly when changing display DPI
Steps to reproduce on Chrome on macOS.

1. Setup up two displays with different DPIs (e.g. a recent macbook pro with a 1080p display). 
2. Run the standard Idetik React component demo.
3. Move the browser window from one display to another without an automatic resize (e.g. by making the window relatively small).

After reverting the changes made in https://github.com/chanzuckerberg/idetik/pull/294 , this bug is not present. However, I'm not sure if that's OS/Browser specific.

I think the better fix is to use [`window.matchMedia`](https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia) to observe changes in the devicePixelRatio with something like the following

```javascript
const pixelRatio = window.devicePixelRatio;
mediaQueryList = window.matchMedia(`(resolution: ${pixelRatio}dppx)`);
mediaQueryList.addEventListener('change', handleDpiChange);
```

but I haven't tried that yet.

Slightly more concerning is that I've just discovered that `window.devicePixelRatio` is [not completely supported by Safari](https://developer.mozilla.org/en-US/docs/Web/API/Window/devicePixelRatio#browser_compatibility), so we might need to rethink this approach more generally.

andy-sweet:
### Summary
This fixes an issue with the reported physical length of the scale bar when switching between displays with different DPIs (see #306) by using `idetik.canvas.clientWidth` instead of `idetik.width` and `window.devicePixelRatio` to handle the scale-bar length calculations.

Using the CSS width rather than the canvas buffer width is conceptually simpler, and is also immune to Safari's partial support for `window.devicePixelRatio`.

### Related Issue
Closes #306 

### Tests & Checks
I manually tested the scale bar example and the React component example with the setup described in #306. The bug is present without these changes and not present after. After these changes, I verified that the reported physical length of a single pixel and the whole image is as expected based on the image metadata.

### DIFF
--- a/packages/core/examples/image2d_from_omezarr5d_u16/scale_bar.ts
+++ b/packages/core/examples/image2d_from_omezarr5d_u16/scale_bar.ts
@@ -27,12 +27,12 @@ export class ScaleBar {
 
     const cameraWidthWorld =
       orthoCamera.transform.scale[0] * orthoCamera.viewportSize[0];
-    const unitPerCanvasPixel = cameraWidthWorld / idetik.width;
+    const unitPerCanvasPixel = cameraWidthWorld / idetik.canvas.clientWidth;
 
     // The use of clientWidth assumes that the lineDiv has no padding,
     // which is true in this example. If similar code is used elsewhere,
     // the lack of padding should be asserted and or enforced.
-    const lineWidth = this.lineDiv_.clientWidth * window.devicePixelRatio;
+    const lineWidth = this.lineDiv_.clientWidth;
     const lineWidthWorld = lineWidth * unitPerCanvasPixel;
 
     this.textDiv_.textContent = `${lineWidthWorld.toFixed(2)} ${this.unit_}`;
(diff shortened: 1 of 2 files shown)


andy-sweet:
@bchu1 : I added you for review because you reviewed the other scale-bar PRs and also wrote #294 , which is tangentially related here. Other comments and reviews welcome.

kyleawayan:
Implemented in OrganelleBox portal, here's the preview link for https://github.com/czbiohub-sf/organelle-box-portal/commit/b1c5dc05f0daa98c70bf93789581055ef5352ab9:

https://organelle-box-portal-au4fpt0zn-czbiohub.vercel.app/gene/2025_03_31_A549_ACTB

bchu1 (approved):

shlomnissan (approved):

PR #310, opened 2025-07-15
(Stacked) Render wireframes on top of meshes [2/2]

### LINKED ISSUE #47: Add debug rendering

shlomnissan:
### Summary
This PR finalizes the wireframe-over-shaded-geometry feature. It introduces
a dedicated wireframe geometry to renderable objects, adds a basic passthrough
shader for rendering flat colors, and completes the necessary integration in WebGLRenderer.

**Next Steps**
I'm calling it done because the functionality is there, but I want to extend it in a follow-up:
- Implement layer-specific logic for displaying debug wireframes.
- Add color option for wireframes - for example, different wireframe colors may be desirable per LOD level.
- Apply polygon offset to ensure wireframes consistently render on top of shaded geometry.

### Related Issue
Closes #47

### Tests & Checks
- Manually verified — see attached video.
- To enable wireframes, set wireframeEnabled = true on renderable objects.
- All existing examples render correctly with no regressions.
- All checks have passed.

https://github.com/user-attachments/assets/6d2d61db-d22a-4d51-988c-28b60d75d353

### DIFF
--- a/packages/core/src/core/geometry.ts
+++ b/packages/core/src/core/geometry.ts
@@ -1,7 +1,6 @@
 import { Node } from "./node";
-import { Logger } from "../utilities/logger";
 
-export type Primitive = "triangles" | "points";
+export type Primitive = "triangles" | "points" | "lines";
 
 type GeometryAttributeType =
   | "position"
@@ -35,11 +34,10 @@ type GeometryAttribute = {
 };
 
 export class Geometry extends Node {
-  private readonly attributes_: GeometryAttribute[];
-  private readonly primitive_: Primitive;
+  protected primitive_: Primitive;
+  protected attributes_: GeometryAttribute[];
   protected vertexData_: Float32Array;
   protected indexData_: Uint32Array;
-  private wireframeIndexData_: Uint32Array;
 
   constructor(
     vertexData: number[] = [],
@@ -50,7 +48,6 @@ export class Geometry extends Node {
     this.vertexData_ = new Float32Array(vertexData);
     this.indexData_ = new Uint32Array(indexData);
     this.primitive_ = primitive;
-    this.wireframeIndexData_ = new Uint32Array();
     this.attributes_ = [];
   }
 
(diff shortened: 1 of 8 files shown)


andy-sweet (approved):
This is much nicer than the previous version, so approving.

I think this is working as expected, so I get slightly strange wireframes at bottom and right sides in the chunked image demo (after waiting for the network to settle) - is that related to the padding?

Screenshot 2025-07-15 at 16 42 54

Also, I think having a `wireframeColor: Color` on `RenderableObject` and passing that through as a uniform to the wireframe fragment shader would be super useful (especially for these white background demos...)!

shlomnissan:
> I get slightly strange wireframes at bottom

@andy-sweet This is weird. I can repro as well. I wonder if it's z-fighting artifacts? I will figure this out before merging.

And yeah, I will add wireframe color for sure.

aganders3:
I think this is due to the padding. I was particularly confused by what looked like a half-chunk  here, but I think that's the diagonal from an LOD-2 chunk. You can see from @phoenixAja's debug overlay (🎉) that it's rendering two LOD-2 chunks.

466707124-cec57e3c-14a2-47b0-b80b-f0aea4a6b660

shlomnissan:
> I think that's the diagonal from an LOD-2 chunk

@aganders3 Ahh that makes sense! I was also confused by what looked like half a chunk, especially when you transition from LOD1 to LOD0.

I will merge this and add some structure around it — support different color wireframes, only show wireframes for current LOD, etc

PR #311, opened 2025-07-15
Fix pan/sticking when dismissing context menu

### LINKED ISSUE #308: Alternating mouse click on the canvas triggers panning
In the viewer, after right/alt-clicking on the canvas and then left/normal-clicking to close the context menu, the camera control sticks to the pointer as if the mouse button is held down.

aganders3:
### Summary
This prevents #308 by changing `PanZoomControls` to ignore non-left-click `pointerdown` events.

I don't notice any ill effects with this change, but there might be a better/more thorough fix we can implement when taking on pinch-to-zoom for example.

### Related Issue
Closes #308 

### Tests & Checks
tested with organellebox data through `npm run components`

### DIFF
--- a/packages/core/src/objects/cameras/controls.ts
+++ b/packages/core/src/objects/cameras/controls.ts
@@ -71,6 +71,10 @@ export class PanZoomControls implements CameraControls {
     target: EventTarget,
     clientToWorld: ClientToWorld
   ) {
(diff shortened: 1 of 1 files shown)


andy-sweet (approved):
This works as a short-term fix for me.

PR #316, opened 2025-07-21
Add WebFileSystemStore

bchu1:
The demo has some hardcoded stuff. I didn't think it was too worth my time to make it generic for now, and a link to a sample Zarr is in the comment.

https://github.com/user-attachments/assets/e22262ca-57b1-4dee-8ff9-aecc04b0c947


aganders3:
This looks good, and is pretty cool functionality! It will be really convenient to reference local files instead of having to run a local http server.

I just want to check with @andy-sweet: will any of the changes here interfere the OME-Zarr 0.4/0.5 work you're doing?

andy-sweet:
> I just want to check with @andy-sweet: will any of the changes here interfere the OME-Zarr 0.4/0.5 work you're doing?

I'm sure there will be a few merge conflicts to resolve, but I don't see any obvious big problems.

aganders3 (approved):
This looks good and works well for me (in your example). I opened another zarr I already had and it also worked.

Thanks for looking into this and hooking it up!

One nit is to consider `import { type Readable } from "@zarrita/storage";` so you can use `group: zarr.Group<Readable>` in typing instead of importing `FetchStore` and also `WebFileSystemStore` at least in some of these files.

`OmeZarrImageSource` might have to stay specific, but the other classes are private and assume the store is already instantiated.

aganders3:
I think this also paves the way for a _writeable_ local annotation store, which is pretty cool!

PR #317, opened 2025-07-21
Support multiple textures in one fragment shader

### LINKED ISSUE #290: Do not call WebGL getError on every render pass for every object
Currently the [WebGL renderer](https://github.com/chanzuckerberg/idetik/blob/2459e46ddb9b3868b01a7624ee0b19ef3cadd431/packages/core/src/renderers/webgl_renderer.ts#L90) calls [WebGL's `getError`](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/getError) at least once on every render pass via [`WebGLShaderProgram.use`](https://github.com/chanzuckerberg/idetik/blob/2459e46ddb9b3868b01a7624ee0b19ef3cadd431/packages/core/src/renderers/webgl_shader_program.ts#L137).

According to [MDN's WebGL best practices](https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API/WebGL_best_practices#avoid_blocking_api_calls_in_production) that is not advisable because it is a blocking call that may take milliseconds.

We should avoid calling `getError` on every render pass.

### LINKED ISSUE #312: Support multiple textures in one fragment shader
Currently, Idetik stores and partially uses multiple textures, but some things are missing to make textures other than the first useful.

1. `RenderableObject` [stores multiple textures](https://github.com/chanzuckerberg/idetik/blob/a84b9f1e6319b4aa05d0c2c5c0a806d204603c2b/packages/core/src/core/renderable_object.ts#L9).
2. `WebGLRenderer` [binds each texture of an object](https://github.com/chanzuckerberg/idetik/blob/a84b9f1e6319b4aa05d0c2c5c0a806d204603c2b/packages/core/src/renderers/webgl_renderer.ts#L90)
3. `WebGLShaderProgram` [skips sampler uniforms when parsing the program](https://github.com/chanzuckerberg/idetik/blob/a84b9f1e6319b4aa05d0c2c5c0a806d204603c2b/packages/core/src/renderers/webgl_shader_program.ts#L78)

Let's say an object has two textures. If we want to use the second texture in a fragment shader, then I believe we'd also need to call something like `gl.activeTexture(gl.TEXTURE1)` before `gl.bindTexture(...)` and we'd also need to call something like `gl.uniform1i(location, 1)` before the actual GL draw call (to bind the location of a sampler to a texture index).

Without making those changes, I found that a fragment shader with two textures failed silently.

The goal of this issue to ensure that Idetik can support a fragment shader with more one than one texture. We may want to wait to work on this until we actually need it.

One important question here is if we should assume that sampler names are arbitrary or whether they will be of the form `texture<x>` where `<x>` is a non-negative integer.

We also need to consider what limit we should put on the number of textures.

andy-sweet:
### Summary
This change allows a renderable object to use multiple textures in its fragment shader. In order to allow these changes, I made a small refactor of how shader programs are linked, validated, and used that may have other positive side-effects (e.g. avoiding error check after use on every render pass).

### Related Issue
Closes #312
Closes #290

### Tests & Checks
I tested that the changes here do not break any of the existing examples.
I also tested that multiple textures and sampler types can be used in PR #320.

### DIFF
--- a/packages/core/src/objects/renderable/image_renderable.ts
+++ b/packages/core/src/objects/renderable/image_renderable.ts
@@ -71,6 +71,7 @@ export class ImageRenderable extends RenderableObject {
       const { color, contrastLimits } =
         this.channels_[0] ?? validateChannel(texture, {});
       return {
+        ImageData: 0,
         Color: color.rgb,
         ValueOffset: -contrastLimits[0],
         ValueScale: 1 / (contrastLimits[1] - contrastLimits[0]),
@@ -93,6 +94,7 @@ export class ImageRenderable extends RenderableObject {
       });
 
       return {
+        ImageData: 0,
         "Visible[0]": visible,
         "Color[0]": color,
         "ValueOffset[0]": valueOffset,
--- a/packages/core/src/objects/renderable/image_renderable.ts
+++ b/packages/core/src/objects/renderable/image_renderable.ts
@@ -11,14 +11,12 @@ import { vec3 } from "gl-matrix";
 import { Shader } from "../../renderers/shaders";
 
 type SingleUniformValues = {
-  ImageData: number;
   Color: vec3;
   ValueOffset: number;
   ValueScale: number;
 };
 
 type ArrayUniformValues = {
-  ImageData: number;
   "Visible[0]": boolean[];
   "Color[0]": number[];
   "ValueOffset[0]": number[];
@@ -71,7 +69,6 @@ export class ImageRenderable extends RenderableObject {
       const { color, contrastLimits } =
         this.channels_[0] ?? validateChannel(texture, {});
       return {
-        ImageData: 0,
         Color: color.rgb,
         ValueOffset: -contrastLimits[0],
         ValueScale: 1 / (contrastLimits[1] - contrastLimits[0]),
@@ -94,7 +91,6 @@ export class ImageRenderable extends RenderableObject {
       });
 
       return {
-        ImageData: 0,
         "Visible[0]": visible,
         "Color[0]": color,
         "ValueOffset[0]": valueOffset,
--- a/packages/core/src/objects/renderable/image_renderable.ts
+++ b/packages/core/src/objects/renderable/image_renderable.ts
@@ -11,12 +11,14 @@ import { vec3 } from "gl-matrix";
 import { Shader } from "../../renderers/shaders";
 
 type SingleUniformValues = {
+  ImageData: number;
   Color: vec3;
   ValueOffset: number;
   ValueScale: number;
 };
 
 type ArrayUniformValues = {
+  ImageData: number;
   "Visible[0]": boolean[];
   "Color[0]": number[];
   "ValueOffset[0]": number[];
--- a/packages/core/src/renderers/shaders/index.ts
+++ b/packages/core/src/renderers/shaders/index.ts
@@ -24,6 +24,7 @@ type ShaderCode = {
   vertexDefines?: ReadonlyMap<string, string>;
   fragment: string;
   fragmentDefines?: ReadonlyMap<string, string>;
+  samplerBindings?: Record<string, number>;
 };
 
 export const shaderCode: Record<Shader, ShaderCode> = {
--- a/packages/core/src/renderers/shaders/index.ts
+++ b/packages/core/src/renderers/shaders/index.ts
@@ -24,7 +24,6 @@ type ShaderCode = {
   vertexDefines?: ReadonlyMap<string, string>;
   fragment: string;
   fragmentDefines?: ReadonlyMap<string, string>;
-  samplerBindings?: Record<string, number>;
 };
 
 export const shaderCode: Record<Shader, ShaderCode> = {
--- a/packages/core/src/renderers/webgl_renderer.ts
+++ b/packages/core/src/renderers/webgl_renderer.ts
@@ -91,12 +91,12 @@ export class WebGLRenderer extends Renderer {
       this.textures_.bindTexture(texture, index);
     });
 
-    const program = this.programs_.get(object.programName).use();
+    const program = this.programs_.use(object.programName);
     this.drawGeometry(object.geometry, object, layer, program);
 
     if (object.wireframeEnabled) {
       this.bindings_.bindGeometry(object.wireframeGeometry);
-      const wireframeProgram = this.programs_.get("wireframe").use();
+      const wireframeProgram = this.programs_.use("wireframe");
       this.drawGeometry(
         object.wireframeGeometry,
         object,
--- a/packages/core/src/renderers/webgl_renderer.ts
+++ b/packages/core/src/renderers/webgl_renderer.ts
@@ -87,8 +87,8 @@ export class WebGLRenderer extends Renderer {
   protected renderObject(layer: Layer, objectIndex: number) {
     const object = layer.objects[objectIndex];
     this.bindings_.bindGeometry(object.geometry);
-    object.textures.forEach((texture) => {
-      this.textures_.bindTexture(texture);
+    object.textures.forEach((texture, index) => {
+      this.textures_.bindTexture(texture, index);
     });
 
     const program = this.programs_.get(object.programName).use();
--- a/packages/core/src/renderers/webgl_renderer.ts
+++ b/packages/core/src/renderers/webgl_renderer.ts
@@ -97,6 +97,7 @@ export class WebGLRenderer extends Renderer {
     if (object.wireframeEnabled) {
       this.bindings_.bindGeometry(object.wireframeGeometry);
       const wireframeProgram = this.programs_.use("wireframe");
+      wireframeProgram.setUniform("u_color", object.wireframeColor.rgb);
       this.drawGeometry(
         object.wireframeGeometry,
         object,
--- a/packages/core/src/renderers/webgl_shader_program.ts
+++ b/packages/core/src/renderers/webgl_shader_program.ts
@@ -128,7 +128,7 @@ export class WebGLShaderProgram {
     return shader;
   }
 
-  private link(): void {
+  private link() {
     this.gl_.linkProgram(this.program_);
     if (!this.getParameter(this.gl_.LINK_STATUS)) {
       const message = this.gl_.getProgramInfoLog(this.program_);
--- a/packages/core/src/renderers/webgl_shader_program.ts
+++ b/packages/core/src/renderers/webgl_shader_program.ts
@@ -1,18 +1,10 @@
 import { mat4, vec2, vec3 } from "gl-matrix";
 
-type ShaderMap = {
-  [type: number]: {
-    shader: WebGLShader;
-    source: string;
-  };
-};
-
 export class WebGLShaderProgram {
   private readonly gl_: WebGL2RenderingContext;
   private readonly program_: WebGLProgram;
   private uniformInfo_: Map<string, [WebGLUniformLocation, WebGLActiveInfo]> =
     new Map();
-  private shaders_: ShaderMap = {};
 
   constructor(
     gl: WebGL2RenderingContext,
@@ -27,9 +19,34 @@ export class WebGLShaderProgram {
     }
     this.program_ = program;
 
-    this.addShader(vertexShaderSource, gl.VERTEX_SHADER);
-    this.addShader(fragmentShaderSource, gl.FRAGMENT_SHADER);
-    this.link();
+    const shaders = [];
+    shaders.push(this.addShader(vertexShaderSource, gl.VERTEX_SHADER));
+    shaders.push(this.addShader(fragmentShaderSource, gl.FRAGMENT_SHADER));
+    this.gl_.linkProgram(this.program_);
+    if (!this.getParameter(this.gl_.LINK_STATUS)) {
+      shaders.forEach((shader) => this.gl_.deleteShader(shader));
+      const message = this.gl_.getProgramInfoLog(this.program_);
+      throw new Error(`Error linking program: ${message}`);
+    }
+    shaders.forEach((shader) => this.gl_.deleteShader(shader));
+
+    this.preprocessUniformLocations();
+  }
+
+  public use() {
+    this.gl_.useProgram(this.program_);
+  }
+
+  public validate() {
+    this.gl_.validateProgram(this.program_);
+    if (!this.getParameter(this.gl_.VALIDATE_STATUS)) {
+      const message = this.gl_.getProgramInfoLog(this.program_);
+      throw new Error(`Error validating program: ${message}`);
+    }
+  }
+
+  public get uniformNames(): string[] {
+    return Array.from(this.uniformInfo_.keys());
   }
 
   public setUniform(name: string, value: unknown) {
@@ -123,50 +140,11 @@ export class WebGLShaderProgram {
     }
 
     this.gl_.attachShader(this.program_, shader);
-    this.shaders_[type] = { shader: shader, source };
-  }
-
-  private link() {
-    this.gl_.linkProgram(this.program_);
-    if (!this.getParameter(this.gl_.LINK_STATUS)) {
-      this.deleteShaders();
-      const message = this.gl_.getProgramInfoLog(this.program_);
-      throw new Error(`Error linking program: ${message}`);
-    }
-
-    this.gl_.validateProgram(this.program_);
-    if (!this.getParameter(this.gl_.VALIDATE_STATUS)) {
-      this.deleteShaders();
-      const message = this.gl_.getProgramInfoLog(this.program_);
-      throw new Error(`Error validating program: ${message}`);
-    }
-
-    this.preprocessUniformLocations();
-    this.deleteShaders();
-  }
-
-  public use() {
-    this.gl_.useProgram(this.program_);
-    const error = this.gl_.getError();
-    if (error !== this.gl_.NO_ERROR) {
-      throw new Error(`Error using WebGL program: ${error}`);
-    }
-    return this;
+    return shader;
   }
 
   private getParameter(parameter: number) {
     return this.gl_.getProgramParameter(this.program_, parameter);
-  }
-
-  private deleteShaders() {
-    for (const idx in this.shaders_) {
-      this.gl_.deleteShader(this.shaders_[idx].shader);
-    }
-    this.shaders_ = {};
-  }
-
-  public get uniformNames(): string[] {
-    return Array.from(this.uniformInfo_.keys());
   }
 }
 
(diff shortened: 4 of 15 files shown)


packages/core/src/renderers/webgl_shader_program.ts
```diff
@@ -175,20 +194,6 @@ const SAMPLER_TYPES: ReadonlySet<GLenum> =
         WebGL2RenderingContext.UNSIGNED_INT_SAMPLER_3D,
         WebGL2RenderingContext.UNSIGNED_INT_SAMPLER_CUBE,
         WebGL2RenderingContext.UNSIGNED_INT_SAMPLER_2D_ARRAY,
-        WebGL2RenderingContext.MAX_SAMPLES,
```
andy-sweet, line 181: I removed this because it's the maximum number of samples that can be used by a sampler (for things like anti-aliasing) and not a sampler type.

packages/core/src/renderers/webgl_shader_program.ts
```diff
@@ -175,20 +194,6 @@ const SAMPLER_TYPES: ReadonlySet<GLenum> =
...
-        WebGL2RenderingContext.SAMPLER_BINDING,
```
andy-sweet, line 182: I removed this because this is constant used to get the sampler that is currently bound to the active texture, and not a sampler type.

packages/core/src/objects/renderable/image_renderable.ts
```diff
@@ -11,12 +11,14 @@ import { vec3 } from "gl-matrix";
 import { Shader } from "../../renderers/shaders";
 
 type SingleUniformValues = {
+  ImageData: number;
```
andy-sweet: Note that because these renderables only have one texture, this doesn't actually need to be specified yet (because the active texture defaults to 0). But I did so for clarity.

andy-sweet:
Asking for a review while in draft to clarify my understanding here.

@aganders3 : assigned you as a reviewer because you had some comments on #312 
@shlomnissan : assigned you as a review because of your GL nous.

andy-sweet:
As it stands, I believe these changes will support more than one texture in a renderable & fragment shader, but only if they use the same precision/type of sampler in the shader.

More changes are needed to support textures with different sampler types, with one potential solution in the following branch: https://github.com/chanzuckerberg/idetik/compare/andy-sweet/multiple-textures...andy-sweet/label-layer?expand=1

packages/core/src/objects/renderable/image_renderable.ts
shlomnissan: That makes sense.. It's best practice to set the active texture unit before binding. I’d recommend renaming this to something like `imageSampler` to make its purpose clearer. We can introduce others later like `segmentationSampler`, etc. Since a sampler is just a reference to a texture unit, a name like `imageData` might misleadingly imply it holds the actual data.

packages/core/src/renderers/webgl_textures.ts
```diff
@@ -20,17 +20,26 @@ export class WebGLTextures {
   private readonly gl_: WebGL2RenderingContext;
   private readonly textures_: Map<Texture, WebGLTexture> = new Map();
   private currentTexture_: Texture | null = null;
+  private readonly maxTextureUnits_: number;
 
   constructor(gl: WebGL2RenderingContext) {
     this.gl_ = gl;
+    this.maxTextureUnits_ = gl.MAX_COMBINED_TEXTURE_IMAGE_UNITS;
```
shlomnissan: `gl.MAX_TEXTURE_IMAGE_UNITS` returns the max supported texture units that can be accessed from the fragment shader. Would it be more suitable to our use case? At this level `MAX_COMBINED_TEXTURE_IMAGE_UNITS ` might be more correct, but I think that `MAX_TEXTURE_IMAGE_UNITS` is more useful at this point since we don't use samplers across shader stages.

shlomnissan:
@andy-sweet Looks good to me!

> only if they use the same precision/type of sampler in the shader

- Is it necessary to include precision qualifiers for samplers in WebGL? I would assume that they have no effect on how the GPU samples or stores the data.
- What do you mean by "they have to be the same type"? Is this an application constraint? As I understand it, different texture units support different sampler types.

andy-sweet:
> * Is it necessary to include precision qualifiers for samplers in WebGL? 

Yes.

From the [OpenGL ES3 GLSL spec](https://registry.khronos.org/OpenGL/specs/es/3.0/GLSL_ES_Specification_3.00.pdf) (section 4.5.4, page 56)

> The fragment language has no default precision qualifier for floating point types. Hence for float, floating
> point vector and matrix variable declarations, either the declaration must include a precision qualifier or
> the default float precision must have been previously declared. Similarly, there is no default precision
> qualifier for the following sampler types in either the vertex or fragment language:
> 
> sampler3D;
> samplerCubeShadow;
> sampler2DShadow;
> sampler2DArray;
> sampler2DArrayShadow;
> isampler2D;
> isampler3D;
> isamplerCube;
> isampler2DArray;
> usampler2D;
> usampler3D;
> usamplerCube;
> usampler2DArray;

If I remove the precision from a sampler declaration in the fragment shader, I get an error like the following

> Uncaught Error: Error compiling shader: ERROR: 0:8: 'usampler2D' : No precision specified

Why do you assume the following?

> I would assume that they have no effect on how the GPU samples or stores the data.

As I [mentioned in a recent PR](https://github.com/chanzuckerberg/idetik/pull/299#discussion_r2211049889), the spec also states that

> For the texture sampling functions, the precision of the return type matches the precision of the
sampler type.
```
uniform lowp sampler2D sampler;
highp vec2 coord;
...
lowp vec4 col = texture (sampler, coord); // texture() returns lowp
```

That being said, I have no idea what the browser implementations of WebGL2 actually do when it comes to pushing data to the GPU. Maybe everything ends up as 32-bit anyway?

andy-sweet:
> What do you mean by "they have to be the same type"? Is this an application constraint? As I understand it, different texture units support different sampler types.

The implementation here breaks once I use sampler uniforms of different types (e.g. usampler2D and sampler2D). The reason is that I don't call `gl.uniform1i` before `gl.validateProgram`.

In OpenGLES 3.10 GLSL, you can use something like

```
layout (binding = 0) uniform highp usampler2D ImageData;
layout (binding = 1) uniform mediump sampler2D ColorCycle;
```

to handle this, but it's not available in GLES 3.00.

I think I have a [reasonable workaround in a WIP branch](https://github.com/chanzuckerberg/idetik/compare/andy-sweet/multiple-textures...andy-sweet/label-layer?expand=1), which I'll discuss today and may bring over to this branch.

packages/core/src/renderers/webgl_textures.ts
andy-sweet: Ah, good catch. I was unaware of this distinction. Change made.

packages/core/src/objects/renderable/image_renderable.ts
andy-sweet: Renamed to `ImageSampler`.

packages/core/src/renderers/webgl_shader_program.ts
```diff
@@ -27,9 +19,34 @@ export class WebGLShaderProgram {
     }
     this.program_ = program;
 
-    this.addShader(vertexShaderSource, gl.VERTEX_SHADER);
-    this.addShader(fragmentShaderSource, gl.FRAGMENT_SHADER);
-    this.link();
```
andy-sweet, line 34: In order to get things working, I needed to break `link` apart for a few reasons.

- We must set the sampler uniform texture unit bindings before calling `validateProgram`.
- We must call `useProgram` before setting the sampler uniform texture unit bindings.

After splitting this up, I made some other related changes that I think are improvements but could be put in a different PR.

packages/core/src/renderers/webgl_shader_programs.ts
```diff
@@ -28,7 +28,20 @@ export class WebGLShaderPrograms {
...
+      const error = this.gl_.getError();
+      if (error !== this.gl_.NO_ERROR) {
+        throw new Error(`Error using WebGL program: ${error}`);
+      }
+      if (code.samplerBindings) {
+        for (const [name, value] of Object.entries(code.samplerBindings)) {
+          program.setUniform(name, value);
+        }
+      }
+      program.validate();
```
andy-sweet: Based on the [`glValidateProgram` docs](https://docs.gl/es3/glValidateProgram), I'm starting to think it might be better to call `validate` in non-production builds in every iteration in the renderer instead of once after construction (because its success depends on GL state).

That way, we could also remove the the explicit sampler bindings above and let `WebGLRenderer` handle them as other uniforms (assuming we called `validate` after setting those uniforms).

packages/core/src/renderers/shaders/index.ts
```diff
@@ -24,6 +24,7 @@ type ShaderCode = {
   vertexDefines?: ReadonlyMap<string, string>;
   fragment: string;
   fragmentDefines?: ReadonlyMap<string, string>;
+  samplerBindings?: Record<string, number>;
```
andy-sweet: I added these here because it's some way to represent the OpenGLES 3.10 way of doing things:

```
layout (binding = 0) uniform highp usampler2D ImageData;
layout (binding = 1) uniform mediump sampler2D ColorCycle;
```

Open to other ideas.

andy-sweet:
Marked this as ready for review because PR #320 is ready for review and is based on this. I think the separation is useful, but let me know if you just want to review them as one PR.

packages/core/src/renderers/shaders/index.ts
andy-sweet: I should probably populate the programs that use fragment shaders with samplers with records like `ImageData: 0` to be explicit, but that works without such changes because the default sampler binding is 0.

shlomnissan:
> I think the separation is useful

I think it's useful too. I will review first thing tomorrow morning

packages/core/src/renderers/webgl_textures.ts
shlomnissan: 👍

When working with multiple texture units across different shader stages, you need to satisfy two constraints: each shader stage must not exceed its individual texture unit limit:
- `GL_MAX_TEXTURE_IMAGE_UNITS` (fragment shader)
- `GL_MAX_VERTEX_TEXTURE_IMAGE_UNITS` (vertex shader)

And the total number of texture units used across all active stages combined must not exceed `GL_MAX_COMBINED_TEXTURE_IMAGE_UNITS`.

A robust system will verify both, but checking `GL_MAX_TEXTURE_IMAGE_UNITS` should be enough for now.

packages/core/src/renderers/webgl_shader_programs.ts
shlomnissan: I typically avoid it altogether. This function is really expensive because it analyzes the entire pipeline state. Also, it's kinda redundant.. If your program links successfully and renders correctly during development and testing, then runtime validation provides little value.

Most modern APIs have no runtime validation for shader programs, they instead move all validation to compile time.

packages/core/src/renderers/webgl_renderer.ts
```diff
@@ -87,16 +87,16 @@ export class WebGLRenderer extends Renderer {
...
     const object = layer.objects[objectIndex];
     this.bindings_.bindGeometry(object.geometry);
-    object.textures.forEach((texture) => {
-      this.textures_.bindTexture(texture);
+    object.textures.forEach((texture, index) => {
+      this.textures_.bindTexture(texture, index);
     });
 
-    const program = this.programs_.get(object.programName).use();
+    const program = this.programs_.use(object.programName);
```
shlomnissan, line 94: 👍

packages/core/src/renderers/webgl_shader_program.ts
```diff
@@ -29,14 +21,36 @@ export class WebGLShaderProgram {
...
+
+  public validate() {
+    this.gl_.validateProgram(this.program_);
+    if (!this.getParameter(this.gl_.VALIDATE_STATUS)) {
+      const message = this.gl_.getProgramInfoLog(this.program_);
+      throw new Error(`Error validating program: ${message}`);
+    }
+  }
+
+  public get uniformNames(): string[] {
```
shlomnissan: As mentioned in my response, I would probably remove this, otherwise, we may forget its there and ship it to production. If we need it for debugging later, we can add it back.

packages/core/src/renderers/webgl_shader_programs.ts
```diff
@@ -28,7 +28,20 @@ export class WebGLShaderPrograms {
...
+      program.use();
+      const error = this.gl_.getError();
+      if (error !== this.gl_.NO_ERROR) {
+        throw new Error(`Error using WebGL program: ${error}`);
+      }
+      if (code.samplerBindings) {
+        for (const [name, value] of Object.entries(code.samplerBindings)) {
+          program.setUniform(name, value);
+        }
+      }
```
shlomnissan: This seems very out of place for me. Why wouldn't you set the uniforms in the renderer? Is this here just so we can run `program.validate()`? If yes, that's another reason to remove it.

shlomnissan (changes requested):
Overall, this looks good. I'm requesting changes because I think we can eliminate the `samplerBindings` map and avoid using implicit sampler uniform bindings in `WebGLShaderPrograms`.

I might be missing something here, but I would have expected something like this:

Replace this code:
```ts
object.textures.forEach((texture, index) => {
  this.textures_.bindTexture(texture, index);
});
```
with something like this:

```ts
if (object.imageMap) {
  this.gl.activeTexture(IMG_SAMPLER_UNIT);
  this.gl.bindTexture(object.imageMap, IMAGE_SAMPLER_UNIT);
}

if (object.segmentationMap) {
  this.gl.activeTexture(SEG_SAMPLER_UNIT);
  this.gl.bindTexture(object.segmentationMap, SEG_SAMPLER_UNIT);
}
```

- Remove the textures array from renderable object and replace it with explicit texture maps members.
- Remove program validation, especially if it complicates the flow.

packages/core/src/renderers/webgl_shader_programs.ts
shlomnissan: From "OpenGL ES 3.0 Programming Guide"

> You really want to use glValidateProgram only for debugging purposes. It is a slow operation and certainly not something you want to check before every render. In fact, you can get away with never using it if your application is successfully rendering. We want to make you aware that this function does exist, though.

packages/core/src/renderers/webgl_shader_programs.ts
andy-sweet: Yes, `validate` is the only reason it's here. Otherwise, the sampler uniform value could be set in the renderer like everything else.

I'm down to remove it entirely for now, and only re-add it if/when we have some kind of Idetik-wide debug mode.

packages/core/src/renderers/webgl_shader_program.ts
andy-sweet: Sorry, I moved some method definitions (including this one) around in this class because I did some refactoring (which is more limited now) and I prefer seeing public methods at the top.

This PR didn't actually add or change this method.

It is needed for the renderer to set uniform values by name. I could be convinced that there's a better approach than what we're currently doing in the renderer right now, but that's out of place in this PR.

shlomnissan: Weird, I think I added this comment to line 42 (`public validate()`)

andy-sweet: Ah makes more sense. I will remove `validate`.

packages/core/src/renderers/webgl_shader_program.ts
```diff
@@ -29,12 +21,18 @@ export class WebGLShaderProgram {
     }
     this.program_ = program;
 
-    this.addShader(vertexShaderSource, gl.VERTEX_SHADER);
-    this.addShader(fragmentShaderSource, gl.FRAGMENT_SHADER);
-    this.link();
-    this.deleteShaders();
-    this.validate();
-    this.preprocessUniformLocations();
+    const shaders = [];
```
andy-sweet, line 24: These changes, as well as those in `webgl_shader_programs`, probably belong in a separate PR since they are performance optimizations that are not strictly needed for this PR.

andy-sweet:
> Overall, this looks good. I'm requesting changes because I think we can eliminate the `samplerBindings` map and avoid using implicit sampler uniform bindings in `WebGLShaderPrograms`.
> 
> I might be missing something here, but I would have expected something like this:
> 
> Replace this code:
> 
> ```ts
> object.textures.forEach((texture, index) => {
>   this.textures_.bindTexture(texture, index);
> });
> ```
> 
> with something like this:
> 
> ```ts
> if (object.imageMap) {
>   this.gl.activeTexture(IMG_SAMPLER_UNIT);
>   this.gl.bindTexture(object.imageMap, IMAGE_SAMPLER_UNIT);
> }
> 
> if (object.segmentationMap) {
>   this.gl.activeTexture(SEG_SAMPLER_UNIT);
>   this.gl.bindTexture(object.segmentationMap, SEG_SAMPLER_UNIT);
> }
> ```
> 
> * Remove the textures array from renderable object and replace it with explicit texture maps members.
> * Remove program validation, especially if it complicates the flow.

I removed the `samplerBindings` map in `shaders/index.ts` and program validation.

But I'm not sold on the suggestion above because it will expand the API of `RenderableObject` in a way that is not beneficial for some of the classes that extend it. This pattern might feel more natural for rendering more standard meshes with a fixed set of possible textures (e.g. for bump mapping, light mapping, etc.), but that's not my goal here (e.g. I want to use a texture for a color-cycle).

Instead, I'd prefer to treat a sampler uniform value just as any other uniform value and return it in `RenderableObject.getUniforms`, as in my [follow up PR to this](https://github.com/chanzuckerberg/idetik/pull/320/files#diff-08eecf40875632fdee13298194a3d4b1caf52aa213ff80cb885ea0560b658a1aR27). That way, I can look in one place (e.g. `ImageRenderable`) and understand the mapping between texture indices and uniform names in the shader. I don't like that `RenderableObject.addTexture` is public because that opens that mapping to outside influence, but it probably shouldn't be public.

Note that `this.textures_.bindTexture` (even before this PR) is a slightly misleading name since it does more than just bind the texture (i.e. it may also configure texture params and upload texture data). This PR also makes it set the active texture unit as in the code above. I don't mind renaming `bindTexture` (if you have some good ideas) or defining `WebGLTextures.activeTexture` and calling it from the renderer.

shlomnissan (approved):
> Instead, I'd prefer to treat a sampler uniform value just as any other uniform value and return it in RenderableObject.getUniforms, as in my [follow up PR to this](https://github.com/chanzuckerberg/idetik/pull/320/files#diff-08eecf40875632fdee13298194a3d4b1caf52aa213ff80cb885ea0560b658a1aR27). That way, I can look in one place (e.g. ImageRenderable) and understand the mapping between texture indices and uniform names in the shader.

Makes sense! And this looks good to me now.

PR #319, opened 2025-07-22
(Small) Add wireframe color to renderable objects

shlomnissan:
### Summary

This PR introduces a wireframe color option for renderable objects. This 
feature is especially helpful for debugging scenarios, such as assigning 
different wireframe colors to individual LOD tiles.  Additionally, a `debugMode`
flag has been added to the base layer class to support this and other
debug-related functionality.

### Notes
I think adding a `debugMode` flag to the layer and letting each layer define  its own behavior is sufficient for now. For example, in the case of `ImageLayer`, enabling `debugMode` simply adds  wireframes to its image renderables. That said, I'm open to a more structured approach if others have suggestions.

### Related Issue
N/A

### Tests & Checks
- Manual verification: See screenshots below
- All checks have passed

| Screenshot 2025-07-22 at 12 57 34 PM | Screenshot 2025-07-22 at 12 57 46 PM | Screenshot 2025-07-22 at 12 57 57 PM |
|--------|--------|--------|

### DIFF
--- a/packages/core/examples/chunk_streaming/main.ts
+++ b/packages/core/examples/chunk_streaming/main.ts
@@ -39,6 +39,8 @@ const chunkInfoOverlay = new ChunkInfoOverlay({
   imageLayer: imageLayer,
 });
 
+imageLayer.debugMode = true;
+
 new Idetik({
   canvas: document.querySelector<HTMLCanvasElement>("#canvas")!,
   camera,
--- a/packages/core/examples/chunk_streaming/main.ts
+++ b/packages/core/examples/chunk_streaming/main.ts
@@ -1,6 +1,5 @@
 import {
   Idetik,
-  AxesLayer,
   ImageLayer,
   OmeZarrImageSource,
   OrthographicCamera,
@@ -28,23 +27,23 @@ const region: Region = [
   { dimension: "x", index: { type: "full" } },
 ];
 const channelProps = [{ contrastLimits: [0, 255] as [number, number] }];
-const imageLayer = new ImageLayer({ source, region, channelProps });
-const axesLayer = new AxesLayer({ length: 1500, width: 0.01 });
 const camera = new OrthographicCamera(left, right, top, bottom);
 const fpsOverlay = new FpsOverlay({
   textDiv: document.querySelector<HTMLDivElement>("#fps-text")!,
 });
+
+const imageLayer = new ImageLayer({ source, region, channelProps });
+imageLayer.debugMode = true;
+
 const chunkInfoOverlay = new ChunkInfoOverlay({
   textDiv: document.querySelector<HTMLDivElement>("#chunk-info")!,
   imageLayer: imageLayer,
 });
 
-imageLayer.debugMode = true;
-
 new Idetik({
   canvas: document.querySelector<HTMLCanvasElement>("#canvas")!,
   camera,
   controls: new PanZoomControls(camera, camera.position),
-  layers: [imageLayer, axesLayer],
+  layers: [imageLayer],
   overlays: [fpsOverlay, chunkInfoOverlay],
 }).start();
--- a/packages/core/src/core/color.ts
+++ b/packages/core/src/core/color.ts
@@ -3,10 +3,16 @@ import { vec3, vec4 } from "gl-matrix";
 export type ColorLike = Color | vec3 | vec4;
 
 export class Color {
+  public static readonly RED: Color = new Color(1.0, 0.0, 0.0);
+  public static readonly GREEN: Color = new Color(0.0, 1.0, 0.0);
+  public static readonly BLUE: Color = new Color(0.0, 0.0, 1.0);
+  public static readonly BLACK: Color = new Color(0.0, 0.0, 0.0);
+  public static readonly WHITE: Color = new Color(1.0, 1.0, 1.0);
+
   // RGBA color values in the range [0, 1]
   private readonly rgba_: readonly [number, number, number, number];
 
-  constructor(r: number = 1.0, g: number = 1.0, b: number = 1.0, a?: number) {
+  constructor(r: number, g: number, b: number, a?: number) {
     if (r < 0 || r > 1 || g < 0 || g > 1 || b < 0 || b > 1) {
       throw new Error("RGB values must be in the range [0, 1]");
     }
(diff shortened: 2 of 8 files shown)


packages/core/src/core/color.ts
```diff
@@ -6,7 +6,7 @@ export class Color {
   // RGBA color values in the range [0, 1]
   private readonly rgba_: readonly [number, number, number, number];
 
-  constructor(r: number, g: number, b: number, a?: number) {
+  constructor(r: number = 1.0, g: number = 1.0, b: number = 1.0, a?: number) {
```
shlomnissan: I'm typically not a fan of default values, but I like having an empty state (white) for colors. Happy to revert if people disagree.

packages/core/src/renderers/webgl_shader_program.ts
```diff
@@ -30,14 +30,16 @@ export class WebGLShaderProgram {
     this.addShader(vertexShaderSource, gl.VERTEX_SHADER);
     this.addShader(fragmentShaderSource, gl.FRAGMENT_SHADER);
     this.link();
+    this.validateProgram();
+    this.preprocessUniformLocations();
+    this.deleteShaders();
```
shlomnissan: I took the opportunity to include a small refactor given the size of this PR. It moves program validation, preprocessing, and shader deletion out of the `link` function, where they didn’t quite belong.

andy-sweet: Snap! Well kind of... https://github.com/chanzuckerberg/idetik/pull/317/files#r2223599051

Here, I think `deleteShaders` can be deleted immediately after `link`, so can be called at the end of that function's definition and removed from here and the error case of `validate`.

packages/core/src/core/color.ts
andy-sweet: I don't know if this is my dark scheme bias showing, but I think of transparent black as my expected default because that's all zeros. This constructor didn't do that before because of the optional alpha that effectively defaults to 1, so I don't feel too strongly about this change.

But I'd probably just use `Color.WHITE` since that disambiguates what's expected from calling `new Color()`.

packages/core/src/layers/image_layer.ts
```diff
@@ -169,6 +177,11 @@ export class ImageLayer extends Layer {
       channelProps
     );
 
+    if (this.debugMode) {
+      image.wireframeEnabled = true;
+      image.wireframeColor = this.wireframeColors_[chunk.lod];
```
andy-sweet: This assumes the `chunk.lod` is less than `this.wireframeColors.length` which is 4 right now. In general, that won't be true. The simplest fix is probably to cycle through wireframe colors by using `chunk.lod % this.wireframeColors.length`.

packages/core/src/core/color.ts
shlomnissan: Ahh I didn't notice the static members. I will use `Color.WHITE` and pull them to the top of the file.

andy-sweet (changes requested):
This is super helpful and should make our upcoming demo of this even better!

My only blocking requested change is to fix indexing the wireframe colors by the chunk LOD in the image layer.

Another thought is that line thickness of the wireframes could be another pretty nice property if that gives us a way to see higher resolution LOD chunk wireframes on top of lower resolution ones. That's probably work for another PR.

packages/core/src/renderers/webgl_shader_program.ts
```diff
@@ -120,16 +122,15 @@ export class WebGLShaderProgram {
       const message = this.gl_.getProgramInfoLog(this.program_);
       throw new Error(`Error linking program: ${message}`);
     }
+  }
 
+  private validateProgram() {
```
andy-sweet: Nit: this is a method of `WebGLShaderProgram` so `Program` feels redundant and is inconsistent with `link` and `use`.

```suggestion
  private validate() {
```

packages/core/src/renderers/webgl_shader_program.ts
shlomnissan: Are you suggesting we move `deleteShaders` back to `link`? I'm happy to pull it up so it's called right after `link` but I'm not convinced it belongs there. It is a separate action, as well as validation.

packages/core/examples/chunk_streaming/main.ts
```diff
@@ -39,6 +39,8 @@ const chunkInfoOverlay = new ChunkInfoOverlay({
   imageLayer: imageLayer,
 });
 
+imageLayer.debugMode = true;
```
andy-sweet: With the wireframes, I think the `axesLayer` becomes slightly less useful and its colors clash with the wireframe colors. I think you could just remove `axesLayer` now, or maybe make its colors customizable and reduce the clashing.

packages/core/src/layers/image_layer.ts
shlomnissan: Sure, I can cycle through them.

packages/core/examples/chunk_streaming/main.ts
shlomnissan: I will remove it for now.

packages/core/src/renderers/webgl_shader_program.ts
andy-sweet: I'm not totally sure what `WebGLShaderProgram.shaders_` is supposed to be used for, but currently I think it's only used for cleanup. I think that cleanup only needs to happen in `addShader` and `link` - i.e. `preprocessUniformLocations` and `validateProgram` don't need the individual shaders now that the entire program is linked.

So my thinking was to call `deleteShaders` ASAP, so I'd also be OK with moving it up in the call chain here.

Please correct my thinking if I missed something here, which is very possible.

shlomnissan:
> Another thought is that line thickness of the wireframes could be another pretty nice property if that gives us a way to see higher resolution LOD chunk wireframes on top of lower resolution ones. That's probably work for another PR.

@andy-sweet Currently, our wireframe geometry is rendered using the `GL_LINES` primitive. However, WebGL appears to restrict line width to 1 in many environments. This is the case on my machine, as confirmed by:

```ts
gl.getParameter(gl.ALIASED_LINE_WIDTH_RANGE); // → [1, 1]
```

So adjusting the line width with `gl.lineWidth()` has no effect. If we want to support custom line widths, we’d need to switch to using projected line geometry or rely on another workaround (@aganders3 ?). At this stage, I don't think the added complexity is necessary.

andy-sweet (approved):
Looks good now!

packages/core/src/renderers/webgl_shader_program.ts
shlomnissan: Yep, it's only used for cleanup at the moment, so strictly speaking, it's not needed. We could replace it with a local variable in the constructor, as you have in the PR you linked above. There are reasons to keep it like deferred linking / relinking or debugging but they are not relevant at this point. Feel free to remove it in your upcoming PR.

Issue #177, opened 2025-02-14
Create a Labels layer


aganders3:
This should be capable of displaying dense segmentation data. The initial version just needs to work with 2D axis-aligned slices (i.e. Image and ImageSeries layers).

PR #320, opened 2025-07-22
Add a label image layer with an example

andy-sweet:
### Summary
This change adds a `LabelImageLayer` along with an associated `LabelImageRenderable`. The associated fragment shader makes use of two textures, one for the segmentation image data and the other for the color cycle.

It also adds an example uses that layer to show a multi-label segmentation image overlaid on a corresponding intensity image from public OrganelleBox data.

Screenshot 2025-07-24 at 10 02 11

### Related Issue
Related to #177 

### Tests & Checks
I manually tested the new example, which works as expected.
I also manually tested the existing examples to check for breakage.


packages/core/examples/image_labels_overlay/index.html
```diff
@@ -0,0 +1,27 @@
...
+    <title>Labels layer</title>
+    <style>
+      html,
+      body {
+        margin: 0;
+        padding: 0;
+        display: flex;
+        width: 100%;
+        height: 100%;
+        background-color: black;
```
andy-sweet, line 15: I chose a black background because this lets us see the labels outside of the extent of the image. If we want to support layer blending on a white background (either the canvas or its container), I think we'd need to some more work in general.

packages/core/src/objects/textures/texture_rgba.ts
```diff
@@ -0,0 +1,35 @@
+import { TextureData, Texture } from "../../objects/textures/texture";
+
+export class TextureRgba extends Texture {
```
andy-sweet: Technically, I don't need this and can use `Texture2D` instead by setting `Texture2D.dataFormat = "rgba"` after construction. But that feels like it breaks the semantics of these texture classes and I'd expect to be able to support more traditional RGBA texture sources in the future.

andy-sweet:
@aganders3 : assigned you as a reviewer as you missed #317. Other reviews and comments are welcome.

andy-sweet:
It might be helpful to look at the follow-up to this for more context about what's coming: https://github.com/chanzuckerberg/idetik/pull/322

packages/core/src/objects/textures/texture_rgba.ts
aganders3: Is 2D implied here or should it be `Texture2DRgba`?

I imagine 3D rgb textures will be rare unless we're doing using the channels for some other purpose.

packages/core/src/layers/label_image_layer.ts
```diff
@@ -0,0 +1,88 @@
...
+    if (this.state !== "initialized") {
+      throw new Error(`Trying to load chunks more than once.`);
+    }
+    this.setState("loading");
+    const loader = await this.source_.open();
+    const attributes = await loader.loadAttributes();
+    const lod = this.lod_ ?? attributes.length - 1;
+    const chunk = await loader.loadRegion(region, lod);
+    this.image_ = this.createImage(chunk);
+    this.addObject(this.image_);
```
aganders3, line 73: Do we want to use the ChunkManager here from the start?

I'm fine with that being added after this PR, but I would hope it's not too hard since this is so similar to an ImageLayer.

packages/core/src/objects/textures/texture_rgba.ts
```diff
@@ -0,0 +1,35 @@
...
+  private data_: TextureData;
+  private readonly width_: number;
+  private readonly height_: number;
+
+  constructor(data: TextureData, width: number, height: number) {
+    super();
+    this.data_ = data;
+    this.width_ = width;
+    this.height_ = height;
```
aganders3: nit: debugging might benefit from an assertion here that `data.length === width * height` or something

packages/core/src/renderers/shaders/label_image_frag.glsl
```diff
@@ -0,0 +1,25 @@
...
+
+uniform float u_opacity;
+
+in vec2 TexCoords;
+
+void main() {
+    uint texel = texture(ImageSampler, TexCoords).r;
+    if (texel == 0u) {
+        fragColor = vec4(0.0, 0.0, 0.0, 0.0);
+        return;
```
aganders3, line 19:
```suggestion
        discard;
```

I think this is effectively the same and might be faster. Looking at #322 it might be irrelevant.

packages/core/src/objects/renderable/image_renderable.ts
```diff
@@ -11,12 +11,14 @@ import { vec3 } from "gl-matrix";
...
 
 type SingleUniformValues = {
+  ImageSampler: number;
   Color: vec3;
   ValueOffset: number;
   ValueScale: number;
 };
 
 type ArrayUniformValues = {
+  ImageSampler: number;
```
aganders3, line 21: Sorry this probably is more of a comment on #317.

This feels like we're encoding the same information or at least having to keep them in sync between the array of textures and the texture unit specified here. I think this can be improved later, along with some other ideas like making `addTexture` private (or protected, or removing it), so this is  non-blocking.

It feels like it should be enough to just set the textures  themselves in the map returned by `getUniforms`, and let the renderer/program decide which texture units to bind them to. Does that make sense?

aganders3 (approved):
This looks good and works well for me. Nice job!

Is there a way to constrain the type of the `ImageChunkSource` for the layer? What will happen if a float image is provided, for example? Will it get all the way to rendering and crash because the texture does not match the sampler type (usampler)?

andy-sweet:
> Is there a way to constrain the type of the `ImageChunkSource` for the layer? What will happen if a float image is provided, for example? Will it get all the way to rendering and crash because the texture does not match the sampler type (usampler)?

Good catch. I decided to add some validation logic to `LabelImageRenderable` since that (through its coupling to the fragment shader) determines the validity of the texture. We can also relax the acceptable pixel types there and in the shader as needed.

I think it would be great to be able to quickly check the properties of the image source to validate much earlier on in the layer's lifetime, but that feels out of scope in this particular PR. I'll create an issue from this comment.

packages/core/src/renderers/shaders/label_image_frag.glsl
andy-sweet: I wasn't aware of `discard` so this comment is helpful regardless of what we do. In this PR, I plan to leave this as is because of #322, like you mentioned.

But I did consider having a special `transparentColors` uniform or texture to handle visibility of certain label values. I avoided it to keep the shader relatively simple, but I'll likely create an issue in case we want to reconsider.

packages/core/src/objects/textures/texture_rgba.ts
andy-sweet: Looking at this again, I am regretting my choice to introduce `TextureRgba` in this PR because I really don't need it. And, in general, I feel like our texture classes could do with a little refinement anyway.

Would you be OK if I just use a `Texture2D` and punt the support of other RGBA texture sources to an issue?

packages/core/src/objects/renderable/image_renderable.ts
andy-sweet: Agreed with the general idea. I really like the idea of letting a specific renderer decided how it actually maps the textures to something like a texture unit. I will do my best to capture these ideas in an issue.

I think one big underlying problem right now is that our renderable objects have a tight coupling to WebGL2 through their associated shaders. Instead, I think they should just collect a few things together (i.e. geometries, textures, properties) then a specific renderer can use to actually create a rendering program for them.

While this problem exists, I'd argue that `getUniforms` is meaningful here (because the uniform is the ImageSampler and its value is the texture unit - ugh!) and having an entry with `ImageSampler: imageDataTexture` would not be, so I'm not planning to make any changes in this PR.

packages/core/src/layers/label_image_layer.ts
andy-sweet: I lean towards not wanting to right now because it makes the logic in the layer hard to follow with some downsides and maybe not that much performance benefit for many label layers (because the images tend to be smaller in overall data volume or chunks bigger in extent).

I think we should prioritize coming up with a design for the chunk manager that would make it easy to use across any layer (instead of adding conditional logic to each), and removes the need for passing in an explicit region for a layer. Once that is done, I would want to use it here.

packages/core/src/objects/textures/texture_rgba.ts
aganders3: Yeah that's fine with me!

packages/core/src/objects/renderable/image_renderable.ts
andy-sweet: Created https://github.com/chanzuckerberg/idetik/issues/326

packages/core/src/renderers/shaders/label_image_frag.glsl
andy-sweet: Created https://github.com/chanzuckerberg/idetik/issues/325

PR #321, opened 2025-07-23
Enable local directory source in <OmeZarrImageViewer>

bchu1:
I originally made `sourceUrl` and `sourceLocalDirectory` a "one of" using a discriminated union, but realized that when using the local file system, initially neither will be set because the user has to select the directory first. So the component has been updated to support an initial state with no source (hide all the stuff on the edges).

Diff looks better when hiding whitespace changes.

Maybe the demo in the base PR doesn't need to exist now that there's a demo in react/?

https://github.com/user-attachments/assets/e542cbb9-d3c4-43a4-b65b-dc5644c1ff52


packages/react/src/components/viewers/OmeZarrImageViewer/OmeZarrImageViewer.tsx
```diff
@@ -92,12 +99,28 @@ export function OmeZarrImageViewer(props: OmeZarrImageViewerProps) {
...
+        sourceLocalDirectory.directory,
+        sourceLocalDirectory.path
+      );
+    } else {
+      return;
+    }
     setSource(newSource);
     setAllSlicesLoaded(false);
-  }, [sourceUrl, resolutionLevel]);
+    // eslint-disable-next-line -- Don't care if sourceLocalDirectory changes, only its properties
```
aganders3: I'm not sure it works all the time, but sometimes you can get rid of the lint error by first destructuring `const { directory, path } = sourceLocalDirectory` and including the individual variables in the dependency array.

The benefit is just to not fully disable the lint here, but it's not a big deal.

aganders3 (approved):
I think this looks good.

I agree you can probably remove the local directory example from the core examples, because this is basically the same thing.

PR #322, opened 2025-07-24
Add an explicit colormap to the label image layer 

andy-sweet:
### Summary
This change adds a colormap to the label image layer, so that certain label values can be mapped to specific colors. This map is passed to the fragment shader as a uint32 scalar texture where the label keys are in the first row and the RGBA colors are packed into the second row. The fragment shader iterates over those keys and unpacks the color when needed.

It modifies an existing example showing multi-label segmentation image overlaid on a corresponding intensity image from public OrganelleBox data, with a specific segmented mapped to a green color outside the other cycled colors.

Screenshot 2025-07-24 at 11 02 51

### Related Issue
Related to #177 

### Tests & Checks
I manually tested the new example, which works as expected.
I also manually tested the existing examples to check for breakage.


andy-sweet:
Converting to draft because this is based on #320 which is still in review

packages/core/src/renderers/shaders/label_image_frag.glsl
```diff
@@ -7,17 +7,34 @@ layout (location = 0) out vec4 fragColor;
...
 
 in vec2 TexCoords;
 
+vec4 unpackRgba(uint packed) {
+    uint r = (packed >> 24u) & 0xFFu;
+    uint g = (packed >> 16u) & 0xFFu;
+    uint b = (packed >> 8u) & 0xFFu;
+    uint a = packed & 0xFFu;
+    return vec4(float(r), float(g), float(b), float(a)) / 255.0;
+}
```
aganders3, line 22: As we get more utility functions like this, we may want to make them reusable across programs. I think neuroglancer uses a glsl "lib" file, and i think vispy includes functions dynamically kind of like templating.

aganders3:
I know this is still a draft, but it looks nice!

We should probably add colormap support to `image_renderable` as well.

packages/core/src/core/color.ts
```diff
@@ -53,6 +54,15 @@ export class Color {
     return `#${this.toHexComponent(this.r)}${this.toHexComponent(this.g)}${this.toHexComponent(this.b)}`;
   }
 
+  public get packed(): number {
+    return (
+      (Math.round(this.r * 255) << 24) |
```
andy-sweet, line 59: Based on the [MDN Number docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number) (see Fixed-width number conversion), I think bit-shifting these numbers is guaranteed to work as intended (because they are 8-bit values left shifted into 32-bit ints), but let me know if there are concerns here and I can do something a bit more explicit with buffer views.

andy-sweet:
> We should probably add colormap support to `image_renderable` as well.

I wasn't planning to do that in this PR.

I interpret that support to be more of a colormap that maps from continuous pixel values to colors with some control points, colors, and interpolation function. Whereas the colormap here is just a direct map from specific values to colors, which is likely not useful when the plain image renderable is used.

Definitely happy to create an issue to track adding such a colormap to image renderable. And maybe an issue concerning making these concepts more integrated (which was avoided as an approach in this first pass [segmentation image support](https://docs.google.com/document/d/1fzP8cobUBpLsHjv8VGgejFnnv-upP9myJM9qWqWWshw/edit?tab=t.0#heading=h.5bk4rpm4lbef)).

andy-sweet:
@aganders3 : re-requested your review because this is non in draft anymore and you've already taken a look.

aganders3:
> > We should probably add colormap support to image_renderable as well.


> I wasn't planning to do that in this PR.

I should have been clear - not in this PR!

> Definitely happy to create an issue to track adding such a colormap to image renderable. And maybe an issue concerning making these concepts more integrated (which was avoided as an approach in this first pass [segmentation image support](https://docs.google.com/document/d/1fzP8cobUBpLsHjv8VGgejFnnv-upP9myJM9qWqWWshw/edit?tab=t.0#heading=h.5bk4rpm4lbef)).

I don't feel strongly about needing an issue for this. I agree the concepts are different, so I also don't think they need to be more integrated. We will remember it when it becomes a need and it should be pretty quick to implement using similar patterns to what you added here.

aganders3 (approved):
Looks good - nice work!

PR #327, opened 2025-07-25
Remove local file example from index

andy-sweet:
### Summary
Clean up missed in https://github.com/chanzuckerberg/idetik/pull/321

### Tests & Checks
The core local file example is not longer shown in the index.


bchu1 (approved):
Thanks!

PR #329, opened 2025-07-25
Refactor effects in <OmeZarrImageViewer>

bchu1:
Previously (4 effects):
1. Effect creates source
2. which simultaneously triggers
    1. effect that creates layer
    2. and effect that fetches metadata
        1. which triggers effect that sets index and loads layer
        2. which triggers callback to add layer to manager
        3. and preload slices (if applicable)
        
Now there is only 1 effect that performs all initialization (and re-initialization when relevant props change).

- I've also added a few `#region` comments to the file which will display headers in the VS Code minimap if you check "Editor › Minimap: Show Region Section Headers".
- The previous initialization logic can now be read linearly in the single effect (there is some parellelism between the 2 fetches on lines 166 and 120 to preserve the previous behavior for performance purposes).
- As a side effect, this PR also fixes the bug in `main` where `shouldAutoLoadAllSlices` doesn't display the correct initial loading state due to a race condition between the effects.
- This also removes the need to use the `state` callback on layer (which was my original motivation for this refactor, as I was adding support for `ImageLayer` whose `state` works differently from `ImageSeriesLayer` (maybe another problem we should address at some point) which was making me add more ugly code than I expected with the way things worked previously).

### DIFF
--- a/packages/react/src/components/viewers/OmeZarrImageViewer/OmeZarrImageViewer.tsx
+++ b/packages/react/src/components/viewers/OmeZarrImageViewer/OmeZarrImageViewer.tsx
@@ -90,7 +90,6 @@ export function OmeZarrImageViewer({
 
   const { isReady: runtimeIsReady, runtime } = useIdetik();
 
-  const [source, setSource] = useState<OmeZarrImageSource | null>(null);
   const [unit, setUnit] = useState<string>();
   const [zRange, setZRange] = useState<[number, number]>([0, 0]);
   const [zValue, setZValue] = useState(0.5);
@@ -99,6 +98,7 @@ export function OmeZarrImageViewer({
   const [extraControlProps, setExtraControlProps] = useState<
     ExtraControlProps[]
   >([]);
+  const sourceRef = useRef<OmeZarrImageSource | null>(null);
   const imageLayerRef = useRef<ImageSeriesLayer | null>(null);
 
   // #region Initialization
@@ -110,7 +110,7 @@ export function OmeZarrImageViewer({
       if (source === undefined) {
         return;
       }
-      setSource(source);
+      sourceRef.current = source;
       setAllSlicesLoaded(false);
       setLoading(true);
       const loadChannelMetadataPromise = loadChannelMetadata(
@@ -128,6 +128,9 @@ export function OmeZarrImageViewer({
       const { channelProps, extraControlProps } =
         await loadChannelMetadataPromise;
       const { xUnit, zRange, zValue } = await loadImageMetadataPromise;
+      if (sourceRef.current !== source) {
+        return;
+      }
       setExtraControlProps(extraControlProps);
       setUnit(xUnit);
       setZRange(zRange);
@@ -142,11 +145,10 @@ export function OmeZarrImageViewer({
       imageLayerRef.current = layer;
       onLayerCreated?.();
       await updateSeriesIndex(zValue);
-      if (imageLayerRef.current !== layer) {
+      if (sourceRef.current !== source) {
         return;
       }
       runtime.layerManager.add(layer);
-      imageLayerRef.current = layer;
       setCameraFrame(layer, runtime);
       setLoading(false);
       onFirstSliceLoaded?.();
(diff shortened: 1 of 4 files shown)


phoenixAja (commented):
This is sooo much cleaner and easier to reason about now that everything is orchestrated via initialize().

One potential blocker: I think there's a race condition risk if initialize() re-runs quickly (e.g., due to prop changes or state oscillation). Since async work like loadChannelMetadata, loadImageMetadata, and createLayer isn't cancellable or tracked per invocation, there's nothing preventing a stale initialize() run from overriding state after a newer one starts.

bchu1:
> This is sooo much cleaner and easier to reason about now that everything is orchestrated via initialize().
> 
> One potential blocker: I think there's a race condition risk if initialize() re-runs quickly (e.g., due to prop changes or state oscillation). Since async work like loadChannelMetadata, loadImageMetadata, and createLayer isn't cancellable or tracked per invocation, there's nothing preventing a stale initialize() run from overriding state after a newer one starts.

Great callout! Previously this was done with various `if (imageLayerRef.current !== layer)` checks throughout the effects, I've cleaned up what was left of those and switched `source` from being a state to ref (we didn't really need that to be state anyways) and using that to check if a newer effect has been triggered (since setting `source` is the actual first step of the sequence) and returning after `await`s

packages/react/src/components/viewers/OmeZarrImageViewer/OmeZarrImageViewer.tsx
```diff
@@ -201,166 +166,72 @@ export function OmeZarrImageViewer({
         imageLayerRef.current = null;
       }
     };
+    // eslint-disable-next-line react-hooks/exhaustive-deps -- These are the only props that matter
```
aganders3: I don't love disabling this lint. I'm sure it's fine but it could miss new props that should be included.

packages/react/src/components/viewers/OmeZarrImageViewer/OmeZarrImageViewer.tsx
```diff
@@ -477,3 +351,170 @@ export function OmeZarrImageViewer({
...
+  return new ImageSeriesLayer({
+    source,
+    region,
+    channelProps,
+    seriesDimensionName,
+    lod: resolutionLevel,
+  });
+}
+
+function setCameraFrame(layer: ImageSeriesLayer, runtime: Idetik) {
```
aganders3: nit: `setCameraFrame` makes this look like a state modifier and doesn't really describe what the function does. I'd suggest we name this `zoomToFit`.

aganders3 (approved):
Thanks! This looks like an improvement to me as well.

I _think_ I understand and agree with your explanation (@bchu1) about the race condition @phoenixAja brought up above. The cleanup function for the `useEffect` does some cleanup (setting the layer reference to null) - should this also do some clean up for the source ref or could that cause a loop?

I think `loadImageMetadata` is still pretty hairy but at least it's factored out into its own function out of the main component logic.

bchu1:
> Thanks! This looks like an improvement to me as well.
> 
> I _think_ I understand and agree with your explanation (@bchu1) about the race condition @phoenixAja brought up above. The cleanup function for the `useEffect` does some cleanup (setting the layer reference to null) - should this also do some clean up for the source ref or could that cause a loop?
> 
> I think `loadImageMetadata` is still pretty hairy but at least it's factored out into its own function out of the main component logic.

I don't think cleaning up `source` should be necessary, since the next time the effect runs it'll set it again immediately. The cleanup function also runs synchronously right before the effect is rerun anyways, so it's basically the same.

There's still a lot that could be improved, but I tried to touch some of the hairier logic as little as possible for now so I didn't break stuff 😅

packages/react/src/components/viewers/OmeZarrImageViewer/OmeZarrImageViewer.tsx
bchu1: Yeah, this is hard. I think this is inevitable, unfortunately. One reason is that, previously, we had the callback props in the dependency array of the effect that creates the layer. That works if the parent component uses `useCallback(() => ..., [])` but let's say their callback actually needs to have something like a state in its dependency array in order to execute correctly, that would trigger layer recreation every time their state changes (this would also apply to helper functions in `<OmeZarrImageViewer>` as well). Another reason is if we want to support not requiring the parent to make things like `region` stable by doing a deep comparison of the array, we'd also have to disable this.

PR #330, opened 2025-07-25
Factor out interface for channel-related layer properties

bchu1:
This is so `<OmeZarrImageViewer>` can support `ImageLayer` as well, since Jonathan's images are 2D.

### DIFF
--- a/packages/core/src/core/layer.ts
+++ b/packages/core/src/core/layer.ts
@@ -1,6 +1,7 @@
 import { IdetikContext } from "../idetik";
 import { RenderableObject } from "./renderable_object";
 import { clamp } from "../utilities/clamp";
+import { ChannelProps } from "@/objects/textures/channel";
 
 export type LayerState = "initialized" | "loading" | "ready";
 export type blendMode = "normal" | "additive" | "subtractive" | "multiply";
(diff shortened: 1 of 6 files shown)


packages/core/src/core/layer.ts
```diff
@@ -1,6 +1,7 @@
 import { IdetikContext } from "../idetik";
 import { RenderableObject } from "./renderable_object";
 import { clamp } from "../utilities/clamp";
+import { ChannelProps } from "@/objects/textures/channel";
```
andy-sweet: Nit: I don't remember exactly why, but I thought there was a reason we were sticking to relative imports right now.

```suggestion
import { ChannelProps } from "../objects/textures/channel";
```

packages/core/src/core/layer.ts
```diff
@@ -107,3 +108,12 @@ export abstract class Layer {
     this.objects_ = [];
   }
 }
+
+/** Layer that exposes channel controls. */
+export interface ChannelsEnabled {
```
andy-sweet: A little nitty, but this feels out of place in the core `layer` module because

1) it will likely only apply to image-like layers;
2) it depends on something outside of core.

For those reasons, I'd personally prefer this to be in `channel.ts`.

It must be exported in the main module, so this doesn't matter too much in practice.

andy-sweet (approved):
Looks good and works for me. I think the interface should live somewhere else, but that's optional.

packages/core/src/core/layer.ts
phoenixAja: yes i think it was related to propagating types for the workspaces

bchu1: we ran into issues with integrating the library as a submodule. Shouldn't be an issue once we make `npm install` the way to integrate

PR #331, opened 2025-07-28
Add an event dispatcher class

shlomnissan:
### Summary
This PR adds a centralized `EventDispatcher` and integrates it into the Idetik runtime.

Centralizing events gives us:
- Consistent event handling across layers and tools
- Fine-grained control over event propagation
- Separation of DOM events from internal interaction logic
- A clean entry point for future extensions like hit testing or tools

#### Notes
- I made `onEvent` an abstract function on the layer's base class. I like the fact the interface is enforced but it does feel like boilerplate at this point. Thoughts?
- The next step is using the event dispatcher to process camera controls.

### Related Issue
N/A

### Tests & Checks
- Verified manually: Add a log for the event in the image layer to see it in action
- All checks have passed

### DIFF
--- a/packages/core/src/core/event_dispatcher.ts
+++ b/packages/core/src/core/event_dispatcher.ts
@@ -1,0 +1,48 @@
+export interface EventContext {
+  readonly type: string;
+  readonly event: Event;
+  readonly propagationStopped: boolean;
+  stopPropagation(): void;
+}
+
+type Listener = (event: EventContext) => void;
+
+export class EventDispatcher {
+  private readonly listeners_: Listener[] = [];
+
+  constructor(canvas: HTMLCanvasElement) {
+    [
+      "pointerdown",
+      "pointermove",
+      "pointerup",
+      "pointercancel",
+      "wheel",
+    ].forEach((type) => {
+      canvas.addEventListener(type, this.handleEvent, { passive: false });
+    });
+  }
+
+  public onEvent(listener: Listener) {
+    this.listeners_.push(listener);
+  }
+
+  private readonly handleEvent = (e: Event) => {
+    let stopped = false;
+
+    const event: EventContext = {
+      type: e.type,
+      event: e,
+      get propagationStopped() {
+        return stopped;
+      },
+      stopPropagation() {
+        stopped = true;
+      },
+    };
+
+    for (const listener of this.listeners_) {
+      listener(event);
+      if (event.propagationStopped) break;
+    }
+  };
+}
--- a/packages/core/src/core/event_dispatcher.ts
+++ b/packages/core/src/core/event_dispatcher.ts
@@ -1,8 +1,35 @@
-export interface EventContext {
-  readonly type: string;
+import { Logger } from "../utilities/logger";
+
+const eventTypes = [
+  "pointerdown",
+  "pointermove",
+  "pointerup",
+  "pointercancel",
+  "wheel",
+] as const;
+
+type EventType = (typeof eventTypes)[number];
+function isEventType(type: string): type is EventType {
+  return (eventTypes as readonly string[]).includes(type);
+}
+
+export class EventContext {
+  readonly type: EventType;
   readonly event: Event;
-  readonly propagationStopped: boolean;
-  stopPropagation(): void;
+  private propagationStopped_: boolean = false;
+
+  constructor(type: EventType, event: Event) {
+    this.type = type;
+    this.event = event;
+  }
+
+  get propagationStopped() {
+    return this.propagationStopped_;
+  }
+
+  stopPropagation() {
+    this.propagationStopped_ = true;
+  }
 }
 
 type Listener = (event: EventContext) => void;
@@ -11,35 +38,22 @@ export class EventDispatcher {
   private readonly listeners_: Listener[] = [];
 
   constructor(canvas: HTMLCanvasElement) {
-    [
-      "pointerdown",
-      "pointermove",
-      "pointerup",
-      "pointercancel",
-      "wheel",
-    ].forEach((type) => {
+    eventTypes.forEach((type) => {
       canvas.addEventListener(type, this.handleEvent, { passive: false });
     });
   }
 
-  public onEvent(listener: Listener) {
+  public addEventListener(listener: Listener) {
     this.listeners_.push(listener);
   }
 
   private readonly handleEvent = (e: Event) => {
-    let stopped = false;
+    if (!isEventType(e.type)) {
+      Logger.error("EventDispatcher", `Unsupported event type ${e.type}`);
+      return;
+    }
 
-    const event: EventContext = {
-      type: e.type,
-      event: e,
-      get propagationStopped() {
-        return stopped;
-      },
-      stopPropagation() {
-        stopped = true;
-      },
-    };
-
+    const event = new EventContext(e.type, e);
     for (const listener of this.listeners_) {
       listener(event);
       if (event.propagationStopped) break;
--- a/packages/core/src/idetik.ts
+++ b/packages/core/src/idetik.ts
@@ -1,6 +1,7 @@
 import { Camera } from "./objects/cameras/camera";
 import { Layer } from "./core/layer";
 import { LayerManager } from "./core/layer_manager";
+import { EventContext, EventDispatcher } from "./core/event_dispatcher";
 import { WebGLRenderer } from "./renderers/webgl_renderer";
 import { CameraControls } from "./objects/cameras/controls";
 import { Logger } from "./utilities/logger";
@@ -29,7 +30,7 @@ export class Idetik {
   public camera: Camera;
   public readonly canvas: HTMLCanvasElement;
   public readonly overlays: Overlay[];
-
+  private readonly eventDispatcher_: EventDispatcher;
   private readonly renderer_: WebGLRenderer;
   private readonly context_: IdetikContext;
   private readonly chunkManager_: ChunkManager;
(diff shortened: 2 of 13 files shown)


packages/core/src/core/event_dispatcher.ts
```diff
@@ -0,0 +1,48 @@
+export interface EventContext {
+  readonly type: string;
```
andy-sweet: As a string, `EventContext.type` doesn't seem useful to me because I could just get `EventContext.event.type` instead.

But if it were a union of literal strings, then it would be more useful because any event handlers could ensure that the event they are handling is handled by Idetik's `EventDispatcher`.

packages/core/src/core/event_dispatcher.ts
```diff
@@ -0,0 +1,48 @@
...
+  stopPropagation(): void;
+}
+
+type Listener = (event: EventContext) => void;
+
+export class EventDispatcher {
+  private readonly listeners_: Listener[] = [];
+
+  constructor(canvas: HTMLCanvasElement) {
+    [
```
andy-sweet: Consider defining this once in the module as a readonly array and using it to define event types that Idetik handles.

packages/core/src/core/event_dispatcher.ts
```diff
@@ -0,0 +1,48 @@
...
+    let stopped = false;
+
+    const event: EventContext = {
+      type: e.type,
+      event: e,
+      get propagationStopped() {
+        return stopped;
+      },
+      stopPropagation() {
+        stopped = true;
```
andy-sweet: After reading the `EventContext` interface where `EventContext.propagationStopped` is `readonly`, this feels a little misleading, but maybe I'm just not used to reading enough TypeScript.

packages/core/src/core/event_dispatcher.ts
```diff
@@ -0,0 +1,48 @@
...
+      },
+    };
+
+    for (const listener of this.listeners_) {
+      listener(event);
```
andy-sweet, line 58: Another pattern I've seen that would require less code here is to return a boolean to indicate that the event has been handled. If more information about the handling is needed a structure can be returned instead.

I.e. do something like `if (listener(event)) break` here.

packages/core/src/idetik.ts
```diff
@@ -73,6 +72,15 @@ export class Idetik {
     }
 
     this.overlays = params.overlays ?? [];
+
+    this.eventDispatcher_ = new EventDispatcher(canvas);
+    this.eventDispatcher_.onEvent((event: EventContext) => {
+      for (const layer of this.layerManager.layers) {
```
andy-sweet, line 78: This implies that only the first layer to handle the event will receive it, which is probably fine for now, as I don't think we have another option. Another pattern I've seen is to have an active/current/selected layer, in which case we might consider passing it to that instead.

packages/core/src/idetik.ts
```diff
@@ -73,6 +72,15 @@ export class Idetik {
     }
 
     this.overlays = params.overlays ?? [];
+
+    this.eventDispatcher_ = new EventDispatcher(canvas);
+    this.eventDispatcher_.onEvent((event: EventContext) => {
```
andy-sweet: I think we need to handle the case when the instance of `LayerManager.layers` changes using  `LayerManager.addLayersChangeCallback`.

packages/core/src/idetik.ts
```diff
@@ -73,6 +72,15 @@ export class Idetik {
...
+      for (const layer of this.layerManager.layers) {
+        layer.onEvent(event);
+        if (event.propagationStopped) return;
+      }
+      // TODO: pass event to camera controls if needed
```
andy-sweet, line 82: To review this PR, I'd find it helpful to see a draft PR of this implemented to understand how you're going to handle `PanZoomControls` adding event listeners after the `pointerdown` event. I see a couple of ways, but want to make sure we have some sensible path forward.

andy-sweet (changes requested):
I have a few comments, but the only blocking feedback is related to the current behavior when layers are added or removed (in which case I think they may not receive events).

andy-sweet:
> * I made `onEvent` an abstract function on the layer's base class. I like the fact the interface is enforced but it does feel like boilerplate at this point. Thoughts?

I don't have strong opinions here. Given that `Layer` is not a pure abstract class right now, I don't mind just defining an empty implementation there and I think this is discoverable enough for a Layer implementor. I also don't think the boilerplate is too offensive here either.

packages/core/src/core/event_dispatcher.ts
shlomnissan: The `propagationStopped` property is declared readonly to indicate that consumers of the event shouldn't assign to it directly. It's intended to be a computed property that reflects internal state, which is only modified by the `stopPropagation()` method. Do you have any suggestions on how to make this clearer?

packages/core/src/core/event_dispatcher.ts
shlomnissan: I considered that route initially (and mentioned it in my Slack thread) but decided to go with `stopPropagation()` instead, as it's more explicit and maps closely to the DOM event model (`event.stopPropagation()`) which developers are already familiar with

phoenixAja:
>  I made onEvent an abstract function on the layer's base class. I like the fact the interface is enforced but it does feel like boilerplate at this point. Thoughts?

I like that onEvent() is standardized across layers, but I wonder if we could provide a default no-op implementation in the base Layer class to avoid requiring boilerplate in layers that don’t care about interaction yet? That would still give us consistency without forcing every subclass to stub it out.

I would rather have the boilerplate than add if ('onEvent' in layer) checks everywhere

packages/core/src/idetik.ts
shlomnissan: > This implies that only the first layer to handle the event will receive it

Yes, assuming you mean the first layer stops propagation.

> Another pattern I've seen is to have an active/current/selected layer

Having an active layer and dispatching only to it would be a different interaction model that might be better, but I suggest we hold off until we better understand how the library will support multiple viewports and layers within these viewports. Either way, I think it is a simple change we can make later

packages/core/src/idetik.ts
shlomnissan: I think this case is already handled. Referencing `layerManager.layers` in every event always reads the current state. Am I missing something?

packages/core/src/idetik.ts
shlomnissan: I don't think the controls impact the validity or design of the event dispatcher pattern itself. The role of it is to standardize event propagation and control, and I think that's a good idea regardless of when listeners are attached by other systems like controls.

For this PR, I think the `EventDispatcher` stands on its own and I don't see a reason to hold off. That said, I'm also not in a rush so if you feel strongly about it, I can put the next PR first.

Currently, the controls class feels a little more complex than I think it should be and there might be an opportunity to refactor it, which is why I didn't want to couple these changes.

packages/core/src/idetik.ts
andy-sweet: Ah, you're right. I misread how the layers are bound to the function.

I'm not sure if had an effect on my read here, but I'd expect a name like `EventDispatcher.addListener` rather than `onEvent`.

packages/core/src/idetik.ts
andy-sweet: > Currently, the controls class feels a little more complex than I think it should be and there might be an opportunity to refactor it, 

This is partly why I asked for more context - it would be another good motivator for this PR. But given this is small and probably sensible independently, I'm good to merge this and move forward.

packages/core/src/core/event_dispatcher.ts
andy-sweet: Agreed on the familiarity, less so on the explicit, but good enough for now.

packages/core/src/core/event_dispatcher.ts
andy-sweet: Your explanation is clear, so the problem is probably on my side.

Another thought is that we could just implement replace the interface with a class

```typescript
export class EventContext {
  readonly type: EventType;
  readonly event: Event;
  private propagationStopped_: boolean = false;

  get propagationStopped() {
    return this.propagationStopped_;
  }

  stopPropagation() {
    this.propagationStopped_ = true;
  }
}
```

which is more clear to me.

But if we actually expect other implementations of `EventContext`, then we should just stick with what we've got.

andy-sweet (approved):
Looks good now after your explanation and adding explicit types. All of my other comments are optional.

packages/core/src/core/event_dispatcher.ts
```diff
@@ -0,0 +1,57 @@
...
+      canvas.addEventListener(type, this.handleEvent, { passive: false });
+    });
+  }
+
+  public onEvent(listener: Listener) {
+    this.listeners_.push(listener);
+  }
+
+  private readonly handleEvent = (e: Event) => {
+    if (!isEventType(e.type)) return;
```
andy-sweet: I think this condition is never intended to true, so this feels like an assertion. I'm not aware of a standard way to define this kind of assertion in TypeScript (i.e. a non-production throw), but maybe we should log an error here to avoid the implementation changing in the future.

packages/core/src/core/event_dispatcher.ts
shlomnissan: Ah yeah, this is much better. Not sure why I was hung up on an interface.

packages/core/src/core/event_dispatcher.ts
shlomnissan: > less so on the explicit

Say more?

PR #347, opened 2025-07-30
Add a label image series layer

### LINKED ISSUE #177: Create a Labels layer

andy-sweet:
### Summary
This introduces a `LabelImageSeriesLayer` to handle loading and rendering label image data in a series across a chosen dimension. Because this new layer overlaps with both the `LabelImageLayer` and `ImageSeriesLayer`, I did some refactoring to limit large chunks of repeated code, though there is still some remaining boiler plate.

The larger refactor is a new non-exported `ImageSeriesLoader` that attempts to encapsulate the complex loading logic of the existing `ImageSeriesLayer`, which is identical in the `LabelImageSeriesLoader`. I expect this class to go away in the future (e.g. with a more sophisticated `ChunkManager`), and I hope its introduction will make that removal easier.

I also refactored the color-cycle and lookup-table into a single `LabelColorMap` class that can map from label image pixel values to colors (though that actually happens in the fragment shader).

Finally, I added a new example that is the same as the label image overlay, but introduces a slider over the time dimension.

https://github.com/user-attachments/assets/999fa362-acbe-4716-9ccd-73e6ffd6c8e5

### Related Issue
Closes #177

### Tests & Checks
I manually tested the new example.


packages/core/examples/image_labels_overlay_with_value_picking/main.ts
```diff
@@ -97,8 +97,10 @@ const labelsLayer = new LabelImageLayer({
   opacity: 0.25,
   blendMode: "normal",
   lod,
-  colorCycle: [Color.YELLOW, Color.MAGENTA, Color.CYAN],
-  colorMap: new Map([[103, Color.GREEN]]),
+  colorMap: {
+    lookUpTable: new Map([[103, Color.GREEN]]),
```
aganders3: Maybe the worlds biggest (tiniest?) nit, but I'd expect this to be `lookupTable`. I know that's commonly abbreviated `LUT` but I think it's only two words when written out like "lookup table".

I guess same with the `ColorLookUpTableSampler` name in the shader.

This is a very weakly held preference though so feel free to not change it.

packages/core/examples/image_series_labels_overlay/main.ts
```diff
@@ -0,0 +1,165 @@
...
+}
+
+async function setLayerIndex(index: number) {
+  tIndexEl!.textContent = "...";
+  const imageResult = await imageLayer.setIndex(index);
+  const labelsResult = await labelsLayer.setIndex(index);
+  if (imageResult.success && labelsResult.success) {
+    tIndexEl!.textContent = `${index}`;
+  }
+}
```
aganders3, line 165: Currently the layers can (at least temporarily) get out of sync pretty easily when lazy-loading the slices. I don't have a great suggestion here because you don't want to wait for the `imageResult` to succeed before kicking off the next promise, and each layer handles setting its own data internally. Given this loading logic is temporary anyway, it may just be something we want to keep in mind for design improvements on the chunk manager.

aganders3 (approved):
This looks good and works as I'd expect in local testing. I think the cleanup and readability are nice. I agree this seems like it will also make it easier to replace the existing loading patterns with the chunk manager.

packages/core/examples/image_labels_overlay_with_value_picking/main.ts
andy-sweet: I went through the same thought and then discarded it. But [the OED is on your side](https://www.oed.com/dictionary/lookup_n?tab=factsheet#38829887), so I'm going to change it.

packages/core/examples/image_series_labels_overlay/main.ts
andy-sweet: Agreed. Like you mentioned, I don't think there's a straightforward way to fix this in the example.

Also agreed that the right solution feels like adapting the ChunkManager or augmenting it with something else to ensure that the data being presented is consistent with the desired coordinates across all layers (e.g. all layers can show some low resolution version of their data).

PR #348, opened 2025-07-30
feat: add point picking support for LabelImageLayer

phoenixAja:
This PR introduces a first-class point picking API for `LabelImageLayer`, enabling user interaction via click-based spatial selection.

### Features

- **New Layer API**
  - Adds `onPickValue?: (info: PointPickingResult) => void` to `LabelImageLayerProps`
  - Accepts `camera` and `clientToClip` injection for coordinate conversion
  - Implements local pointer event handling (click detection, drag suppression, etc.)

- **Interaction Flow**
  - Captures pointer down/up, filters by movement threshold
  - Computes world coordinates via: `client → clip → world`
  - Resolves value (stubbed for now) via `getValueAtWorld(world)`
  - Calls user-provided `onPickValue` callback if a value is present
  - Stops event propagation on successful pick

- **Example**
  - Adds `/examples/point-picking/` demo
  - Displays client/world/value/layer metadata in a side info box

- **Infrastructure**
  - Introduces shared `ClientToClip` type in `core/types`
  - Passes picking context entirely via constructor injection
  - Moves away from `setClientToClip()` pattern (eliminated)

https://github.com/user-attachments/assets/8227dce3-b415-44a2-ae88-bc80f00484d4


packages/core/examples/index.html
```diff
@@ -54,6 +54,12 @@ <h1>Examples</h1>
       <li>
         <a href="/image_mask_overlay/index.html">Image mask overlay</a>
       </li>
+      <li>
+        <a href="/point-picking/index.html">Point Picking</a>
+      </li>
+      <li>
+        <a href="/local_files/index.html">Local File System</a>
+      </li>
```
andy-sweet: This example was moved into the react package

```suggestion
```

packages/core/examples/point-picking/main.ts
```diff
@@ -0,0 +1,127 @@
...
+  Color,
+  ImageLayer,
+} from "@";
+import {
+  LabelImageLayer,
+  PointPickingResult,
+} from "@/layers/label_image_layer";
+import { PanZoomControls } from "@/objects/cameras/controls";
+
+// These roughly correspond in terms of content and the number of time-points.
```
andy-sweet: As this is largely similar to the `image_labels_overlay` example but with the added bonus of point picking, I think you should remove that example and keep this one. Puts slightly less maintenance burden on these examples, especially as they are main way of testing right now.

packages/core/src/core/types.ts
```diff
@@ -0,0 +1,3 @@
+import { vec2, vec3 } from "gl-matrix";
+
+export type ClientToClip = (clientPos: vec2, depth?: number) => vec3;
```
andy-sweet: I don't like general purpose modules like `types`, `constants`, `utils` because the name describes the types of things that are defined and exported there, rather than what those things mean.

I think a better home for this export would either be the existing `transforms.ts` or a new module named something like `coordinate_transforms` or `coordinate_spaces`.

packages/core/src/layers/label_image_layer.ts
```diff
@@ -100,4 +167,9 @@ export class LabelImageLayer extends Layer {
     image.transform.setTranslation([chunk.offset.x, chunk.offset.y, 0]);
     return image;
   }
+
+  public getValueAtWorld(world: vec3): vec3 /* TODO: label value */ | null {
+    // TODO: replace with actual sampling from renderable data (e.g. texture buffer)
+    return world; // stub - currently returns world coordinates instead of label value
```
andy-sweet, line 152: Showing the world coordinate instead of the value confused me in the demo. Given that the type here is `unknown | null`, I think you could put a `"TODO"` string or just `null` instead.

packages/core/src/layers/label_image_layer.ts
```diff
@@ -31,13 +44,21 @@ export class LabelImageLayer extends Layer {
...
+  private readonly dragThreshold_ = 3;
 
   constructor({
     source,
     region,
     colorCycle = DEFAULT_COLOR_CYCLE,
     colorMap = new Map(),
+    onPickValue,
+    camera,
+    clientToClip,
```
andy-sweet: Maybe I need to take another look at the spec, but passing through the camera and client through to the layer feels very unexpected and potentially fragile (because the layer can now mutate the camera). As the layers are the main API a user deals with, this also puts an onus on the user to pass these things through.

Instead, I wonder if we can transform the event before it gets to the layer (e.g. in `Idetik` which already has access to all these things).

A simple change could be to add optional properties to `EventContext` that store things like `clientCoords`, `clipCoords`, `worldCoords` that would be added when appropriate. Or only storing `clientPos` with the `clientToClip` transform so that it only needs to applied when needed.

A more complex change is to start at different types of `EventContext` to reflect special types of event (e.g. `PointerEventContext`), maybe along with other `Layer` methods (e.g. `onPointerEvent`), but it feels a little premature for that.

andy-sweet (changes requested):
Some minor comments, but my main blocking comment is around the need for passing camera and clientToClip through to the layer. Sorry the reasoning behind that is justified in the updated spec - I'll take another look.

packages/core/examples/point-picking/main.ts
```diff
@@ -0,0 +1,127 @@
...
+  layers: [imageLayer],
+  controls: new PanZoomControls(camera),
+});
+
+// Create label image layer with picking functionality
+const labelsLayer = new LabelImageLayer({
+  source: labelsSource,
+  region: labelsRegion,
+  transparent: true,
+  opacity: 0.5,
```
andy-sweet: Nit: maybe I need to calibrate my display, but I find toning this down a little makes the demo a little more pleasant.

```suggestion
  opacity: 0.25,
```

packages/core/examples/point-picking/main.ts
phoenixAja: yeah i was thinking about combining them, but i didn't know if we want to keep a strict separation of functionality between the demos. Happy to condense!

packages/core/src/layers/label_image_layer.ts
phoenixAja: mover the camera actions to the eventDispatcher, let me know what you think

packages/core/src/layers/label_image_layer.ts
phoenixAja: this will be resolved in the next PR when i implement this , so it's very temporary

packages/core/src/idetik.ts
```diff
@@ -75,6 +75,22 @@ export class Idetik {
...
     this.eventDispatcher_.addEventListener((event: EventContext) => {
+      if (event.event instanceof PointerEvent) {
+        const client = vec2.fromValues(
+          event.event.clientX,
+          event.event.clientY
+        );
+        Object.defineProperties(event, {
+          clipPos: {
+            get: () => this.clientToClip(client, 0),
+            configurable: true,
```
andy-sweet: I haven't used `Object.defineProperties` before. Is this configurable because it's overwriting the existing `EventContext.clipPos` property? In practice, I found I could set this false or equivalently omit it here without any issue.

packages/core/examples/image_labels_overlay_with_value_picking/main.ts
```diff
@@ -88,12 +107,21 @@ const labelsLayer = new LabelImageLayer({
...
 });
 
-const camera = new OrthographicCamera(0, xStopPoint, 0, yStopPoint);
-new Idetik({
-  canvasSelector: "canvas",
-  camera,
-  layers: [imageLayer, labelsLayer],
-  controls: new PanZoomControls(camera, camera.position),
-}).start();
+// Add the labels layer after creating the Idetik instance
```
andy-sweet: Is there a reason to add the layers after creation (i.e. does something not work if you add them on creation)?

packages/core/src/layers/label_image_layer.ts
```diff
@@ -70,8 +100,45 @@ export class LabelImageLayer extends Layer {
...
+        if (!this.pointerDownPos_) break;
+
+        const e = event.event as PointerEvent;
+        const pointerUpPos = vec2.fromValues(e.clientX, e.clientY);
+        const dist = vec2.distance(this.pointerDownPos_, pointerUpPos);
+
+        this.pointerDownPos_ = null;
+
+        if (dist < this.dragThreshold_) {
+          const client = event.clipPos;
```
andy-sweet: This should be `clip` or `clipPos` instead of `client`.

Thinking a bit more, while the clip coordinates are maybe useful for debugging purposes, I can't think of an obvious use for them right now (or for point value picking), so I might consider removing them from the `EventContext`.

packages/core/src/layers/label_image_layer.ts
andy-sweet: I definitely prefer it, but I have a few more comments and suggestions.

Could also be wise to wait for other feedback before doing anything to avoid churn.

packages/core/src/layers/label_image_layer.ts
```diff
@@ -31,13 +44,21 @@ export class LabelImageLayer extends Layer {
   private readonly lod_?: number;
   private readonly colorCycle_: ReadonlyArray<Color>;
   private readonly colorMap_: ReadonlyMap<number, Color>;
+  private readonly onPickValue_?: (info: PointPickingResult) => void;
+  private readonly camera_?: Camera;
+  private readonly clientToClip_?: ClientToClip;
```
andy-sweet: These can be removed now right?

```suggestion
```

packages/core/src/layers/label_image_layer.ts
```diff
@@ -6,12 +6,25 @@ import { PlaneGeometry } from "../objects/geometry/plane_geometry";
...
 import { EventContext } from "../core/event_dispatcher";
+import { vec2, vec3 } from "gl-matrix";
+import { Camera } from "../objects/cameras/camera";
+import { ClientToClip } from "../core/transforms";
+
+export interface PointPickingResult {
+  client: vec3;
+  world: vec3;
+  value: unknown | null;
+  layer: LabelImageLayer | null;
```
andy-sweet: I think you're including this in case the callback needs access to the layer because the callback is passed in on construction. Are you anticipating an important instance of this callback that needs a reference to the layer? If not, I might consider removing it in the spirit of keeping things simple and avoiding nullables/optionals.

packages/core/src/idetik.ts
```diff
@@ -75,6 +75,22 @@ export class Idetik {
 
     this.eventDispatcher_ = new EventDispatcher(canvas);
     this.eventDispatcher_.addEventListener((event: EventContext) => {
+      if (event.event instanceof PointerEvent) {
+        const client = vec2.fromValues(
+          event.event.clientX,
+          event.event.clientY
+        );
+        Object.defineProperties(event, {
```
andy-sweet, line 83: I think this approach is good enough for now. My other thought is to just set something like `EventContext.clipToWorld` and then define a `Event.getWorldPos` method that calls that (and maybe caches the result for other consumers). Hold off on implementing that until someone else actually thinks that a good idea...

packages/core/src/idetik.ts
```diff
@@ -75,6 +75,22 @@ export class Idetik {
...
+          clipPos: {
+            get: () => this.clientToClip(client, 0),
+            configurable: true,
+          },
+          worldPos: {
+            get: () => this.camera.clipToWorld(this.clientToClip(client, 0)),
```
andy-sweet, line 85: Not specific to this PR, but we should start to thinking about all the places we are executing this transform or similar (e.g. `ChunkManager.computeVisibleBounds` and think about some kind of cache to avoid burning the CPU too much in the Idetik animation loop.

andy-sweet (changes requested):
There are still some changes required for me before I can approve (e.g. removing the now unneeded `LabelImageLayer.camera_`), but otherwise I think I'm OK with the general approach and the other comments are minor/optional.

packages/core/examples/image_labels_overlay_with_value_picking/main.ts
phoenixAja: nope, i condensed it

packages/core/src/layers/label_image_layer.ts
phoenixAja: removed them. at first i thought they would be nice to display, but it is adding more code bloat which makes me feel like it's not actually worth it

packages/core/src/layers/label_image_layer.ts
phoenixAja: i'll remove it

packages/core/src/idetik.ts
phoenixAja: i'll remove configurable, we don't need it for the state of things at the moment anyways (and we probably won't need it in the future either). Thanks for the catch

packages/core/src/idetik.ts
phoenixAja: yeah i was trying to keep the design as lightweight as possible. But I can add that if people want. I would prefer to get this PR out the door sooner rather than later though

packages/core/examples/image_labels_overlay_with_value_picking/main.ts
```diff
@@ -88,12 +99,22 @@ const labelsLayer = new LabelImageLayer({
...
+      World: (${world[0].toFixed(1)}, ${world[1].toFixed(1)}, ${world[2].toFixed(1)})<br/>
+      Value: ${value ?? "null"}<br/>
+    `;
+  },
 });
 
-const camera = new OrthographicCamera(0, xStopPoint, 0, yStopPoint);
-new Idetik({
-  canvasSelector: "canvas",
+// Create Idetik instance
```
phoenixAja:
```suggestion
```

packages/core/src/core/event_dispatcher.ts
```diff
@@ -17,6 +18,8 @@ export class EventContext {
   readonly type: EventType;
   readonly event: Event;
   private propagationStopped_: boolean = false;
+  public clipPos?: vec3;
```
phoenixAja:
```suggestion
```

packages/core/examples/image_labels_overlay_with_value_picking/main.ts
```diff
@@ -88,12 +99,21 @@ const labelsLayer = new LabelImageLayer({
...
-const camera = new OrthographicCamera(0, xStopPoint, 0, yStopPoint);
-new Idetik({
-  canvasSelector: "canvas",
+const idetik = new Idetik({
+  canvas,
   camera,
   layers: [imageLayer, labelsLayer],
-  controls: new PanZoomControls(camera, camera.position),
-}).start();
+  controls: new PanZoomControls(camera),
```
andy-sweet: This is required to avoid a bug where sometimes panning will cause flickering, I think because the planes are going out of the camera's depth bounds.

```suggestion
  controls: new PanZoomControls(camera, camera.position),
```

I don't remember all the details of this bug, but think @aganders3 does. We should probably write an issue somewhere to capture this and potential improvements.

Maybe one very small improvement is to use `camera.position` as the target by default instead of `(0, 0, 0)`, so we don't have to pass both on every example that uses this.

phoenixAja: thanks!

packages/core/src/idetik.ts
andy-sweet: I'm fine with this for now.

Nit: this could be `Object.defineProperty` now.

packages/core/src/objects/cameras/controls.ts
```diff
@@ -1,7 +1,7 @@
 import { vec2, vec3 } from "gl-matrix";
 import { Camera } from "./camera";
+import { ClientToClip } from "../../core/transforms";
 
-type ClientToClip = (clientPos: vec2, depth: number) => vec3;
 type ClientToWorld = (clientPos: vec2, depth: number) => vec3;
```
andy-sweet: Sorry for the extra ask, but if we're moving `ClientToClip` to transforms then I think we should move this along with it for consistency.

andy-sweet (approved):
Looks good to me now. All other comments are optional.

PR #349, opened 2025-07-31
Support images with no series in <OmeZarrImageViewer>

bchu1:
- Adds `ImageLayer` support.
- This switches from using `layer.extent` to adjust the camera (which doesn't work with `ImageLayer` right now) to just grabbing it from the existing `loadAttributes()` call. It would've been nice to just use `extent` but I'm not sure if/where something like that should belong given the current chunk manager work (I'm not sure if chunk manager really cares what the extent is, only what needs to be loaded? And we were already fetching attributes anyways in `loadImageAttributes()` for other things).
- This also fixes a bug where I referenced `zRange` but forgot that state isn't updated immediately.

### DIFF
--- a/packages/react/examples/local-files/App.tsx
+++ b/packages/react/examples/local-files/App.tsx
@@ -3,7 +3,6 @@ import { OmeZarrImageViewer } from "../../src/components/viewers/OmeZarrImageVie
 import { useState } from "react";
 
 const region: Region = [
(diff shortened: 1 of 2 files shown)


aganders3 (approved):
This looks good. I'd like a more standard way to get the "extent" of a layer, especially before the initial load as this does. I think that can come at a later time though with some more consideration.

PR #350, opened 2025-08-03
Refactor camera controls

shlomnissan:
### Summary

This PR refactors the camera controls with the following changes:
- Rewrites the `PanZoomControls` logic for clarity and simplicity.
- Moves camera controls out of the renderer and into the runtime, improving separation of concerns.
- Replaces internal event handling with the centralized event dispatcher workflow.
- Removes unused functions and types for a leaner implementation.

### Notes

- @aganders3 I wasn’t entirely sure about the original rationale behind some of the design choices, so I took the liberty of rewriting the implementation in a way that seemed more straightforward to me. This refactor significantly reduced code complexity without breaking any functionality (I think?). I’d appreciate a close review to make sure I didn’t miss any edge cases or design constraints.
- FWIW zooming toward the cursor feels unnatural to me, probably because it's uncommon in typical 3D navigation controls. I generally expect zooming to be centered on the screen, which would also simplify some of this logic.

### Tests & Checks

- Manual verification
- All checks are passing

https://github.com/user-attachments/assets/117d70f2-1a9b-4840-ab6a-16aaf6087896

### DIFF
--- a/packages/core/examples/image2d_from_omezarr4d_hcs/main.ts
+++ b/packages/core/examples/image2d_from_omezarr4d_hcs/main.ts
@@ -34,11 +34,11 @@ wellPaths.forEach((path) => {
 });
 
 const camera = new OrthographicCamera(0, 840, 0, 360);
-const controls = new PanZoomControls(camera, camera.position);
+const controls = new PanZoomControls(camera);
 const app = new Idetik({
   canvas: document.querySelector<HTMLCanvasElement>("canvas")!,
   camera,
-  controls,
+  cameraControls: controls,
 }).start();
 
 const region: Region = [
@@ -84,7 +84,6 @@ const onImageChange = async () => {
     if (state === "ready" && newLayer.extent) {
       camera.setFrame(0, newLayer.extent.x, 0, newLayer.extent.y);
       camera.update();
-      controls.panTarget = camera.position;
     }
   });
 };
--- a/packages/core/examples/image2d_from_omezarr5d_u8/main.ts
+++ b/packages/core/examples/image2d_from_omezarr5d_u8/main.ts
@@ -1,4 +1,3 @@
-import { vec3 } from "gl-matrix";
 import {
   ChannelProps,
   Color,
@@ -6,7 +5,6 @@ import {
   ImageLayer,
   OmeZarrImageSource,
   OrthographicCamera,
-  PerspectiveCamera,
   Region,
 } from "@";
 import { AxesLayer } from "@/layers/axes_layer";
@@ -14,9 +12,7 @@ import { PanZoomControls } from "@/objects/cameras/controls";
 
 const url =
   "https://public.czbiohub.org/royerlab/ultrack/multi-color/image.zarr/";
-const orthoCam = new OrthographicCamera(-2000, 2000, -2000, 2000);
-const cameraPos = vec3.fromValues(0, 0, 5000);
-const perspectiveCam = new PerspectiveCamera({ fov: 60, position: cameraPos });
+const camera = new OrthographicCamera(-2000, 2000, -2000, 2000);
 
 // Source is technically 5D (even though Z is unitary),
 // so provide indices at 3 dimensions to project to 2D.
@@ -36,28 +32,11 @@ const channelProps: ChannelProps[] = [
 ];
 const layer = new ImageLayer({ source, region, channelProps });
 const axes = new AxesLayer({ length: 1920, width: 0.01 });
-const orthoCamControls = new PanZoomControls(orthoCam);
-const perspectiveCamControls = new PanZoomControls(perspectiveCam);
+const cameraControls = new PanZoomControls(camera);
 
-const app = new Idetik({
+new Idetik({
   canvas: document.querySelector<HTMLCanvasElement>("canvas")!,
-  camera: perspectiveCam,
-  cameraControls: perspectiveCamControls,
+  camera,
+  cameraControls,
   layers: [layer, axes],
 }).start();
-
-document.addEventListener("keydown", (event) => {
-  if (event.key === " ") {
-    toggleCamera();
-  }
-});
-
-function toggleCamera() {
-  if (app.camera === orthoCam) {
-    app.camera = orthoCam;
-    app.cameraControls = orthoCamControls;
-  } else {
-    app.camera = perspectiveCam;
-    app.cameraControls = perspectiveCamControls;
-  }
-}
--- a/packages/core/examples/image2d_from_omezarr5d_u8/main.ts
+++ b/packages/core/examples/image2d_from_omezarr5d_u8/main.ts
@@ -42,7 +42,7 @@ const perspectiveCamControls = new PanZoomControls(perspectiveCam);
 const app = new Idetik({
   canvas: document.querySelector<HTMLCanvasElement>("canvas")!,
   camera: perspectiveCam,
-  controls: perspectiveCamControls,
+  cameraControls: perspectiveCamControls,
   layers: [layer, axes],
 }).start();
 
@@ -53,8 +53,11 @@ document.addEventListener("keydown", (event) => {
 });
 
 function toggleCamera() {
-  app.camera = app.camera === orthoCam ? perspectiveCam : orthoCam;
-  app.setControls(
-    app.camera === orthoCam ? orthoCamControls : perspectiveCamControls
-  );
+  if (app.camera === orthoCam) {
+    app.camera = orthoCam;
+    app.cameraControls = orthoCamControls;
+  } else {
+    app.camera = perspectiveCam;
+    app.cameraControls = perspectiveCamControls;
+  }
 }
--- a/packages/core/examples/image_labels_overlay_with_value_picking/main.ts
+++ b/packages/core/examples/image_labels_overlay_with_value_picking/main.ts
@@ -113,7 +113,7 @@ const idetik = new Idetik({
   canvas,
   camera,
   layers: [imageLayer, labelsLayer],
-  controls: new PanZoomControls(camera, camera.position),
+  cameraControls: new PanZoomControls(camera),
 });
 
 idetik.start();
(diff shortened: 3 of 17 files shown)


packages/core/examples/image2d_from_omezarr5d_u8/main.ts
```diff
@@ -53,8 +53,11 @@ document.addEventListener("keydown", (event) => {
...
-  app.setControls(
-    app.camera === orthoCam ? orthoCamControls : perspectiveCamControls
-  );
+  if (app.camera === orthoCam) {
+    app.camera = orthoCam;
+    app.cameraControls = orthoCamControls;
+  } else {
+    app.camera = perspectiveCam;
+    app.cameraControls = perspectiveCamControls;
+  }
```
andy-sweet: Conditional logic was flipped when refactoring

```suggestion
  if (app.camera === orthoCam) {
    app.camera = perspectiveCam;
    app.cameraControls = perspectiveCamControls;
  } else {
    app.camera = orthoCam;
    app.cameraControls = orthoCamControls;
  }
```

packages/core/examples/image2d_from_omezarr5d_u8/main.ts
```diff
@@ -42,7 +42,7 @@ const perspectiveCamControls = new PanZoomControls(perspectiveCam);
 const app = new Idetik({
   canvas: document.querySelector<HTMLCanvasElement>("canvas")!,
   camera: perspectiveCam,
-  controls: perspectiveCamControls,
+  cameraControls: perspectiveCamControls,
```
andy-sweet: The perspective pan-zoom controls don't seem to support panning because trying to pan by clicking and dragging with my mouse has no effect. They zoom at `(0, 0, 0)` which is fine for the perspective controls for now, since they are not super useful for 2D-presentation anyway.

The ortho pan-zoom controls work fine.

packages/core/examples/image2d_from_omezarr4d_hcs/main.ts
```diff
@@ -34,11 +34,11 @@ wellPaths.forEach((path) => {
 });
 
 const camera = new OrthographicCamera(0, 840, 0, 360);
-const controls = new PanZoomControls(camera, camera.position);
```
andy-sweet, line 37: Love this simplification.

packages/core/examples/image_labels_overlay_with_value_picking/main.ts
```diff
@@ -113,7 +113,7 @@ const idetik = new Idetik({
   canvas,
   camera,
   layers: [imageLayer, labelsLayer],
-  controls: new PanZoomControls(camera, camera.position),
+  cameraControls: new PanZoomControls(camera),
```
andy-sweet, line 116: I found a bug in this example where after clicking on a segment, the camera starts panning. I cannot reliably get out of pan mode this, so it's quite severe.

packages/core/src/idetik.ts
```diff
@@ -75,22 +72,20 @@ export class Idetik {
...
-            get: () => this.camera.clipToWorld(this.clientToClip(client, 0)),
-          },
-        });
+      if (
+        event.event instanceof PointerEvent ||
+        event.event instanceof WheelEvent
+      ) {
+        const { clientX, clientY } = event.event;
+        const client = vec2.fromValues(clientX, clientY);
+        event.clipPos = this.clientToClip(client, 0);
```
andy-sweet, line 81: The number of overlapping/redundant matrix mults and inversions we do on every render loop iteration is starting to grow. Until we have standard profile/benchmark processes/tooling in place, it's hard to be confident about how important it is to optimize this.

But I'm generally of the opinion that for fast polling, we should avoid repeated computation when easy enough (which it seems to be in this case).

No need to do anything in this PR, but I would like to capture and maybe prioritize this work somehow.

packages/core/src/idetik.ts
```diff
@@ -102,16 +97,18 @@ export class Idetik {
...
-  public setControls(controls: CameraControls) {
-    this.renderer_.setControls(controls);
+  public set cameraControls(controls: CameraControls | undefined) {
+    this.cameraControls_ = controls;
   }
 
-  // TODO: this can be moved directly to this class, but will need access to (and possibly expose)
-  // the canvas element
-  // let's do it at the same time `setControls` is moved to this class
   public clientToClip(position: vec2, depth: number = 0): vec3 {
```
andy-sweet, line 104: Not directly related to this PR since `Idetik.clientToClip` is already public, but this feels a bit out of place here because this is the top-level runtime, where `clientToClip` feels like a rendering/coordinate-space detail. Does some external dependency need it? Could we at least consider making it private (maybe in another PR?).

andy-sweet (changes requested):
Generally looks good to me, with some nice simplifications - who doesn't love a big red number in the diff :)

But I found at least one (and maybe two) important enough bugs to request changes here.

packages/core/examples/image_labels_overlay_with_value_picking/main.ts
andy-sweet: A simple fix for me was to remove the call to `EventContext.stopPropagation` in `LabelImageLayer.onEvent`.

@phoenixAja should have a better idea than me if that is fine after this refactor.

packages/core/src/idetik.ts
aganders3: I'm pretty sure this is only public for the Try CryoLens project, but that predates the event manager and is registering its own mouse event listener directly on the canvas element. That should be ported to use the event manager + similar mechanism as the segmentation ID lookup (i.e. logic owned by a Particles layer).

packages/core/examples/image2d_from_omezarr5d_u8/main.ts
aganders3: I think perspective panning may also be weird without the clipDepth, at least when the image is not at z=0 it will have like a parallax effect where the pan does not feel like direct manipulation. I think often this is solved by picking, so you pan relative to the object you click on. I'd be fine with restricting the panzoom controls to the orthographic camera though. I think perspective cameras are useful in 3D but we'll probably want arcball or orbit controls then.

aganders3:
Overall this looks good! I'll take a closer look after you respond to @andy-sweet's comments, but I don't expect any issues getting this in quickly.

shlomnissan:
Thanks for the reviews @andy-sweet. I'm going to focus on prefetching today, but I will address your comments first thing tomorrow.

packages/core/examples/image2d_from_omezarr5d_u8/main.ts
shlomnissan: Ah, that explains some of the complexity. I wasn’t sure why `clipDepth` was needed.. I thought I had tested with a perspective camera, but clearly not thoroughly. For now, I think I’ll remove perspective controls and make the “pan zoom” controls expect an orthographic camera. I can easily add an arcball-style implementation later when needed.

Something else I’ve been thinking about is generalizing controls. For example, what if I want to keep pan, but have the scroll wheel translate along Z instead of zooming like in Neuroglancer? Would that require a separate control class, or is there a more flexible abstraction we can use? Maybe the traditional “camera controls” model isn’t the right long-term approach.

packages/core/src/idetik.ts
shlomnissan: I think we can start tackling this once we begin proper benchmarking, which should happen soon likely as part of the prefetching work.

How would you suggest limiting the repeated computation in this case? As I mentioned in my summary, my preference would be removing this altogether and simply zooming from the center.

packages/core/src/idetik.ts
shlomnissan: I totally agree!

@aganders3 Are you planning to prioritize this work? If not, can you add a ticket that captures the necessary changes to CryoLens?

packages/core/examples/image_labels_overlay_with_value_picking/main.ts
shlomnissan: Good catch! Removing `event.stopPropagation();` from `LabelImageLayer::onEvent` seems to maintain the behavior so I added it to this PR, but I will wait for @phoenixAja confirmation before merging.

shlomnissan:
@andy-sweet I requested another review.

@phoenixAja waiting for you to unblock.

packages/core/src/idetik.ts
aganders3: Yeah I'll prioritize it. I'd like for the Try CryoLens app to be built against `main` of Idetik when it ships (maybe late this month?). I'll open some tickets here and on [the try-cryolens repo](https://github.com/chanzuckerberg/try-cryolens) to track the work required.

packages/core/src/idetik.ts
andy-sweet: Outside of caching the matrix ops (especially the inversions), I haven't thought about other solutions. In general, I like to avoid special-case caching unless its easy to make guarantees about correctness or if there are benchmark/profiling results to motivate it.

> As I mentioned in my summary, my preference would be removing this altogether and simply zooming from the center.

I personally like zoom from the cursor position, and if we want to support touch controls (which we probably do at some point), I'm assuming we'd need to similar calculations.

We're also doing similar calculations elsewhere, so I don't think this is a long term solution.

andy-sweet (approved):
Looks good to me now. Approval is conditioned on finding some way to fix the picking example bug.

shlomnissan:
> Approval is conditioned on finding some way to fix the picking example bug.

@andy-sweet To clarify, my last commit included your fix which seems to work without impacting the original behavior, so this is conditioned on @phoenixAja confirmation that stopping propagation doesn't impact the intended behavior.

packages/core/examples/image2d_from_omezarr4d_hcs/main.ts
phoenixAja: agreed

phoenixAja (approved):
point picking example still working as expected for me!

PR #351, opened 2025-08-04
(Small) Fix incorrect visible bounds comparison in ChunkManager

shlomnissan:
### Summary

This PR resolves a bug where the `ChunkManager` always detected changes in
visible bounds, even when none occurred. The issue was caused by comparing
deep-copied `vec2` objects by reference instead of by value.

### Tests & Checks
- Manual verification
- All checks have passed


andy-sweet (approved):
Good catch and a small nice improvement in readability.

PR #352, opened 2025-08-05
OME-Zarr v0.5 support prototype

### LINKED ISSUE #304: Support loading of OME-Zarr v0.5 image data
Currently, Idetik supports loading data from OME-Zarr sources of version 0.4.

The goal of this task is to support loading data from [OME-Zarr version 0.5](https://ngff.openmicroscopy.org/0.5/index.html). The biggest change from v0.4 to v0.5 is that v0.5 assumes zarr v3, but the [change list](https://ngff.openmicroscopy.org/0.5/index.html#history) should be inspected for other things to consider.

This task could be as simple as [re-running zod](https://github.com/chanzuckerberg/idetik/blob/97ad312e90e24170b418a69ef5d147539f3d2497/packages/core/src/data/ome_ngff/0.4/README.md) to generate the types for the v0.5 schema and then updating the zarrita loading logic to be more flexible about the zarr version type.

However, we may also need to adapt some of our handling of metadata/attributes in case those have changed too.

The following task list represents what I think we need to do to close this issue based on exploratory work in #352.

andy-sweet:
### Summary
This is a prototype branch that implements support for OME-Zarr v0.5.

It is intended to be a reference for smaller PRs and is not intended to be merged directly.

### Related Issue
Closes #304 

### Tests & Checks
Ran the existing examples and the new example.


andy-sweet:
Closing, as almost all the necessary PRs related to this are merged or are in review.

PR #357, opened 2025-08-05
Fix chunk visibility bounds check

### LINKED ISSUE #355: Fix chunk manager chunk offsets and visibility handling

andy-sweet:
### Summary
This fixes the visibility bounds check in the chunk manager, as well as a bug in how we handle translations in OME-Zarr metadata and the ultimate conversion to integer indices for an array at some multiscale level.

### Related Issue
Closes #355 

### Tests & Checks
I tested the existing examples to ensure they continue to work as expected.
I modified one of the existing examples to use OME-Zarr data that includes a translation to test that this fix also works when the underlying data contains an offset.

### DIFF
--- a/packages/core/examples/image2d_from_omezarr5d_u8/main.ts
+++ b/packages/core/examples/image2d_from_omezarr5d_u8/main.ts
@@ -10,7 +10,8 @@ import {
 import { AxesLayer } from "@/layers/axes_layer";
 import { PanZoomControls } from "@/objects/cameras/controls";
 
-const url = "https://ome-zarr-scivis.s3.us-east-1.amazonaws.com/v0.4/96x0/marmoset_neurons.ome.zarr";
+const url =
+  "https://ome-zarr-scivis.s3.us-east-1.amazonaws.com/v0.4/96x0/marmoset_neurons.ome.zarr";
 const source = new OmeZarrImageSource(url);
 const loader = await source.open();
 const attributes = await loader.loadAttributes();
@@ -45,12 +46,15 @@ const channelProps: ChannelProps[] = [
   },
 ];
 const layer = new ImageLayer({ source, region, channelProps });
-const axes = new AxesLayer({ length: 0.75 * xInfo.scale * xInfo.size, width: 0.01 });
+const axes = new AxesLayer({
+  length: 0.75 * xInfo.scale * xInfo.size,
+  width: 0.01,
+});
 const camera = new OrthographicCamera(
   xInfo.offset,
   xInfo.offset + xInfo.scale * xInfo.size,
   yInfo.offset,
-  yInfo.offset + yInfo.scale * yInfo.size,
+  yInfo.offset + yInfo.scale * yInfo.size
 );
 
 new Idetik({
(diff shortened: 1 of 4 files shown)


andy-sweet:
@phoenixAja : assigned you for review because you wrote the chunk manager code.
@shlomnissan : assigned you because i think you ran into the same bug this fixes.

phoenixAja (approved):

andy-sweet:
Interesting aside is that the new data in the modified example highlights a maybe-bug, which you can see when you enable wireframes (the chunks are not as well aligned as I would expect).

I think this is because of a well known image visualization dichotomy of deciding where the origin should be placed relative to the top-left pixel of otherwise untransformed image. The visualization default tends to be the top-left corner of that pixel (in the typical 2D image presentation where the y-dimension goes down the screen), whereas the image processing default tends to be the center. I think that's what's happening here because the issue gets worse with different multiscale levels.

I'll write up an issue to describe this in more detail. I feel like we might need a fix that allows this to be configured per source because this is such an ongoing feud.

aganders3:
Nice spot @andy-sweet!
I think your diagnosis is correct. I tried a quick fix to confirm, and at least for that marmoset neurons example this better aligns the ~~chunks~~ tiles:
```diff
diff --git a/packages/core/src/layers/image_layer.ts b/packages/core/src/layers/image_layer.ts
index 0a6ab29..6abe2c8 100644
--- a/packages/core/src/layers/image_layer.ts
+++ b/packages/core/src/layers/image_layer.ts
@@ -205,7 +205,11 @@ export class ImageLayer extends Layer implements ChannelsEnabled {
     }
 
     image.transform.setScale([chunk.scale.x, chunk.scale.y, 1]);
-    image.transform.setTranslation([chunk.offset.x, chunk.offset.y, 0]);
+    image.transform.setTranslation([
+      chunk.offset.x - chunk.scale.x / 2,
+      chunk.offset.y - chunk.scale.y / 2,
+      0,
+    ]);
     return image;
   }
 }
```

> I'll write up an issue to describe this in more detail. I feel like we might need a fix that allows this to be configured per source because this is such an ongoing feud.

When you put it this way I picture setting `imageSource.alignment = "hatfield"` or `imageSource.alignment = "mccoy"`.

shlomnissan:
> The visualization default tends to be the top-left corner of that pixel (in the typical 2D image presentation where the y-dimension goes down the screen), whereas the image processing default tends to be the center.

@andy-sweet Are you referring to sub-pixel positioning, like whether the origin of a pixel is at (0.0, 0.0) vs (0.5, 0.5)? Or by “origin” do you mean the position of the entire image in space?

shlomnissan (approved):
LGTM!

andy-sweet:
> @andy-sweet Are you referring to sub-pixel positioning, like whether the origin of a pixel is at (0.0, 0.0) vs (0.5, 0.5)? Or by “origin” do you mean the position of the entire image in space?

Yes to both, without getting too into the weeds (the phrase "origin of a pixel" could start some kind of war!), the latter assuming that the image has no transforms associated with it.

The following simple ITK docs + diagram describe things better than I can: https://simpleitk.readthedocs.io/en/master/fundamentalConcepts.html#images

andy-sweet:
I created #369 to track the rendering offset issue.

PR #359, opened 2025-08-05
Add prefetching on X/Y axes

shlomnissan:
### Summary

This PR introduces prefetching along the X and Y axes by checking for
intersections with padded visible bounds, allowing the system to prefetch
chunk data that isn't currently visible but is likely to be soon. It also refactors
the chunk info overlay to include details about prefetched chunks. The approach
is designed to scale to the Z axis once chunk manager support is added.

### Tests & Checks
- Manual verification, added "prefetched" information to the chunk overlay
- All checks have passed

### DIFF
--- a/packages/core/src/core/chunk_manager.ts
+++ b/packages/core/src/core/chunk_manager.ts
@@ -130,7 +130,7 @@ export class ChunkManagerSource {
     if (this.visibleBoundsChanged(visibleBounds)) {
       this.updateChunkVisibility(visibleBounds);
     }
-    this.loadVisibleChunks();
+    this.loadPendingChunks();
   }
 
   public get lodCount() {
@@ -145,7 +145,7 @@ export class ChunkManagerSource {
     return this.currentLOD_;
   }
 
-  private loadVisibleChunks() {
+  private loadPendingChunks() {
     this.loadLowResChunks();
 
     for (const chunk of this.chunks_) {
@@ -154,7 +154,7 @@ export class ChunkManagerSource {
         chunk.state === "unloaded" &&
         (chunk.visible || chunk.prefetch)
       ) {
-        this.processChunkData(chunk);
+        this.loadChunkData(chunk);
       }
     }
   }
@@ -162,12 +162,12 @@ export class ChunkManagerSource {
   private loadLowResChunks(): void {
     for (const chunk of this.chunks_) {
       if (chunk.lod === this.lowestResLOD_ && chunk.state === "unloaded") {
-        this.processChunkData(chunk);
+        this.loadChunkData(chunk);
       }
     }
   }
 
-  private processChunkData(chunk: ImageChunk): void {
+  private loadChunkData(chunk: ImageChunk): void {
     chunk.state = "loading";
     this.loader_
       .loadChunkDataFromRegion(chunk, this.region_)
@@ -177,7 +177,7 @@ export class ChunkManagerSource {
       .catch((error) => {
         Logger.error(
           "ChunkManager",
-          `Error loading chunk (${chunk.chunkIndex?.x},${chunk.chunkIndex?.y}): ${error}`
+          `Error loading chunk (${chunk.chunkIndex.x},${chunk.chunkIndex.y}): ${error}`
         );
         chunk.state = "unloaded";
       });
--- a/packages/core/src/core/chunk_manager.ts
+++ b/packages/core/src/core/chunk_manager.ts
@@ -12,6 +12,10 @@ import { Logger } from "../utilities/logger";
 
 type Bounds = { min: vec2; max: vec2 };
 
+// Number of chunks to extend beyond the visible bounds in each direction (x/y/z)
+// These additional chunks are prefetched to improve responsiveness when panning.
+const PREFETCH_PADDING_CHUNKS = 1;
+
 export class ChunkManagerSource {
   private readonly chunks_: ImageChunk[];
   private readonly loader_;
@@ -75,6 +79,7 @@ export class ChunkManagerSource {
             state: "unloaded",
             lod,
             visible: false,
+            prefetch: false,
             shape: {
               x: chunkWidth,
               y: chunkHeight,
@@ -120,7 +125,19 @@ export class ChunkManagerSource {
     return [...lowResChunks, ...currentLODChunks];
   }
 
-  public get allChunks(): ImageChunk[] {
+  public update(lodFactor: number, visibleBounds: Bounds) {
+    this.setLOD(lodFactor);
+    if (this.visibleBoundsChanged(visibleBounds)) {
+      this.updateChunkVisibility(visibleBounds);
+    }
+    this.loadVisibleChunks();
+  }
+
+  public get lodCount() {
+    return this.lowestResLOD_ + 1;
+  }
+
+  public get chunks(): ImageChunk[] {
     return this.chunks_;
   }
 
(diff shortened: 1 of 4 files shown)


packages/core/src/core/chunk_manager.ts
```diff
@@ -206,28 +212,15 @@ export class ChunkManagerSource {
         continue;
       }
 
-      const chunkVirtualWidth = chunk.shape.x * chunk.scale.x;
-      const chunkVirtualHeight = chunk.shape.y * chunk.scale.y;
+      // reset state
+      chunk.visible = false;
+      chunk.prefetched = false;
```
andy-sweet: Nit: I'm guessing that if `!chunk.chunkIndex` implies that `!chunk.prefetched`, so I wonder if we could just always do the reset at the beginning of the for loop body?

andy-sweet (approved):
Looks good to me.

The only other thought I had here is if the padding should instead be a percentage of the field of view. That better matches the idea of prefetching these chunks to anticipate small pans because "small" is relative to the current field of view. Whereas pre-fetching a fixed number of chunks is more dependent on the chunking of the underlying source. In practice, I don't think it will make much difference, so this is fine for now.

packages/core/src/core/chunk_manager.ts
aganders3: I could be missing something, but when is `chunk.chunkIndex` falsy anyway? It looks to me like we could get rid of the conditional above.

Likewise there's an error log that uses
```
`Error loading chunk (${chunk.chunkIndex?.x},${chunk.chunkIndex?.y}): ${error}`
```
but I don't think we need the question marks in it.

packages/core/src/core/chunk_manager.ts
```diff
@@ -206,28 +212,15 @@ export class ChunkManagerSource {
...
-      chunk.visible =
-        x >= minChunkIndexX &&
-        x <= maxChunkIndexX &&
-        y >= minChunkIndexY &&
-        y <= maxChunkIndexY;
+      chunk.visible = this.isChunkWithinBounds(chunk, visibleBounds);
+      if (!chunk.visible) {
+        const paddedBounds = this.getPaddedBounds(visibleBounds);
+        chunk.prefetched = this.isChunkWithinBounds(chunk, paddedBounds);
+      }
```
aganders3: Calculation of `paddedBounds` can be moved outside the loop.

packages/core/src/core/chunk_manager.ts
```diff
@@ -247,6 +240,25 @@ export class ChunkManagerSource {
...
+    const maxY = minY + chunk.shape.y * chunk.scale.y;
+
+    return (
+      minX <= boundsMaxX &&
+      maxX >= boundsMinX &&
+      minY <= boundsMaxY &&
+      maxY >= boundsMinY
+    );
+  }
+
```
aganders3, line 255: To reduce double-checking of chunks outside of `visibleBounds` what do you think about replacing `isChunkWithinBounds` with a signed-distance calculation, like this:

```typescript
  private updateChunkVisibility(visibleBounds: Bounds): void {
    if (this.chunks_.length === 0) {
      Logger.warn(
        "ChunkManager",
        "updateChunkVisibility called with no chunks initialized"
      );
      return;
    }

    const prefetchDistance = this.getPrefetchDistance();

      const distance = this.getChunkDistanceFromBounds(chunk, visibleBounds);

      chunk.visible = distance <= 0;
      chunk.prefetched = !chunk.visible && distance <= prefetchDistance;
    }
  }

  private getChunkDistanceFromBounds(
    chunk: ImageChunk,
    bounds: Bounds
  ): number {
    const boundsMinX = bounds.min[0];
    const boundsMaxX = bounds.max[0];
    const boundsMinY = bounds.min[1];
    const boundsMaxY = bounds.max[1];

    const chunkMinX = chunk.offset.x;
    const chunkMaxX = chunkMinX + chunk.shape.x * chunk.scale.x;
    const chunkMinY = chunk.offset.y;
    const chunkMaxY = chunkMinY + chunk.shape.y * chunk.scale.y;

    // Distance to bounds (negative = overlapping, positive = gap distance)
    const xDistance = Math.max(boundsMinX - chunkMaxX, chunkMinX - boundsMaxX);
    const yDistance = Math.max(boundsMinY - chunkMaxY, chunkMinY - boundsMaxY);

    return Math.max(xDistance, yDistance);
  }

  private getPrefetchDistance(): number {
    const chunkWidth =
      this.attrs_[this.currentLOD_].chunks[this.xIdx_] *
      this.attrs_[this.currentLOD_].scale[this.xIdx_];
    const chunkHeight =
      this.attrs_[this.currentLOD_].chunks[this.yIdx_] *
      this.attrs_[this.currentLOD_].scale[this.yIdx_];
    return Math.max(chunkWidth, chunkHeight) * PREFETCH_PADDING_CHUNKS;
  }

```

packages/core/src/core/chunk_manager.ts
```diff
@@ -130,11 +147,10 @@ export class ChunkManagerSource {
     this.loadLowResChunks();
 
     for (const chunk of this.chunks_) {
-      // Only load chunks for current LOD
       if (
         chunk.lod === this.currentLOD_ &&
         chunk.state === "unloaded" &&
-        chunk.visible
+        (chunk.visible || chunk.prefetched)
```
aganders3: With this change (also given this calls `loadLowResChunks`) maybe this fn should be renamed from `loadVisibleChunks`. I don't have a better name right now...maybe something like `loadNecessaryChunkData`.

packages/core/src/data/image_chunk.ts
```diff
@@ -36,6 +36,7 @@ export type ImageChunk = {
   state: "unloaded" | "loading" | "loaded";
   lod: number;
   visible: boolean;
+  prefetched: boolean;
```
aganders3: nit: I'd probably use `prefetch` here instead of the past-tense, since this does not indicate the chunk is actually loaded.

packages/core/src/core/chunk_manager.ts
aganders3: This collapses the distance to a scalar, but doesn't have to. We could make the margin distance 4d to include tzyx.

aganders3 (approved):
This looks good to me! I have some mostly nitpick comments and ideas, but happy to see this go in as-is since it works and we can iterate on it when expanding to other dimensions.

packages/core/src/core/chunk_manager.ts
shlomnissan: Ahh good catch! I didn't dig too dip into the existing implementation. I just assumed it's correct and necessary.

packages/core/src/core/chunk_manager.ts
shlomnissan: Great catch! To improve clarity, I will rename:
- `loadVisibleChunks` to `loadPendingChunks`
- `processChunkData` to `loadChunkData`

I like the idea of "pending chunks" or "scheduled chunks". It's not great, but I can't think of a better description at this time.

shlomnissan: This thread was triggered by the comment
https://czi-sci.slack.com/archives/C07GHGSJSQP/p1754498905130989

packages/core/src/core/chunk_manager.ts
shlomnissan: I think it's a good idea. I might explore it in another PR, once I have XYZ prefetching. I'll create an issue for it

shlomnissan: https://github.com/chanzuckerberg/idetik/issues/364

PR #362, opened 2025-08-06
Make label layers' colormap mutable

### LINKED ISSUE #342: Dense segmentation overlay rendering
This is possibly a duplicate of #177 but keeping for tracking purposes.

andy-sweet:
### Summary
This makes the two label layer's colormap gettable and settable. It also handles the necessary texture updates when that is set.

In order to dispose of old textures that will never be used again, it adds a new core functionality to renderable objects to mark some textures as stale when they are replaced. The label renderable uses this core functionality to replace the color map textures. The WebGL renderer then disposes of any stale textures in the core rendering loop.

Rather than add new examples, I modified the existing examples that use label image layers to allow modification of their colormap with some basic interactions.
- Label picking on a 2D image: click a label to override it to be white.
- Label series image: click on a button to choose a different color cycle.

### Related Issue
Closes #342

### Tests & Checks
I manually tested the modified examples.

### DIFF
--- a/packages/core/examples/image_labels_overlay_with_value_picking/main.ts
+++ b/packages/core/examples/image_labels_overlay_with_value_picking/main.ts
@@ -98,7 +98,7 @@ const colorMaps = [
   {
     lookupTable: new Map([[103, Color.YELLOW]]),
     cycle: [Color.RED, Color.GREEN, Color.BLUE],
-  }
+  },
 ];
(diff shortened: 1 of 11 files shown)


andy-sweet:
@bchu1 : added you as a review, because I think this should be enough for the CZII HITL segmentation application. Other reviews and comments are welcome.

bchu1 (approved):
🎉

PR #363, opened 2025-08-06
[small] Add support for color + layer transparency in Points renderable

aganders3:
### Summary
This adds basic support for transparency in the `Points` renderable object. The shader also includes `u_opacity`, which corresponds to a layer-level opacity value. This also allows for alpha-based anti-aliasing depending on how the marker atlas is drawn or sampled. For any of this to work, a layer using this will need to set its `transparent` attribute to `true`.

### Related Issue
N/A

### Tests & Checks
Tested primarily in the Try CryoLens app, but the points example here still works.

### DIFF
--- a/packages/core/src/objects/renderable/points.ts
+++ b/packages/core/src/objects/renderable/points.ts
@@ -37,6 +37,7 @@ export class Points extends RenderableObject {
       point.color.r,
       point.color.g,
       point.color.b,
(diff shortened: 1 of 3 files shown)


aganders3:
Assigning @phoenixAja because you worked on the initial layer blending and transparency

shlomnissan (approved):
👍

PR #365, opened 2025-08-06
Update tools and docs for generating zod OME-Zarr classes

### LINKED ISSUE #356: Dereference JSON refs using @apidevtools/json-schema-ref-parser instead of json-refs

andy-sweet:
### Summary
This refreshes our docs and tooling around how we automatically generate zod classes, which can be used to validate JSON, from the OME-Zarr JSON schemata. The main motivation for this change is to support the automatic generation for OME-Zarr v0.5 where the current tooling was found not to work (see #352 for more context).

- Used `@apidevtools/json-schema-ref-parser` instead of `json-refs` which more completely handles dereferencing, including from URLs which allows us to avoid a git clone.
- Added the required node packages as dev-dependencies.
- Removed OME-Zarr classes that we don't currently use.
- Added a portable shell script that can be re-run with future OME-Zarr versions and schemata.
- Updated the README accordingly.

### Related Issue
Closes #356

### Tests & Checks
I ran the following commands from the top of the repo, which produced the same zod class definitions as we have on main.

```shell
packages/core/src/data/ome_ngff/make_zod_types.sh 0.4 image plate well
npm run format:all
```


andy-sweet:
@phoenixAja : assigned you for review because github said you worked on these recently. Feel free to pass it to someone else!

packages/core/src/data/ome_ngff/resolve_schema.cjs
```diff
@@ -0,0 +1,21 @@
+const $RefParser = require("@apidevtools/json-schema-ref-parser");
```
andy-sweet: Maybe this should be MJS instead of CJS, in which case it might be more readable and this PR diff might become smaller?

phoenixAja (approved):
works for me!

PR #370, opened 2025-08-06
feat: Implement getValueAtWorld method for LabelImageLayer value picking

phoenixAja:
Summary

  Implements the `getValueAtWorld` method in `LabelImageLayer` to enable pixel value picking on segmentation layers. This allows users to click on label images and retrieve the underlying label ID values.

This approach will need to be adjusted if we end up enabling progressive loading for the `LabelImageLayer`

  Changes Made

  - LabelImageLayer: Implemented `getValueAtWorld` method that was previously a stub
  - Coordinate transformation: Converts world coordinates to texture coordinates using the image transform's inverse matrix
  - Value sampling: Samples label values directly from the original chunk data for accuracy
  - Data storage: Added imageChunk_ property to store raw chunk data alongside the rendered texture
  - Example update: Enhanced the value picking example to display "Label Value" instead of generic "Value"

  Testing

  - Updated existing point picking example demonstrates the functionality
  - Clicking on segmented regions now displays the correct label ID values
  - Proper handling of out-of-bounds clicks and missing data

  This enables interactive exploration of segmentation data, allowing users to identify specific labeled regions by clicking on them.

https://github.com/user-attachments/assets/9ba14801-6964-46c5-ab4a-ba8c750f279a

### DIFF
--- a/packages/core/src/layers/label_image_layer.ts
+++ b/packages/core/src/layers/label_image_layer.ts
@@ -154,7 +154,7 @@ export class LabelImageLayer extends Layer {
       return null;
     }
 
-    const pixelIndex = y * this.imageChunk_.shape.x + x;
+    const pixelIndex = y * this.imageChunk_.rowStride + x;
     const data = this.imageChunk_.data;
     return data[pixelIndex];
   }
--- a/packages/core/src/layers/label_image_layer.ts
+++ b/packages/core/src/layers/label_image_layer.ts
@@ -154,17 +154,8 @@ export class LabelImageLayer extends Layer {
       return null;
     }
 
-    // Sample pixel value from chunk data
     const pixelIndex = y * this.imageChunk_.shape.x + x;
     const data = this.imageChunk_.data;
-
-    if (
-      data instanceof Uint8Array ||
-      data instanceof Uint16Array ||
-      data instanceof Uint32Array
-    ) {
-      return data[pixelIndex];
-    }
-    return null;
+    return data[pixelIndex];
   }
 }
--- a/packages/core/src/layers/label_image_layer.ts
+++ b/packages/core/src/layers/label_image_layer.ts
@@ -29,6 +29,7 @@ export class LabelImageLayer extends Layer {
   private readonly colorMap_: LabelColorMap;
   private readonly onPickValue_?: (info: PointPickingResult) => void;
   private image_?: LabelImageRenderable;
+  private imageChunk_?: ImageChunk;
   private pointerDownPos_: vec2 | null = null;
   private readonly dragThreshold_ = 3;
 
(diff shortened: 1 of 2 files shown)


packages/core/src/layers/label_image_layer.ts
```diff
@@ -128,8 +130,41 @@ export class LabelImageLayer extends Layer {
...
+      x < 0 ||
+      x >= this.imageChunk_.shape.x ||
+      y < 0 ||
+      y >= this.imageChunk_.shape.y
+    ) {
+      return null;
+    }
+
+    // Sample pixel value from chunk data
+    const pixelIndex = y * this.imageChunk_.shape.x + x;
```
andy-sweet: Nit: if the data comes from OME-Zarr, I think it always gets compacted (i.e. the row-stride is the same as the width), but this should work for sources that don't do that (there aren't many...).

```suggestion
    const pixelIndex = y * this.imageChunk_.rowStride + x;
```

packages/core/src/layers/label_image_layer.ts
```diff
@@ -128,8 +130,41 @@ export class LabelImageLayer extends Layer {
...
+    const data = this.imageChunk_.data;
+
+    if (
+      data instanceof Uint8Array ||
+      data instanceof Uint16Array ||
+      data instanceof Uint32Array
+    ) {
```
andy-sweet: Why do we need to check this is unsigned here?

I think non-unsigned ints won't work with label image layers for other reasons (i.e. the fragment shader doesn't support them yet), but I feel like this doesn't need to be here given this returns a number.

packages/core/src/layers/label_image_layer.ts
```diff
@@ -117,6 +118,7 @@ export class LabelImageLayer extends Layer {
   }
 
   private createImage(chunk: ImageChunk) {
+    this.imageChunk_ = chunk; // Store chunk for value picking
```
andy-sweet, line 121: I think you could get the equivalent information from `this.image.textures[0]`, but I think I prefer this.

andy-sweet (approved):
Demo works well for me and no important/blocking feedback, so approved.

Once the label image layer uses the chunk manager this will get a little more complicated, but not too much more. But we can worry about that later.

phoenixAja:
> Once the label image layer uses the chunk manager this will get a little more complicated, but not too much more. But we can worry about that later.

Yes, this solution will need to be adjusted once we integrated the chunk manager, but i didn't want to prematurely optimize.

packages/core/src/layers/label_image_layer.ts
phoenixAja: removed

PR #371, opened 2025-08-07
Define non-null point picking result value type

andy-sweet:
### Summary
I think this was missed in #370. If point picking results are returned by other layers they may need to relax the interface or make it generic to handle arrays of numbers for multi-channel images.

### Tests & Checks
I manually tested the point picking example.

### DIFF
--- a/packages/core/src/layers/label_image_layer.ts
+++ b/packages/core/src/layers/label_image_layer.ts
@@ -13,7 +13,7 @@ import { vec2, vec3 } from "gl-matrix";
 
 export interface PointPickingResult {
   world: vec3;
-  value: number | null;
+  value: number;
 }
 
 export type LabelImageLayerProps = LayerOptions & {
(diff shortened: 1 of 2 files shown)


packages/core/src/layers/label_image_layer.ts
```diff
@@ -13,7 +13,7 @@ import { vec2, vec3 } from "gl-matrix";
 
 export interface PointPickingResult {
   world: vec3;
-  value: unknown | null;
+  value: number;
```
andy-sweet, line 16: Right now, `LabelImageLayer.onEvent` protects against returning `null`, so I removed null here in the spirit of avoiding null when it's not needed.

Alternatively, I think it would be fine to include null here, but not guard it in `onEvent`.

But doing both seems to complicate the implementation, while also requiring the caller to handle the possibility of null anyway even though it's not possible yet.

andy-sweet:
@phoenixAja: I assigned you because I think this may have been missed in #370 

I'm not totally sure of the intended semantics here, so please reject or close if these changes are not consistent with what you had in mind.

shlomnissan (approved):

phoenixAja:
thanks for patching

PR #373, opened 2025-08-07
Rename ImageChunk to Chunk

shlomnissan:
### Summary

This PR renames `ImageChunk` to `Chunk`. In upcoming work, a chunk will hold
3D data, which will be sliced at the layer (or image renderable) level. This
rename reflects that change and aligns with the terminology we agreed on.

### Tests & Checks
- All checks have passed
- All examples run as expected

### DIFF
--- a/packages/core/src/objects/textures/texture_2d.ts
+++ b/packages/core/src/objects/textures/texture_2d.ts
@@ -3,7 +3,7 @@ import {
   Texture,
   bufferToDataType,
 } from "../../objects/textures/texture";
-import { ImageChunk } from "../../data/image_chunk";
+import { Chunk } from "../../data/chunk";
 
 export class Texture2D extends Texture {
   private data_: DataTextureTypedArray;
@@ -41,7 +41,7 @@ export class Texture2D extends Texture {
     return this.height_;
   }
 
-  public static createWithImageChunk(chunk: ImageChunk) {
+  public static createWithChunk(chunk: Chunk) {
     if (!chunk.data) {
       throw new Error(
         "Unabled to create texture, chunk data is not initialized."
--- a/packages/core/examples/chunk_streaming/chunk_info_overlay.ts
+++ b/packages/core/examples/chunk_streaming/chunk_info_overlay.ts
@@ -1,6 +1,6 @@
 import { Idetik } from "../../src/idetik";
 import { ImageLayer } from "../../src/layers/image_layer";
-import { ImageChunk } from "../../src/data/image_chunk";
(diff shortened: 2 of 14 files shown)


packages/core/src/objects/textures/texture_2d.ts
```diff
@@ -41,7 +41,7 @@ export class Texture2D extends Texture {
     return this.height_;
   }
 
-  public static createWithImageChunk(chunk: ImageChunk) {
+  public static createWithChunk(chunk: Chunk) {
```
andy-sweet, line 44: As these are 2D textures, I guess they may throw if the chunk is not actually a 2D slab. Or maybe this will get replaced entirely?

andy-sweet (approved):
I don't have strong opinions here, so approved if this enables other work.

Though my understanding was previously that `Chunk` would be used in a similar way to [Neuroglancer's chunk manager](https://github.com/google/neuroglancer/tree/master/src/chunk_manager), covering partial/complete data from things other n-dimensional image sources (e.g. a point cloud).

If and when we need that, I'm sure we can find another name. How about `Blob` as a slightly less refined `Chunk`?

packages/core/src/objects/textures/texture_2d.ts
shlomnissan: I suspect we’ll end up removing this or changing the interface entirely. Right now, when using the chunk manager, we slice the data before storing it in the chunk. Once "z" support is added to the chunk manager, the layer will handle slicing based on the region it covers. At that point, this function will just take width, height, and data.

This is the mental model I’m currently working with, though it’s still a bit in flux.

shlomnissan:
> If and when we need that, I'm sure we can find another name. How about Blob as a slightly less refined Chunk?

@andy-sweet Yeah, I think you're right that the next evolution of our logical "chunk" will be somewhere in between a pure data container and a stateful object. In our current model, a chunk carries priority, loading state, and also holds the actual data and metadata, since right now we’re only dealing with volumetric data.

But I agree that once we start supporting other data types (like meshes, line segments, or point clouds, similar to what Neuroglancer does), it probably makes sense to rethink the abstraction. At that point, we could generalize "chunk" into a more neutral state container that tracks loading and priority, while referencing an underlying payload. That payload might be a blob (as you suggested), or something more structured, like a multiscale volume tile, a mesh fragment, etc.

So “chunk” may remain a useful name for our existing approach, and we could introduce something like “blob” or another term for the more generic form when we need it.

What do you think?

PR #375, opened 2025-08-08
Manually add identity transform and rdefs to OME-Zarr v0.4 zod Image

andy-sweet:
### Summary
This manually adds two things to the OME-Zarr v0.4 zod Image automatically generated from the corresponding JSON schema.

1. Allows an identity transformation in the top-level coordinateTransformations.
2. Allows optional rdefs in the transitional omero metadata.

The reasons for manually adding these are given inline, one important one being that iohub writes these.

This allows us to remove a hacky workaround that mutated a shallow copy of the attributes, and a lint check skip.

I think this is a reasonable approach as the OME-Zarr specification evolves. I.e. we automatically and deterministically generate the zod object from the JSON schema, then make intentional and explained modifications as needed.

### Related Issue
Closes #360
Closes #361

### Tests & Checks
I manually tested all the existing examples.
I also manually tested the React component example.

PR #376, opened 2025-08-08
Support OME-Zarr v0.5

### LINKED ISSUE #367: Add an example with an OME-Zarr v0.5 image

### LINKED ISSUE #368: Add zod types for OME-Zarr v0.5

andy-sweet:
### Summary
This adds support for OME-Zarr v0.5 by doing a few things.

1. Automatically generates new zod objects from the v0.5 JSON schemata
2. Manually modifies the zod Image in the way as for v0.4.
3. Implement v0.4 -> v0.5 adaptors for image, plate, and well objects.
4. Wrap zarrita open functions to cleanly choose Zarr v2/v3 when possible (e.g. from the OME-Zarr version).
5. Update Idetik code based on changes from OME-Zarr v0.5 to v0.4.

It also modifies one of the existing example to use v0.5 as a simple way to test things are working.

### Related Issue
Closes #367
Closes #368
Related to #366 

### Tests & Checks
I manually tested all the dev examples.


packages/core/src/data/ome_zarr_image_loader.ts
```diff
@@ -42,12 +47,15 @@ export class PromiseQueue<T> {
...
+  private readonly metadata_: OmeNgffImage["ome"];
+  private readonly omeVersion_: OmeZarrVersion;
   private readonly lods_: number;
 
   constructor(root: zarr.Group<Readable>) {
     this.root_ = root;
-    this.metadata_ = OmeNgffImage.parse(this.root_.attrs);
+    const adaptedOmeImage = parseOmeNgffImage(this.root_.attrs);
+    this.metadata_ = adaptedOmeImage;
+    this.omeVersion_ = adaptedOmeImage.originalVersion;
```
andy-sweet: I should probably just separate these out as `ome` and `originalVersion`, instead of the flat approach used currently (though that does make some things more convenient). Another thought was to avoid returning `originalVersion` and instead return a function that can be used for opening the array.

andy-sweet:
@aganders3 : assigning to you as a follow up to #375

packages/core/src/data/ome_zarr_image_loader.ts
```diff
@@ -42,12 +47,15 @@ export class PromiseQueue<T> {
 // https://ngff.openmicroscopy.org/0.4/#image-layout
 export class OmeZarrImageLoader {
   private readonly root_: zarr.Group<Readable>;
-  private readonly metadata_: OmeNgffImage;
+  private readonly metadata_: OmeNgffImage["ome"];
+  private readonly omeVersion_: OmeZarrVersion;
```
aganders3: This might be unavoidable or by design, but we should try to be consistent about OmeNgff vs OmeZarr. I'm still not 100% clear on the difference and see them used interchangeably.

packages/core/src/data/ome_zarr_image_loader.ts
```diff
@@ -89,6 +94,11 @@ export class OmeZarrImageLoader {
     chunk.data = data.slice(zOffset, zOffset + sliceSize);
   }
 
+  private async openArray(path: string) {
+    const zarrVersion = this.omeVersion_ === "0.4" ? "v2" : "v3";
+    return openArray(this.root_.resolve(path), zarrVersion);
```
aganders3: I'm a little confused here because the version is being tested in more than once place. Can we leave any explicit mentions of the ome or zarr version out of this file, and keep that to the utility functions you defined in `zarrita/open.ts`?

packages/core/src/data/zarrita/open.ts
aganders3, line 1: I've got mixed feelings about naming this dir `zarrita` but I don't feel strongly and don't have a better suggestion at the moment. Maybe `zarritaWrappers` or `zarrCompat` but I don't love either.

packages/core/src/data/ome_zarr_hcs_metadata_loader.ts
```diff
@@ -1,41 +1,193 @@
 import * as zarr from "zarrita";
-import { Plate } from "../data/ome_ngff/0.4/plate";
-import { Well } from "../data/ome_ngff/0.4/well";
-import { Image } from "../data/ome_ngff/0.4/image";
+import { Image as ImageV04 } from "./ome_ngff/0.4/image";
+import { Plate as PlateV04 } from "./ome_ngff/0.4/plate";
+import { Well as WellV04 } from "./ome_ngff/0.4/well";
+import { Image } from "./ome_ngff/0.5/image";
+import { Plate } from "./ome_ngff/0.5/plate";
+import { Well } from "./ome_ngff/0.5/well";
```
aganders3, line 7: For readability it might help to add barrel exports here such that we could `import * as omeV04Schemas from "./ome_ngff/0.4"` or something.

(totally optional - this is also good)

packages/core/src/data/ome_zarr_hcs_metadata_loader.ts
aganders3, line 1: Nice job with these convenience wrappers!

aganders3 (approved):
This looks good to me. I'm not sure this technique will scale that well beyond OME-Zarr v0.5, but we've already discussed possible alternatives.

Looking forward to our apps (like the tomogram quality app) supporting zarr v3 w/ progressive chunk streaming and this is a big step in that direction!

packages/core/src/data/ome_zarr_image_loader.ts
andy-sweet: I did move this to `OmeZarrImageSource` because of other changes on `main` but I don't think that answers your question.

This is testing the original OME-Zarr version (i.e. 0.4 vs 0.5) to be explicit about which Zarr version to read when calling `openArray` (or potentially `openGroup`) later. I.e. it, or something derived from it, is passed to those functions to avoid the guesswork in zarrita's version-agnostic `open` function which may result in unnecessary fetches.

The only other place the OME-Zarr version is tested is when adapting the OME-Zarr metadata to match the latest supported version.

I could store the Zarr version as the result of parsing/adapting the OME-Zarr instead, but that just moves this comparison.

packages/core/src/data/ome_zarr_image_loader.ts
andy-sweet: Agreed. I tend to prefer OME-Zarr since that is more specific. I may follow this PR up with a general refactor and renaming (#366), which could handle this as well the name of the `zarrita` directory and `ome_zarr_hcs_metadata_loader` module.

packages/core/src/data/zarrita/open.ts
andy-sweet: I'll probably address this with other naming related things in a follow up (#366).

packages/core/src/data/ome_zarr_hcs_metadata_loader.ts
andy-sweet: If we need to support more of the schemata or more versions, this is probably worth doing. I may also consider addressing this in #366.

packages/core/src/data/ome_zarr_image_loader.ts
andy-sweet: Another point worth making here is that while I can update the OME-Zarr metadata with the adaptor functions, I can't easily update the Zarr version of the groups/arrays within it.

PR #377, opened 2025-08-10
Move "compute visible bounds" to camera and refactor

shlomnissan:
### Summary

This PR moves `computeVisibleBounds` from the chunk manager into the camera
class, where it more logically belongs. In the process, the temporary `Bounds`
type was replaced with a dedicated `Box2` type that encapsulates AABB
intersection checks. These changes are another step toward enabling the chunk
manager to prefetch data across all spatial dimensions.

### Tests & Checks

- Adds tests to validate the `Box2` functionality
-  All checks have passed

### DIFF
--- a/packages/core/examples/image2d_from_omezarr5d_u16/scale_bar.ts
+++ b/packages/core/examples/image2d_from_omezarr5d_u16/scale_bar.ts
@@ -19,7 +19,7 @@ export class ScaleBar {
 
   public update(idetik: Idetik, _timestamp?: DOMHighResTimeStamp): void {
     const camera = idetik.camera;
-    if (camera.type !== "OrthographicCamera") {
+    if (camera.type !== "Orthographic") {
       console.error("ScaleBar can only be used with OrthographicCamera");
       return;
     }
--- a/packages/core/examples/image2d_from_omezarr5d_u16/scale_bar.ts
+++ b/packages/core/examples/image2d_from_omezarr5d_u16/scale_bar.ts
@@ -19,7 +19,7 @@ export class ScaleBar {
 
   public update(idetik: Idetik, _timestamp?: DOMHighResTimeStamp): void {
     const camera = idetik.camera;
-    if (camera.type !== "Orthographic") {
+    if (camera.type !== "OrthographicCamera") {
       console.error("ScaleBar can only be used with OrthographicCamera");
       return;
     }
--- a/packages/core/src/core/chunk_manager.ts
+++ b/packages/core/src/core/chunk_manager.ts
@@ -291,8 +291,11 @@ export class ChunkManager {
   }
 
   public update(camera: OrthographicCamera, bufferWidth: number) {
-    if (camera.type !== "Orthographic") {
-      throw new Error("");
+    if (camera.type !== "OrthographicCamera") {
+      throw new Error(
+        "ChunkManager currently supports only orthographic cameras. " +
+          "Update the implementation before using a perspective camera."
+      );
     }
 
     const visibleBounds = camera.getWorldViewRect2D();
--- a/packages/core/src/core/chunk_manager.ts
+++ b/packages/core/src/core/chunk_manager.ts
@@ -5,12 +5,11 @@ import {
   LoaderAttributes,
 } from "../data/chunk";
 import { Region } from "../data/region";
-import { Camera } from "../objects/cameras/camera";
-import { vec2, vec4, mat4 } from "gl-matrix";
+import { vec2 } from "gl-matrix";
+import { Box2 } from "../math/box2";
 import { almostEqual } from "../utilities/almost_equal";
 import { Logger } from "../utilities/logger";
-
-type Bounds = { min: vec2; max: vec2 };
+import { OrthographicCamera } from "@/objects/cameras/orthographic_camera";
 
 // Number of chunks to extend beyond the visible bounds in each direction (x/y/z)
 // These additional chunks are prefetched to improve responsiveness when panning.
@@ -26,7 +25,7 @@ export class ChunkManagerSource {
   private readonly xIdx_: number;
   private readonly yIdx_: number;
   private readonly channelIdx_: number;
-  private lastVisibleBounds_: Bounds | null = null;
+  private lastVisibleBounds_: Box2 | null = null;
 
   constructor(loader: ChunkLoader, attrs: LoaderAttributes[], region: Region) {
     this.loader_ = loader;
@@ -121,7 +120,7 @@ export class ChunkManagerSource {
     return [...lowResChunks, ...currentLODChunks];
   }
 
-  public update(lodFactor: number, visibleBounds: Bounds) {
+  public update(lodFactor: number, visibleBounds: Box2) {
     this.setLOD(lodFactor);
     if (this.visibleBoundsChanged(visibleBounds)) {
       this.updateChunkVisibility(visibleBounds);
@@ -195,7 +194,7 @@ export class ChunkManagerSource {
     }
   }
 
-  private updateChunkVisibility(visibleBounds: Bounds): void {
+  private updateChunkVisibility(visibleBounds: Box2): void {
     if (this.chunks_.length === 0) {
       Logger.warn(
         "ChunkManager",
@@ -230,26 +229,18 @@ export class ChunkManagerSource {
     }
   }
 
-  private isChunkWithinBounds(chunk: Chunk, bounds: Bounds): boolean {
-    const boundsMinX = bounds.min[0];
-    const boundsMaxX = bounds.max[0];
-    const boundsMinY = bounds.min[1];
-    const boundsMaxY = bounds.max[1];
-
-    const minX = chunk.offset.x;
-    const maxX = minX + chunk.shape.x * chunk.scale.x;
-    const minY = chunk.offset.y;
-    const maxY = minY + chunk.shape.y * chunk.scale.y;
-
-    return (
-      minX <= boundsMaxX &&
-      maxX >= boundsMinX &&
-      minY <= boundsMaxY &&
-      maxY >= boundsMinY
+  private isChunkWithinBounds(chunk: Chunk, bounds: Box2): boolean {
+    const chunkBounds = new Box2(
+      vec2.fromValues(chunk.offset.x, chunk.offset.y),
+      vec2.fromValues(
+        chunk.offset.x + chunk.shape.x * chunk.scale.x,
+        chunk.offset.y + chunk.shape.y * chunk.scale.y
+      )
     );
+    return Box2.intersects(chunkBounds, bounds);
   }
 
-  private visibleBoundsChanged(newBounds: Bounds): boolean {
+  private visibleBoundsChanged(newBounds: Box2): boolean {
     const prev = this.lastVisibleBounds_;
     const changed =
       prev === null ||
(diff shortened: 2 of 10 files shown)


packages/core/src/core/chunk_manager.ts
```diff
@@ -299,34 +290,18 @@ export class ChunkManager {
     return existing;
   }
 
-  public update(camera: Camera, _bufferWidth: number, _bufferHeight: number) {
-    const visibleBounds = this.computeVisibleBounds(camera);
+  public update(camera: OrthographicCamera, bufferWidth: number) {
+    if (camera.type !== "Orthographic") {
+      throw new Error("");
```
andy-sweet: Currently, you're throwing here and in Idetik's render function. I think you should just throw here with a reasonable message because I think keeping Idetik's render function concise and readable is important.

packages/core/examples/image2d_from_omezarr5d_u16/scale_bar.ts
```diff
@@ -19,7 +19,7 @@ export class ScaleBar {
 
   public update(idetik: Idetik, _timestamp?: DOMHighResTimeStamp): void {
     const camera = idetik.camera;
-    if (camera.type !== "OrthographicCamera") {
+    if (camera.type !== "Orthographic") {
```
andy-sweet: Nit-ish: we use `type` as a property somewhat frequently to identify a class (e.g. `ImageLayer`), at least partly so we can identify it as such after minification. For that reason, I prefer the longer `OrthographicCamera` over `Orthographic` here.

packages/core/src/math/box2.ts
```diff
@@ -0,0 +1,25 @@
+import { vec2 } from "gl-matrix";
+
+export class Box2 {
+  public min: vec2;
+  public max: vec2;
+
+  constructor(min?: vec2, max?: vec2) {
```
andy-sweet, line 14: Do we need these to be optional? I would have expected a default box to be all zeros rather than edges infinitely far away. I understand that all zeros wouldn't be empty the way it is defined here, but I'm not sure I love that definition anyway (since the area of the box is 0).

Since this is quite a primitive type, I think it might just be simpler to always require these values for now, which I think Idetik can do right now.

packages/core/test/box2.test.ts
```diff
@@ -0,0 +1,62 @@
...
+  const b = new Box2();
+  expect(b.isEmpty()).toBe(true);
+});
+
+test("non-empty when min <= max on both axes", () => {
+  const b = new Box2(vec2.fromValues(0, 0), vec2.fromValues(1, 1));
+  expect(b.isEmpty()).toBe(false);
+});
+
+test("isEmtpy detects inverted bounds", () => {
```
andy-sweet:
```suggestion
test("isEmpty detects inverted bounds", () => {
```

packages/core/src/math/box2.ts
```diff
@@ -0,0 +1,25 @@
+import { vec2 } from "gl-matrix";
+
+export class Box2 {
```
andy-sweet, line 3: Definitely approve of this new class!

packages/core/test/box2.test.ts
```diff
@@ -0,0 +1,62 @@
...
+
+test("intersects: touching edges count as intersecting", () => {
+  const a = new Box2(vec2.fromValues(0, 0), vec2.fromValues(2, 2));
+  const b = new Box2(vec2.fromValues(2, 0), vec2.fromValues(4, 2));
+  const c = new Box2(vec2.fromValues(0, 2), vec2.fromValues(2, 4));
+  expect(Box2.intersects(a, b)).toBe(true);
+  expect(Box2.intersects(a, c)).toBe(true);
+});
+
+test("intersects: empty boxes never intersect", () => {
```
andy-sweet: Nit: this test name is quite strong (never) even though it only tests one case. Also the way it's worded makes me think both boxes are empty, which is not the case here. The test body is fine, but consider a different name/description.

```suggestion
test("intersects: empty and non-empty box do not intersect", () => {
```

packages/core/src/math/box2.ts
```diff
@@ -0,0 +1,25 @@
...
+  constructor(min?: vec2, max?: vec2) {
+    this.min = min ? vec2.clone(min) : vec2.fromValues(+Infinity, +Infinity);
+    this.max = max ? vec2.clone(max) : vec2.fromValues(-Infinity, -Infinity);
+  }
+
+  public clone() {
+    return new Box2(this.min, this.max);
+  }
+
+  public isEmpty(): boolean {
```
andy-sweet, line 23: Optional.

The definition of `isEmpty` is a little unexpected since a zero-area box is not empty. We don't actually use this in Idetik yet, so I'm wondering if we can just remove it?

andy-sweet (changes requested):
The only blocking request is to throw once (ideally in the chunk manager). All other comments are optional/nits.

packages/core/src/math/box2.ts
aganders3: +1!

packages/core/src/math/box2.ts
```diff
@@ -0,0 +1,25 @@
+import { vec2 } from "gl-matrix";
+
+export class Box2 {
+  public min: vec2;
+  public max: vec2;
```
aganders3, line 5: Since we're not doing any vector algebra with these (yet?) I'd find it more ergonomic to have the object structured like:

`{ x: { min: number, max, number}, y { min: number, max: number } }`

This prevents the unintuitive (IMO) accessing by indices.

Something like this could probably be done with computed properties though if there is value in keeping these as `vec2`.

packages/core/src/core/chunk_manager.ts
shlomnissan: Ah double throw was not intentional. I wanted to move it from the runtime to the chunk manager, but I forgot to complete the change.

packages/core/src/objects/cameras/orthographic_camera.ts
```diff
@@ -59,6 +60,26 @@ export class OrthographicCamera extends Camera {
     this.transform.addScale([inverseFactor, inverseFactor, 1.0]);
   }
 
+  public getWorldViewRect2D(): Box2 {
```
aganders3, line 63: nit: I prickle a little when I see a getter for `Rect2D` return a `Box2` because they're such similar terms. I'd align them or somehow be clear what the difference is.

packages/core/examples/image2d_from_omezarr5d_u16/scale_bar.ts
shlomnissan: My bias against longer names comes from C++ where scoped enums force an explicit name (i.e. `CameraType::Orthographic`). This is irrelevant here and I don't have a strong opinion so I'm happy to revert.

packages/core/src/idetik.ts
```diff
@@ -124,11 +125,19 @@ export class Idetik {
...
       }
-      this.chunkManager_.update(
-        this.camera,
-        this.renderer_.width,
-        this.renderer_.height
-      );
+
+      if (this.camera.type === "Orthographic") {
+        this.chunkManager_.update(
+          this.camera as OrthographicCamera,
```
aganders3: In the spirit of fully decoupling the ChunkManager from the Camera - this could take the `visibleBounds` (e.g. as a `Box2`).

packages/core/src/math/box2.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  constructor(min?: vec2, max?: vec2) {
+    this.min = min ? vec2.clone(min) : vec2.fromValues(+Infinity, +Infinity);
+    this.max = max ? vec2.clone(max) : vec2.fromValues(-Infinity, -Infinity);
```
aganders3, line 16: Should we do anything here to enforce a valid specification? Like checking `min[0]` < `max[0]` for example. If not, what does it mean if one or both axes are "inverted" like that?

aganders3 (commented):
LGTM pending Andy's requested changes!

packages/core/src/math/box2.ts
shlomnissan: This is a very common pattern (e.g., [Three.js::Box2](https://github.com/mrdoob/three.js/blob/dev/src/math/Box2.js)) called "empty-by-sentinel". It initializes a primitive with extreme values so any expansion (point or box union) will automatically shrink it to valid bounds without needing special-case logic. Also, a zero box at the origin is strictly not empty, it's a valid degenerate box that can leak into logic (e.g., intersects, unions) and subtly bias results toward the origin.

I’d like to stick with the empty-by-sentinel approach. it’s pretty standard in geometry/math code, and the simplicity it gives us is worth the slightly less obvious default state. I know we're not using the empty state right now but having it in place means if someone extends this later, they’ll be following a good convention.

I’ll add a comment explaining the reasoning so it’s clear to future readers.

packages/core/src/math/box2.ts
shlomnissan: I added a comment explaining my reasoning.

packages/core/src/objects/cameras/orthographic_camera.ts
shlomnissan: Ah good catch! I initially called `Box2` -> `Rect2` but decided to change it because I might introduce a `Box3` type soon, and I forgot to rename this function.

packages/core/src/math/box2.ts
aganders3: Sorry - I just read your comment above and understand now that min<max means "empty". I'm happy to follow established conventions here.

shlomnissan: You could validate here, but most geometry/math libraries don’t because sometimes you intentionally create an “inverted” box as a sentinel (empty-by-sentinel pattern) or as a temporary state before expansion.

packages/core/src/idetik.ts
shlomnissan: I suspect this will change soon. As mentioned in my Slack thread, I think the layer may be responsible for obtaining the visible bounds from the camera and updating the region.

packages/core/src/math/box2.ts
shlomnissan: I’d like to keep min/max as `vec2` for now because we use we already use it all over. keeping `vec2` avoids wrappers/conversions. Also, this mirrors existing math libraries where index access is the norm. I agree that indexing isn’t the most readable. If you feel strongly about it, I can add named accessors for ergonomics?

aganders3: Nah, don't change it now. I think working _within_ this class it's fine to use the indexing. Outside the class we can see where it's used and make a decision to add accessors or just use local assignments to disambiguate.

packages/react/src/components/viewers/OmeZarrImageViewer/components/ScaleBar/ScaleBar.tsx
```diff
@@ -87,7 +87,7 @@ class ScaleBarOverlay {
     if (textDiv === null) return;
 
     const camera = idetik.camera;
-    if (camera.type !== "OrthographicCamera") {
+    if (camera.type !== "Orthographic") {
```
andy-sweet: Need to revert this back.

```suggestion
    if (camera.type !== "OrthographicCamera") {
```

andy-sweet (approved):
Looks good to me now - you just missed one revert in the `ScaleBar` component.

PR #378, opened 2025-08-11
Add Z-axis fields to Chunk type

shlomnissan:
### Summary

This PR adds Z-axis support to the `Chunk` type and related code paths. It
introduces z-dimension indices, shapes, scales, and offsets for all LODs,
ensuring the type can represent full 3D chunks. Visibility and loading logic
remain 2D-only for now, with nonzero Z slices gated until `Box3` intersection
tests are implemented. **This is a schema-only change**, preparing for future
3D visibility and prefetching support.

### Tests & Checks
- All checks have passed

### DIFF
--- a/packages/core/src/core/chunk_manager.ts
+++ b/packages/core/src/core/chunk_manager.ts
@@ -36,30 +36,31 @@ export class ChunkManagerSource {
     this.currentLOD_ = 0;
 
     this.xIdx_ = region.findIndex(
-      (entry) => entry.dimension.toLocaleLowerCase() === "x"
+      (entry) => entry.dimension.toLowerCase() === "x"
     );
     this.yIdx_ = region.findIndex(
-      (entry) => entry.dimension.toLocaleLowerCase() === "y"
+      (entry) => entry.dimension.toLowerCase() === "y"
     );
     this.zIdx_ = region.findIndex(
-      (entry) => entry.dimension.toLocaleLowerCase() === "z"
+      (entry) => entry.dimension.toLowerCase() === "z"
     );
     this.channelIdx_ = region.findIndex(
-      (entry) => entry.dimension.toLocaleLowerCase() === "c"
+      (entry) => entry.dimension.toLowerCase() === "c"
     );
 
-    if (this.xIdx_ === -1 || this.yIdx_ === -1 || this.zIdx_ === -1) {
-      throw new Error("Missing required spatial axis x/y/z");
+    if (this.xIdx_ === -1 || this.yIdx_ === -1) {
+      throw new Error("Missing required spatial axis x/y");
     }
 
-    this.validateScaleRatios(this.xIdx_, this.yIdx_, this.zIdx_);
+    this.validateXYScaleRatios(this.xIdx_, this.yIdx_);
 
     // generate chunks for each LOD without loading data
     this.chunks_ = [];
     for (let lod = 0; lod < this.attrs_.length; ++lod) {
       const chunkWidth = this.attrs_[lod].chunks[this.xIdx_];
       const chunkHeight = this.attrs_[lod].chunks[this.yIdx_];
-      const chunkDepth = this.attrs_[lod].chunks[this.zIdx_];
+      const chunkDepth =
+        this.zIdx_ !== -1 ? this.attrs_[lod].chunks[this.zIdx_] : 1;
 
       const chunksX = Math.ceil(
         this.attrs_[lod].shape[this.xIdx_] / chunkWidth
@@ -67,9 +68,10 @@ export class ChunkManagerSource {
       const chunksY = Math.ceil(
         this.attrs_[lod].shape[this.yIdx_] / chunkHeight
       );
-      const chunksZ = Math.ceil(
-        this.attrs_[lod].shape[this.zIdx_] / chunkDepth
-      );
+      const chunksZ =
+        this.zIdx_ !== -1
+          ? Math.ceil(this.attrs_[lod].shape[this.zIdx_] / chunkDepth)
+          : 1;
 
       const channels =
         this.channelIdx_ >= 0 ? this.attrs_[lod].shape[this.channelIdx_] : 1;
@@ -96,13 +98,17 @@ export class ChunkManagerSource {
               scale: {
                 x: scale[this.xIdx_],
                 y: scale[this.yIdx_],
-                z: scale[this.zIdx_],
+                z: this.zIdx_ !== -1 ? scale[this.zIdx_] : 1,
               },
               offset: {
                 x: translation[this.xIdx_] + x * chunkWidth * scale[this.xIdx_],
                 y:
                   translation[this.yIdx_] + y * chunkHeight * scale[this.yIdx_],
-                z: translation[this.zIdx_] + z * chunkDepth * scale[this.zIdx_],
+                z:
+                  this.zIdx_ !== -1
+                    ? translation[this.zIdx_] +
+                      z * chunkDepth * scale[this.zIdx_]
+                    : 0,
               },
             });
           }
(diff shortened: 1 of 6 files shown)


packages/core/src/core/chunk_manager.ts
```diff
@@ -40,58 +41,71 @@ export class ChunkManagerSource {
...
       );
+      const chunksZ = Math.ceil(
+        this.attrs_[lod].shape[this.zIdx_] / chunkDepth
+      );
+
       const channels =
-        this.attrs_[lod].shape.length === 3
-          ? this.attrs_[lod].shape[this.channelIdx_]
-          : 1;
+        this.channelIdx_ >= 0 ? this.attrs_[lod].shape[this.channelIdx_] : 1;
```
shlomnissan, line 77: Unrelated change. I think this check is more robust than `shape.length === 3`.

packages/core/src/data/ome_zarr_image_loader.ts
```diff
@@ -158,13 +159,18 @@ export class OmeZarrImageLoader {
       shape: {
         x: subarray.shape[subarray.shape.length - 1],
         y: subarray.shape[subarray.shape.length - 2],
+        z: subarray.shape[subarray.shape.length - 3],
```
andy-sweet: What if the OME-Zarr does not have a z dimension?

packages/core/src/core/chunk_manager.ts
```diff
@@ -213,17 +235,21 @@ export class ChunkManagerSource {
...
       const curr = availableScales[i];
       const rx = curr[xIdx] / prev[xIdx];
       const ry = curr[yIdx] / prev[yIdx];
+      const rz = curr[zIdx] / prev[zIdx];
 
-      if (!almostEqual(rx, 2) || !almostEqual(ry, 2)) {
+      if (!almostEqual(rx, 2) || !almostEqual(ry, 2) || !almostEqual(rz, 2)) {
         throw new Error(
-          `Scales must be separated by factors of 2. Got ratio (${rx}, ${ry}) between scales ${prev} and ${curr}`
+          `Invalid scale ratio between LOD ${i - 1} and LOD ${i}: ` +
```
andy-sweet: This is thrown unexpectedly in the following example: http://localhost:5173/image2d_from_omezarr5d_u8/index.html

That's because the downsampling is not the same in all dimensions in that case, which in turn is because each slice in the z-stack is spaced relatively far apart.

packages/core/src/core/chunk_manager.ts
```diff
@@ -40,58 +41,71 @@ export class ChunkManagerSource {
...
+    this.zIdx_ = region.findIndex(
+      (entry) => entry.dimension.toLocaleLowerCase() === "z"
+    );
+    this.channelIdx_ = region.findIndex(
+      (entry) => entry.dimension.toLocaleLowerCase() === "c"
     );
-    if (channelIdx === -1) {
-      channelIdx = 0;
+
+    if (this.xIdx_ === -1 || this.yIdx_ === -1 || this.zIdx_ === -1) {
```
andy-sweet: I would only expect z to be required if the source has z.

shlomnissan:
@andy-sweet The 3D assumption is an oversight. Updated to support OME-Zarr datasets without a Z dimension. The chunk manager code, especially the constructor is getting messy and could use a refactor, but that can be done
in a follow-up.

packages/core/src/data/ome_zarr_image_loader.ts
```diff
@@ -146,8 +146,20 @@ export class OmeZarrImageLoader {
...
+
+    const xIdx = region.findIndex(
+      (entry) => entry.dimension.toLowerCase() === "x"
+    );
+    const yIdx = region.findIndex(
+      (entry) => entry.dimension.toLowerCase() === "y"
+    );
+    const zIdx = region.findIndex(
+      (entry) => entry.dimension.toLowerCase() === "z"
+    );
```
shlomnissan: This is becoming repetitive. A better approach would be to refactor `Region` into a class that encapsulates this functionality, but that’s out of scope for this PR.

andy-sweet: This confuses the region which the user provides (whose order is not currently relevant) and `indices` which are ordered consistently with the Zarr array dimensions. This change causes a bug in the `points` example, and perhaps others.

The simple fix in this PR is to revert these changes to use `indices.length - 1` and `indices.length - 2` for x and y, which is implied by the OME-Zarr spec (though we should be more explicit about it here.

For z, I think you need to something a bit more complex here.
1. Check if it is a point-type index in the region. If it is, then I believe the subarray will not include it (because zarrita removes singleton dimensions when indexing with a number), and you can just set dummy values.
2. If it isn't a point-type index and it exists in the region, then it has some depth in the subarray, and it should be found as `indices.length - 3`.

There's some further nuance here, because even if z is a single number, I might expect the zOffset to tell me where that plane/slab is in z, rather than being 0. But that can be left for future work.

Another thought is that `loadRegion` could throw if it detects a non-point z-slice, because we're not actually using that anywhere.

> A better approach would be to refactor Region into a class that encapsulates this functionality, but that’s out of scope for this PR.

Agreed that some somewhat significant refactoring is needed here and in chunk manager. That's the bulk of what I've been thinking through over the last day or two, and think I have something that should more explicit and solid than what we currently have.

packages/core/src/data/ome_zarr_image_loader.ts
```diff
@@ -158,13 +170,18 @@ export class OmeZarrImageLoader {
       shape: {
         x: subarray.shape[subarray.shape.length - 1],
         y: subarray.shape[subarray.shape.length - 2],
+        z: zIdx !== -1 ? subarray.shape[subarray.shape.length - 3] : 1,
         c: subarray.shape.length === 3 ? subarray.shape[0] : 1,
```
andy-sweet, line 163: Does this check for c need to be more complex now since that z could be present in the subarray?

packages/core/src/core/chunk_manager.ts
```diff
@@ -35,63 +36,82 @@ export class ChunkManagerSource {
...
-            scale: {
-              x: scale[this.xIdx_],
-              y: scale[this.yIdx_],
-            },
-            offset: {
-              x: translation[this.xIdx_] + x * chunkWidth * scale[this.xIdx_],
-              y: translation[this.yIdx_] + y * chunkHeight * scale[this.yIdx_],
-            },
-          });
+          for (let z = 0; z < chunksZ; ++z) {
```
andy-sweet, line 83: No actionable feedback, but this extra for loop could multiply the total number of chunks to be quite large for some data (e.g. the zebrahub or DCA data come to mind). I'm not too worried about the memory associated with empty chunk object (though even that could become an issue with an extension to t), but we may need to start thinking about freeing up memory once they are not empty anymore (e.g. a short term way to handle this could be to set `chunk.data = null` in `ImageLayer` when they are not intended for prefetch.

packages/core/src/data/ome_zarr_image_loader.ts
shlomnissan: Already confirmed with @andy-sweet over Slack, posting here for completeness:

I believe these changes might be unnecessary. `loadRegion` operates outside the chunk manager. The simplest approach would be to revert these changes and assign default values to the new z-properties. This should preserve the current behavior. If there’s an issue with how this function handles the non-point z-index, it’s likely an existing problem.

I manually verified that all existing examples work fine with this change.

packages/core/src/core/chunk_manager.ts
shlomnissan: > short term way to handle this could be to set chunk.data = null in ImageLayer

I agree. I think it's best to address this when we experiment with time prefetching because we might hit memory limits at that point. Setting `chunk.data = null` might be enough, but I think we also need to dispose memory stored in GPU buffers

packages/core/src/data/ome_zarr_image_loader.ts
andy-sweet: Well, now it doesn't! Resolved!

andy-sweet (approved):
This looks good to me now.

PR #379, opened 2025-08-11
feat: add value picking support to ImageLayer

phoenixAja:
###  Summary

  Adds value picking functionality to ImageLayer similar to existing LabelImageLayer implementation, with support for progressive loading.

  Changes

  - Added value picking to ImageLayer: Supports clicking to get pixel values at world coordinates
  - Progressive loading compatibility: Works with chunk manager and multiple visible chunks
  - Multi-channel support: Returns first channel value for multi-channel images
  - Updated uint8 example: Added point picking overlay demo with visual feedback
  - Shared utility function: Created handlePointPickingEvent to DRY out common pointer logic between layers
  - Type definitions: Added ImagePointPickingResult interface

  Key Features

  - Works with both single and multi-channel images (returns first channel)
  - Supports progressive loading with multiple chunks
  - Provides world coordinates and pixel values in callback
  - Includes visual overlay example for testing

  Test Plan

  - Point picking works in image2d_from_omezarr5d_u8 example
  - Shared pointer event logic works in both ImageLayer and LabelImageLayer

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -15,7 +15,6 @@ export { AxesLayer } from "./layers/axes_layer";
 export { ProjectedLineLayer } from "./layers/projected_line_layer";
 export { TracksLayer } from "./layers/tracks_layer";
 export { ImageLayer } from "./layers/image_layer";
-export type { ImagePointPickingResult } from "./layers/image_layer";
 export { LabelImageLayer } from "./layers/label_image_layer";
 export { ImageSeriesLayer } from "./layers/image_series_layer";
 export { OmeZarrImageSource } from "./data/ome_zarr_image_source";
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -15,8 +15,8 @@ export { AxesLayer } from "./layers/axes_layer";
 export { ProjectedLineLayer } from "./layers/projected_line_layer";
 export { TracksLayer } from "./layers/tracks_layer";
 export { ImageLayer } from "./layers/image_layer";
-export type { ImagePointPickingResult } from "./layers/image_layer";
 export { LabelImageLayer } from "./layers/label_image_layer";
+export type { PointPickingResult } from "./layers/label_image_layer";
 export { ImageSeriesLayer } from "./layers/image_series_layer";
 export { OmeZarrImageSource } from "./data/ome_zarr_image_source";
 
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -15,6 +15,7 @@ export { AxesLayer } from "./layers/axes_layer";
 export { ProjectedLineLayer } from "./layers/projected_line_layer";
 export { TracksLayer } from "./layers/tracks_layer";
 export { ImageLayer } from "./layers/image_layer";
+export type { ImagePointPickingResult } from "./layers/image_layer";
 export { LabelImageLayer } from "./layers/label_image_layer";
 export { ImageSeriesLayer } from "./layers/image_series_layer";
 export { OmeZarrImageSource } from "./data/ome_zarr_image_source";
--- a/packages/core/src/layers/image_layer.ts
+++ b/packages/core/src/layers/image_layer.ts
@@ -9,11 +9,19 @@ import { Texture2DArray } from "../objects/textures/texture_2d_array";
 import { PlaneGeometry } from "../objects/geometry/plane_geometry";
 import { Logger } from "../utilities/logger";
 import { Color } from "../core/color";
+import { EventContext } from "../core/event_dispatcher";
+import { vec2, vec3 } from "gl-matrix";
+
+export interface ImagePointPickingResult {
+  world: vec3;
+  value: number;
+}
 
 export type ImageLayerProps = LayerOptions & {
   source: ChunkSource;
   region: Region;
   channelProps?: ChannelProps[];
+  onPickValue?: (info: ImagePointPickingResult) => void;
 };
 
 // Loads data from an image source into renderable objects.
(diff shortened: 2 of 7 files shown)


packages/core/src/layers/image_layer.ts
```diff
@@ -9,11 +9,20 @@ import { Texture2DArray } from "../objects/textures/texture_2d_array";
 import { PlaneGeometry } from "../objects/geometry/plane_geometry";
 import { Logger } from "../utilities/logger";
 import { Color } from "../core/color";
+import { EventContext } from "../core/event_dispatcher";
+import { vec2, vec3 } from "gl-matrix";
+import { handlePointPickingEvent } from "../utilities/point_picking";
+
+export interface ImagePointPickingResult {
+  world: vec3;
+  value: number;
```
andy-sweet: I would find it slightly more natural for `value: number | number[]` so that we can return values across all channels, because I think of a multi-channel image as being a vector valued image. But don't feel too strongly until we actually need that functionality.

packages/core/src/layers/image_layer.ts
```diff
@@ -208,4 +233,29 @@ export class ImageLayer extends Layer implements ChannelsEnabled {
     image.transform.setTranslation([chunk.offset.x, chunk.offset.y, 0]);
     return image;
   }
+
+  public getValueAtWorld(world: vec3): number | null {
+    // Iterate through all visible chunks to find the one containing the world position
+    for (const [chunk, image] of this.visibleChunks_) {
```
andy-sweet, line 234: This will only work for chunk manager based implementations, which might be fine, because that is the long term plan. But you could easily extend this to the other implementation path by adding the single `ImageRenderable` to `visibleChunks` in `load()`.

packages/core/src/utilities/point_picking.ts
```diff
@@ -0,0 +1,55 @@
...
+        const world = event.worldPos;
+        if (world) {
+          const value = getValueAtWorld(world);
+          if (value !== null) {
+            if (onPickValue) {
+              onPickValue({ world, value });
+            } else {
+              Logger.warn(
+                layerName,
+                "Point picking attempted but no onPickValue callback provided"
```
andy-sweet: I'm seeing this warning in the label image example, but not the intensity image example. I'm not sure why.

packages/core/src/layers/image_layer.ts
andy-sweet: I also wonder if this type could be shared across the image-like layers.

packages/core/src/utilities/point_picking.ts
```diff
@@ -0,0 +1,55 @@
+import { EventContext } from "../core/event_dispatcher";
+import { vec2, vec3 } from "gl-matrix";
+import { Logger } from "./logger";
+
+type LoggerModule = "ImageLayer" | "LabelImageLayer";
+
+export function handlePointPickingEvent<T>(
```
andy-sweet, line 4: Nit: I would probably expect this to live in the `layers` directory (because it is intended to be used by layers) rather than `utilities` (which look like very general purpose utils).

packages/core/src/layers/image_layer.ts
```diff
@@ -130,6 +144,17 @@ export class ImageLayer extends Layer implements ChannelsEnabled {
     }
   }
 
+  public onEvent(event: EventContext) {
+    this.pointerDownPos_ = handlePointPickingEvent(
```
andy-sweet, line 144: This plain function call is somewhat hard to read, so if you did feel like trying out a mix-in now to improve that, I'd be happy to review it.

Another thought is that if some of these arguments are always the same (e.g. dragThreshold), then maybe we should create an instance of `PointPickingEventHandler`  in the constructor and call some method of it here. This might be a way to improve readability here without complicating the typing/inheritance of the class (i.e. preferring composition over mixing-in or standard inheritance).

andy-sweet (changes requested):
All comments are optional, except that I get a warning in the label image picking example though everything seems to work as expected, which should be resolved before merging.

phoenixAja:
> All comments are optional, except that I get a warning in the label image picking example though everything seems to work as expected, which should be resolved before merging.

Had to remove the logging - now that the ImageLayer has value picking enabled, you signal to the layer that you don't want support for value picking by not passing in the onPickValue callback. So the warning doesn't actually make sense

packages/core/src/layers/image_layer.ts
phoenixAja: yeah i think let's add number[] once we need it (otherwise it feels misleading). Also am now sharing the interface!

packages/core/src/utilities/point_picking.ts
phoenixAja: it's because the label Image also has point picking, but no callback, i removed the logging, (explained in a response below) - thanks for catching!

packages/core/src/layers/image_layer.ts
phoenixAja: so i added this because handlePointPickingEvent is now shared between labelImageLayer and ImageLayer. I think a mixin might be less readable than a shared utility function like this?

andy-sweet: I think it's fine for now.

packages/core/src/index.ts
```diff
@@ -16,6 +16,7 @@ export { ProjectedLineLayer } from "./layers/projected_line_layer";
 export { TracksLayer } from "./layers/tracks_layer";
 export { ImageLayer } from "./layers/image_layer";
 export { LabelImageLayer } from "./layers/label_image_layer";
+export type { PointPickingResult } from "./layers/label_image_layer";
```
andy-sweet, line 19: Nit: given we have a module called `point_picking`, I would expect to see this type defined and exported there instead of the `label_image_layer` module.

packages/core/src/layers/image_layer.ts
```diff
@@ -26,12 +31,15 @@ export class ImageLayer extends Layer implements ChannelsEnabled {
...
   private readonly initialChannelProps_?: ChannelProps[];
+  private readonly onPickValue_?: (info: PointPickingResult) => void;
   private readonly channelChangeCallbacks_: Array<() => void> = [];
   private readonly visibleChunks_: Map<Chunk, ImageRenderable> = new Map();
   private chunkManagerSource_?: ChunkManagerSource;
   private channelProps_?: ChannelProps[];
   private image_?: ImageRenderable;
   private extent_?: { x: number; y: number };
+  private pointerDownPos_: vec2 | null = null;
+  private readonly dragThreshold_ = 3;
```
andy-sweet, line 42: Optional.

Since this is a readonly property here and in `LabelImageLayer` (i.e. it's effectively a constant), I think it makes more sense to make this a argument of `handlePointPickingEvent` with a default value of 3.

andy-sweet (approved):
Looks good now. Thanks!

phoenixAja:
@andy-sweet will address feedback in a followup PR, merging this now to unblock bryan!

PR #380, opened 2025-08-13
Consolidate OME-Zarr loader attributes and array opens

andy-sweet:
### Summary
This is a refactor that consolidates OME-Zarr attribute types, as well calls to opening the underlying Zarr arrays that store the multi-scale image chunks.

Before this change, `ImageAttributes` and `LoaderAttributes` were storing very similar information. Since `LoaderAttributes` is publicly exposed and stores a little more information (i.e. the overall shape and chunk shape for a multi-scale level), I kept that here.

I also consolidated opening all the underlying Zarr arrays in the call to `OmeZarrImageSource.open` rather than scattered across various functions in `OmeZarrImageLoader`. These arrays are then passed through to the loader, which then allows us to populate the loader attributes synchronously on construction. The main motivation behind this change is simplification, though there may be some small performance gains. Note that any call to `loadAttributes` (e.g. in `ChunkManager.addSource`) was already opening all the arrays anyway, so this should be no worse in that case.

### Tests & Checks

### DIFF
--- a/packages/core/src/data/chunk.ts
+++ b/packages/core/src/data/chunk.ts
@@ -60,9 +60,9 @@ export type ChunkSource = {
 };
 
 export type LoaderAttributes = {
-  chunks: readonly number[];
   dimensionNames: string[];
   dimensionUnits: (string | undefined)[];
+  chunks: readonly number[];
   shape: readonly number[];
   scale: readonly number[];
   translation: readonly number[];
@@ -77,5 +77,5 @@ export type ChunkLoader = {
 
   loadChunkDataFromRegion(chunk: Chunk, region: Region): Promise<void>;
 
-  loadAttributes(): Promise<LoaderAttributes[]>;
+  getAttributes(): ReadonlyArray<LoaderAttributes>;
 };
--- a/packages/core/src/data/ome_zarr_image_source.ts
+++ b/packages/core/src/data/ome_zarr_image_source.ts
@@ -27,12 +27,12 @@ export class OmeZarrImageSource {
         : new Location(new WebFileSystemStore(source), path);
   }
 
-  async open(): Promise<OmeZarrImageLoader> {
+  public async open(): Promise<OmeZarrImageLoader> {
     const root = await zarritaOpen.v2(this.location, { kind: "group" });
     const images = OmeNgffImage.parse(root.attrs).multiscales;
-    if (images.length > 1) {
+    if (images.length !== 1) {
       throw new Error(
-        `Currently only a single multiscale image is supported. Found ${images.length} images.`
+        `Exactly one multiscale image is supported. Found ${images.length} images.`
       );
     }
     const metadata = images[0];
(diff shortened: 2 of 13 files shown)


andy-sweet:
@shlomnissan : I assigned this to you for review because this likely overlaps with some of your previous and upcoming work on the chunk manager. Other reviews and comments are welcome.

packages/core/src/data/chunk.ts
```diff
@@ -60,9 +60,9 @@ export type ChunkSource = {
 };
 
 export type LoaderAttributes = {
```
andy-sweet, line 62: I'd love a more descriptive name than `LoaderAttributes`, but I can't think of one. An instance of this stores the attributes of a multi-scale level / LOD of the larger thing a chunk lives in (e.g. a tensor), but `TensorLodAttributes` doesn't feel particularly readable to me.

andy-sweet:
For more context, see the [virtual-camera](https://github.com/chanzuckerberg/idetik/compare/main...andy-sweet/virtual-camera) branch, which is where I initially made some similar changes. I created this smaller PR from that because I thought this was a small, independently useful change.

packages/core/src/data/ome_zarr_image_source.ts
```diff
@@ -26,8 +27,36 @@ export class OmeZarrImageSource {
...
+    if (images.length !== 1) {
+      throw new Error(
+        `Exactly one multiscale image is supported. Found ${images.length} images.`
+      );
+    }
+    const metadata = images[0];
+    if (metadata.datasets.length === 0) {
+      throw new Error(`No datasets found in the multiscale image.`);
+    }
+    const arrays = await Promise.all(
```
andy-sweet, line 42: Note this is very similar to the concept of [consolidated metadata in Zarr](https://zarr.readthedocs.io/en/main/user-guide/consolidated_metadata.html#). 

[zarrita](https://zarrita.dev/cookbook.html#open-a-consolidated-store) seems to have some support for that, so it might be worth trying to fetch the consolidated metadata first (so we only do one fetch), and then falling back to this if it's missing.

packages/core/src/data/chunk.ts
shlomnissan: Maybe something like `SourceAttributes`? It's not much more descriptive but I find it slightly clearer as "loader attributes" might be confused with application-level loading attributes.

shlomnissan (approved):
LGTM!

packages/core/src/data/ome_zarr_image_source.ts
andy-sweet: I quickly tried using `tryWithConsolidated` here, but it was not as straightforward as hoped. We may need to bump the zarrita version for reasonable support, so I'll create a separate issue.

packages/core/src/data/chunk.ts
andy-sweet: I don't like that much more, so I'll leave it for now, and maybe a better name will materialize as we work through some of the upcoming changes.

PR #381, opened 2025-08-13
Add 3D visibility support to the chunk manager

shlomnissan:
### Summary

This change extends `ChunkManager` to support Z-dimension visibility,
enabling intersection checks in full 3D. A new `makeViewBounds3D()` method
builds a `Box3` view volume from the camera’s 2D bounds and the current Z
index in the region, mapping it to the correct chunk in world space. Chunk
visibility and prefetch checks now operate against 3D bounds using
`Box3.intersects()`.

### Notes
- Prefetching on Z remains disabled (see _`TEMP`_ comments in class). With the current zebra fish dataset, extending Z in both directions would load the entire dataset. Enabling this confidently will require:
  1. Prioritized loading so non-prefetched chunks load first (otherwise this change will slow things down).
  2. Fixed-size logical chunks or temporary logic to prefetch Z only when the point is near a chunk boundary.
- I’m still building confidence in this change. I plan to run more manual tests and possibly add unit tests before merging, but I think it’s ready for initial review.

### Tests & Checks
- manually verified that all examples are working correctly
- manually verified using debug overlay that chunk intersection works as before
- all checks have passed

### DIFF
--- a/packages/core/src/core/chunk_manager.ts
+++ b/packages/core/src/core/chunk_manager.ts
@@ -5,8 +5,9 @@ import {
   LoaderAttributes,
 } from "../data/chunk";
 import { Region } from "../data/region";
-import { vec2 } from "gl-matrix";
+import { vec2, vec3 } from "gl-matrix";
 import { Box2 } from "../math/box2";
+import { Box3 } from "../math/box3";
 import { almostEqual } from "../utilities/almost_equal";
 import { Logger } from "../utilities/logger";
 import { OrthographicCamera } from "@/objects/cameras/orthographic_camera";
@@ -26,7 +27,7 @@ export class ChunkManagerSource {
   private readonly yIdx_: number;
   private readonly zIdx_: number;
   private readonly channelIdx_: number;
-  private lastVisibleBounds_: Box2 | null = null;
+  private lastViewBounds2D_: Box2 | null = null;
 
   constructor(
     loader: ChunkLoader,
@@ -144,10 +145,10 @@ export class ChunkManagerSource {
     return [...lowResChunks, ...currentLODChunks];
   }
 
-  public update(lodFactor: number, visibleBounds: Box2) {
+  public update(lodFactor: number, viewBounds2D: Box2) {
     this.setLOD(lodFactor);
-    if (this.visibleBoundsChanged(visibleBounds)) {
-      this.updateChunkVisibility(visibleBounds);
+    if (this.viewBounds2DChanged(viewBounds2D)) {
+      this.updateChunkVisibility(viewBounds2D);
     }
     this.loadPendingChunks();
   }
@@ -223,7 +224,7 @@ export class ChunkManagerSource {
     }
   }
 
-  private updateChunkVisibility(visibleBounds: Box2): void {
+  private updateChunkVisibility(viewBounds2D: Box2): void {
     if (this.chunks_.length === 0) {
       Logger.warn(
         "ChunkManager",
@@ -232,13 +233,14 @@ export class ChunkManagerSource {
       return;
     }
 
-    const paddedBounds = this.getPaddedBounds(visibleBounds);
+    const viewBounds3D = this.makeViewBounds3D(viewBounds2D);
+    const paddedBounds = this.getPaddedBounds(viewBounds3D);
     for (const chunk of this.chunks_) {
       // TEMP: visibility is 2D (Box2). Ignore nonzero Z until Box3 intersects() exists.
       if (chunk.chunkIndex.z !== 0) continue;
 
       chunk.prefetch = false;
-      chunk.visible = this.isChunkWithinBounds(chunk, visibleBounds);
+      chunk.visible = this.isChunkWithinBounds(chunk, viewBounds3D);
       if (!chunk.visible) {
         chunk.prefetch = this.isChunkWithinBounds(chunk, paddedBounds);
       }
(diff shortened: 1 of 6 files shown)


packages/core/src/core/chunk_manager.ts
```diff
@@ -263,26 +265,65 @@ export class ChunkManagerSource {
...
+
+    const index = this.region_[this.zIdx_].index;
+    if (index.type !== "point") {
+      return new Box3(
+        vec3.fromValues(bounds.min[0], bounds.min[1], 0),
+        vec3.fromValues(bounds.max[0], bounds.max[1], 0)
+      );
+    }
+
+    const shapeZ = this.attrs_[this.currentLOD_].shape[this.zIdx_];
```
andy-sweet: Nit: `zShape` is slightly more readable to me and is consistent with the existing `zIdx`, `zMin`,  & `zMax`. Same for the other names here.

packages/core/src/core/chunk_manager.ts
```diff
@@ -267,26 +269,65 @@ export class ChunkManagerSource {
...
+  private makeViewBounds3D(bounds: Box2): Box3 {
+    if (this.zIdx_ === -1) {
+      return new Box3(
+        vec3.fromValues(bounds.min[0], bounds.min[1], 0),
+        vec3.fromValues(bounds.max[0], bounds.max[1], 0)
+      );
+    }
+
+    const index = this.region_[this.zIdx_].index;
+    if (index.type !== "point") {
```
andy-sweet, line 291: This is the case when z has some range (i.e. it's an interval or it's full), right?

It feels like we could implement this either by using the start/end of an interval or using the full range defined by the attrs. Though I'm not sure we actually hit this code path in any of our examples, so we may have no way to test it.

So it may be fine to leave it. If so, we may want to error for our future selves to recognize that we're not really handling this case properly.

packages/core/src/core/chunk_manager.ts
```diff
@@ -267,26 +269,65 @@ export class ChunkManagerSource {
...
+      return new Box3(
+        vec3.fromValues(bounds.min[0], bounds.min[1], 0),
+        vec3.fromValues(bounds.max[0], bounds.max[1], 0)
+      );
+    }
+
+    const shapeZ = this.attrs_[this.currentLOD_].shape[this.zIdx_];
+    const scaleZ = this.attrs_[this.currentLOD_].scale[this.zIdx_];
+    const transZ = this.attrs_[this.currentLOD_].translation[this.zIdx_];
+    const chunkDepth = this.attrs_[this.currentLOD_].chunks[this.zIdx_];
```
andy-sweet, line 299: Like I mentioned on Slack, in general this is not the `zIdx` we want (since it is an index in the Region not in the Zarr array). But this is consistently incorrect in the chunk manager and should work in the current examples, so no need to change it in this PR.

packages/core/src/core/chunk_manager.ts
```diff
@@ -267,26 +269,65 @@ export class ChunkManagerSource {
...
+        Math.floor(pointZ / chunkDepth),
+        Math.ceil(shapeZ / chunkDepth) - 1
+      )
+    );
+
+    const zMin = transZ + chunkZ * chunkDepth * scaleZ;
+    const zMax = transZ + (chunkZ + 1) * chunkDepth * scaleZ;
+
+    return new Box3(
+      vec3.fromValues(bounds.min[0], bounds.min[1], zMin),
```
andy-sweet: Optional: most of the code here is dedicated to finding `zMin` and `zMax`, but the return statement is effectively the same and is quite verbose, so you could define a function like `getZBounds` that handles the different conditions with this function have a single return. This would make this function much shorter, to the point where you could even inline it as it's only called once.

This is also fine as is though.

andy-sweet (approved):
Looks good and no issues found when testing. All comments and questions are optional.

packages/core/src/core/chunk_manager.ts
shlomnissan: The initial motivation for this check was making TypeScript happy without casting. The "z-interval" code path will never reach this function because chunk manager usage depends on "no intervals" at the moment:

```
const hasIntervals = region.some((r) => r.index.type === "interval");
this.useChunkManager_ =
  !hasIntervals &&
  x?.index.type === "full" &&
  y?.index.type === "full";
```

> we may want to error for our future selves to recognize that we're not really handling this case properly

I will implement this suggestion for correctness, making it clear this case is not supported.

Is there a clear case for supporting intervals in `z` and `t`? I would like to hear your thoughts on how the `Region` type should be structured in the future. I'm not sure this is the right abstraction in the context of the chunk manager or at least we can limit its flexibility to simplify some code paths.

packages/core/src/core/chunk_manager.ts
shlomnissan: Noted. We may want to prioritize correcting this but we can do it in a follow-up.

packages/core/src/core/chunk_manager.ts
shlomnissan: Good call! I was focusing on correctness without thinking too much about the shape of this function.

packages/core/src/core/chunk_manager.ts
andy-sweet: I'm planning to address it soon - may have a PR up tomorrow.

packages/core/src/core/chunk_manager.ts
andy-sweet: > Is there a clear case for supporting intervals in z and t? 

In the ultrack prototype, we had a need to only load data around some time point (i.e. a small subset of the entire time range of a dataset), so it was useful there. In general, I think of the interval as being a virtual crop, potentially being useful for x, y, and z.

One potential small benefit in the context of the chunk manager and prefetching is that a virtual crop could prevent unnecessary prefetching when a chunk is virtually cropped.

I don't think it should be a priority for the chunk manager to support that (e.g. before the end of the year), but it might not be that hard to implement either.

packages/core/src/core/chunk_manager.ts
```diff
@@ -232,13 +233,14 @@ export class ChunkManagerSource {
       return;
     }
 
-    const paddedBounds = this.getPaddedBounds(visibleBounds);
+    const viewBounds3D = this.makeViewBounds3D(viewBounds2D);
+    const paddedBounds = this.getPaddedBounds(viewBounds3D);
     for (const chunk of this.chunks_) {
       // TEMP: visibility is 2D (Box2). Ignore nonzero Z until Box3 intersects() exists.
```
aganders3: I think this is intentionally left for another PR - just a bit confusing because Box3 intersects() does exist.

aganders3 (approved):
Nice work! As mentioned in Slack I still have some questions about where some of these functions will/should live, and I share some of @andy-sweet's comments on naming, etc. but I grok the method here and agree with the direction. Understanding this is all work in active development I have nothing blocking.

packages/core/src/core/chunk_manager.ts
aganders3: I think we're using "region" in two ways as we move to the chunk manager:
* In the layer (as passed to the constructor) it's used as this immutable "virtual crop" to constrain available data
* The chunk manager is using a mutable region to control a currently visible subset of the data (a 2D slice)

I'm not totally convinced we _need_ the virtual crop, but I do think it's a nice-to-have. I still think it feels natural to me to keep these separate concepts and use the chunk manager to compute the intersection of: camera bounds, "virtual crop", and "slice".

shlomnissan: > One potential small benefit in the context of the chunk manager and prefetching is that a virtual crop could prevent unnecessary prefetching when a chunk is virtually cropped

Maybe a better way to address this is making the prefetching range configurable.

> it might not be that hard to implement either

I agree it might not be hard to implement but we want to avoid implementing things that don't provide clear value to reduce complexity, and I'm not convinced a virtual crop is needed as well.

> I still think it feels natural to me to keep these separate concepts and use the chunk manager to compute the intersection of: camera bounds, "virtual crop", and "slice".

I agree with this in the short-term, but I see the chunk manager evolving into something simpler and more generic in the long-term responsible for loading, prioritizing and disposing chunks, similar to Neuroglancer. I think that something else should be responsible for telling the chunk manager what chunks to load, but we can deal with this when we introduce a new type of source data.

packages/core/src/core/chunk_manager.ts
shlomnissan: And that's why I dislike comments so much. It's difficult to keep them up-to-date even with the best of intentions.

PR #382, opened 2025-08-14
Add Stats.js to measure memory usage

shlomnissan:
### Summary

This PR integrates Stats.js to measure memory usage. It is gated by a runtime flag that's disabled by default.

### Notes

- The runtime flag approach is okay for now. A better approach would be generalizing the event dispatcher and extending it to support non-DOM events. This way the runtime can emit `onBeforeRender` and `onAfterRender` events and adding something like `Stats.js` can be the responsibility of the application. This change is non-trivial so I started with a simple flag but I will add a GitHub Issue as a reminder to address this in the future.
- @andy-sweet It looks like `fpsOverlay` is not used in other examples so I removed it. I'm happy to revert this if you want to keep it for now. 

### Tests & Checks

- Manually verified that all the existing examples work as expected.
- I didn't verify the reported memory usage corresponds to the size of the image, but I did verify the memory usage grows as expected when more chunks are stored in memory by enabling prefetching in the z-axis.

https://github.com/user-attachments/assets/511632ed-7a7c-4f6a-b231-5583929ca897


aganders3:
I know I just turned 40, but can we make this a little bigger 😆

packages/core/src/utilities/stats.ts
```diff
@@ -0,0 +1,8 @@
+import Stats from "stats.js";
+
+export function createStats(panel: 0 | 1 | 2 = 0) {
```
phoenixAja: what does 0, 1, and 2 mean? Not immediately clear to me

shlomnissan:
> I know I just turned 40, but can we make this a little bigger

@aganders3 Have you pulled the PR and tested it? It’s small but clear. There’s no simple way to scale it, applying `stats.dom.style.transform = 'scale(2)'` just makes it look pixelated. I could inject CSS for each element or use some hacky workarounds, but I don’t think it’s worth the effort. I’m not sure this request should be a blocker.

packages/core/src/utilities/stats.ts
shlomnissan: It's used for selecting the starting panel (FPS, MS, MB). I removed it from the API for now and added a comment.

aganders3:
It's not a blocker. The scale is also not great because (at least for me) it completely cuts off the text on top.

shlomnissan:
> The scale is also not great because (at least for me) it completely cuts off the text on top

@aganders3 Ah I didn't remove the scaling thing I tried to do. It should be okay now.

aganders3 (approved):
> It’s small but clear.

Don't rub it in!

I'm mostly kidding. It does feel small, but this looks good and works for me.

For what it's worth it's such a small (lol) library that vendoring it doesn't seem crazy, and a quick edit let me scale it 2x while still looking decent.
Image

package.json
```diff
@@ -36,5 +37,8 @@
     "vite-plugin-eslint": "^1.8.1",
     "vite-plugin-glsl": "^1.3.0",
     "vitest": "^3.0.9"
+  },
+  "dependencies": {
```
bchu1: this should be installed in core/package.json?

andy-sweet:
> @andy-sweet It looks like fpsOverlay is not used in other examples so I removed it. I'm happy to revert this if you want to keep it for now.

Totally fine to remove this.

To repeat what I said in the sync, I don't think it would be too hard to introduce before/after callbacks - overlays are really just after callbacks already and could be renamed as such. This would then allow us to avoid introducing a dependency in core Idetik (vendored or not). But this feature does seem useful as a quick way to profile (before getting more detail profiling information from browser dev tools), so I don't want to block it waiting for that approach.

shlomnissan:
@aganders3  I'm going to hold off on merging until you submit your patch for resizing the view

aganders3:
I pushed to this branch in [281ebd4](https://github.com/chanzuckerberg/idetik/pull/382/commits/281ebd4ba1129c1b101685eb2c22c431332c4a7c)

Here's the diff of the changes I made to `stats.js` (plus whitespace changes):
```diff
2a3,4
>  *
>  * Modified to add resizing support via the `scale` parameter.
5c7
< var Stats = function () {
---
> var Stats = function(scale = 1) {
43,44c45,46
< 	var fpsPanel = addPanel( new Stats.Panel( 'FPS', '#0ff', '#002' ) );
< 	var msPanel = addPanel( new Stats.Panel( 'MS', '#0f0', '#020' ) );
---
>   var fpsPanel = addPanel(new Stats.Panel('FPS', '#0ff', '#002', scale));
>   var msPanel = addPanel(new Stats.Panel('MS', '#0f0', '#020', scale));
48c50
< 		var memPanel = addPanel( new Stats.Panel( 'MB', '#f08', '#201' ) );
---
>     var memPanel = addPanel(new Stats.Panel('MB', '#f08', '#201', scale));
112c114
< Stats.Panel = function ( name, fg, bg ) {
---
> Stats.Panel = function(name, fg, bg, scale) {
115c117
< 	var PR = round( window.devicePixelRatio || 1 );
---
>   var PR = scale * round(window.devicePixelRatio || 1);
125c127
< 	canvas.style.cssText = 'width:80px;height:48px';
---
>   canvas.style.cssText = `width:${scale * 80}px;height:${scale * 48}px`;
```

PR #384, opened 2025-08-15
Make box intersection test half-open

shlomnissan:
### Summary

This PR makes box intersections half-open instead of closed intervals.

While working on layer slicing, I removed the temporary skip for non-zero Z-index chunks and disabled Z-axis prefetching. This unexpectedly increased the number of loaded chunks. The cause was our closed-interval intersection test, which marked face-touching chunks as intersecting. I switched to a half-open intersection test instead of adding offsets to the view bounds or chunk bounds since half-open checks are the more common convention in graphics math libraries. The only reason I used closed intervals before is that it was part of the initial implementation. I'm happy to reconsider if there's a good reason to.

### Tests & Checks

- All checks have passed
- Box intersection tests are updated accordingly
- Manual verification of chunk loading across Z

### DIFF
--- a/packages/core/src/math/box2.ts
+++ b/packages/core/src/math/box2.ts
@@ -20,7 +20,7 @@ export class Box2 {
   }
 
   public isEmpty(): boolean {
-    return this.max[0] < this.min[0] || this.max[1] < this.min[1];
+    return this.max[0] <= this.min[0] || this.max[1] <= this.min[1];
   }
 
   // Half-open interval intersection: returns true only if boxes overlap.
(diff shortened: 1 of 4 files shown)


aganders3 (approved):
LGTM

What do you think about changing `isEmpty` to match? That is, `isEmpty` check for either dimension to <= 0. This feels like it would semantically match the intersection test to me (touching edges makes a 0-width "overlap region", which is "empty") - for example if `Box2` gains an `intersectionWith(other: Box2) -> Box2` method.

shlomnissan:
> What do you think about changing isEmpty to match? 

@aganders3 Great catch! I updated `isEmpty` as well.

PR #385, opened 2025-08-15
Move Z-plane slicing to the layer

shlomnissan:
### Summary

This PR modifies the chunk loading logic by storing the entire data set for
the chunk in the chunk data structure and moving the Z-plane slicing from the
loader to the image layer. I also refactored the loader now that I understand
the schema a little better by removing getDimInfoMap and writing a cleaner
helper.

### Tests & Checks

- Manually verified all existing examples work as expected
- Memory reflects the entire chunk size (~2GB without chunk disposal)
- All checks have passed

### DIFF
--- a/packages/core/src/core/chunk_manager.ts
+++ b/packages/core/src/core/chunk_manager.ts
@@ -183,10 +183,6 @@ export class ChunkManagerSource {
     for (const chunk of this.chunks_) {
       if (chunk.lod !== this.lowestResLOD_ || chunk.state !== "unloaded")
         continue;
-
-      // TEMP: visibility is 2D. Ignore nonzero Z until chunk prioritization exists.
-      if (chunk.chunkIndex.z !== 0) continue;
-
       this.loadChunkData(chunk);
     }
   }
@@ -194,7 +190,7 @@ export class ChunkManagerSource {
   private loadChunkData(chunk: Chunk): void {
     chunk.state = "loading";
     this.loader_
-      .loadChunkDataFromRegion(chunk, this.region_)
+      .loadChunkData(chunk, this.region_)
       .then(() => {
         chunk.state = "loaded";
       })
@@ -240,9 +236,6 @@ export class ChunkManagerSource {
 
     const paddedBounds = this.getPaddedBounds(viewBounds3D);
     for (const chunk of this.chunks_) {
-      // TEMP: visibility is 2D. Ignore nonzero Z until chunk prioritization exists.
-      if (chunk.chunkIndex.z !== 0) continue;
-
       chunk.prefetch = false;
       chunk.visible = this.isChunkWithinBounds(chunk, viewBounds3D);
       if (!chunk.visible) {
@@ -338,13 +331,11 @@ export class ChunkManagerSource {
       this.attrs_[this.currentLOD_].chunks[this.yIdx_] *
       this.attrs_[this.currentLOD_].scale[this.yIdx_];
 
-    const chunkDepth =
-      this.attrs_[this.currentLOD_].chunks[this.zIdx_] *
-      this.attrs_[this.currentLOD_].scale[this.zIdx_];
-
     const padX = chunkWidth * PREFETCH_PADDING_CHUNKS;
     const padY = chunkHeight * PREFETCH_PADDING_CHUNKS;
-    const padZ = chunkDepth * PREFETCH_PADDING_CHUNKS;
+
+    // Disable prefetching in Z until chunk prioritization exists.
+    const padZ = 0;
 
     return new Box3(
       vec3.fromValues(
--- a/packages/core/src/data/ome_zarr_image_loader.ts
+++ b/packages/core/src/data/ome_zarr_image_loader.ts
@@ -49,33 +49,62 @@ export class OmeZarrImageLoader {
     this.lods_ = this.metadata_.datasets.length;
   }
 
-  public async loadChunkDataFromRegion(chunk: Chunk, region: Region) {
-    const attrs = this.loaderAttributes_[chunk.lod];
-    const array = this.arrays_[chunk.lod];
+  public buildChunkCoords(
+    chunk: Chunk,
+    region: Region,
+    attrs: LoaderAttributes
+  ) {
+    const dims = attrs.dimensionNames.map((d) => d.toLowerCase());
+    const coords = new Array(dims.length).fill(0);
+    const indices = this.regionToIndices(region, attrs);
 
-    const dimInfo = this.getDimInfoMap(region, chunk, attrs);
-    if (!dimInfo.has("x") || !dimInfo.has("y")) {
-      throw new Error("Missing required spatial axis x/y");
+    for (let dimIdx = 0; dimIdx < dims.length; dimIdx++) {
+      const dim = dims[dimIdx];
+      const chunkSize = attrs.chunks[dimIdx];
+      let coord: number;
+      switch (dim) {
+        case "x":
+          coord = chunk.chunkIndex.x;
+          break;
+        case "y":
+          coord = chunk.chunkIndex.y;
+          break;
+        case "z":
+          coord = chunk.chunkIndex.z ?? 0;
+          break;
+        default: {
+          // Non-spatial dimension (e.g., "t", "c", or any other).
+          // Pull from indices[dimIdx] if present otherwise default to 0.
+          const value = indices[dimIdx];
+          if (typeof value === "number") {
+            coord = Math.floor(value / chunkSize);
+          } else {
+            coord =
+              value.start === null ? 0 : Math.floor(value.start / chunkSize);
+          }
+          break;
+        }
+      }
+      const dimSize = attrs.shape[dimIdx];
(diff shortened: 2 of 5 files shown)


packages/core/src/core/chunk_manager.ts
```diff
@@ -183,18 +183,14 @@ export class ChunkManagerSource {
...
-
       this.loadChunkData(chunk);
     }
   }
 
   private loadChunkData(chunk: Chunk): void {
     chunk.state = "loading";
     this.loader_
-      .loadChunkDataFromRegion(chunk, this.region_)
+      .loadChunkData(chunk, this.region_)
```
andy-sweet, line 193: Great minds [think alike](https://github.com/chanzuckerberg/idetik/pull/387/files#diff-e353e0d834a632ae97f9c7d667271e67826e1c195d7a6d6f497a5021e22f327aR97) :D

packages/core/src/core/chunk_manager.ts
```diff
@@ -338,13 +331,11 @@ export class ChunkManagerSource {
...
-    const chunkDepth =
-      this.attrs_[this.currentLOD_].chunks[this.zIdx_] *
-      this.attrs_[this.currentLOD_].scale[this.zIdx_];
-
     const padX = chunkWidth * PREFETCH_PADDING_CHUNKS;
     const padY = chunkHeight * PREFETCH_PADDING_CHUNKS;
-    const padZ = chunkDepth * PREFETCH_PADDING_CHUNKS;
+
+    // Disable prefetching in Z until chunk prioritization exists.
+    const padZ = 0;
```
andy-sweet, line 338: I noticed that z is not quite handled correctly above (especially when z is not present in the source/region) in #387 so this temporary fix looks good to me!

packages/core/src/layers/image_layer.ts
```diff
@@ -209,12 +209,28 @@ export class ImageLayer extends Layer implements ChannelsEnabled {
     return this.chunkManagerSource_;
   }
 
+  private slicePlane(chunk: Chunk, zValue: number) {
+    const zLocal = Math.round((zValue - chunk.offset.z) / chunk.scale.z);
+    const sliceSize = chunk.shape.x * chunk.shape.y;
+    const offset = sliceSize * (zLocal % chunk.shape.z);
+    return chunk.data?.slice(offset, offset + sliceSize);
```
andy-sweet: If `chunk.data` is undefined, should we just return early rather than do the calculations above?

packages/core/src/data/ome_zarr_image_loader.ts
```diff
@@ -49,33 +49,62 @@ export class OmeZarrImageLoader {
     this.lods_ = this.metadata_.datasets.length;
   }
 
-  public async loadChunkDataFromRegion(chunk: Chunk, region: Region) {
-    const attrs = this.loaderAttributes_[chunk.lod];
-    const array = this.arrays_[chunk.lod];
+  public buildChunkCoords(
```
andy-sweet, line 52: Agreed this is more clear than before. I went for a different approach in #387 that takes a step towards replacing `Region`, and has some overlap with the structure of this function.

andy-sweet (approved):
This overlaps quite a bit with #387, but this importantly handles slicing on the layer. Changes look good and all the examples worked for me, so approved.

packages/core/src/core/chunk_manager.ts
aganders3, line 338: I don't really like the names here that use width/height/depth. I think it's confusing to add yet-more terms for spatial dimensions. Depth in particular may also be confused with channel/color depth.

Sorry though this unrelated to this PR - I just didn't catch it in the previous one. No need to change here but something to think about.

packages/core/src/data/ome_zarr_image_loader.ts
aganders3: Agreed I can follow this much more easily than `dimInfoMap`!

I think there's still an opportunity to clear up some naming here - especially between indices vs. coords, but also between size/length/shape.

I see zarrita's `getChunk` uses `coords` (in fact `chunk_coords`), so maybe we settle on that, but in that case should `chunk.chunkIndex` change?

Sorry to be such a name-nit-picker on this PR!

packages/core/src/layers/image_layer.ts
```diff
@@ -209,12 +209,28 @@ export class ImageLayer extends Layer implements ChannelsEnabled {
     return this.chunkManagerSource_;
   }
 
+  private slicePlane(chunk: Chunk, zValue: number) {
```
aganders3, line 210: It's not blocking, but I'm hesitant about all of this being hard-coded to slice Z. Thinking about orthoslices we will want to be able to configure the slice direction, but once this is working for Z perhaps generalizing will become more clear.

aganders3 (approved):
I think there's a lot of work left to generalize, but this is a step in the right direction for sure. I'm happy to see the slicing on the layer!

packages/core/src/core/chunk_manager.ts
shlomnissan: > I noticed that z is not quite handled correctly above

Did you note any specific issues? I disabled this to temporarily prevent prefetching in Z until we implement prioritized chunk loading but I didn't run into any issues when I tested this.

packages/core/src/layers/image_layer.ts
shlomnissan: I added an early return if `chunk.data` is undefined. We probably want to gate `createImage` as well, but this require updating other code paths and I want to skip a broader refactor.

packages/core/src/core/chunk_manager.ts
shlomnissan: Feedback on naming is important for improving clarity and maintainability. It would be especially helpful if you could share alternative suggestions or examples. “I don’t really like the names here” or  "should chunk.chunkIndex change?" is hard to act on.

As for `chunkDepth` in particular, I think it's clear when located underneath `chunkWidth` and `chunkHeight` but I can't think of better terms for the chunk dimensions. What would you suggest?

packages/core/src/layers/image_layer.ts
shlomnissan: I consider this function as temporary. I haven’t put much thought into the orthoslice implementation. Without a clearer picture, it’s hard to see an obvious path to generalizing this right now. In my view, the challenge with an orthoslice view isn’t rendering multiple viewports (it's a well understood problem with common solutions), but how we integrate into our existing API, and translating/slicing data is a part of that.

packages/core/src/core/chunk_manager.ts
andy-sweet: The calculation of `chunkDepth` will fail when z is not present since indexing with -1 will give a RHS of `undefined * undefined`. I'm really hoping javascript errors at that point...

But as it's gone now, we don't need to worry about it

andy-sweet: Some ideas for more consistent naming are `xChunkSize` or maybe `xChunkLength`.

shlomnissan: Ahh of course! We can gate it properly when we bring it back :+1:

shlomnissan: > Some ideas for more consistent naming are xChunkSize or maybe xChunkLength

We use `chunkWidth`, `chunkHeight`, and `chunkDepth` in the constructor as well. These variables mean the same thing but in local space (we don't multiply them by the scale). We also have `chunksX`, `chunksY`, `chunksZ`. We probably want to rename them to `xChunks`, `yChunks`, for consistency.

I'm going to merge this PR as is. Once #387 is merged, we can put out a small patch for renaming.

shlomnissan: Also, I'm not convinced that `xChunkSize` or `xChunkLength` are better. I think what is missing is the distinction between chunk dimension in local space vs. chunk dimension in world space.

Can you explain how the existing terminology is confusing?

> Depth in particular may also be confused with channel/color depth.

We typically use explicit names for channels, and in the context of intersection or slicing, I assume channels/colors are transparent.

aganders3: > “I don’t really like the names here” or "should chunk.chunkIndex change?" is hard to act on.

Fair enough - I meant more that we should have a discussion about it. It's not specific to this PR either, hence why I don't think it's blocking. I think these terms are inconsistent across the code. As an initial proposal:
* I prefer to use `shape` to indicate number of pixels in any dimension (rather than `size` or `length`). This is biased by my experience with numpy, but they are pretty consistent with it.
* `xChunkShape`, `yChunkShape`, and `zChunkShape` (or `chunkXShape` or `chunkShapeX` but those may be less readable) then would replace `chunkWidth`, `chunkHeight`, and `chunkDepth`. At the moment we're considering the chunks aligned to our world-space axes. Perhaps width/height/depth become useful if that changes, but I'm not sure it will.
* on that note, we should be consistent with using the dimension as a prefix or suffix. I think I prefer prefix.
* `chunk.chunkIndex` should be `chunk.chunkCoords` I actually personally prefer `index` to `coords` here because `index` feels more like an integral value, but it seems reasonable to use `coords` to match zarrita.

PR #386, opened 2025-08-15
refactor: point picking utilities and fix getValueAtWorld

phoenixAja:
Summary

  Addresses PR feedback items by refactoring point picking utilities and fixing value picking functionality for non-chunk manager implementations.

  Changes Made

  - Fix getValueAtWorld for non-chunk manager implementation: Add single ImageRenderable to visibleChunks in load() method so getValueAtWorld can find it regardless
   of which loading path is used
  - Move point_picking.ts to layers directory: Relocated from utilities/ to layers/ since it's specifically intended for use by layers, updating all import paths
  accordingly
  - Centralize PointPickingResult type: Moved from label_image_layer.ts to point_picking.ts module where it logically belongs.
  - Simplify dragThreshold handling: Converted from readonly properties in both ImageLayer and LabelImageLayer to a default parameter (3) in
  handlePointPickingEvent, eliminating code duplication


andy-sweet (approved):
Looks great! Thanks!

PR #387, opened 2025-08-15
Refactor dimension mapping in chunk manager

andy-sweet:
### Summary
This is mostly a refactor that attempts to clarify and fix the dimension mapping in the chunk manager and the OME-Zarr loader. It also brings some validation and mapping on construction of the chunk manager source, which allows us to skip some checks later.

### Tests & Checks
I manually tested all the examples that use the chunk manager.


andy-sweet:
@shlomnissan : I assigned this to you because I'm sure it overlaps a lot with what you're working on right now.

For more context on where this might be going see: https://github.com/chanzuckerberg/idetik/compare/main...andy-sweet/dimension-mapping

andy-sweet:
I could also remove almost all of the "full" region dimensions in the examples and components here if desired (see a [branch](https://github.com/chanzuckerberg/idetik/compare/andy-sweet/chunk-manager-indices-no-full?expand=1) to see what this looks like).

packages/core/src/core/chunk_manager.ts
```diff
@@ -35,57 +32,43 @@ export class ChunkManagerSource {
...
-    this.yIdx_ = region.findIndex(
-      (entry) => entry.dimension.toLowerCase() === "y"
-    );
-    this.zIdx_ = region.findIndex(
-      (entry) => entry.dimension.toLowerCase() === "z"
-    );
-    this.channelIdx_ = region.findIndex(
-      (entry) => entry.dimension.toLowerCase() === "c"
-    );
+    this.dimensions_ = getDimensionMapping(region, this.attrs_[0]);
```
shlomnissan: If the data we need from the attributes object is LOD-independent, it might be better to pass the entire object to avoid confusion, or even better, remove this parameter altogether and access `this.attrs_[0]` from within `getDimensionMapping` so it becomes an implementation detail.

packages/core/src/core/chunk_manager.ts
```diff
@@ -35,57 +32,43 @@ export class ChunkManagerSource {
...
-        this.zIdx_ !== -1
-          ? Math.ceil(this.attrs_[lod].shape[this.zIdx_] / chunkDepth)
-          : 1;
-
-      const channels =
-        this.channelIdx_ >= 0 ? this.attrs_[lod].shape[this.channelIdx_] : 1;
-
-      const scale = this.attrs_[lod].scale;
-      const translation = this.attrs_[lod].translation;
+      const lodAttrs = this.attrs_[lod];
```
shlomnissan: Since you are already creating variables, consider destructuring the object:

```ts
const { chunks, scale, shape, translate } = this.attrs_[lod];
```

packages/core/src/core/chunk_manager.ts
```diff
@@ -384,3 +360,73 @@ export class ChunkManager {
     }
   }
 }
+
+function getDimensionMapping(
+  region: Region,
+  attrs: LoaderAttributes
+): DimensionMap {
+  const names = attrs.dimensionNames;
+  const xIndex = findDimensionIndex(names, "x");
```
shlomnissan: I don't think `findDimensionIndex` and `findDimensionIndexSafe` belong to the chunk manager. I'm not sure what your long-term plan is for `LoaderAttributes`, but if we keep it, we can make it a class type with both functions as class members.

Alternatively, we could change the signatures (e.g., `findDimensionIndex(attrib: LoaderAttributes, target: string)`) and move it the chunk module.

packages/core/src/data/ome_zarr/image_loader.ts
```diff
@@ -249,3 +235,17 @@ function getLoaderAttributes(
...
   return output;
 }
+
+function sliceChunkIndex(dim: SliceDimension, attrs: LoaderAttributes): number {
+  const { sourceIndex } = dim;
+  const dataIndex = sliceDataIndex(dim, attrs);
+  return Math.floor(dataIndex / attrs.chunks[sourceIndex]);
+}
+
+function sliceDataIndex(dim: SliceDimension, attrs: LoaderAttributes): number {
```
shlomnissan: - Consider inlining this function since it's only called once and uses the same parameters as `sliceChunkIndex`. I don't think that separating these functions adds clarity.
- I wonder how these changes in general impact the image loader. `regionToIndices` is an example to a function that share some logic and can benefit from the explicit assumptions you set in this PR around individual axis.

packages/core/src/data/chunk.ts
```diff
@@ -55,6 +55,25 @@ export type Chunk = {
...
+
+export type SliceDimension = {
+  name: string;
+  sourceIndex: number;
+  pointWorld: number;
+};
+
+export type DimensionMap = {
+  x: VisibleDimension;
+  y: VisibleDimension;
```
shlomnissan, line 71: We can probably inline `VisibleDimension` if it's not exported.

shlomnissan (commented):
Looks great! Just a couple of nits/questions. Solid improvement overall!

shlomnissan (approved):

packages/core/src/core/chunk_manager.ts
andy-sweet: Ah, much nicer!

packages/core/src/data/ome_zarr/image_loader.ts
andy-sweet: > Consider inlining this function since it's only called once and uses the same parameters as sliceChunkIndex. I don't think that separating these functions adds clarity.

Done. I used to use it in one other place, but no longer.

> I wonder how these changes in general impact the image loader. regionToIndices is an example to a function that share some logic and can benefit from the explicit assumptions you set in this PR around individual axis.

I considered this, but the chunk manager loading logic still has some restrictions that the loadRegion method does not, so I was planning to leave `loadRegion` alone for now as the relatively simple path that should work and building the chunk manager up to be the more complex path that also works, but much better.

packages/core/src/core/chunk_manager.ts
andy-sweet: After thinking a little about this, I moved `getDimensionMapping` (and the functions it calls) to `OmeZarrImageLoader.getDimensionMap`, so this logic is contained to the OME-Zarr image loader (as some of it is specific to that  - e.g. the default names).

packages/core/src/core/chunk_manager.ts
andy-sweet: I moved this function to a method on the OME-Zarr image loader, which can make its own more specific determination about how to get dimension names that should be consistent across LODs.

PR #388, opened 2025-08-18
Rename and move OME-Zarr and Zarr modules and types

### LINKED ISSUE #366: Refactor OME-Zarr code to handle 3rd party calls and errors

andy-sweet:
### Summary
This change renames existing modules and types to prefer OME-Zarr over OME-NGFF. Before this change, we used both names interchangeably which could be confusing. I decided to use OME-Zarr because it is more specific and consistent with our usage (e.g. `loadOmeZarrImage` assumes the URL points to a Zarr group).

Some of these naming concerns came from comments in #376.

### Related Issue
Closes #366 

### Tests & Checks
I manually tested the examples and React components, which work as they do on the main branch.


packages/core/src/data/ome_zarr/metadata_loaders.ts
```diff
@@ -144,15 +144,15 @@ export async function loadOmeroChannels(
...
+  const image = parseOmeZarrImage(group.attrs);
   return image.omero?.channels ?? [];
 }
 
 export async function loadOmeroDefaultZ(
   source: OmeZarrImageSource
 ): Promise<number> {
   const group = await openGroup(source.location);
-  const image = parseOmeNgffImage(group.attrs);
+  const image = parseOmeZarrImage(group.attrs);
```
andy-sweet, line 155: I was tempted to do some refactoring to the OME-Zarr source/loader so that these metadata are just available because it's silly to fetch and parse them again.

But I think that requires a bit more thought and is out of place in this PR. The fetch and parse is probably quite fast too.

andy-sweet:
@aganders3 : assigned you for review because many of these changes are related to your comments in #376

I renamed the `zarrita` directory to `zarr`, which at least abstracts away from the specific dependency we're using right now. Very open to other ideas.

packages/core/src/data/ome_zarr/metadata_loaders.ts
aganders3: I agree it would be useful, but is out of place here.

Right now we use this for a bunch of complex logic in the React components that I don't like. With slicing moving to the ImageLayer it may make sense to incorporate this there or something.

aganders3 (approved):
This looks great - thanks!

PR #390, opened 2025-08-19
Fix series layer preload to be awaitable

andy-sweet:
### Summary
This change fixes the image and image label series layers `preloadSeries` method to be await-able by returning the promise that the underlying series loader returns.

This bug was introduced by the series loading refactor included in #347.

### Tests & Checks
I manually tested the core examples that use series.
I manually tested the React component examples.

### DIFF
--- a/packages/core/src/layers/image_series_layer.ts
+++ b/packages/core/src/layers/image_series_layer.ts
@@ -96,7 +96,7 @@ export class ImageSeriesLayer extends Layer implements ChannelsEnabled {
   }
 
   public async preloadSeries() {
-    this.seriesLoader_.preloadAllChunks();
+    return this.seriesLoader_.preloadAllChunks();
   }
 
   public get extent(): { x: number; y: number } | undefined {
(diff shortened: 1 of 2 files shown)


andy-sweet:
@bchu1 : added you as a reviewer, because this is more noticeable in the React components which rely more heavily on this method being await-able.

I considered adding some linter checks (e.g. `require-await`), but I'm not sure if they would have actually helped in this case. Feel free to ideas you have used elsewhere.

bchu1 (approved):

bchu1:
> @bchu1 : added you as a reviewer, because this is more noticeable in the React components which rely more heavily on this method being await-able.
> 
> I considered adding some linter checks (e.g. `require-await`), but I'm not sure if they would have actually helped in this case. Feel free to ideas you have used elsewhere.

`require-await` probably would've caught it (either that or something else, I think this is avoidable)

andy-sweet:
> require-await probably would've caught it (either that or something else, I think this is avoidable)

I tried a few lint configs, but they generated quite a few other warnings and errors, so will make a separate issue to set this up.

PR #393, opened 2025-08-19
Add dynamic slicing along Z to chunk manager

shlomnissan:
### Summary

This PR adds dynamic slicing along Z in the context of the chunk manager.

### Notes

- Prefetching along Z remains disabled. I think we'll need to address it separately, since enabling it would significantly increase memory usage and could introduce order-dependent rendering issues.
- The region is currently used not only for cropping but also to trigger reslicing and visible chunk recomputation. While this works as a temporary solution, in the medium-to-long term we’ll need to separate these into two distinct concepts: an initial data crop and a virtual camera.
- Similar to previous PRs, the focus here is on slicing along Z. Eventually, we’ll need to generalize this to support orthoslices across multiple planes, but the right time to address that will be when we integrate orthoslices into Idetik.

### Tests & Checks
- All examples are working as expected
- Manual verification, dynamic slicing works as expected
- Pull the branch and test on your end before reviewing the PR in case I missed anything.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,5 +1,5 @@
 {
-  "name": "idetik",
+  "name": "web-viewer",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
@@ -9,6 +9,9 @@
         "packages/react",
         "packages/ultrack-prototype"
       ],
+      "dependencies": {
+        "lil-gui": "^0.20.0"
+      },
       "devDependencies": {
         "@rollup/plugin-typescript": "^12.1.2",
         "@types/webgl2": "^0.0.11",
@@ -5616,6 +5619,11 @@
         "type": "opencollective",
         "url": "https://opencollective.com/parcel"
       }
+    },
+    "node_modules/lil-gui": {
+      "version": "0.20.0",
+      "resolved": "https://registry.npmjs.org/lil-gui/-/lil-gui-0.20.0.tgz",
+      "integrity": "sha512-k7Ipr0ztqslMA2XvM5z5ZaWhxQtnEOwJBfI/hmSuRh6q4iMG9L0boqqrnZSzBR1jzyJ28OMl47l65ILzRe1TdA=="
     },
     "node_modules/lilconfig": {
       "version": "3.1.3",
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,5 +1,5 @@
 {
-  "name": "web-viewer",
+  "name": "idetik",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
@@ -9,9 +9,6 @@
         "packages/react",
         "packages/ultrack-prototype"
       ],
-      "dependencies": {
-        "lil-gui": "^0.20.0"
-      },
       "devDependencies": {
         "@rollup/plugin-typescript": "^12.1.2",
         "@types/webgl2": "^0.0.11",
@@ -25,6 +22,7 @@
         "eslint-plugin-react-refresh": "^0.4.19",
         "globals": "^15.9.0",
         "jsdom": "^24.1.1",
+        "lil-gui": "^0.20.0",
         "playwright": "1.46.0",
         "prettier": "^3.3.2",
         "rollup": "^4.34.9",
@@ -5623,7 +5621,8 @@
     "node_modules/lil-gui": {
       "version": "0.20.0",
       "resolved": "https://registry.npmjs.org/lil-gui/-/lil-gui-0.20.0.tgz",
-      "integrity": "sha512-k7Ipr0ztqslMA2XvM5z5ZaWhxQtnEOwJBfI/hmSuRh6q4iMG9L0boqqrnZSzBR1jzyJ28OMl47l65ILzRe1TdA=="
+      "integrity": "sha512-k7Ipr0ztqslMA2XvM5z5ZaWhxQtnEOwJBfI/hmSuRh6q4iMG9L0boqqrnZSzBR1jzyJ28OMl47l65ILzRe1TdA==",
+      "dev": true
     },
     "node_modules/lilconfig": {
       "version": "3.1.3",
--- a/package.json
+++ b/package.json
@@ -36,5 +36,8 @@
     "vite-plugin-eslint": "^1.8.1",
     "vite-plugin-glsl": "^1.3.0",
     "vitest": "^3.0.9"
+  },
+  "dependencies": {
+    "lil-gui": "^0.20.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
     "eslint-plugin-react-refresh": "^0.4.19",
     "globals": "^15.9.0",
     "jsdom": "^24.1.1",
+    "lil-gui": "^0.20.0",
     "playwright": "1.46.0",
     "prettier": "^3.3.2",
     "rollup": "^4.34.9",
@@ -36,8 +37,5 @@
     "vite-plugin-eslint": "^1.8.1",
     "vite-plugin-glsl": "^1.3.0",
     "vitest": "^3.0.9"
-  },
-  "dependencies": {
-    "lil-gui": "^0.20.0"
   }
 }
--- a/packages/core/examples/chunk_streaming/index.html
+++ b/packages/core/examples/chunk_streaming/index.html
@@ -49,7 +49,6 @@
 
   <body>
     <canvas id="canvas"></canvas>
-    <div id="chunk-info">Chunks: Loading...</div>
     <script type="module" src="./main.ts"></script>
   </body>
 </html>
--- a/packages/core/examples/chunk_streaming/index.html
+++ b/packages/core/examples/chunk_streaming/index.html
@@ -35,8 +35,8 @@
 
       #chunk-info {
         position: absolute;
-        top: 1em;
-        right: 1em;
+        top: 4.5em;
+        right: 0.3em;
         width: 40%;
         margin: 1em;
         padding: 1em;
@@ -49,6 +49,7 @@
 
   <body>
     <canvas id="canvas"></canvas>
+    <div id="chunk-info">Chunks: Loading...</div>
     <script type="module" src="./main.ts"></script>
   </body>
 </html>
--- a/packages/core/examples/chunk_streaming/main.ts
+++ b/packages/core/examples/chunk_streaming/main.ts
@@ -5,8 +5,9 @@ import {
   OrthographicCamera,
   Region,
 } from "@";
+import { Point } from "@/data/region";
 import { PanZoomControls } from "@/objects/cameras/controls";
-import { ChunkInfoOverlay } from "./chunk_info_overlay";
+import GUI from "lil-gui";
 
 const url =
   "https://public.czbiohub.org/royerlab/zebrahub/imaging/single-objective/ZSNS001.ome.zarr/";
@@ -18,6 +19,7 @@ const bottom = 900;
 // Source is 5D, so provide indices at 3 dimensions to project to 2D.
 // Also specify a subregion in x and y to exercise that part of the API.
 const source = new OmeZarrImageSource(url);
+
 const region: Region = [
   { dimension: "t", index: { type: "point", value: 400 } },
   { dimension: "c", index: { type: "point", value: 0 } },
@@ -25,22 +27,22 @@ const region: Region = [
   { dimension: "y", index: { type: "full" } },
   { dimension: "x", index: { type: "full" } },
 ];
+
+const zIndex = region[2].index as Point;
+const zRange = { min: 0, max: 448 }; // max value taken from source
+const gui = new GUI({ width: 500 });
+gui.add(zIndex, "value", zRange.min, zRange.max, 1).name("Z-index");
+
 const channelProps = [{ contrastLimits: [0, 255] as [number, number] }];
 const camera = new OrthographicCamera(left, right, top, bottom);
-
 const imageLayer = new ImageLayer({ source, region, channelProps });
 imageLayer.debugMode = true;
-
-const chunkInfoOverlay = new ChunkInfoOverlay({
-  textDiv: document.querySelector<HTMLDivElement>("#chunk-info")!,
-  imageLayer: imageLayer,
-});
 
 new Idetik({
   canvas: document.querySelector<HTMLCanvasElement>("#canvas")!,
   camera,
   cameraControls: new PanZoomControls(camera),
   layers: [imageLayer],
-  overlays: [chunkInfoOverlay],
+  overlays: [],
   showStats: true,
 }).start();
--- a/packages/core/examples/chunk_streaming/main.ts
+++ b/packages/core/examples/chunk_streaming/main.ts
@@ -7,6 +7,7 @@ import {
 } from "@";
 import { Point } from "@/data/region";
 import { PanZoomControls } from "@/objects/cameras/controls";
+import { ChunkInfoOverlay } from "./chunk_info_overlay";
 import GUI from "lil-gui";
 
 const url =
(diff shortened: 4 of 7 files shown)


packages/core/examples/chunk_streaming/index.html
```diff
@@ -49,7 +49,6 @@
 
   <body>
     <canvas id="canvas"></canvas>
-    <div id="chunk-info">Chunks: Loading...</div>
```
shlomnissan: I temporarily removed the overlay from this example. I'll bring it back when I'm working on prefetching if needed.

packages/core/examples/chunk_streaming/main.ts
```diff
@@ -18,29 +19,30 @@ const bottom = 900;
...
 const region: Region = [
   { dimension: "t", index: { type: "point", value: 400 } },
   { dimension: "c", index: { type: "point", value: 0 } },
   { dimension: "z", index: { type: "point", value: 300 } },
   { dimension: "y", index: { type: "full" } },
   { dimension: "x", index: { type: "full" } },
 ];
+
+const zIndex = region[2].index as Point;
+const zRange = { min: 0, max: 448 }; // max value taken from source
```
shlomnissan: I hardcoded this value for now. I don't think applications should have access to the loader. A reasonable approach for having access to the shape of the data is using a lifetime hook on the image layer.

packages/core/examples/chunk_streaming/index.html
phoenixAja: you don't like the overlay?

shlomnissan: The overlay is GREAT! I added a slider using `lil-gui` and it was in the way so I removed it temporarily. I will bring it back when I enable prefetching along Z.

Screenshot 2025-08-20 at 9 51 04 AM

package.json
```diff
@@ -36,5 +36,8 @@
     "vite-plugin-eslint": "^1.8.1",
     "vite-plugin-glsl": "^1.3.0",
     "vitest": "^3.0.9"
+  },
+  "dependencies": {
+    "lil-gui": "^0.20.0"
```
andy-sweet: I'd prefer this to be a devDependency either in this package.json or the core package.json.

packages/core/examples/chunk_streaming/index.html
andy-sweet: I've found the chunk info overlay useful for debugging recently and would miss it, so why not just move it to the bottom right or left?

packages/core/examples/chunk_streaming/main.ts
andy-sweet: I don't mind hard-coding in the examples, but this value is incorrect, since the coordinates take into account the scale (and translation) of each multi-resolution level. Since the z-scale is 1.24, this should be [0, 1.24 * 448] and the slider step should be 1.24.

packages/core/src/layers/image_layer.ts
```diff
@@ -210,21 +231,19 @@ export class ImageLayer extends Layer implements ChannelsEnabled {
   private slicePlane(chunk: Chunk, zValue: number) {
     if (!chunk.data) return;
     const zLocal = Math.round((zValue - chunk.offset.z) / chunk.scale.z);
+    const zClamped = Math.max(0, Math.min(zLocal, chunk.shape.z - 1));
```
andy-sweet: I don't think `zLocal` should need to be clamped (or modded for that matter). Did you find some cases when it needed to be?

packages/core/src/core/chunk_manager.ts
```diff
@@ -122,16 +125,29 @@ export class ChunkManagerSource {
 
   public update(lodFactor: number, viewBounds2D: Box2) {
     this.setLOD(lodFactor);
-    if (this.viewBounds2DChanged(viewBounds2D)) {
+    this.dimensions_ = this.loader_.getDimensionMap(this.region_);
```
andy-sweet, line 128: It took me a little while to understand why you were recreating this here, but I guess it's because you're mutating the region object from the main example.

This is probably fine as a short term work around in this PR, but definitely motivates me to think of a different design.

packages/core/src/core/chunk_manager.ts
```diff
@@ -122,16 +125,29 @@ export class ChunkManagerSource {
...
+
+    const zBounds = this.getZBounds();
+
+    if (
+      this.viewBounds2DChanged(viewBounds2D) ||
+      this.zBoundsChanged(zBounds)
+    ) {
       this.updateChunkVisibility(viewBounds2D);
+      this.lastZBounds_ = zBounds;
```
andy-sweet: For symmetry with `viewBounds2DChanged` this assignment should happen in `zBoundsChanged`.

packages/core/src/layers/image_layer.ts
```diff
@@ -210,21 +231,19 @@ export class ImageLayer extends Layer implements ChannelsEnabled {
...
     let data = chunk.data;
-    if (chunk.shape.z > 1) {
-      const zIdx = this.region_.find((r) => r.dimension.toLowerCase() === "z");
-      if (zIdx?.index.type !== "point") {
-        throw new Error("Expected Z index to be a point");
-      }
-      data = this.slicePlane(chunk, zIdx.index.value);
+    const dimensions = this.chunkManagerSource?.dimensions;
+    if (dimensions?.z) {
+      data = this.slicePlane(chunk, dimensions.z.pointWorld);
```
andy-sweet, line 255: I was worried that this would be incorrect (i.e. could be stale), but I think I'm wrong because the dimensions instance is recreated in the  `ChunkManager`, all the chunk visibility is updated accordingly (if anything changed), and the layer polls the visible chunks all the same sync execution. Right?

packages/core/src/core/chunk_manager.ts
phoenixAja: I think i still need to understand how often region will be changing? and the region will only be changing in z? If so could we get a new dimension map only when z changes?

packages/core/src/layers/image_layer.ts
```diff
@@ -119,6 +122,24 @@ export class ImageLayer extends Layer implements ChannelsEnabled {
...
+    const pointWorld = dimensions?.z?.pointWorld;
+    if (pointWorld === undefined || this.zPrevPointWorld_ === pointWorld) {
+      return;
+    }
+
+    for (const [chunk, image] of this.visibleChunks_) {
+      if (chunk.state !== "loaded" || !chunk.data) continue;
+      const data = this.slicePlane(chunk, pointWorld);
+      if (data) {
+        image.textures[0].data = data;
```
phoenixAja, line 136: do we need to worry about deleting old data before reassigning?

packages/core/examples/chunk_streaming/index.html
shlomnissan: 👍

packages/core/src/core/chunk_manager.ts
andy-sweet: Unfortunately we need the map to get the Z-bounds right now, so I think it needs to be here. But in the long term, I think we separate the mapping of dimensions (which should be static) and the slice coordinates of those dimensions (which should be dynamic).

shlomnissan: > This is probably fine as a short term work around in this PR, but definitely motivates me to think of a different design

Yes. It's definitely a short term workaround.

> I think i still need to understand how often region will be changing? and the region will only be changing in z? If so could we get a new dimension map only when z changes?

The region changes when the z-index scroll bar is adjusted.  We can create a `Region` type that is observable and have the chunk manager listen to changes, but I thinks it's overkill because:
- The current structure of `Region` is temporary, as mentioned in PR notes.
- Creating a dimension map from a region takes `<0.1 ms`. At once per frame, this cost is negligible relative to our frame budget, so there’s no need to optimize this path further.

packages/core/src/layers/image_layer.ts
shlomnissan: The old array will be garbage-collected once unreferenced. On the GPU side, this assignment invokes a mutator
```ts
public set data(data: DataTextureTypedArray) {
  this.data_ = data;
  this.needsUpdate = true;
}
```
the next upload overwrites the buffer, so there’s no need to explicitly delete anything.

packages/core/src/layers/image_layer.ts
shlomnissan: Yes.
- chunk manager `update()` is called
- dimensions instance is recreated
- chunk visibility logic is updated accordingly
- the renderer `render()` is called. for each layer:
  - `update()` is called, layer polls for visible chunks
  - `render()` is called

all in the same frame execution.

packages/core/src/core/chunk_manager.ts
shlomnissan: Good catch!

phoenixAja (approved):
besides andy's notes this looks good to me and works on my side, I do want the chunk info overlay back though if you don't mind!

packages/core/src/layers/image_layer.ts
shlomnissan: I haven’t hit a concrete case where zLocal was out of range. The clamp was more of a defensive measure against floating-point roundoff or slider jitter. In theory it shouldn’t be necessary if zValue always comes from within the chunk’s extent.

I'll drop it and add a lightweight sanity check without clamping.

package-lock.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "idetik",
+  "name": "web-viewer",
```
andy-sweet: Nit: I'm not sure I care what the name is in the top-level package-lock, but also unsure why it changed.

packages/core/src/layers/image_layer.ts
```diff
@@ -209,22 +230,23 @@ export class ImageLayer extends Layer implements ChannelsEnabled {
 
   private slicePlane(chunk: Chunk, zValue: number) {
     if (!chunk.data) return;
-    const zLocal = Math.round((zValue - chunk.offset.z) / chunk.scale.z);
     const sliceSize = chunk.shape.x * chunk.shape.y;
-    const offset = sliceSize * (zLocal % chunk.shape.z);
+    const zLocal = Math.round((zValue - chunk.offset.z) / chunk.scale.z);
+    if (zLocal < 0 || zLocal >= chunk.shape.z) {
+      console.warn("zLocal out of bounds", { zLocal, zValue, chunk });
```
andy-sweet:
```suggestion
      Logger.warn("zLocal out of bounds", { zLocal, zValue, chunk });
```

packages/core/src/layers/image_layer.ts
```diff
@@ -209,22 +230,23 @@ export class ImageLayer extends Layer implements ChannelsEnabled {
...
+    }
+
+    const offset = sliceSize * zLocal;
     return chunk.data.slice(offset, offset + sliceSize);
   }
 
   private createImage(chunk: Chunk, channelProps?: ChannelProps[]) {
     const geometry = new PlaneGeometry(chunk.shape.x, chunk.shape.y, 1, 1);
```
aganders3, line 250: I think the geometry should be translated in Z to the world-point of the slice location. This will help coordinate where the planes appear in a 3D orthoslices view, unless we plan to make those separate renderable objects. What do you think?

This can be done below, but I can't comment on the appropriate line (~262) as it's not close enough to a change in this PR:
```
```

This may also affect whether the camera near/far planes and/or position should also change when slicing. If I had to make a decision on this now I think I'd set the camera's Z-position to coincide with the slice position, and set the near/far planes to `+/- zChunkSize/2`.

Sorry I haven't had time to review the full PR, so I'm leaving this as a non-blocking comment for now.

packages/core/src/layers/image_layer.ts
```diff
@@ -209,22 +230,23 @@ export class ImageLayer extends Layer implements ChannelsEnabled {
 
   private slicePlane(chunk: Chunk, zValue: number) {
     if (!chunk.data) return;
-    const zLocal = Math.round((zValue - chunk.offset.z) / chunk.scale.z);
     const sliceSize = chunk.shape.x * chunk.shape.y;
-    const offset = sliceSize * (zLocal % chunk.shape.z);
+    const zLocal = Math.round((zValue - chunk.offset.z) / chunk.scale.z);
+    if (zLocal < 0 || zLocal >= chunk.shape.z) {
```
andy-sweet: I do actually hit this condition when jumping to a larger value of z such that zLocal is non-zero on some non-zero LOD (it's always zero in LOD-0).

I scratched my head for a bit, but I think it's a non-visible slice. The problem is that we slice before updating `ImageLayer.visibleChunks_` based on the new chunk visibility.

packages/core/src/core/chunk_manager.ts
```diff
@@ -296,6 +311,16 @@ export class ChunkManagerSource {
     return changed;
   }
 
+  private zBoundsChanged(newBounds: [number, number]): boolean {
+    const prev = this.lastZBounds_;
+    const changed =
+      !prev || prev[0] !== newBounds[0] || prev[1] !== newBounds[1];
```
andy-sweet: Nit: more concise and easier to read?

```suggestion
    const changed = !prev || !vec2.equals(prev, newBounds);
```

packages/core/src/layers/image_layer.ts
```diff
@@ -94,6 +95,8 @@ export class ImageLayer extends Layer implements ChannelsEnabled {
       this.setState("ready");
     }
 
+    this.resliceIfZChanged();
```
andy-sweet: I think this needs to come after `visibleChunks_` has been updated. See other comment for details.

andy-sweet (changes requested):
Generally looks good and works well. Only blocking comment is around the z-slice and `visibleChunks_` update.

package-lock.json
shlomnissan: Not sure what happened. I couldn't find any references to `web-viewer`.

packages/core/src/layers/image_layer.ts
shlomnissan: I followed your suggestion: clamping but check for delta to avoid hiding potential bugs. I'm comfortable with this approach for now.

packages/core/src/layers/image_layer.ts
```diff
@@ -229,11 +229,24 @@ export class ImageLayer extends Layer implements ChannelsEnabled {
...
+
+    // Allow for small overshoot to account for rounding and
+    // floating point error. Anything larger means the requested
+    // zValue is truly outside the chunk extent.
+    if (zIdx !== zClamped) {
+      const delta = Math.abs(zLocal - zClamped);
+      const eps = 1e-6;
+      if (delta > 1 + eps) {
+        Logger.error("ImageLayer", "slicePlane zValue outside extent");
+      }
```
aganders3: We have an `almostEqual` function in utilities for comparing floats :)

shlomnissan: Ah I forgot about that.

packages/core/src/layers/image_layer.ts
```diff
@@ -209,22 +230,28 @@ export class ImageLayer extends Layer implements ChannelsEnabled {
 
   private slicePlane(chunk: Chunk, zValue: number) {
     if (!chunk.data) return;
-    const zLocal = Math.round((zValue - chunk.offset.z) / chunk.scale.z);
+    const zLocal = (zValue - chunk.offset.z) / chunk.scale.z;
+    const zIdx = Math.round(zLocal);
+    const zClamped = Math.max(0, Math.min(zIdx, chunk.shape.z - 1));
```
andy-sweet: Nit: we have a `clamp` function in `utilities/clamp.ts`

packages/core/src/layers/image_layer.ts
```diff
@@ -209,22 +230,28 @@ export class ImageLayer extends Layer implements ChannelsEnabled {
 
   private slicePlane(chunk: Chunk, zValue: number) {
     if (!chunk.data) return;
-    const zLocal = Math.round((zValue - chunk.offset.z) / chunk.scale.z);
+    const zLocal = (zValue - chunk.offset.z) / chunk.scale.z;
+    const zIdx = Math.round(zLocal);
```
andy-sweet, line 235: After some more thought, I'm pretty confident this should be a `floor` and not a round. But I think that mistake is also present in the `OmeZarrImageLoader`, so this is non-blocking for this PR. I.e. we should fix it clearly and consistently everywhere in another PR, and this is good enough for now to unblock progress.

andy-sweet (approved):
Looks good enough now - thanks for your patience!

PR #394, opened 2025-08-21
Stop Idetik runtime when canvas unmounts

### LINKED ISSUE #314: React Idetik is never stopped
Coming from a bug from the OrganelleBox portal [reported through Sentry](https://czbiohub-sf.sentry.io/issues/6739602056/?environment=vercel-production&project=4509114094321664&query=is%3Aunresolved%20issue.category%3A%5Berror%2Coutage%5D&referrer=issue-stream).

If you look at the detail of that bug, the URL is https://organellebox.sf.czbiohub.org/methods

One way to reproduce the bug is to navigate away from a page showing the Idetik React OME-Zarr viewer (e.g. https://organellebox.sf.czbiohub.org/gene/2025_03_31_A549_ACTB) then navigate to the methods URL by clicking on the link at the top.

The [IdetikProvider](https://github.com/chanzuckerberg/idetik/blob/bab0c632be634dddd18dc2af07720a171d9e6730/packages/react/src/components/providers/IdetikProvider.tsx#L20) effectively calls `Idetik.start` when the canvas is mounted, but a corresponding call to `Idetik.stop` does not seem to be made.

While I would expect a corresponding call to `stop`, I'm not sure when to do that, or how significant that bug is (e.g maybe something else is effectively cancelling the animation loop).

andy-sweet:
### Summary
This changes how the Idetik provider and context works in a way that allows the idetik runtime to be stopped and set to null when its canvas is unmounted.

### Related Issue
Closes #314

### Tests & Checks
The existing component examples work as expected.
Wrote new React component unit tests for IdetikCanvas.
I tested commit 6a0c55be31adfc5c7fe8a396e1d49e3633781daf integrated into a local dev version of the [organelle-box-portal](https://github.com/czbiohub-sf/organelle-box-portal).


andy-sweet:
@bchu1 : I could do with your React expertise here, especially because I'm a little unfamiliar with the provider mechanism and logic. This is a draft because I'm not sure this is the right way to handle this issue and because I haven't tested that this fixes the bug.

packages/core/src/utilities/logger.ts
```diff
@@ -28,7 +28,7 @@ type Module =
 
 export class Logger {
   private static logLevel_: LogLevel =
-    import.meta.env.MODE === "production" ? "warn" : "debug";
+    import.meta.env.MODE === "production" ? "info" : "debug";
```
andy-sweet: I'm not convinced we need to filter out logs here at all, but if we do I'd expect info logs to pass the default level in production. Unrelated to this PR, though the start/stop logs are useful for a quick a sanity check.

bchu1: Yeah, I don't really see the point of this logger, since we're a library. Would probably be more useful for us to expose callbacks for different severity events or something so clients can monitor our errors.

packages/react/src/components/providers/IdetikProvider.tsx
```diff
@@ -25,6 +26,15 @@ export const IdetikProvider = ({ children }: PropsWithChildren) => {
     },
   });
 
+  useEffect(() => {
+    return () => {
+      console.debug("IdetikProvider unmounting");
+      if (idetikContext.runtime) {
```
bchu1: can also do `idetikContext.runtime?.stop()`

bchu1 (approved):
Kinda surprised we didn't have this 😅 

probably makes more sense to be the cleanup function of the `ref` callback in `<IdetikCanvas>` since it's common to put providers in layouts or places where they will never get unmounted.

bchu1:
> Kinda surprised we didn't have this 😅
> 
> probably makes more sense to be the cleanup function of the `ref` callback in `<IdetikCanvas>` since it's common to put providers in layouts or places where they will never get unmounted.

Actually I'm not sure if React 18 has the cleanup function for `ref`s... https://18.react.dev/reference/react-dom/components/common#ref-callback-parameters

andy-sweet:
> Actually I'm not sure if React 18 has the cleanup function for refs... https://18.react.dev/reference/react-dom/components/common#ref-callback-parameters

Yeah, that's what I was looking for, but looks like it's in canary for 18. Maybe in 19?

andy-sweet:
Would another option be to call `Idetik.stop` in the [callback](https://github.com/chanzuckerberg/idetik/blob/759b8df49dbf65ca5d30e030ce82491099325a08/packages/react/src/components/IdetikCanvas.tsx#L19) that binds to the canvas ref in `IdetikCanvas`, in the case that the canvas is null and the runtime isn't?

From the React docs

> When that \<div\> DOM node is removed, React will call your ref callback with null.

bchu1:
> > Actually I'm not sure if React 18 has the cleanup function for refs... https://18.react.dev/reference/react-dom/components/common#ref-callback-parameters
> 
> Yeah, that's what I was looking for, but looks like it's in canary for 18. Maybe in 19?

No idea how canaries work in React 😞

bchu1:
> Would another option be to call `Idetik.stop` in the [callback](https://github.com/chanzuckerberg/idetik/blob/759b8df49dbf65ca5d30e030ce82491099325a08/packages/react/src/components/IdetikCanvas.tsx#L19) that binds to the canvas ref in `IdetikCanvas`, in the case that the canvas is null and the runtime isn't?
> 
> From the React docs
> 
> > When that <div> DOM node is removed, React will call your ref callback with null.

Ooh, yeah that sounds good!

bchu1:
> > Would another option be to call `Idetik.stop` in the [callback](https://github.com/chanzuckerberg/idetik/blob/759b8df49dbf65ca5d30e030ce82491099325a08/packages/react/src/components/IdetikCanvas.tsx#L19) that binds to the canvas ref in `IdetikCanvas`, in the case that the canvas is null and the runtime isn't?
> > From the React docs
> > > When that  DOM node is removed, React will call your ref callback with null.
> 
> Ooh, yeah that sounds good!

Actually... if the parent of `<IdetikCanvas>` gets unmounted I'm not sure that works, does it still get called?

andy-sweet:
> Actually... if the parent of `<IdetikCanvas>` gets unmounted I'm not sure that works, does it still get called?

I was in the middle of manually testing this, and I think it does not get called. But I got distracted by something else... Will confirm tomorrow.

.github/workflows/lint-test-build.yml
```diff
@@ -29,7 +29,10 @@ jobs:
           npx playwright install chromium
 
       - name: Compile TypeScript
-        run: npm run compile --workspace=@idetik/core-prerelease
+        run: |
+          npm run compile --workspace=@idetik/core-prerelease
+          npm run build --workspace=@idetik/core-prerelease
```
andy-sweet, line 34: This is needed for compiling the react package and for running the tests. It's a big ugly here, but I think these two packages are going to split soon anyway, so I think we can live with it.

packages/react/.gitignore
```diff
@@ -0,0 +1 @@
+dist
```
andy-sweet, line 1: This seems to be needed for prettier to ignore dist in the react package. I'm not sure why the top-level .gitignore doesn't handle this, but it's the same for the core package.

packages/react/test/IdetikCanvas.test.tsx
```diff
@@ -0,0 +1,81 @@
+import { describe, it, expect, beforeEach, vi } from "vitest";
+import { render, cleanup } from "@testing-library/react";
+import { IdetikCanvas } from "../src/components/IdetikCanvas";
+import { IdetikProvider } from "../src/components/providers/IdetikProvider";
+import { Idetik } from "@idetik/core-prerelease";
+
+vi.mock("@idetik/core-prerelease", () => ({
```
andy-sweet, line 7: I used a mock for a few reasons, but I'm open to changing.

1. It's the easiest way to check Idetik.stop/start. I think I could use some custom layer or overlay to check them too, but I'm not convinced it's worth it here.
2. Keeps these tests fast and lightweight.

packages/react/vite.config.js
```diff
@@ -77,5 +77,12 @@ export default defineConfig(() => {
         include: [path.resolve(_dirname, "src/**")],
       },
     },
+    test: {
+      environment: "jsdom",
+      coverage: {
```
andy-sweet, line 82: I can add a browser config here, but it means the test runner hangs with the following warning

```
close timed out after 10000ms
Tests closed successfully but something prevents the main process from exiting
You can try to identify the cause by enabling "hanging-process" reporter. See https://vitest.dev/config/#reporters
```

The tests pass, and the core tests give the same warning. So I could add the browser config in if desired.

andy-sweet:
@bchu1 : re-requested a review because quite a lot changed here, and I think this is the first time I've written React tests (so may need some guidance!). I left some inline comments to highlight a few points of low confidence and/or potential confusion.

> > > Would another option be to call `Idetik.stop` in the [callback](https://github.com/chanzuckerberg/idetik/blob/759b8df49dbf65ca5d30e030ce82491099325a08/packages/react/src/components/IdetikCanvas.tsx#L19) that binds to the canvas ref in `IdetikCanvas`, in the case that the canvas is null and the runtime isn't?
> > > From the React docs
> > > > When that  DOM node is removed, React will call your ref callback with null.
> > 
> > 
> > Ooh, yeah that sounds good!
> 
> Actually... if the parent of `<IdetikCanvas>` gets unmounted I'm not sure that works, does it still get called?

I believe my automated and manual tests confirm this approach works as intended. Note that in the OrganelleBox, the provider is actually a [component](https://github.com/czbiohub-sf/organelle-box-portal/blob/c871aea85029f762b93b1f1449ec1c8aea319303/components/OmeZarrImageViewer/GeneImageViewer.tsx#L68) that gets mounted and unmounted, so this also covers that case.

packages/react/src/components/providers/IdetikProvider.tsx
```diff
@@ -6,31 +6,36 @@ import {
...
 
 /** Global Idetik state provider that you must wrap your application in. */
 export const IdetikProvider = ({ children }: PropsWithChildren) => {
-  const [idetikContext, setIdetikContext] = useState<IdetikContextValue>({
-    isReady: false,
-    runtime: null,
-    initializeWithCanvas: (canvas: HTMLCanvasElement) => {
+  const [runtime, setRuntime] = useState<Idetik | null>(null);
+
+  const onCanvasChange = (canvas: HTMLCanvasElement | null) => {
```
bchu1: would help to name this something that indicates it's supposed to be passed to a `ref`, maybe `canvasRefCallback` or something?

bchu1 (approved):

packages/react/src/components/providers/IdetikProvider.tsx
andy-sweet: Done.

andy-sweet:
This is back to draft to prevent accidental merging because I still don't understand the possible orderings here. For example, I want to error in the case that two `IdetikCanvas` components are used under the same provider (not supported), but that causes other issues.

PR #395, opened 2025-08-21
Fix ImageLayer not marking any chunks visible for 2D images

bchu1:
`isChunkWithinBounds()` was comparing chunk z range `[0, 1]` to visible bounds `[0, 0]`. Since we're assuming absence of `z` means the shape is `1`, make default z bounds `[0, 1]`.


phoenixAja:
did you test with the examples?

bchu1:
> did you test with the examples?

Yep

phoenixAja (approved):

PR #396, opened 2025-08-21
Fix image layer object render order

### LINKED ISSUE #389: ImageLayer is not guaranteed to present visible and loaded higher-resolution chunks
`ChunkManager.getChunks` returns chunks ordered by the LOD in increasing value (i.e. low-res before high-res).

`ImageLayer.updateChunks` processes the chunks is that order. However, the order in which it creates renderable objects is dependent on when the chunks are asynchronously loaded.

I think that if low-resolution chunks are loaded after higher-resolution chunks, then their renderable objects are added to the layer later and will effectively clobber the render in the related regions of the viewport.

When the chunks have been previously fetched and cached by the browser, I will sometimes see this bug.

Image

andy-sweet:
### Summary
This fixes a render issue related to visible chunks at different LODs described in the related issue. Credit to @aganders3, who effectively suggested this fix in Slack.

### Related Issue
Closes #389

### Tests & Checks
I manually tested the chunk streaming example before and after this fix with the browser cache enabled.

### DIFF
--- a/packages/core/src/layers/image_layer.ts
+++ b/packages/core/src/layers/image_layer.ts
@@ -103,9 +103,8 @@ export class ImageLayer extends Layer implements ChannelsEnabled {
     // creating new texture objects. Note: GPU resources are not currently being
     // released, so this will also need to be addressed soon.
     const currentChunks = new Set(this.chunkManagerSource_.getChunks());
-    this.visibleChunks_.forEach((image, chunk) => {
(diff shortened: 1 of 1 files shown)


andy-sweet:
@shlomnissan : added you for review since this is partly @aganders3 's code and I think you had another solution here, so want to check this was sufficient.

shlomnissan (approved):
LGTM 👍

PR #400, opened 2025-08-25
Dynamic slice coordinates and a fixed dimension map for chunked images

andy-sweet:
### Summary
This change further splits the layers and data structures used for chunk streaming loading, away from the older/cruder approach of loading a single fixed region.

The main motivation behind this change is to separate the concerns of `DimensionMap`, which currently maps from Idetik dimensions to OME-Zarr array indices and also stores slicing coordinates.

The approach of this change is to define a new `SliceCoordinates` type that optionally stores the z, c, and t coordinates that should be used to index a particular point or index in each of those dimensions. This effectively replaces the use of `Region` in the examples that use chunk streaming.

Then also define a `SourceDimensionMap` that handles both the dimension mapping and loader attributes, and is intended to be immutable.

### Tests & Checks
I manually tested the existing components.


packages/core/src/core/chunk_manager.ts
```diff
@@ -20,54 +19,47 @@ const PREFETCH_PADDING_CHUNKS = 1;
 export class ChunkManagerSource {
   private readonly chunks_: Chunk[];
   private readonly loader_;
-  private readonly attrs_: ReadonlyArray<LoaderAttributes>;
```
andy-sweet, line 23: `ChunkDimensionMap` stores all the information that `LoaderAttributes` in a slightly more ergonomic way for most usage (i.e. dimension first, rather than LOD-first), so we can remove the attributes here. We could consider removing it everywhere, but it is currently used in multiple applications.

packages/core/src/data/chunk.ts
```diff
@@ -55,23 +55,33 @@ export type Chunk = {
...
+  translation: number;
 };
 
-export type DimensionMap = {
-  x: VisibleDimension;
-  y: VisibleDimension;
-  z?: SliceDimension;
-  c?: SliceDimension;
-  t?: SliceDimension;
+export type SliceCoordinates = {
```
andy-sweet, line 88: I think the simplest way to extend this to work with some relatively simple ortho-slices approach (viewports or not) is to add x and y here also as optional numbers. Then it would be the layer/viewport that would be responsible for pulling out the desired coordinate need for the loading it wants to trigger.

packages/core/src/data/chunk.ts
```diff
@@ -94,9 +104,9 @@ export type ChunkLoader = {
     scheduler?: PromiseScheduler
   ): Promise<Chunk>;
 
-  getDimensionMap(region: Region): DimensionMap;
+  getDimensionMap(): ChunkDimensionMap;
 
-  loadChunkData(chunk: Chunk, mapping: DimensionMap): Promise<void>;
+  loadChunkData(chunk: Chunk, region: SliceCoordinates): Promise<void>;
```
aganders3: Maybe rename this for clarity? Sorry I'm still reading so not sure of the best name here.
```suggestion
  loadChunkData(chunk: Chunk, sliceCoords: SliceCoordinates): Promise<void>;
```

packages/core/src/data/chunk.ts
aganders3: As long as slices are aligned with the (spatial) axes, this might we well represented by a tagged union like:
```
export type xySlice = { orientation: "XY"; z?: SliceDimension, c?: SliceDimension, t?: SliceDimension };
export type yzSlice = { orientation: "YZ"; x?: SliceDimension, c?: SliceDimension, t?: SliceDimension };
export type xzSlice = { orientation: "XZ"; y?: SliceDimension, c?: SliceDimension, t?: SliceDimension };
export type SliceCoordinates = xySlice | yzSlice | xzSlice;
```

packages/core/src/core/chunk_manager.ts
```diff
@@ -20,54 +19,47 @@ const PREFETCH_PADDING_CHUNKS = 1;
...
-  ) {
+  constructor(loader: ChunkLoader, sliceCoords: SliceCoordinates) {
     this.loader_ = loader;
-    this.attrs_ = attrs;
-    this.region_ = region;
-    this.lowestResLOD_ = attrs.length - 1;
+    this.dimensions_ = this.loader_.getDimensionMap();
+    this.lowestResLOD_ = this.dimensions_.numLods - 1;
     this.currentLOD_ = 0;
+    this.sliceCoords_ = sliceCoords;
```
aganders3, line 34: I know we said `ChunkManagerSource` is probably doing too much right now, so maybe this is temporary, but I was not expecting it to maintain `sliceCoords_` at all.

I would think separate layers sharing this `ChunkManagerSource` would own separate (though perhaps dependent in the case of linked OrthoSlices) `SliceCoordinates` and pass them in as-needed when requesting chunks.

I think either model works assuming `ChunkManagerSource` gets split or the initial empty chunk creation/ownership gets transferred to another class (existing -- like `ChunkManager` -- or new).

aganders3: Reading some more, I think this is only used to (conditionally) slice in `c` and `t` in `loadChunkData` (on the _loader_). Should this information live on each chunk, then? I'm thinking ahead to sharing across viewports with different `t`.

packages/core/src/core/chunk_manager.ts
```diff
@@ -352,12 +336,11 @@ export class ChunkManagerSource {
 export class ChunkManager {
   private readonly sources_ = new Map<ChunkSource, ChunkManagerSource>();
 
-  public async addSource(source: ChunkSource, region: Region) {
+  public async addSource(source: ChunkSource, sliceCoords: SliceCoordinates) {
```
aganders3, line 355: This is repeating what I've said above, but I think it's well exhibited here:
I find it odd that the `ChunkManager` stores these keyed only by `source`, but this function and the `ChunkManagerSource` constructor both take `sliceCoords`.

packages/core/src/data/ome_zarr/image_loader.ts
```diff
@@ -279,13 +257,57 @@ function getLoaderAttributes(
...
 }
 
-function sliceChunkIndex(dim: SliceDimension, attrs: LoaderAttributes): number {
-  const { sourceIndex, pointWorld } = dim;
-  const { scale, translation } = attrs;
-  const dataIndex = Math.round(
-    (pointWorld - translation[sourceIndex]) / scale[sourceIndex]
-  );
-  return Math.floor(dataIndex / attrs.chunks[sourceIndex]);
+function getChunkDimensionMap(
```
aganders3: It might be premature but I'd call this `inferChunkDimensionMap` and leave room for the loader to take an explicit (optional) `ChunkDimensionMap` instead.

packages/core/src/data/chunk.ts
```diff
@@ -55,23 +55,33 @@ export type Chunk = {
...
+  unit?: string;
+  lods: ChunkDimensionLod[];
+};
+
+export type ChunkDimensionLod = {
+  size: number;
+  chunkSize: number;
+  scale: number;
+  translation: number;
 };
```
aganders3: I like this! Do we need `Chunk` in the names here? In my head I think of these as `DimensionMap` and `SourceDimension` but that's a nitpick.

packages/core/src/data/chunk.ts
```diff
@@ -55,23 +55,33 @@ export type Chunk = {
...
 
-export type DimensionMap = {
-  x: VisibleDimension;
-  y: VisibleDimension;
-  z?: SliceDimension;
-  c?: SliceDimension;
-  t?: SliceDimension;
+export type SliceCoordinates = {
+  z?: number;
+  c?: number;
```
aganders3, line 90: Does this constrain chunks in general to be 3D (single channel, single timepoint)? If so we can remove `Chunk.shape.c`, right?

packages/core/src/objects/textures/channel.ts
aganders3, line 1: This is great, but I suspect not necessary for the other changes and could be in a separate PR? I don't mind reviewing at the same time if it's how the `ImageChunkLayer` is implemented. There's no need to go backwards then forwards there. I'm just checking my understanding.

aganders3 (commented):
I left a few comments/questions but this looks a lot like what I had in mind. Thanks!

packages/core/src/core/chunk_manager.ts
aganders3: I like this. I think I created `LoaderAttributes` but it was a quick-and-dirty pretty direct map to the OME-Zarr metadata, and I agree `ChunkDimensionMap` is more ergonomic (and feels more generic/universal).

aganders3:
One more thing to clarify here:
I think `SliceCoords` is now in Idetik world space, is that correct? This is a difference between `Region`, which uses arbitrary dimensions based on source data dimension names.

packages/core/src/objects/textures/channel.ts
andy-sweet: Yes, I got a bit carried away with the refactoring here, which is really to assist the ImageLayer split. That's why this is a draft that's not intended to be directly reviewed then merged.

packages/core/src/data/ome_zarr/image_loader.ts
andy-sweet: Yeah, I would something like that to handle the case when the axes are not named as expected. I've never run into that, but it is possible.

There is a naming problem here (of course!) because the map here is between Idetik dimensions (xyztc) and zarr array indices, whereas the map you're describing is between Idetik dimensions and OME-Zarr image axes.

packages/core/src/data/chunk.ts
andy-sweet: Sorry, I missed this one!

packages/core/src/data/chunk.ts
andy-sweet: Maybe not. The only piece of metadata that is chunk-specific is chunkSize.

packages/core/src/core/chunk_manager.ts
andy-sweet: Agreed.

Currently, we need the ortho-camera xy-bounds, LOD, and slice coordinates to determine what chunks to load. The Idetik runtime passes the camera and canvas buffer width through to ChunkManager.update, which then passes the inferred LOD and ortho-camera xy-bounds through to ChunkManagerSource.update. But Idetik does not know about the slice coordinates, so cannot pass them through similarly.

The current solution for chunk streaming is to have the slider component own the region, which it then mutates to update z. The layer takes a reference to that region on construction, then passes it on to the ChunkManagerSource here on construction via `ChunkManager.addSource`. The ChunkManagerSource updates the dimension map with the new mutated values of region to work out what chunks to fetch. This has at least two downsides: 1) the in-place mutation is distant and hard to reason about; 2) the dimension map is mutable.

This PR improves one of these (2) by defining the light-weight `SliceCoordinates` and allows for an immutable dimension map. However, it does not improve (1).

One solution to improve (1) from @shlomnissan was to create `SliceCoordinates` in the ChunkManagerSource and then mutate it from the example via the layer when it's available. But I think this does not solve your concern here.

Another solution is along the lines of what you described, which is pass everything through to ChunkManagerSource (or the thing that loads and caches chunks). This is my preferred solution right now.

andy-sweet: > Reading some more, I think this is only used to (conditionally) slice in `c` and `t` in `loadChunkData` (on the _loader_). Should this information live on each chunk, then? I'm thinking ahead to sharing across viewports with different `t`.

That's right. If the source has dimensions that are not handled by the chunk manager source (i.e. c and t), then we need to index/integrate them out when loading. Storing them in the chunk info sounds reasonable to me. I think this is also similar to your idea for store the z-translation in the chunk and slice so that we can position the corresponding plane in space accordingly.

packages/core/src/core/chunk_manager.ts
andy-sweet: Agreed. This PR does not attempt to solve that problem *yet*.

packages/core/src/data/chunk.ts
andy-sweet: If it is present, then it does constrain chunks as you described. If it is not present, but the source has C, then chunk streaming path will error since it cannot handle that case right now - I might expect that to not error in the future, depending on related design choices.

But more importantly, the old `loadRegion` path also returns a `Chunk` that may be multi-channel.

Also, in the future I might expect the type of this value to be something like `number | number[]` to handle selecting multiple channels - again depending on related design choices.

andy-sweet:
> One more thing to clarify here:
> I think `SliceCoords` is now in Idetik world space, is that correct? This is a difference between `Region`, which uses arbitrary dimensions based on source data dimension names.

Correct. That is another key difference between being the single chunk `loadRegion` loading path and the chunk manager loading path.

andy-sweet:
@aganders3 , @shlomnissan 

This is ready for review now since it's now smaller and I think represents an improvement over main. That improvement is limited, so I think it is fine to push back on this for now if we think there are related changes that would make ortho-slices much easier.

I'm going to work off this branch to see there's a way to satisfy some of the other concerns/potential improvements (e.g. ownership of slice coordinates, splitting concerns of chunk manager source).

packages/core/src/core/chunk_manager.ts
```diff
@@ -194,7 +184,7 @@ export class ChunkManagerSource {
   }
 
   private setLOD(lodFactor: number): void {
-    const maxLOD = this.attrs_.length - 1;
+    const maxLOD = this.dimensions_.numLods - 1;
```
aganders3:
```suggestion
    const maxLOD = this.lowestResLod_;
```

packages/core/src/core/chunk_manager.ts
```diff
@@ -20,54 +19,47 @@ const PREFETCH_PADDING_CHUNKS = 1;
...
-    for (let lod = 0; lod < this.attrs_.length; ++lod) {
-      const { chunks, scale, shape, translation } = this.attrs_[lod];
-      const chunkWidth = chunks[xIdx];
-      const chunkHeight = chunks[yIdx];
-      const chunkDepth = chunks[zIdx] ?? 1;
+    for (let lod = 0; lod < this.dimensions_.numLods; ++lod) {
+      const xDim = this.dimensions_.x.lods[lod];
+      const yDim = this.dimensions_.y.lods[lod];
+      const zDim = this.dimensions_.z?.lods[lod];
+      const cDim = this.dimensions_.c?.lods[lod];
```
aganders3: nit: below when accessing the `ChunkDimensionLod` this way you use `zLod` - consider matching the naming here

packages/core/src/core/chunk_manager.ts
```diff
@@ -286,15 +275,14 @@ export class ChunkManagerSource {
...
-    const zScale = lodAttrs.scale[zIdx];
-    const zTran = lodAttrs.translation[zIdx];
-    const zPoint = Math.floor((zDim.pointWorld - zTran) / zScale);
-    const chunkDepth = lodAttrs.chunks[zIdx];
+    const zLod = zDim.lods[this.currentLOD_];
+    const zShape = zLod.size;
+    const zScale = zLod.scale;
+    const zTran = zLod.translation;
+    const zPoint = Math.floor((this.sliceCoords_.z - zTran) / zScale);
+    const chunkDepth = zLod.chunkSize;
```
aganders3, line 285: Take-or-leave idea to help readability (and check my understanding)

edit: suggestion looks wrong here, but the idea is just to bind `const zSlice = this.sliceCoords_.z;` for clarity and reuse.

packages/core/src/core/chunk_manager.ts
```diff
@@ -337,15 +325,11 @@ export class ChunkManagerSource {
...
 
   private getPaddedBounds(bounds: Box3): Box3 {
-    const xIdx = this.dimensions_.x.sourceIndex;
-    const yIdx = this.dimensions_.y.sourceIndex;
-
-    const attrs = this.attrs_[this.currentLOD_];
-    const chunkWidth = attrs.chunks[xIdx] * attrs.scale[xIdx];
-    const chunkHeight = attrs.chunks[yIdx] * attrs.scale[yIdx];
+    const xDim = this.dimensions_.x.lods[this.currentLOD_];
+    const yDim = this.dimensions_.y.lods[this.currentLOD_];
```
aganders3: Again when indexing into `lods` I'd consider naming these `xLod` and `yLod`.

packages/core/src/core/chunk_manager.ts
aganders3: I created [an issue](https://github.com/chanzuckerberg/idetik/issues/417) for this because I think it's a good idea to remove it everywhere, but if it's used in other applications it makes sense to do that more carefully in a separate PR.

packages/core/src/data/ome_zarr/image_loader.ts
```diff
@@ -45,74 +47,50 @@ type OmeZarrImageLoaderProps = {
...
-  private readonly lods_: number;
   private readonly loaderAttributes_: ReadonlyArray<LoaderAttributes>;
+  private readonly dimensions_: ChunkDimensionMap;
 
   constructor(props: OmeZarrImageLoaderProps) {
     this.metadata_ = props.metadata;
     this.arrays_ = props.arrays;
     this.loaderAttributes_ = getLoaderAttributes(this.metadata_, this.arrays_);
-    this.lods_ = this.metadata_.datasets.length;
+    this.dimensions_ = getChunkDimensionMap(this.loaderAttributes_);
```
aganders3: Not blocking, but consider renaming this to avoid confusion with the below `getDimensionMap` member function. I still like "infer" here, but maybe `getDimensionsFromSourceMetadata` is a better description of what it's doing.

packages/core/src/data/ome_zarr/image_loader.ts
```diff
@@ -45,74 +47,50 @@ type OmeZarrImageLoaderProps = {
...
-      };
-    }
-
-    return mapping;
+  public getDimensionMap(): ChunkDimensionMap {
+    return this.dimensions_;
   }
 
-  public async loadChunkData(chunk: Chunk, mapping: DimensionMap) {
+  public async loadChunkData(chunk: Chunk, sliceIndices: SliceCoordinates) {
```
aganders3: nit: rename `sliceIndices` as `sliceCoords` to match its type

packages/core/src/data/ome_zarr/image_loader.ts
```diff
@@ -45,74 +47,50 @@ type OmeZarrImageLoaderProps = {
...
     const array = this.arrays_[chunk.lod];
```
aganders3: super-nit: I think it would be more readable to move this line below the construction of `chunkCoords` and just above `const subarray = await array.getChunk(chunkCoords);`

packages/core/src/data/ome_zarr/image_loader.ts
```diff
@@ -45,74 +47,50 @@ type OmeZarrImageLoaderProps = {
...
-    const attrs = this.loaderAttributes_[chunk.lod];
 
     const chunkCoords: number[] = [];
-    chunkCoords[mapping.x.sourceIndex] = chunk.chunkIndex.x;
-    chunkCoords[mapping.y.sourceIndex] = chunk.chunkIndex.y;
-    if (mapping.z) {
-      chunkCoords[mapping.z.sourceIndex] = chunk.chunkIndex.z;
+    chunkCoords[this.dimensions_.x.index] = chunk.chunkIndex.x;
```
aganders3, line 66: nit: In other functions `this.dimensions_.x.index` et al. are bound to `xIdx`, `yIdx`, and `zIdx`, which I like.

aganders3 (approved):
This looks good, with lots of nits for naming. Any of them I'm happy to push off to a later cleanup/refactor. For now I'd really prefer these changes or similar - very open to other ideas here:

`ChunkDimensionMap` -> `DimensionMap`
`ChunkDimension` -> `SourceDimension` or `DataDimension`
`ChunkDimensionLod` -> `SourceDimensionLod` or `DimensionLodMetadata`

It may also be nice to add comments (docstrings?) to these types to explain what they do/intended use. Here's my crack at that:

```
/**
 * Maps Idetik spatial (world) dimensions (XYZ) and non-spatial dimensions (C and T) to dimensions 
 * in the source data by name and index, including dimension attributes and LOD metadata.
 */
DimensionMap

/**
 * Describes a single dimension in the source data, including its name, index position,
 * optional unit suffix, and metadata for each level of detail in the image pyramid.
 */
DataDimension

/**
 * Metadata for a single dimension at a specific level of detail (LOD).
 * For example, combines zarr array metadata (size, chunkSize) with OME-zarr coordinate 
 * transforms (scale, translation).
 */
DimensionLodMetadata

/**
 * Coordinates for selecting a specific slice (2D array) in Idetik's coordinate space (XYZCT).
 * Used to specify which z-plane, channel, and/or timepoint to display.
 * Values are transformed coordinates (scaled and translated), not array indices.
 */
SliceCoordinates
```

Otherwise this looks great! I'm approving since as-is I think this is a step in the right direction, but I'm happy to chat about any of this before merging if you'd like.

shlomnissan:
I skimmed through the changes and they look good overall. I will take a deeper look later but don't let it block.

packages/core/src/core/chunk_manager.ts
andy-sweet: Done.

packages/core/src/core/chunk_manager.ts
andy-sweet: Done.

packages/core/src/data/ome_zarr/image_loader.ts
andy-sweet: Infer is great here. Done.

packages/core/src/data/ome_zarr/image_loader.ts
andy-sweet: Done.

packages/core/src/data/ome_zarr/image_loader.ts
andy-sweet: That's not the case anymore (due to this PR), and I don't think giving these local names improves readability enough to warrant it. Vertical space matters too!

packages/core/src/data/ome_zarr/image_loader.ts
andy-sweet: Done.

packages/core/src/core/chunk_manager.ts
andy-sweet: Leave. `zSlice` is more ambiguous (it could be the result of slicing) whereas `sliceCoords.z` is more explicit.

andy-sweet:
> This looks good, with lots of nits for naming.

I mostly used your names and docstrings with some adaptations: `SourceDimensionMap`, `SourceDimension`, `SourceDimensionLod`. I think the inclusion of `Source` because it helps clarify what the name/index refer to. I don't love `Data` and `Metadata` in class names, since they usually don't tell me much.

packages/core/src/data/chunk.ts
andy-sweet: Resolving as I mostly adopted your name suggestions.

andy-sweet:
@shlomnissan : I merged this since I'm out next week, and I think it could be helpful for work while I'm out. Feel to leave comments and I'll address them when I'm back. Or feel free to make changes yourself!

PR #402, opened 2025-08-25
fix: prioritize current LOD chunks in getValueAtWorld()

phoenixAja:
## Summary
Fix pixel value picking to return high-resolution values when both high-res and low-res chunks contain the same world position

  ## Problem
  The `getValueAtWorld()` method in `ImageLayer` was returning low-resolution pixel values even when high-resolution chunks were  available. This happened because the method iterated through `visibleChunks_` and returned the first matching chunk, but low-res  fallback chunks could be encountered before current LOD chunks.

  ## Solution
  - Split chunk iteration into two passes: current LOD chunks first, then fallback to low-res chunks
  - Extract common chunk value extraction logic into `getValueFromChunk()` helper method
  - Maintain the same fallback behavior while prioritizing accuracy

## Test:
- enabled point picking in chunk streaming example to verify that point picking worked at multiple LODs and that the correct chunks were being used to extract a value.

### DIFF
--- a/packages/core/src/layers/image_layer.ts
+++ b/packages/core/src/layers/image_layer.ts
@@ -277,33 +277,56 @@ export class ImageLayer extends Layer implements ChannelsEnabled {
     return image;
   }
 
-  public getValueAtWorld(world: vec3): number | null {
-    // Iterate through all visible chunks to find the one containing the world position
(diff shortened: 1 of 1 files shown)


andy-sweet (approved):
Looks good. Thanks!

PR #403, opened 2025-08-26
Define distinct ChunkedImageLayer for chunk streaming

andy-sweet:
### Summary
This splits the existing image layer into two classes.

1. `ImageLayer`: this encapsulates the pre-chunk-streaming image layer, which is crude but simple. This is currently used by the main React component and the applications that use that.

2. `ChunkedImageLayer`: this uses chunk-streaming for a better experience and performance, as well as built-in support for image series, but at the cost of some complexity. This does not support some features of the `ImageLayer` (e.g. multiple channels). This will eventually replace at least the `ImageLayer` and `ImageSeriesLayer`.

The main motivation of this change is clarity. Since `ChunkedImageLayer` has a very different loading pattern, the existing image layer had become hard to read and maintain. This also allows us to move more quickly on developing chunk streaming without breaking existing usage of older layers.

Going forward, we intend to leave `ImageLayer` and `ImageSeriesLayer` alone except for important maintenance, and will eventually delete them in favor of `ChunkedImageLayer`. At the point, we may want to rename `ChunkedImageLayer`.

I updated all the examples that I'm aware of using chunk streaming to use the `ChunkedImageLayer`.

### Tests & Checks
I manually tested the existing examples and React components.

### DIFF
--- a/packages/core/src/layers/chunked_image_layer.ts
+++ b/packages/core/src/layers/chunked_image_layer.ts
@@ -1,0 +1,224 @@
+import { Layer, LayerOptions } from "../core/layer";
+import { IdetikContext } from "../idetik";
+import { Chunk, ChunkSource } from "../data/chunk";
+import { ChunkManagerSource } from "../core/chunk_manager";
+import { ChannelProps } from "../objects/textures/channel";
+import { ImageRenderable } from "../objects/renderable/image_renderable";
+import { Texture2DArray } from "../objects/textures/texture_2d_array";
+import { PlaneGeometry } from "../objects/geometry/plane_geometry";
+import { Logger } from "../utilities/logger";
+import { Color } from "../core/color";
+import { EventContext } from "../core/event_dispatcher";
+import { vec2, vec3 } from "gl-matrix";
+import { handlePointPickingEvent, PointPickingResult } from "./point_picking";
+import { almostEqual } from "../utilities/almost_equal";
+import { clamp } from "../utilities/clamp";
+import { Region } from "../data/region";
+
+export type ChunkedImageLayerProps = LayerOptions & {
+  source: ChunkSource;
+  region: Region;
+  channelProps?: ChannelProps;
+  onPickValue?: (info: PointPickingResult) => void;
+};
+
+export class ChunkedImageLayer extends Layer {
+  public readonly type = "ChunkedImageLayer";
+
+  private readonly source_: ChunkSource;
+  private readonly region_: Region;
+  private readonly onPickValue_?: (info: PointPickingResult) => void;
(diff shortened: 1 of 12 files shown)


shlomnissan:
> I updated all the examples that I'm aware of using chunk streaming to use the ChunkedImageLayer.

I thought the chunk streaming example is the only one

packages/core/src/layers/chunked_image_layer.ts
```diff
@@ -0,0 +1,224 @@
...
+      const y = Math.floor(localPos[1]);
+
+      // Check if this chunk contains the requested position
+      if (x >= 0 && x < chunk.shape.x && y >= 0 && y < chunk.shape.y) {
+        const dimensions = this.chunkManagerSource_.dimensions;
+
+        const data =
+          dimensions?.z !== undefined
+            ? this.slicePlane(chunk, dimensions.z.pointWorld)!
+            : chunk.data;
```
shlomnissan: With this change, we create a copy of the sliced data:
https://github.com/chanzuckerberg/idetik/pull/404#discussion_r2301645176

Is slicing twice necessary? Can we use the underlying texture data via the renderable object?

This is a non-blocking comment because it's not directly related to the change, but if there's a way to improve this, let's add a comment or create an issue before moving on.

shlomnissan (approved):
LGTM! I left one non-blocking comment :)

packages/core/src/layers/chunked_image_layer.ts
andy-sweet: Ideally, we could use the texture data here, but that requires typecasting, which I'd like to avoid. A simple fix is to store the sliced chunk data in this layer (that's what I do in the plain ImageLayer for similar reasons). There are likely better fixes (e.g. properly typing the first texture in the ImageRenderable which is other just a plain Texture). There is no change from main here, so I'll create an issue to make sure we don't lose this.

shlomnissan: Why does it require type casting? Textures are accessible through the base renderable class and each texture has a `data()` getter

andy-sweet:
> I thought the chunk streaming example is the only one

I think there are three examples that ended up using chunk streaming based on the logic in the old ImageLayer's constructor, which are the three I updated here to explicitly use the new ChunkImageLayer.

packages/core/src/layers/chunked_image_layer.ts
andy-sweet: That data could be a `TexImageSource`, but I guess we can just skip if that's the case (which it won't be with any of our layers). I'll still leave this for a separate PR to minimize and expedite changes here.

PR #404, opened 2025-08-26
Delete unused chunk data from memory

shlomnissan:
### Summary
This PR updates chunk disposal logic in `updateChunkVisibility`
to ensure only the current LOD and fallback LOD keep CPU data.
All other LODs are disposed even if they are visible. Prefetching
is now limited to the current LOD and unloaded chunks. Fallback
LOD chunks remain pinned and never disposed.

### Tests & Checks
- Examples are working as expected
- Manual verification:
    - Use stats to track memory
    - Use chunk overlay to track number of "ready chunks"
    - Add log when chunk is disposed and ensure correct behavior
    - **NOTE** It may take the GC 10-20 seconds before it releases memory

https://github.com/user-attachments/assets/0a3bc656-6d66-40e0-b8fd-eacc5576ea9e

### DIFF
--- a/packages/core/src/core/chunk_manager.ts
+++ b/packages/core/src/core/chunk_manager.ts
@@ -161,7 +161,7 @@ export class ChunkManagerSource {
     for (const chunk of this.chunks_) {
       if (
         chunk.lod === this.currentLOD_ &&
-        chunk.state === "unloaded" &&
+        (chunk.state === "unloaded" || chunk.state === "disposed") &&
         (chunk.visible || chunk.prefetch)
       ) {
         this.loadChunkData(chunk);
@@ -226,10 +226,32 @@ export class ChunkManagerSource {
 
     const paddedBounds = this.getPaddedBounds(viewBounds3D);
     for (const chunk of this.chunks_) {
-      chunk.prefetch = false;
-      chunk.visible = this.isChunkWithinBounds(chunk, viewBounds3D);
-      if (!chunk.visible) {
-        chunk.prefetch = this.isChunkWithinBounds(chunk, paddedBounds);
+      const isVisible = this.isChunkWithinBounds(chunk, viewBounds3D);
+      const inPrefetchRegion =
+        !isVisible && this.isChunkWithinBounds(chunk, paddedBounds);
+
+      const isCurrent = chunk.lod === this.currentLOD_;
+      const isFallback = chunk.lod === this.lowestResLOD_;
+      const isLoaded = chunk.state === "loaded";
+
+      chunk.visible = isVisible;
+      chunk.prefetch = isCurrent && inPrefetchRegion && !isLoaded;
+
+      /*
+        Disposal rules:
+        - Keep fallback LOD pinned.
+        - For current LOD dispose if not visible AND not in prefetch region.
+        - For any other LOD dispose regardless.
+      */
+      if (isLoaded && !isFallback) {
+        const shouldDispose =
+          !isCurrent || (isCurrent && !isVisible && !inPrefetchRegion);
+
+        if (shouldDispose) {
+          chunk.data = undefined;
+          chunk.state = "disposed";
+          Logger.debug("ChunkManager", `Disposing chunk in LOD ${chunk.lod}`);
+        }
       }
     }
   }
--- a/packages/core/src/core/chunk_manager.ts
+++ b/packages/core/src/core/chunk_manager.ts
@@ -161,7 +161,7 @@ export class ChunkManagerSource {
     for (const chunk of this.chunks_) {
       if (
         chunk.lod === this.currentLOD_ &&
-        (chunk.state === "unloaded" || chunk.state === "disposed") &&
+        chunk.state === "unloaded" &&
         (chunk.visible || chunk.prefetch)
       ) {
         this.loadChunkData(chunk);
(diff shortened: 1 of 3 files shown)


packages/core/src/layers/image_layer.ts
```diff
@@ -248,7 +248,7 @@ export class ImageLayer extends Layer implements ChannelsEnabled {
 
     const sliceSize = chunk.shape.x * chunk.shape.y;
     const offset = sliceSize * zClamped;
-    return chunk.data.subarray(offset, offset + sliceSize);
+    return chunk.data.slice(offset, offset + sliceSize);
```
shlomnissan: Turns out `subarray` creates a reference to the view and `slice` creates a copy.

I learned this the hard way, finally got Chrome to max memory.

andy-sweet: Ah, I thought of the view as an advantage in terms of no extra memory needed. But I think I understand the need for the change here - because of the ImageRenderable's texture holding keeping the data alive, right?

In some cases (i.e. when the chunk has some z-depth), the slice should be an order of magnitude smaller than the chunk, so this is good for now.

packages/core/src/core/chunk_manager.ts
```diff
@@ -226,10 +226,32 @@ export class ChunkManagerSource {
 
     const paddedBounds = this.getPaddedBounds(viewBounds3D);
     for (const chunk of this.chunks_) {
-      chunk.prefetch = false;
-      chunk.visible = this.isChunkWithinBounds(chunk, viewBounds3D);
-      if (!chunk.visible) {
-        chunk.prefetch = this.isChunkWithinBounds(chunk, paddedBounds);
+      const isVisible = this.isChunkWithinBounds(chunk, viewBounds3D);
+      const inPrefetchRegion =
+        !isVisible && this.isChunkWithinBounds(chunk, paddedBounds);
```
andy-sweet: Nit: the RHS is a combo of visibility and being in the padded bounds, so consider a different name.

```suggestion
      const shouldPrefetch =
        !isVisible && this.isChunkWithinBounds(chunk, paddedBounds);
```

packages/core/src/core/chunk_manager.ts
```diff
@@ -226,10 +226,32 @@ export class ChunkManagerSource {
...
+        - Keep fallback LOD pinned.
+        - For current LOD dispose if not visible AND not in prefetch region.
+        - For any other LOD dispose regardless.
+      */
+      if (isLoaded && !isFallback) {
+        const shouldDispose =
+          !isCurrent || (isCurrent && !isVisible && !inPrefetchRegion);
+
+        if (shouldDispose) {
+          chunk.data = undefined;
```
andy-sweet, line 245: Nit: we might want to consider renaming the method this is in from `updateChunkVisibility` to `updateChunkState` or similar given this does more than updating the visibility. That was true before this PR with the potential prefetch change though, so not necessary to address it in this PR.

I guess the changes are always related to visibility, so maybe the names usefully indicates that though.

packages/core/src/core/chunk_manager.ts
```diff
@@ -226,10 +226,32 @@ export class ChunkManagerSource {
...
+        !isVisible && this.isChunkWithinBounds(chunk, paddedBounds);
+
+      const isCurrent = chunk.lod === this.currentLOD_;
+      const isFallback = chunk.lod === this.lowestResLOD_;
+      const isLoaded = chunk.state === "loaded";
+
+      chunk.visible = isVisible;
+      chunk.prefetch = isCurrent && inPrefetchRegion && !isLoaded;
+
+      /*
```
andy-sweet: Nit: with the boolean variable names, I found it relatively easy just to read the conditions, so I didn't find this comment particularly helpful and would consider removing. Don't feel strongly though.

packages/core/src/data/chunk.ts
```diff
@@ -26,7 +26,7 @@ export function isChunkData(value: unknown): value is ChunkData {
 
 export type Chunk = {
   data?: ChunkData;
-  state: "unloaded" | "loading" | "loaded";
+  state: "unloaded" | "loading" | "loaded" | "disposed";
```
andy-sweet: The "disposed" state currently has no functional difference to "unloaded" now, right? I.e. it distinguishes a chunk from one that was previously "loaded" from one that was "never" loaded. I don't mind that semantic distinction, but am curious if you had some thoughts about how those states might be treated differently somewhere.

andy-sweet (approved):
No blocking comments and works well for me!

andy-sweet (changes requested):
Actually, I think I just caught an issue here in `ChunkManagerSource.loadLowResChunks`, which checks `"unloaded"` but not `"disposed"`, so requesting changes to block a merge with that in while I/we think about it.

andy-sweet:
> Actually, I think I just caught an issue here in `ChunkManagerSource.loadLowResChunks`, which checks `"unloaded"` but not `"disposed"`, so requesting changes to block a merge with that in while I/we think about it.

Ah wait, we never dispose the low res chunks, so I think this doesn't matter in practice, though it's slightly confusing. Unblocking shortly.

andy-sweet (approved):
Approving again now that I'm not concerned about the low-res chunks needing to be reloaded after disposal (since they cannot be disposed).

packages/core/src/layers/image_layer.ts
shlomnissan: > because of the ImageRenderable's texture holding keeping the data alive, right?

That's right. The texture will hold a reference to it, and since we're not explicitly deleting textures, the entire data array will never be deleted in the `.subarray` case.

packages/core/src/core/chunk_manager.ts
shlomnissan: I don't think `shouldPrefetch` is right. Prefetching is now predicated on the chunk not being loaded and in the current LOD:
```ts
chunk.prefetch = isCurrent && inPrefetchRegion && !isLoaded;
```
I renamed it to `eligibleForPrefetch` and reordered the booleans in the condition above.

packages/core/src/core/chunk_manager.ts
shlomnissan: > I guess the changes are always related to visibility

I considered renaming it and reached a similar conclusion.

I think `updateChunkState` reduces clarity in this case. Maybe the solution is refactoring: separating visibility test and disposal logic. I'm going to leave it as is for now, but I will continue to think about this.

packages/core/src/data/chunk.ts
shlomnissan: I thought of using "disposed" as a flag to move the image renderable back to the pool but it might be better to check for "unloaded" and avoid redundancy in this stage, or at least add it when it becomes necessary. I will revert this change.

PR #407, opened 2025-08-26
Multiple Viewports: WebGLState and Renderer updates for `gl.viewport` and `gl.scissor`

aganders3:
### Summary
This PR adds viewport and scissor test management to the WebGL renderer, along with utilities to support rendering to specific regions of the canvas.

1. Added `setViewport(box: Box2)` method to `WebGLState` that accepts a `Box2` region
2. Added `setScissor(box: Box2)` method to `WebGLState` to set the scissor test box
3. Added `setScissorTest(bool)` method to `WebGLState` to enable the scissor test
4. Modified `Renderer.render()` interface and `WebGLRenderer.render()` to require a `viewportBox: Box2` parameter for region-specific rendering. Currently passing a box that matches the full canvas.
5. Added `Box2.floor()` and `Box2.toRect()` methods to convert `Box2` from min/max format to integer x/y/width/height format needed by WebGL APIs

The render loop in the Idetik runtime does not yet take advantage of these new features to render into sub-regions of the the canvas. The runtime API has not changed, so examples and applications using the runtime do not need to be updated, but the Ultrack example (to be removed) needed a small update.

This is step one of the implementation plan in the [Multiple Viewports Tech Spec](https://docs.google.com/document/d/1OwGL7KW34hxFAENyOm2WiQ71E58XGu9Y7jKkJlyRx84/edit?tab=t.0#heading=h.g0elbcs8pp3f).

Tagging @phoenixAja for review because you've probably worked on WebGLState the most - also you said you have some time :).

### Related Issue
N/A

### Tests & Checks
Manually tested examples
Tests pass, with some tests added for new functionality


phoenixAja (approved):
looks good!

packages/core/src/core/renderer.ts
```diff
@@ -26,7 +27,12 @@ export abstract class Renderer {
     }
   }
 
-  public abstract render(layerManager: LayerManager, camera: Camera): void;
+  // TODO: make this just take a Viewport instance once implemented
```
shlomnissan, line 30: What's preventing us from introducing a `Viewport` type at this point?

packages/core/src/math/box2.ts
```diff
@@ -29,4 +29,12 @@ export class Box2 {
     if (a.max[1] <= b.min[1] || a.min[1] >= b.max[1]) return false;
     return true;
   }
+
+  public asXYWH(): { x: number; y: number; width: number; height: number } {
```
shlomnissan: In my experience `getSize()` or `toRect()` are typically more common.

packages/core/src/renderers/WebGLState.ts
```diff
@@ -1,3 +1,8 @@
+import { Box2 } from "../math/box2";
+import { vec2 } from "gl-matrix";
+
+type XYWH = { x: number; y: number; width: number; height: number };
```
shlomnissan: This is out of place and defined twice. A better approach would be creating a `Rect` type at a higher level and use it across the board. Or you might be able to get away with using `Box2` and `getSize()`.

packages/core/src/renderers/WebGLState.ts
```diff
@@ -13,11 +18,19 @@ export class WebGLState {
...
+  private currentViewport_: XYWH | null = null;
+  private currentScissor_: XYWH | null = null;
 
   constructor(gl: WebGL2RenderingContext) {
     this.gl_ = gl;
   }
 
+  private static xywhEquals(a: XYWH, b: XYWH): boolean {
+    return (
+      a.x === b.x && a.y === b.y && a.width === b.width && a.height === b.height
```
shlomnissan: This is also out of place and could be a member function of `Rect`.

packages/core/src/core/renderer.ts
```diff
@@ -26,7 +27,12 @@ export abstract class Renderer {
...
+  public abstract render(
+    layerManager: LayerManager,
+    camera: Camera,
+    viewportBox?: Box2
```
shlomnissan: Can we make the `viewportBox` non-optional? The runtime can be responsible for creating it.

packages/core/src/renderers/WebGLState.ts
```diff
@@ -94,4 +107,73 @@ export class WebGLState {
...
+  public setViewport(viewportBox?: Box2) {
+    const viewport =
+      viewportBox ??
+      new Box2(
+        vec2.fromValues(0, 0),
+        vec2.fromValues(this.gl_.canvas.width, this.gl_.canvas.height)
+      );
+
+    const { x, y, width, height } = viewport.asXYWH();
+    const flooredXYWH = {
```
shlomnissan: Following previous comments, consider `viewport.clamp().getSize()` or `viewport.getRect().clamp()`.

packages/core/src/renderers/WebGLState.ts
```diff
@@ -94,4 +107,73 @@ export class WebGLState {
...
+    this.gl_.viewport(
+      flooredXYWH.x,
+      flooredXYWH.y,
+      flooredXYWH.width,
+      flooredXYWH.height
+    );
+    this.currentViewport_ = flooredXYWH;
+  }
+
+  public setScissor(scissorBox?: Box2) {
```
shlomnissan: I think we should be more explicit. Maybe add `disableScissor` or instead expect clients to pass `null`?

shlomnissan (changes requested):
I think `XYWH` logic in the `WebGLState` is misplaced. Extending `Box2` with `getSize` or introducing a simple `Rect` type might be a better approach. I would also like to see if we can make things more explicit by removing optional parameters.

andy-sweet:
Will take a look tomorrow, but don't wait for me if you have the feedback you need to merge!

packages/core/src/core/renderer.ts
aganders3: It's just to make the PR smaller. I intend for this to take a non-optional `Viewport` as the only parameter once that is introduced.

packages/core/src/renderers/WebGLState.ts
aganders3: Thanks for this idea - `clamp()` returning a new `Box2` and `getSize()` seems nice to me. I was trying to isolate the clamping here as it's not used elsewhere, but I like this better.

packages/core/src/core/renderer.ts
aganders3: My intention is to introduce a `Viewport` object that comprises a `LayerManager`, a `Camera`, and an HTML element (that gives the `viewportBox` needed here). That will be a required parameter, so this is only temporary.

packages/core/src/math/box2.ts
aganders3: I guess I prefer `toRect` of those names since it's both position and size.

packages/core/src/renderers/WebGLState.ts
aganders3: I was trying to reuse `Box2` to not create another very similar abstraction (`Rect`) but maybe they're different enough to warrant both. `Rect` doesn't necessarily imply "integer valued" to me but I suppose neither does this.

aganders3:
Thanks for the feedback - I think it's cleaner now. I removed the optional parameters and added some functionality directly to `Box2` instead of maintaining it in WebGLState.

packages/core/src/core/renderer.ts
aganders3: Though I expect this to change, I made the parameter non-optional and manage a single `viewportBox` in the runtime now.

packages/core/src/idetik.ts
```diff
@@ -149,7 +150,10 @@ export class Idetik {
         this.renderer_.updateSize();
         this.needsResize_ = false;
       }
-      this.renderer_.render(this.layerManager, this.camera);
+
+      // TEMP: in the future, the renderer will manage a list of dynmaically-sized viewports
```
```suggestion
      // TEMP: in the future, the renderer will manage a list of dynamically-sized viewports
```

packages/core/vite.config.js
```diff
@@ -69,8 +69,12 @@ export default defineConfig(({ mode }) => {
       browser: {
         enabled: true,
         provider: "playwright",
-        name: "chromium", // browser name is required
         headless: true,
+        instances: [
```
andy-sweet, line 73: Was this change required? Or did you temporarily test other browsers out?

packages/ultrack-prototype/frontend/components/Renderer.tsx
```diff
@@ -86,8 +86,16 @@ export default function Renderer({
     let lastRequestId = 0;
     const canvas = document.querySelector<HTMLCanvasElement>(`#${canvasId}`)!;
     const renderer = new WebGLRenderer(canvas);
+    let needsResize = false;
+    new ResizeObserver(() => {
+      needsResize = true;
+    }).observe(canvas);
```
andy-sweet: I realize this is a fix in ultrack that is mostly unrelated to the changes in this PR, but I given it's here, I think you should also store the observer and call disconnect or unobserve in the cleanup function that's returned.

packages/core/src/math/box2.ts
```diff
@@ -29,4 +29,28 @@ export class Box2 {
...
+  public static equals(a: Box2, b: Box2): boolean {
+    return (
+      a.min[0] === b.min[0] &&
+      a.min[1] === b.min[1] &&
+      a.max[0] === b.max[0] &&
+      a.max[1] === b.max[1]
+    );
+  }
+
+  public clamp(): Box2 {
```
andy-sweet: Why isn't this called floor? If I think of a box with a clamp method, I think it would take another box as an argument and clamps its values according to the min/max here.

packages/core/test/webgl_state.test.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  const viewport = new Box2(vec2.fromValues(10, 20), vec2.fromValues(110, 220));
+
+  webglState.setViewport(viewport);
+
+  const actualViewport = gl.getParameter(gl.VIEWPORT);
+  expect(Array.from(actualViewport)).toEqual([10, 20, 100, 200]);
+});
+
+test("setViewport with floating point values floors to integers", () => {
```
andy-sweet, line 29: Is there any noticeable effect at the borders of the divs related to this behavior? E.g. if the divs effectively define floating point coordinates, do we see unrendered/previously-rendered gaps between the canvas and the div border?

packages/core/src/core/renderer.ts
```diff
@@ -62,6 +69,13 @@ export abstract class Renderer {
     return this.height_;
   }
 
+  public get box(): Box2 {
```
andy-sweet: This feels relatively silly and unneeded as a public property on a relatively high-level important object like the renderer. Also, it's only really used in `WebGLRenderer.render`, so why not just create it there from width/height?

Optional: I assume this is anticipating future changes, so no need to change if you think this is worth keeping.

packages/core/src/math/box2.ts
```diff
@@ -29,4 +29,28 @@ export class Box2 {
...
     return true;
   }
+
+  public static equals(a: Box2, b: Box2): boolean {
+    return (
+      a.min[0] === b.min[0] &&
+      a.min[1] === b.min[1] &&
+      a.max[0] === b.max[0] &&
+      a.max[1] === b.max[1]
+    );
```
andy-sweet: Nit: marginally more readable.

```suggestion
    return vec2.exactEquals(a.min, b.min) && vec2.exactEquals(a.max, b.max);
```

packages/core/src/renderers/webgl_renderer.ts
```diff
@@ -52,11 +53,19 @@ export class WebGLRenderer extends Renderer {
...
     this.textures_ = new WebGLTextures(this.gl);
-    this.resize(this.canvas.width, this.canvas.height);
     this.state_ = new WebGLState(this.gl);
+    this.resize(this.canvas.width, this.canvas.height);
   }
 
-  public render(layerManager: LayerManager, camera: Camera) {
+  public render(layerManager: LayerManager, camera: Camera, viewportBox: Box2) {
+    this.state_.setViewport(viewportBox);
+    if (Box2.equals(viewportBox.clamp(), this.box.clamp())) {
```
andy-sweet: Is it possible for the renderer's box to have non-integer values? Maybe due to the device pixel ratio?

packages/core/src/core/renderer.ts
andy-sweet: I really want to see a Viewport type, but also OK with this PR as a smaller setup for that.

andy-sweet (approved):
Looks good - just a few optional comments and requests.

andy-sweet:
Oh, and update the PR description please!

packages/core/src/math/box2.ts
shlomnissan: I agree `clamp` is not a good name. It was my suggestion, but I didn't think too much about it.

I would prefer something like `toIntegerBounds` but `floor` or `snapToGrid` also works.

packages/core/src/core/renderer.ts
shlomnissan: 100% agree! Avoid it altogether if possible. If it is worth keeping, I would suggest renaming it since `box` is too generic. `bounds` or `extent` might be better names.

packages/core/src/core/renderer.ts
shlomnissan: Same! I don't expect a minimal `Viewport` type to add complexity since the renderer is scoped to the runtime, but I'm okay with leaving it.

shlomnissan (approved):
LGTM :+1:

packages/core/vite.config.js
aganders3: I should have commented - it's in the commit message there but this is just to silence a deprecation warning from vitest. I think it's mostly a name change in the config as far as we're concerned.

packages/core/src/renderers/webgl_renderer.ts
aganders3: Unfortunately yes.

Clamping is kind of serving two purposes here:
1. it makes comparison simple (but we could use `almostEqual`)
2. WebGL expects integers or it will just truncate. I think this is the more important point, and I'd like to investigate the effects more in depth once we have a complete implementation.

PR #408, opened 2025-08-27
feat: add outline mode for selected segments in LabelImageLayer

phoenixAja:
### Summary
  Adds an optional outline rendering mode for selected segments in `LabelImageLayer`, providing an alternative to the existing  fill-based selection highlighting.

  ### Changes
  - **LabelImageLayer**: Added `outlineSelected` option to enable outline mode for selected segments
  - **LabelImageRenderable**: Added support for outline uniforms (`u_outlineSelected`, `u_selectedValue`)
  - **Shader**: Enhanced label image fragment shader with edge detection and white outline rendering
  - **Example**: Updated image labels example with toggle button to switch between Fill and Outline modes

  ### Features
  - White outline rendering around selected segment boundaries using edge detection
  - Toggle between traditional fill mode (entire segment turns white) and new outline mode
  - Selected segments become slightly transparent (90% opacity) in outline mode to enhance outline visibility
  - Integrated toggle UI that matches existing info box styling

  ### Technical Details
  - Edge detection checks 8-connected neighbors to identify segment boundaries
  - Uses float uniforms for WebGL compatibility (boolean uniforms converted to 0.0/1.0)
  - Maintains backward compatibility - outline mode is opt-in via `outlineSelected` parameter

  ### Testing
  - Updated example demonstrates both modes with interactive toggle
  - Works with existing label picking and color mapping functionality

https://github.com/user-attachments/assets/900d1886-b1c4-4dc7-9a18-c7a8107f4f59

### DIFF
--- a/packages/core/examples/image_labels_overlay_with_value_picking/main.ts
+++ b/packages/core/examples/image_labels_overlay_with_value_picking/main.ts
@@ -72,6 +72,21 @@ const canvas = document.querySelector<HTMLCanvasElement>("canvas")!;
 // Get the info div for displaying pick results
 const pickInfoDiv = document.querySelector<HTMLDivElement>("#pick-info")!;
 
+// Add outline mode toggle
+let outlineMode = false;
+const toggleButton = document.createElement("button");
+toggleButton.textContent = "Mode: Fill";
+toggleButton.style.position = "absolute";
+toggleButton.style.top = "10px";
+toggleButton.style.right = "10px";
+toggleButton.style.padding = "8px 16px";
+toggleButton.style.backgroundColor = "#007acc";
+toggleButton.style.color = "white";
+toggleButton.style.border = "none";
+toggleButton.style.borderRadius = "4px";
+toggleButton.style.cursor = "pointer";
+document.body.appendChild(toggleButton);
+
 // Create base image layer
 const imageLayer = new ChunkedImageLayer({
   source: imageSource,
@@ -80,32 +95,64 @@ const imageLayer = new ChunkedImageLayer({
   channelProps: { contrastLimits: phaseContrastLimits },
 });
 
-// Create label image layer with picking functionality
-const labelsLayer = new LabelImageLayer({
-  source: labelsSource,
-  region: labelsRegion,
-  transparent: true,
-  opacity: 0.25,
-  blendMode: "normal",
-  lod,
-  onPickValue: (info: PointPickingResult) => {
-    const { world, value } = info;
-    pickInfoDiv.innerHTML = `
-      <strong>Pick Result:</strong><br/>
-      World: (${world[0].toFixed(1)}, ${world[1].toFixed(1)}, ${world[2].toFixed(1)})<br/>
-      Label Value: ${value}<br/>
-    `;
-    console.debug(`Setting color for label ${value} to transparent`);
-    labelsLayer.setColorMap({
-      cycle: Array.from(labelsLayer.colorMap.cycle),
-      lookupTable: new Map([[value, Color.WHITE]]),
-    });
-  },
-});
+// Function to create label layer with current mode
+function createLabelsLayer() {
+  return new LabelImageLayer({
+    source: labelsSource,
+    region: labelsRegion,
+    transparent: true,
+    opacity: 0.25,
+    blendMode: "normal",
+    lod,
+    outlineSelected: outlineMode,
+    onPickValue: (info: PointPickingResult) => {
+      const { world, value } = info;
+      pickInfoDiv.innerHTML = `
+        <strong>Pick Result:</strong><br/>
+        World: (${world[0].toFixed(1)}, ${world[1].toFixed(1)}, ${world[2].toFixed(1)})<br/>
+        Label Value: ${value}<br/>
+        Mode: ${outlineMode ? "Outline" : "Fill"}
+      `;
+      
+      if (outlineMode) {
+        // In outline mode, the layer handles the selection internally
+      } else {
+        // In fill mode, use the old white fill behavior
+        labelsLayer.setColorMap({
+          cycle: Array.from(labelsLayer.colorMap.cycle),
+          lookupTable: new Map([[value, Color.WHITE]]),
+        });
+      }
+    },
+  });
+}
 
-new Idetik({
+// Create initial label layer
+let labelsLayer = createLabelsLayer();
+
+// Create Idetik instance
+const idetik = new Idetik({
   canvas,
   camera,
   layers: [imageLayer, labelsLayer],
   cameraControls: new PanZoomControls(camera),
-}).start();
+});
+
+// Add toggle button functionality
+toggleButton.addEventListener("click", () => {
+  outlineMode = !outlineMode;
+  toggleButton.textContent = `Mode: ${outlineMode ? "Outline" : "Fill"}`;
+  
+  // Remove old layer and create new one with updated mode
+  idetik.layerManager.remove(labelsLayer);
+  labelsLayer = createLabelsLayer();
+  idetik.layerManager.add(labelsLayer);
+  
+  // Clear pick info
+  pickInfoDiv.innerHTML = `
+    <strong>Mode changed to:</strong> ${outlineMode ? "Outline" : "Fill"}<br/>
+    Click on a label to see the effect.
+  `;
+});
+
+idetik.start();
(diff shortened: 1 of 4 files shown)


packages/core/src/renderers/shaders/label_image_frag.glsl
```diff
@@ -10,6 +10,8 @@ uniform mediump sampler2D ColorCycleSampler;
 uniform highp usampler2D ColorLookupTableSampler;
 
 uniform float u_opacity;
+uniform float u_outlineSelected;
```
andy-sweet: I think we can currently handle this being a bool, which would be a little clearer than a float here.

packages/core/src/renderers/shaders/label_image_frag.glsl
```diff
@@ -10,6 +10,8 @@ uniform mediump sampler2D ColorCycleSampler;
...
+uniform float u_selectedValue;
```
andy-sweet, line 14: Label image currently only supports uint, so this could also be a uint which would avoid the cast in the program itself.

At some point, this shader should at least support int, and maybe float too? So this is optional.

packages/core/examples/image_labels_overlay_with_value_picking/main.ts
```diff
@@ -80,32 +89,58 @@ const imageLayer = new ChunkedImageLayer({
...
+// Create Idetik instance
+const idetik = new Idetik({
   canvas,
   camera,
   layers: [imageLayer, labelsLayer],
   cameraControls: new PanZoomControls(camera),
-}).start();
+});
+
+// Add toggle functionality
```
andy-sweet, line 134: I also don't mind if you want to delete the fill behavior to keep this example simpler. But you already handled it, so also fine to leave as is.

packages/core/src/renderers/shaders/label_image_frag.glsl
```diff
@@ -21,22 +23,71 @@ vec4 unpackRgba(uint packed) {
     return vec4(float(r), float(g), float(b), float(a)) / 255.0;
 }
 
+bool isEdgePixel(uint centerValue, vec2 texCoords) {
+    vec2 texSize = vec2(textureSize(ImageSampler, 0));
```
andy-sweet, line 27: You might be able to use textureOffset to avoid this lookup, but I think that would require inlining the for loop below and may changing how to boundaries are handled. So probably not worth it.

packages/core/src/renderers/shaders/label_image_frag.glsl
```diff
@@ -21,22 +23,71 @@ vec4 unpackRgba(uint packed) {
...
+}
+
 void main() {
     uint texel = texture(ImageSampler, TexCoords).r;
+    
+    // Check if we should outline this selected segment
+    if (u_outlineSelected > 0.5 && u_selectedValue >= 0.0 && float(texel) == u_selectedValue) {
+        if (isEdgePixel(texel, TexCoords)) {
+            // Draw outline in bright white with full opacity
+            fragColor = vec4(1.0, 1.0, 1.0, 1.0);
```
andy-sweet: I might expect the outline to use the layer opacity defined in `u_opacity`.

packages/core/src/renderers/shaders/label_image_frag.glsl
```diff
@@ -21,22 +23,71 @@ vec4 unpackRgba(uint packed) {
     return vec4(float(r), float(g), float(b), float(a)) / 255.0;
 }
 
+bool isEdgePixel(uint centerValue, vec2 texCoords) {
```
andy-sweet: TexCoords is available to this function as a uniform, so no need to pass it in.

packages/core/src/renderers/shaders/label_image_frag.glsl
```diff
@@ -21,22 +23,71 @@ vec4 unpackRgba(uint packed) {
...
+        }
+    }
+    return false;
+}
+
 void main() {
     uint texel = texture(ImageSampler, TexCoords).r;
+    
+    // Check if we should outline this selected segment
+    if (u_outlineSelected > 0.5 && u_selectedValue >= 0.0 && float(texel) == u_selectedValue) {
```
andy-sweet: Define this bool once, maybe as `bool isSelectedValue` since it's used here and two more times below.

packages/core/src/layers/label_image_layer.ts
```diff
@@ -28,16 +29,19 @@ export class LabelImageLayer extends Layer {
   private readonly lod_?: number;
   private colorMap_: LabelColorMap;
   private readonly onPickValue_?: (info: PointPickingResult) => void;
+  private readonly outlineSelected_: boolean;
   private image_?: LabelImageRenderable;
   private imageChunk_?: Chunk;
   private pointerDownPos_: vec2 | null = null;
+  private selectedValue_: number | null = null;
```
andy-sweet, line 36: Since null is a possible value for the selectedValue, could we just use that to infer `outlineSelected_`? I'm wondering if we need the two values defined on the layer like this.

packages/core/src/renderers/shaders/label_image_frag.glsl
```diff
@@ -21,22 +23,71 @@ vec4 unpackRgba(uint packed) {
...
+        if (isEdgePixel(texel, TexCoords)) {
+            // Draw outline in bright white with full opacity
+            fragColor = vec4(1.0, 1.0, 1.0, 1.0);
+            return;
+        }
+        // For non-edge pixels of selected segment, use normal coloring but more transparent
```
andy-sweet: Since no code immediately follows this, it's a little confusing to read this comment. Since the code that handles the selection transparency is also commented below, I think removing this makes things more readable.

andy-sweet (approved):
Generally works well and looks good, so approved. But I think some of my requests are worth implementing.

The bigger question is that I thought the segmentation app is using regular image layers for the segmentations, so this won't work there. Is that not correct?

phoenixAja:
> The bigger question is that I thought the segmentation app is using regular image layers for the segmentations, so this won't work there. Is that not correct?

This is correct, though we were thinking, if time permits, of moving over to the LableImageLayer. I thought of implementing this in ImageLayer, but i don't think this makes sense there at all.

packages/core/src/renderers/shaders/label_image_frag.glsl
phoenixAja: it's not a bool becuase of [this](https://github.com/chanzuckerberg/idetik/blob/9f9c46da2b1893ed26ed92fd0b06cf8593348d51/packages/core/src/renderers/webgl_shader_program.ts#L45) comment (no dedicated uniform1b in WebGL)

packages/core/examples/image_labels_overlay_with_value_picking/main.ts
phoenixAja: i like having it as an option so people can see

packages/core/src/renderers/shaders/label_image_frag.glsl
phoenixAja: really? I thought it would be nice for the outline to be as solid as possible?

packages/core/src/renderers/shaders/label_image_frag.glsl
andy-sweet: But we have the bool array in the scalar image array fragment shader (for channel visibility), so I think it should work (i.e. defining a bool in the fragment shader and using uniform1f to pass it through). Maybe we shouldn't rely on that though? Anyway, not blocking at all!

packages/core/src/renderers/shaders/label_image_frag.glsl
phoenixAja: thank you!

packages/core/src/renderers/shaders/label_image_frag.glsl
andy-sweet: Well, I think of the outline as part of the layer (since it is configured through it), so I expect it to pick up the opacity from there too. But I don't feel too strongly, and in the working example, the outline was quite faint.

phoenixAja: that's a good point

phoenixAja: i added your suggestion

PR #410, opened 2025-08-28
Remove ultrack-prototype workspace

aganders3:
### Summary
This removes the ultrack-prototype workspace.

Our other applications all live in other repos, which will be a more maintainable model for us as we move towards releasing and versioning Idetik.

I pushed two tags to make it easier to find this code if/when we need it in the future:
* **[last-ultrack-prototype](https://github.com/chanzuckerberg/idetik/releases/tag/last-ultrack-prototype)** - Last commit with ultrack-prototype
* **[removed-ultrack-prototype](https://github.com/chanzuckerberg/idetik/releases/tag/removed-ultrack-prototype)** - Commit removing ultrack-prototype workspace

Now that I think about it, searching for this PR is probably the easiest. ¯\\\_(ツ)\_/¯

### Related Issue
N/A

### Tests & Checks
Project still builds, no remaining references to the ultrack prototype in the repo


aganders3:
I'm not sure if this weird CodeQL failure - fails to detect Python code - will block merging. It looks like we can maybe still merge as long as the PR is approved. I hope the error then just goes away because it looks like it's basing which languages to scan from the default branch, not the PR itself.

andy-sweet (changes requested):
Everything looks good. The only blocking change is to remove the Ultrack part of the README. It would also be good to remove the Ultrack specific things in the top-level .gitignore file (i.e. pixi and slurm at the end).

andy-sweet (approved):
Looks good now, thanks!

PR #411, opened 2025-08-28
Add image renderable pool

shlomnissan:
### Summary

This PR introduces a renderable pool to recycle image renderables,
reducing allocation overhead. It also refactors the chunked image
layer to eliminate repetitive logic and adds defensive checks when
updating texture data.

### Tests & Checks
- Chunk info overlay shows:
  - Number of textures never exceeds the total number of tiles
  - Texture memory usage is bounded by the same limit
- Basic unit tests for verifying the pool's functionality
- Console logs when items are acquired or released from the pool
- Verified that all existing examples are working as expected

### DIFF
--- a/packages/core/src/layers/chunked_image_layer.ts
+++ b/packages/core/src/layers/chunked_image_layer.ts
@@ -14,6 +14,10 @@ import { handlePointPickingEvent, PointPickingResult } from "./point_picking";
 import { almostEqual } from "../utilities/almost_equal";
 import { clamp } from "../utilities/clamp";
 import { Region } from "../data/region";
+import {
+  RenderablePool,
+  poolKeyForImageRenderable,
+} from "../utilities/renderable_pool";
 
 export type ChunkedImageLayerProps = LayerOptions & {
   source: ChunkSource;
@@ -29,7 +33,8 @@ export class ChunkedImageLayer extends Layer {
   private readonly region_: Region;
   private readonly onPickValue_?: (info: PointPickingResult) => void;
   private readonly visibleChunks_: Map<Chunk, ImageRenderable> = new Map();
-  private channelProps_?: ChannelProps;
+  private readonly channelProps_?: ChannelProps;
+  private readonly pool_ = new RenderablePool<ImageRenderable>();
   private chunkManagerSource_?: ChunkManagerSource;
   private pointerDownPos_: vec2 | null = null;
   private zPrevPointWorld_?: number;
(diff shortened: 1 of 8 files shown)


packages/core/src/utilities/renderable_pool.ts
```diff
@@ -0,0 +1,40 @@
...
+    let bin = this.bins_.get(key);
+    if (!bin) {
+      bin = [];
+      this.bins_.set(key, bin);
+    }
+    bin.push(item);
+    Logger.debug("RenderablePool", "Renderable object released");
+  }
+
+  clearAll(disposer?: (t: T) => void) {
```
shlomnissan, line 26: Currently not in use. I considered adding `Layer.dispose` and use it to clear resources but decided to skip as I want to think about it more. Meanwhile, I kept this function for completeness.

packages/core/test/renderable_pool.test.ts
```diff
@@ -0,0 +1,82 @@
...
+});
+
+describe("poolKeyForImageRenderable", () => {
+  test("builds a stable key string from chunk layout fields", () => {
+    const chunk = {
+      lod: 2,
+      shape: { x: 256, y: 128 },
+      rowStride: 256,
+      rowAlignmentBytes: 4,
+    } as unknown as import("../src/data/chunk").Chunk;
```
shlomnissan, line 75: This cast trick lets me construct a chunk without specifying all required fields. It's also scoped which is nice. There might be a cleaner way to do this but it works for the test.

packages/core/src/objects/textures/texture_2d.ts
```diff
@@ -41,13 +41,36 @@ export class Texture2D extends Texture {
     return this.height_;
   }
 
-  public static createWithChunk(chunk: Chunk) {
-    if (!chunk.data) {
+  public updateWithChunk(chunk: Chunk, data?: ChunkData) {
```
shlomnissan, line 44: I'm not convinced this should be a member function. Lower level texture types should be agnostic to the source, but we already generate textures using chunks in this module (statically) and we can always move it later.

packages/core/src/layers/chunked_image_layer.ts
```diff
@@ -71,39 +76,27 @@ export class ChunkedImageLayer extends Layer {
 
   private updateChunks() {
     if (!this.chunkManagerSource_) return;
+    if (this.state !== "ready") this.setState("ready");
 
-    // Temporary until we decide how to approach state management
-    if (this.state !== "ready") {
-      this.setState("ready");
-    }
+    const ordered = this.chunkManagerSource_.getChunks();
```
andy-sweet: Nit: I like the idea of giving this a more informative name, but `ordered` doesn't tell us how this array is ordered, so maybe something like this?

```suggestion
    const orderedByLod = this.chunkManagerSource_.getChunks();
```

packages/core/src/layers/chunked_image_layer.ts
```diff
@@ -71,39 +76,27 @@ export class ChunkedImageLayer extends Layer {
...
+    const current = new Set(ordered);
```
andy-sweet: Nit: I believe a JS Set is ordered according to entry order, so we could just use `Set` for ordered. But I don't mind keeping the separation for clarity since the array is just a reference.

packages/core/src/layers/chunked_image_layer.ts
```diff
@@ -71,39 +76,27 @@ export class ChunkedImageLayer extends Layer {
...
+      if (!current.has(chunk)) {
+        this.visibleChunks_.delete(chunk);
+        this.pool_.release(poolKeyForImageRenderable(chunk), image);
       }
     });
 
-    // Add all objects anew so that they respect the chunk order, which may
-    // capture details important for rendering, such as LOD, instead of the
-    // creation order, which is dependent on when the chunks finished loading.
+    // Rebuild draw list in correct order
```
andy-sweet: Nit: both of these one line comments are less informative that the code, which is quite concise and readable now. So I think they should just be removed.

packages/core/test/renderable_pool.test.ts
andy-sweet: I have considered if we want to distinguish between a chunk (that may be 4D pretty soon) and 2D chunk that is a slice through that, which could be useful here. But this is fine for now.

packages/core/src/utilities/renderable_pool.ts
```diff
@@ -0,0 +1,40 @@
...
+    if (disposer) for (const bin of this.bins_.values()) bin.forEach(disposer);
+    this.bins_.clear();
+  }
+}
+
+export function poolKeyForImageRenderable(chunk: Chunk) {
```
andy-sweet: Nit: this feels a little out of place because this key is quite specific to the `ChunkedImageLayer`. Maybe it is better off there? Don't feel strongly.

packages/core/src/layers/chunked_image_layer.ts
```diff
@@ -117,7 +110,8 @@ export class ChunkedImageLayer extends Layer {
       if (chunk.state !== "loaded" || !chunk.data) continue;
       const data = this.slicePlane(chunk, pointWorld);
       if (data) {
-        image.textures[0].data = data;
+        const texture = image.textures[0] as Texture2DArray;
```
andy-sweet, line 108: This cast is a bit of code smell. It's not too hard (at the moment) to understand that this static cast is correct, so this doesn't block merging this PR. But it suggests a change needed somewhere else with regards to renderable/texture/chunk/data typing.

andy-sweet (approved):
Looks good and works well for me. All comments are optional/nits.

packages/core/src/utilities/renderable_pool.ts
aganders3: I had the same reaction and agree it makes more sense to me in `ChunkedImageLayer`.

packages/core/src/layers/chunked_image_layer.ts
shlomnissan: > I believe a JS Set is ordered according to entry order

I wasn't sure when I reviewed the code so I looked into it and found that it is ordered, but decided to iterate through the array to make it clear.

packages/core/src/layers/chunked_image_layer.ts
shlomnissan: I agree. I added these comments before refactoring and at the time, it wasn't clear to me. Removing these comments now makes sense.

packages/core/src/utilities/renderable_pool.ts
shlomnissan: I agree. I moved it to `ChunkedImageLayer` but kept the test in the renderable pool test suite for now.

packages/core/src/layers/chunked_image_layer.ts
shlomnissan: I agree. I added `updateWithChunk` to `Texture2D` and `Texture2DArray` so technically we can add an abstract member to the base class, but as you said, it suggests a change somewhere else and I don't want to hide it. This is related to the comment I left earlier in this PR: https://github.com/chanzuckerberg/idetik/pull/411#discussion_r2307995884

PR #414, opened 2025-08-28
Add assets for publishing to NPM

bchu1:
https://www.npmjs.com/package/@idetik/core-prerelease
https://www.npmjs.com/package/@idetik/react-prerelease

- The packages for now are called `[core|react]-prerelease` so that current clients can have some degree of safety with semver while allowing us to reserve `@idetik/[core|react]@1.0.0` for our true launch (at which point we can delete these temporary packages).
- Next step: Split the repos and delete the top-level NPM project.
- Added a Claude-generated README.md to `core` so that something shows up on the NPM page, but I could delete that if we don't think it helps.
- Adds disclaimers that the packages are under active development.

### DIFF
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -1,6 +1,6 @@
 # @idetik/core-prerelease
 
-🚨 **This package is under active development and is subject to frequent changes.** Use at your own risk in production environments.
+🚨 **This package is under active development.** Use at your own risk in production environments.
 
 A layer-based visualization abstraction for interactive rendering of large datasets, with particular focus on scientific imaging and bioinformatics data formats.
 
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -1,0 +1,53 @@
+# @idetik/core-prerelease
+
+🚨 **This package is under active development and is subject to frequent changes.** Use at your own risk in production environments.
+
+A layer-based visualization abstraction for interactive rendering of large datasets, with particular focus on scientific imaging and bioinformatics data formats.
+
+## Key Components
+
+### Core Architecture
+- `Idetik`: Main runtime
+- `LayerManager`: Manages layer composition and rendering order
+- `WebGLRenderer`: GPU-accelerated rendering engine
+- `Layer`: Base class for all visualization layers
+
+### Layer Types
+- `ImageLayer`: Static image rendering
+- `ChunkedImageLayer`: Streaming large image data
+- `ImageSeriesLayer`: Time-series and multi-dimensional images
+- `LabelImageLayer`: Segmentation masks and labels
+- `TracksLayer`: Object tracking visualization
+- `ProjectedLineLayer`: Line and curve rendering
+- `AxesLayer`: Coordinate system visualization
+
+## Examples
+
+The `/examples` directory contains comprehensive demonstrations:
+
+- **Basic image visualization**: 2D image rendering from OME-Zarr
+- **Multi-dimensional data**: 5D datasets (x, y, z, channel, time)
+- **High Content Screening**: Plate/well data visualization  
+- **Label overlays**: Segmentation masks with value picking
+- **Image series**: Time-lapse and z-stack navigation
+- **Streaming data**: Chunk-based loading for large datasets
+- **Layer composition**: Blending multiple visualization layers
+- **Interactive features**: Point selection and data sampling
+
+## Development
+
+### Building
+```bash
+npm run build  # Compile TypeScript and bundle
+```
+
+### Testing
+```bash
+npm test              # Run tests with watch mode
+npm run test-with-coverage  # Generate coverage report
+```
+
+### Development Server
+```bash
+npm run dev           # Start development server with examples
+```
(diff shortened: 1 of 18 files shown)


aganders3:
> Added a Claude-generated README.md to core so that something shows up on the NPM page, but I could delete that if we don't think it helps.

I think it's nice to have something there, but the README mentions the examples and I don't think they ship with the npm package.

edit: also the emoji does not render for me on the npm page

packages/react/package.json
```diff
@@ -46,16 +46,19 @@
   ],
   "author": "",
   "license": "MIT",
+  "keywords": [
+    "idetik"
+  ],
   "devDependencies": {
-    "@idetik/core": "^0.0.1",
+    "@idetik/core-prerelease": "^0.0.1",
```
aganders3: Do we need this in both `devDependencies` as well as `peerDependencies`?

I know this is not new but I'm realizing here.

packages/react/README.md
```diff
@@ -1,14 +1,22 @@
...
 ```bash
-npm install @idetik/react @idetik/core
+npm i @idetik/react-prerelease @idetik/core-prerelease
+```
+
+Add
+```css
+@import "@idetik/react-prerelease/style.css";
 ```
+to you global CSS file.
```
andy-sweet:
```suggestion
to your global CSS file.
```

packages/react/README.md
```diff
@@ -1,14 +1,22 @@
...
+
 React components for interactive visualization of large datasets, built on the Idetik runtime.
 
 ## For Integrators: Using @idetik/react Components
 
 ### Installation
 
 ```bash
-npm install @idetik/react @idetik/core
+npm i @idetik/react-prerelease @idetik/core-prerelease
```
andy-sweet, line 12: After this PR, we don't need to explicitly install core, right?

```suggestion
npm i @idetik/react-prerelease
```

packages/core/README.md
```diff
@@ -0,0 +1,53 @@
...
+The `/examples` directory contains comprehensive demonstrations:
+
+- **Basic image visualization**: 2D image rendering from OME-Zarr
+- **Multi-dimensional data**: 5D datasets (x, y, z, channel, time)
+- **High Content Screening**: Plate/well data visualization  
+- **Label overlays**: Segmentation masks with value picking
+- **Image series**: Time-lapse and z-stack navigation
+- **Streaming data**: Chunk-based loading for large datasets
+- **Layer composition**: Blending multiple visualization layers
+- **Interactive features**: Point selection and data sampling
```
andy-sweet: I'd probably just scrap all of this. I wouldn't expect the examples to be included. And I don't feel a strong need to document/advertise the architecture or layer types, which are likely to change a little soon anyway.

packages/react/package.json
```diff
@@ -1,15 +1,15 @@
 {
-  "name": "@idetik/react",
-  "private": true,
-  "version": "0.0.1",
+  "name": "@idetik/react-prerelease",
+  "version": "0.0.3",
```
andy-sweet: Where did this version number come from? I think @aganders3 hope was that we would start at 1.0.0 especially if this has prerelease in the name.

andy-sweet:
> edit: also the emoji does not render for me on the npm page

Same for me.

bchu1:
> > Added a Claude-generated README.md to core so that something shows up on the NPM page, but I could delete that if we don't think it helps.
> 
> I think it's nice to have something there, but the README mentions the examples and I don't think they ship with the npm package.
> 
> edit: also the emoji does not render for me on the npm page

Scrapped most of the README. The NPM page was old, just updated it to `1.0.0`.

packages/react/package.json
bchu1: Just published `1.0.0`!

packages/react/README.md
bchu1: Just learned about this but apparently if you only install `@idetik/react`, `@idetik/core` will go into the lockfile so everything will work, but you still won't be able to directly use `@idetik/core` like with `new ImageLayer()`. I think Yarn actually forces you to install both manually.

I'm fine if we want to have an NPM command for just `@idetik/react` and a Yarn one with both, but I figured we might as well give them both? which I noticed is common for other libraries too https://mui.com/material-ui/getting-started/installation/#default-installation

packages/react/package.json
aganders3: I think the npm site encouraged publishing at 1.0.0 but I promise I don't care _that_ much!

aganders3 (commented):
I think I'm good with this - I'm going to try installing/using in the CryoLens app before approving if that's okay!

packages/react/package.json
bchu1: I think we need it if we're going to split the repos and develop `react`? Not sure about the patterns for this though

bchu1:
> I think I'm good with this - I'm going to try installing/using in the CryoLens app before approving if that's okay!

CryoLens demo using the new NPM packages has been merged!

aganders3 (approved):
Looks good!

The working apps in VCP is proof-enough that this works for me.

PR #418, opened 2025-08-29
Replace all console use in core library

andy-sweet:
### Summary
This change replaces any existing usage of the console with the Logger. It also introduces a new lint rule to prevent future use of the console directly.

### Tests & Checks
I manually tested the core examples.
I manually tested the React examples.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -43,5 +43,11 @@ export default [
       ],
     }
   },
+  {
+    files: ["packages/core/src/**/*.{js,ts,jsx,tsx}"],
+    rules: {
+      'no-console': 'error'
+    }
+  },
   { ignores: ["node_modules", "**/coverage", "**/dist"] },
 ];
--- a/packages/core/src/core/layer.ts
+++ b/packages/core/src/core/layer.ts
@@ -1,6 +1,7 @@
 import { IdetikContext } from "../idetik";
 import { RenderableObject } from "./renderable_object";
 import { clamp } from "../utilities/clamp";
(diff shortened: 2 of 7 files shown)


andy-sweet:
@shlomnissan : I assigned you for review because I think you introduced the logger and wanted to make sure you thought the level bump is reasonable.

aganders3:
Can we prevent new instances from creeping in with a lint rule like [no-console](https://eslint.org/docs/latest/rules/no-console)?

I'm happy to hold that off because we have a few other issues pending about adding lint rules.

aganders3:
@andy-sweet @shlomnissan 

I merged main and added the `no-console` lint. Usage of `console` is still allowed in examples, config/build files, and tests.

I think we should get this in!

aganders3 (approved):

andy-sweet:
Actually, I'm going to mark this as draft just because I'm not sure we should show Idetik core info logs in applications that depend on it, so don't want it to be accidentally merged.

andy-sweet:
Will update it shortly though, as the log removals and lint check are valuable.

eslint.config.mjs
```diff
@@ -43,5 +43,11 @@ export default [
       ],
     }
   },
+  {
+    files: ["packages/core/src/**/*.{js,ts,jsx,tsx}"],
```
andy-sweet, line 47: This only covers the core package, but we likely want to consider a similar rule for the React package too. Can do that in a separate PR though because we'll have to decide if it should use the core Logger or its own.

PR #419, opened 2025-08-29
Multiple Viewports: Introduce Viewport class and update Renderer.render() to take viewport

aganders3:
### Summary

This PR introduces the Viewport abstraction and updates the Renderer API to accept viewport objects instead of separate parameters.

- Added `Viewport` class that encapsulates camera, HTML element, layer manager, and coordinate transformations
- Added `ViewportConfig` interface and `parseViewportConfigs()` utility function with validation
- Modified `Renderer.render()` interface from `render(layerManager, camera, viewportBox)` to `render(viewport)`
- Updated `WebGLRenderer` to implement the new render signature
- `activeCamera` removed from base `Renderer`, as viewports now handle resizing and camera updates
- Modified runtime to use viewport internally while maintaining backward compatibility through getter/setter proxies
- Modified runtime to start layout observers for resize events on all viewports and device pixel ratio

The runtime still operates with a single viewport only - this change is purely an API refactor to prepare for multiple viewport support. All existing applications and examples continue to work without modification.

This is step two of the implementation plan in the [Multiple Viewports Tech Spec](https://docs.google.com/document/d/1OwGL7KW34hxFAENyOm2WiQ71E58XGu9Y7jKkJlyRx84/edit?tab=t.0).

### Related Issue
N/A

### Tests & Checks
Manually tested examples
Added basic tests for viewport + instantiation

### DIFF
--- a/packages/core/src/core/renderer.ts
+++ b/packages/core/src/core/renderer.ts
@@ -1,15 +1,12 @@
-import { LayerManager } from "./layer_manager";
-import { Camera } from "../objects/cameras/camera";
 import { Color, ColorLike } from "./color";
 import { Layer } from "./layer";
-import { Box2 } from "../math/box2";
+import { Viewport } from "./viewport";
 
 export abstract class Renderer {
   private readonly canvas_: HTMLCanvasElement | null;
   private width_ = 0;
   private height_ = 0;
   private backgroundColor_: Color = new Color(0, 0, 0, 0);
-  private activeCamera_: Camera | null = null;
 
   protected abstract resize(width: number, height: number): void;
   protected abstract renderObject(layer: Layer, objectIndex: number): void;
@@ -20,19 +17,7 @@ export abstract class Renderer {
     this.updateRendererSize();
   }
 
-  protected set activeCamera(camera: Camera) {
-    if (this.activeCamera_ !== camera) {
-      this.activeCamera_ = camera;
-      this.updateActiveCamera();
-    }
-  }
-
-  // TODO: make this just take a Viewport instance once implemented
-  public abstract render(
-    layerManager: LayerManager,
-    camera: Camera,
-    viewportBox: Box2
-  ): void;
+  public abstract render(viewport: Viewport): void;
 
   public updateSize(): void {
     this.updateRendererSize();
@@ -45,15 +30,6 @@ export abstract class Renderer {
 
     if (this.canvas.width !== this.width_) this.canvas.width = this.width_;
     if (this.canvas.height !== this.height_) this.canvas.height = this.height_;
-
-    this.updateActiveCamera();
-  }
-
-  private updateActiveCamera() {
-    const canvasAspectRatio = this.width_ / this.height_;
-    if (this.activeCamera_) {
-      this.activeCamera_.setAspectRatio(canvasAspectRatio);
-    }
   }
 
   protected get canvas() {
@@ -74,14 +50,5 @@ export abstract class Renderer {
 
   public set backgroundColor(color: ColorLike) {
     this.backgroundColor_ = Color.from(color);
-  }
-
-  protected get activeCamera() {
-    if (this.activeCamera_ === null) {
-      throw new Error(
-        "Attempted to access the active camera before it was set."
-      );
-    }
-    return this.activeCamera_;
   }
 }
--- a/packages/core/src/core/renderer.ts
+++ b/packages/core/src/core/renderer.ts
@@ -1,3 +1,4 @@
+import { Camera } from "../objects/cameras/camera";
 import { Color, ColorLike } from "./color";
 import { Layer } from "./layer";
 import { Viewport } from "./viewport";
@@ -9,7 +10,11 @@ export abstract class Renderer {
   private backgroundColor_: Color = new Color(0, 0, 0, 0);
 
   protected abstract resize(width: number, height: number): void;
-  protected abstract renderObject(layer: Layer, objectIndex: number): void;
+  protected abstract renderObject(
+    layer: Layer,
+    objectIndex: number,
+    camera: Camera
+  ): void;
   protected abstract clear(): void;
 
   constructor(canvas: HTMLCanvasElement) {
--- a/packages/core/src/core/viewport.ts
+++ b/packages/core/src/core/viewport.ts
@@ -5,6 +5,7 @@ import { CameraControls } from "../objects/cameras/controls";
 import { Box2 } from "../math/box2";
 import { vec2, vec3 } from "gl-matrix";
 import { generateUUID } from "../utilities/uuid_generator";
+import { Logger } from "../utilities/logger";
 
 export interface ViewportConfig {
   id?: string;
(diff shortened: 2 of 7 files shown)


packages/core/test/viewport.test.ts
```diff
@@ -0,0 +1,145 @@
...
+    element,
+    camera,
+  };
+
+  const canvas = createTestCanvas();
+  const viewport = new Viewport(config, layerManager, canvas);
+  expect(viewport.id).toBe("element-id");
+});
+
+test("Viewport constructor generates UUID when no ID available", () => {
```
andy-sweet: Nit: while the implementation may use a UUID, this does not test that.

```suggestion
test("Viewport constructor generates some ID when no ID available", () => {
```

packages/core/test/viewport.test.ts
```diff
@@ -0,0 +1,145 @@
...
+  const canvas = createTestCanvas();
+  const viewport = new Viewport(config, layerManager, canvas);
+  expect(viewport.id).toBeDefined();
+  expect(viewport.id.length).toBeGreaterThan(0);
+});
+
+test("parseViewportConfigs creates viewports with validation", () => {
+  const element1 = createTestElement("viewport1");
+  const element2 = createTestElement("viewport2");
+  const camera1 = new OrthographicCamera(-1, 1, -1, 1, 0.1, 1000);
```
andy-sweet: It might be worth defining a `createTestCamera` since the camera's type and parameters are not relevant to these tests (yet), and their construction is quite noisy.

packages/core/src/renderers/webgl_renderer.ts
```diff
@@ -34,6 +34,7 @@ export class WebGLRenderer extends Renderer {
   private readonly bindings_: WebGLBuffers;
   private readonly textures_: WebGLTextures;
   private readonly state_: WebGLState;
+  private activeCamera_: Camera | null = null;
```
andy-sweet: It looks like this is only used in `render` and subroutines it calls (assigned from `viewport.camera`). Could we remove this and use the local non-nullable instead, and pass it through as needed? Then we could also remove the check and throw.

packages/core/src/core/viewport.ts
```diff
@@ -0,0 +1,151 @@
...
+    this.camera = config.camera;
+    this.layerManager = layerManager;
+    this.cameraControls = config.cameraControls ?? null;
+    this.updateAspectRatio();
+
+    for (const layer of config.layers ?? []) {
+      this.layerManager.add(layer);
+    }
+  }
+
```
andy-sweet, line 39:
```suggestion
```

packages/core/src/core/renderer.ts
```diff
@@ -20,19 +17,7 @@ export abstract class Renderer {
...
-    }
-  }
-
-  // TODO: make this just take a Viewport instance once implemented
-  public abstract render(
-    layerManager: LayerManager,
-    camera: Camera,
-    viewportBox: Box2
-  ): void;
+  public abstract render(viewport: Viewport): void;
```
andy-sweet, line 25: I like all this red.

packages/core/src/core/viewport.ts
```diff
@@ -0,0 +1,151 @@
...
+
+  public updateSize(): void {
+    this.cachedViewportBox_ = null;
+    this.updateAspectRatio();
+  }
+
+  public getViewportBox(): Box2 {
```
andy-sweet: Nit: `Viewport.getViewportBox` is a little repetitive. I think would help to clarify what coordinate space this box lives in - maybe with something like `getCanvasBox`, `getCanvasRect`, or `getCanvasBounds`.

andy-sweet (commented):
Sorry, I can't approve/request because I ran out of time to look at this today. Hopefully my initial comments are useful.

packages/core/src/renderers/webgl_renderer.ts
aganders3: Thanks - I tried this change but decided against it in the initial implementation because it just felt like a lot of changes. You're right though it would be nice to use a non-nullable and not have to check/throw. It's a required parameter here anyway, just passed through this private attribute, so passing it through is more clear.

packages/core/src/core/viewport.ts
aganders3: Is `getBoxRelativeToCanvas` too long? I'm concerned `getCanvasX` sounds like it's getting the size/position of the canvas itself. I think it makes sense to use "Box" in the name since that's the return type.

aganders3: I opted for `getBoxRelativeTo(canvas: HTMLCanvas)` for now, but definitely open to ideas on that.

packages/core/src/renderers/webgl_renderer.ts
```diff
@@ -57,37 +57,43 @@ export class WebGLRenderer extends Renderer {
     this.resize(this.canvas.width, this.canvas.height);
   }
 
-  public render(layerManager: LayerManager, camera: Camera, viewportBox: Box2) {
+  public render(viewport: Viewport) {
```
andy-sweet, line 60: I love this new function signature.

packages/core/src/renderers/webgl_renderer.ts
```diff
@@ -57,37 +57,43 @@ export class WebGLRenderer extends Renderer {
...
       vec2.fromValues(this.width, this.height)
     );
     if (Box2.equals(viewportBox.floor(), rendererBox.floor())) {
       this.state_.setScissorTest(false);
-    } else {
+    } else if (Box2.intersects(viewportBox, rendererBox)) {
       this.state_.setScissor(viewportBox);
       this.state_.setScissorTest(true);
+    } else {
+      Logger.debug(
```
andy-sweet: Nit: I'd bump this to a warning because it seems unlikely that a user wants the viewport to be outside the canvas bounds, and this means that this log won't be filtered out by our logger or the browser's console by default.

```suggestion
      Logger.warn(
```

packages/core/src/idetik.ts
```diff
@@ -175,4 +173,40 @@ export class Idetik {
...
+    });
+
+    resizeObserver.observe(this.canvas);
+    for (const viewport of this.viewports_) {
+      if (viewport.element !== this.canvas) {
+        resizeObserver.observe(viewport.element);
+      }
+    }
+
+    const startDevicePixelRatioObserver = () => {
```
andy-sweet, line 189: Optional

At first, I was unsure why you made this call itself and used `once: true`, or why you didn't just inline the code and not use `once: true`. I think that's because this approach will ensure every sequential change to the pixel ratio is accounted for, right? I think a short inline comment could help others in the future.

andy-sweet (approved):
This looks good to me now. Just a few optional comments.

packages/core/src/renderers/webgl_renderer.ts
aganders3: That makes sense. I was thinking about perhaps using HTML layout to intentionally show/hide viewports, but maybe that should be a separate check (i.e. check `display: none` or something).

PR #420, opened 2025-08-29
Split chunk manager source into cache and view

andy-sweet:
### Summary
This is a draft PR that I'm pushing up for reference only. It show one way to approach different views of the same underlying cache, that could be useful for things like multiple viewports. I'm not convinced it's the right approach yet though.


andy-sweet:
Closing since I have no intention of merging this, but will not delete the branch for now, so this can be used as a reference for related future work.

PR #421, opened 2025-09-03
Move fetching to chunk manager

shlomnissan:
### Summary

This PR moves fetching from chunk manager source to the chunk
manager itself. It also adds a basic priority system which is needed
now that we fetch from multiple sources in a single location.

### Notes

This is one of three PRs I have planned. The following two will include:
- Fetch queue that supports aborting requests
- Priority based on chunk distance from the center

### Tests & Checks

- I verified that all the existing examples work as expected.


phoenixAja (approved):
lg to me

Issue #372, opened 2025-08-07
ChunkedImageLayer does not support multi-channel images


andy-sweet:
When using the ChunkManager with a multi-channel image, nothing is rendered and the following error is repeatedly thrown

```
[2025-08-07T14:35:09.926Z][ERROR][ChunkManager] Error loading chunk (0,0): Error: Expected numeric index for C
log	@	logger.ts:92
error	@	logger.ts:65
(anonymous)	@	chunk_manager.ts:178
```

To reproduce you can modify the [latest value picking example](https://github.com/chanzuckerberg/idetik/blob/ba5420e1c7d78952858a0d8a39406f9d0dde0716/packages/core/examples/image_labels_overlay_with_value_picking/main.ts) with the following git diff

```
diff --git a/packages/core/examples/image_labels_overlay_with_value_picking/main.ts b/packages/core/examples/image_labels_overlay_with_value_picking/main.ts
index a5f111b..7476716 100644
--- a/packages/core/examples/image_labels_overlay_with_value_picking/main.ts
+++ b/packages/core/examples/image_labels_overlay_with_value_picking/main.ts
@@ -28,10 +28,6 @@ const loader = await imageSource.open();
 const attributes = await loader.loadAttributes();
 const attributesAtLod = attributes[lod];
 
-// Phase contrast limits were chosen qualitatively.
-const phaseChannelIndex = 0;
-const phaseContrastLimits: [number, number] = [20, 200];
-
 const tStartPoint = 0;
 
 const dimensionExtent = (dimensionName: string) => {
@@ -53,7 +49,7 @@ const yStopPoint = yExtent.size * yExtent.scale;
 
 const imageRegion: Region = [
   { dimension: "T", index: { type: "point", value: tStartPoint } },
-  { dimension: "C", index: { type: "point", value: phaseChannelIndex } },
+  { dimension: "C", index: { type: "full" } },
   { dimension: "Z", index: { type: "point", value: zMidPoint } },
   { dimension: "Y", index: { type: "full" } },
   { dimension: "X", index: { type: "full" } },
@@ -79,13 +75,6 @@ const imageLayer = new ImageLayer({
   source: imageSource,
   region: imageRegion,
   transparent: true,
-  channelProps: [
-    {
-      visible: true,
-      color: Color.WHITE,
-      contrastLimits: phaseContrastLimits,
-    },
-  ],
   lod,
 });
```

andy-sweet:
[...]

PR #422, opened 2025-09-04
Implement `ChannelsEnabled` for `ChunkedImageLayer`

aganders3:
### Summary
Our current method for integrating with channel controls relies on implementing the `ChannelsEnabled` interface. `ChunkedImageLayer` currently only supports a single-channel (see #372), but implementing this interface allows us to add dynamic contrast limit (window/level) controls for these layers. This is needed for porting some existing applications (Try Cryolens, Tomogram Quality) to use the `ChunkedImagelayer` for progressive chunk streaming.

It's a little weird to have to pass a list of channel props when the limit is 1. So we may want to change the interface later if we decide to only support single-channel images in this layer (relying instead on layer blending for multiple channels). I think this can be revisited when we are ready to remove `ImageLayer` and `ImageSeriesLayer`.

Tagging @phoenixAja for review, since I think this is needed for porting the Tomogram Quality app to use `ChunkedImageLayer`.

### Related Issue
#372 (does not close)

### Tests & Checks
Tested with current examples
Added window/level controls to chunk streaming example

### DIFF
--- a/packages/core/src/core/chunk_manager.ts
+++ b/packages/core/src/core/chunk_manager.ts
@@ -10,7 +10,7 @@ import { Box2 } from "../math/box2";
 import { Box3 } from "../math/box3";
 import { almostEqual } from "../utilities/almost_equal";
 import { Logger } from "../utilities/logger";
-import { OrthographicCamera } from "@/objects/cameras/orthographic_camera";
+import { OrthographicCamera } from "../objects/cameras/orthographic_camera";
 
 // Number of chunks to extend beyond the visible bounds in each direction (x/y/z)
 // These additional chunks are prefetched to improve responsiveness when panning.
--- a/packages/core/src/layers/chunked_image_layer.ts
+++ b/packages/core/src/layers/chunked_image_layer.ts
@@ -167,7 +167,7 @@ export class ChunkedImageLayer extends Layer implements ChannelsEnabled {
     const image = new ImageRenderable(
       geometry,
       Texture2DArray.createWithChunk(chunk, this.getDataForImage(chunk)),
-      this.channelProps_
+      this.channelProps_ ?? [{}]
     );
     this.updateImageChunk(image, chunk);
     return image;
--- a/packages/core/src/layers/chunked_image_layer.ts
+++ b/packages/core/src/layers/chunked_image_layer.ts
@@ -31,7 +31,7 @@ export class ChunkedImageLayer extends Layer implements ChannelsEnabled {
   private readonly visibleChunks_: Map<Chunk, ImageRenderable> = new Map();
   private readonly pool_ = new RenderablePool<ImageRenderable>();
   private readonly initialChannelProps_?: ChannelProps[];
-  private readonly channelChangeCallbacks_: Array<() => void> = [];
+  private readonly channelChangeCallbacks_:  (() => void)[] = [];
   private channelProps_?: ChannelProps[];
   private chunkManagerSource_?: ChunkManagerSource;
   private pointerDownPos_: vec2 | null = null;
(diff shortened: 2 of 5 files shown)


packages/core/src/core/chunk_manager.ts
```diff
@@ -10,7 +10,7 @@ import { Box2 } from "../math/box2";
 import { Box3 } from "../math/box3";
 import { almostEqual } from "../utilities/almost_equal";
 import { Logger } from "../utilities/logger";
-import { OrthographicCamera } from "@/objects/cameras/orthographic_camera";
+import { OrthographicCamera } from "../objects/cameras/orthographic_camera";
```
aganders3, line 13: Unrelated but I was getting a warning from the build for this.

packages/core/src/layers/chunked_image_layer.ts
```diff
@@ -55,6 +57,7 @@ export class ChunkedImageLayer extends Layer {
     this.source_ = source;
     this.sliceCoords_ = sliceCoords;
     this.channelProps_ = channelProps;
+    this.initialChannelProps_ = channelProps;
```
aganders3, line 60: We could check the length and fail early, but I think it will throw an informative enough error when it actually tries to validate/set the channels.

packages/core/src/core/chunk_manager.ts
phoenixAja: thanks for patching i was running to this as well when building for the review app!

packages/core/src/layers/chunked_image_layer.ts
```diff
@@ -161,7 +167,7 @@ export class ChunkedImageLayer extends Layer {
     const image = new ImageRenderable(
       geometry,
       Texture2DArray.createWithChunk(chunk, this.getDataForImage(chunk)),
-      this.channelProps_ ? [this.channelProps_] : [{}]
```
phoenixAja, line 164: is it safe to remove this default?

aganders3: Ah, you may be right. Typescript doesn't catch it because the `ImageRenderable` constructor has a default, but that's an empty array rather than an array with one empty object. I'll take a closer look.

aganders3: It's not safe, but it fails reasonably with this error:

> Number of channels (0) must match depth of texture (1).

This is the same for the `ImageLayer` and `ImageSeriesLayer` so I think we could leave it that way, but I'll put it back.

The right fix may be to eventually make the channel props required in more places instead of optional.

packages/core/src/layers/chunked_image_layer.ts
```diff
@@ -18,19 +18,21 @@ import { RenderablePool } from "../utilities/renderable_pool";
...
   public readonly type = "ChunkedImageLayer";
 
   private readonly source_: ChunkSource;
   private readonly sliceCoords_: SliceCoordinates;
   private readonly onPickValue_?: (info: PointPickingResult) => void;
   private readonly visibleChunks_: Map<Chunk, ImageRenderable> = new Map();
-  private readonly channelProps_?: ChannelProps;
   private readonly pool_ = new RenderablePool<ImageRenderable>();
+  private readonly initialChannelProps_?: ChannelProps[];
+  private readonly channelChangeCallbacks_: Array<() => void> = [];
```
phoenixAja: nit: I kind of like this syntax better:
```suggestion
  private readonly channelChangeCallbacks_:  (() => void)[] = [];
```

packages/core/src/layers/chunked_image_layer.ts
phoenixAja: agreed

phoenixAja (approved):
look good and works locally

PR #424, opened 2025-09-05
Don't include core-prerelease in rollup of react-prerelease

bchu1:
Find/replace was missed during the migration.

Tested that:
 - `yalc add @idetik/core-prerelease` works now.
 - VCP with `react-prerelease@1.1.1` works.

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@idetik/react-prerelease",
-  "version": "1.1.0",
+  "version": "1.1.1",
   "type": "module",
   "scripts": {
     "dev": "vite",
--- a/packages/react/vite.config.js
+++ b/packages/react/vite.config.js
@@ -57,7 +57,7 @@ export default defineConfig(() => {
             "react-dom/client": "ReactDOM.client",
             "@czi-sds/components": "SDS",
             "@mui/material": "MaterialUI",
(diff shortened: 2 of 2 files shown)


packages/react/vite.config.js
```diff
@@ -27,60 +27,55 @@ const plugins = [
...
-            '@czi-sds/components': 'SDS',
-            '@mui/material': 'MaterialUI',
-            '@idetik/core': 'IdetikCore',
+            react: "React",
+            "react-dom": "ReactDOM",
+            "react/jsx-runtime": "React.jsxRuntime",
+            "react-dom/client": "ReactDOM.client",
+            "@czi-sds/components": "SDS",
+            "@mui/material": "MaterialUI",
+            "@idetik/core": "IdetikCore",
```
aganders3:
```suggestion
            "@idetik/core-prerelease": "IdetikCorePre",
```

aganders3: I think this is only for usage without a build system or something.

packages/react/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@idetik/react-prerelease",
-  "version": "1.1.0",
+  "version": "1.1.1",
```
aganders3, line 3: Did you publish this version as well?

aganders3 (approved):
LGTM - thanks!

PR #425, opened 2025-09-05
fix: submodule compatibility with Next: replace vite env, add vendored stats types, adjust hook import

phoenixAja:
### Context
After switching back to consuming idetik/core via git submodule, Next.js started compiling core source directly. This surfaced build/type issues tied to vite-specific APIs and missing type declarations.

### Changes

Logger: replaced import.meta.env.MODE (vite-only) with process.env.NODE_ENV (works in both vite and next).

Stats: removed @ts-expect-error and added a proper declaration module for ./vendored/stats.js, re-exporting Stats/StatsPanel types. createStats now returns the declared type.

React component: changed useIdetik import from @idetik/core-prerelease to the local hooks path, avoiding bundling the entire core package in the app.

### Why

Next doesn’t define import.meta.env, so the build crashed.

TypeScript needs declarations for plain JS modules; without them, the build failed.

The broad hook import caused Next to resolve through the package entrypoint (and its problematic exports). Using a relative import keeps the build clean.

### Impact

No runtime logic changes.

Core and app both build successfully under Next after these adjustments.

Logging and stats behave the same; only typing and compatibility changed.

### DIFF
--- a/packages/core/src/utilities/logger.ts
+++ b/packages/core/src/utilities/logger.ts
@@ -29,15 +29,6 @@ type Module =
   | "WebGLTexture"
   | "WireframeGeometry";
 
-// function getMode(): "production" | "development" | "test" {
-//   const nodeEnv = process?.env?.NODE_ENV;
-//   return nodeEnv === "production"
-//     ? "production"
-//     : nodeEnv === "test"
-//       ? "test"
-//       : "development";
-// }
-
 function getMode(): "production" | "development" | "test" {
   if (typeof process !== "undefined" && process.env && process.env.NODE_ENV) {
     return process.env.NODE_ENV === "production"
--- a/packages/core/src/utilities/logger.ts
+++ b/packages/core/src/utilities/logger.ts
@@ -30,24 +30,32 @@ type Module =
   | "WebGLTexture"
   | "WireframeGeometry";
 
-function getMode(): "production" | "development" | "test" {
-  if (typeof process !== "undefined" && process.env && process.env.NODE_ENV) {
-    return process.env.NODE_ENV === "production"
-      ? "production"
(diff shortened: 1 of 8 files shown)


packages/core/src/utilities/vendored/stats.js.d.ts
```diff
@@ -0,0 +1,28 @@
+export type StatsPanel = {
```
phoenixAja, line 2: i realize that this is duplicate code, but we need both for the build

bchu1: could we add a comment?

packages/core/src/utilities/logger.ts
```diff
@@ -30,9 +30,29 @@ type Module =
   | "WebGLTexture"
   | "WireframeGeometry";
 
+function getMode(): "production" | "development" | "test" {
```
bchu1: since this is only for the log level, maybe we only need production and not production? Or just directly make it `getLogLevel()`

packages/core/src/utilities/logger.ts
```diff
@@ -30,9 +30,29 @@ type Module =
...
+      : process.env.NODE_ENV === "test"
+        ? "test"
+        : "development";
+  }
+  // Fallback for browser/test environments, type-safe version
+  if (
+    typeof window !== "undefined" &&
+    typeof (window as { NODE_ENV?: string }).NODE_ENV === "string"
+  ) {
+    const env = (window as { NODE_ENV?: string }).NODE_ENV;
```
bchu1: seems like this return is doing the same thing as the if above, could both have the same logic

bchu1 (approved):

PR #430, opened 2025-09-09
Adjust LOD factor based on source data scale

### LINKED ISSUE #426: LOD calculation is not correct or universal
I believe there is an issue with LOD calculation that depends on the image source. I think I understand the root cause here, which is that the LOD calculation is based on virtual units per screen pixel, but that assumes 1:1 ratio of virtual units to image pixels (at LOD0). 

When creating geometry for our image tiles, we set the coordinates based on `chunk.size * chunk.scale`. This has different effects depending on the values in the OME-Zarr metadata for image scale. For CryoET data for example each pixel at LOD0 is ~5 or ~10 virtual units. Zebrahub data on the other hand (used in the chunk streaming example) will have each pixel at LOD0 representing 0.439 virtual units.

So, for CryoET data we have to zoom in wayyyy to far to get LOD0, and for Zebrahub we may switch to higher resolution tiles too early. I need to do some more testing, because experimentally this still doesn't quite match for me. Maybe it has to do with the size of the pixels at the lowest resolution LOD instead of LOD0?

One way to see this: when zooming, there should not be an appreciable quality change at LOD steps, but here you can clearly see the quality degrade when moving from LOD1 to LOD2:

https://github.com/user-attachments/assets/b0deed3e-509b-4dc4-9ff6-81ea6d52af20
(this is from the ome zarr v0.5 example, which uses some [public marmoset neuron data](https://ome-zarr-scivis.s3.us-east-1.amazonaws.com/v0.5/96x2/marmoset_neurons.ome.zarr/zarr.json
))

This will also benefit from some configuration as we've discussed (e.g. bias term, LOD limits) - I'm not sure where such a configuration API should be (on the layer? on the source?).

So to summarize, I think we have two tasks:
* Confirm LOD calculation is _correct_
* Provide a good API to make it do what we _want_ (per-application performance/UX)

aganders3:
### Summary
This PR updates the LOD calculation system to incorporate the image scale (per-source).

This also introduces an LOD "bias" term on the source, that can be used to nudge the calculation towards more or less LOD. It's currently hard-coded to 0.5, which nudges to a slightly lower-resolution than "necessary". This PR does not currently add an API to change this bias, so I added a TODO comment. I'm not sure if this bias should be per-source, per-layer, or perhaps per-viewport, and what the user API should look like.

Note: all our LOD calculation is currently based on width/x. We will need to make this more general for anisotropic resolutions and additional slice orientations.

### Related Issue
Closes #426 

### Tests & Checks
I tested by making a custom example page that syncs cameras, and loads test zarr files with three different scales: 10, 1, and 0.1. All three have the exact same image content with a 4-level pyramid from 2048x2048 to 256x256 (chunk size 512x512, or 256x256 at LOD3).

On `main` you can see the images display at different LODs even though they are the same underlying number of pixels shown at the same size (canvases are about 600px square):

https://github.com/user-attachments/assets/e7ec0b64-3202-42eb-80bb-4c98246c9925

With these proposed changes, you can see the LOD is kept in sync across images with different scales:

https://github.com/user-attachments/assets/5976ff13-41b8-4b05-a139-93da6c27951e

### DIFF
--- a/packages/core/src/core/chunk_manager.ts
+++ b/packages/core/src/core/chunk_manager.ts
@@ -28,6 +28,9 @@ export class ChunkManagerSource {
   private readonly sliceCoords_: SliceCoordinates;
   private readonly dimensions_: SourceDimensionMap;
   private currentLOD_: number = 0;
+  // TODO: make LOD bias configurable per-source or per-layer
+  // positive values nudge towards coarser resolution (higher LOD number)
+  private lodBias_: number = 0.5;
   private lastViewBounds2D_: Box2 | null = null;
   private lastZBounds_?: [number, number];
 
@@ -154,10 +157,13 @@ export class ChunkManagerSource {
   }
 
   private setLOD(lodFactor: number): void {
+    const sourceAdjustment =
+      this.lodBias_ - Math.log2(this.dimensions_.x.lods[0].scale);
+    const sourceAdjustedLodFactor = sourceAdjustment - lodFactor;
     const maxLOD = this.lowestResLOD_;
     const targetLOD = Math.max(
       0,
-      Math.min(maxLOD, Math.floor(maxLOD - lodFactor))
+      Math.min(maxLOD, Math.floor(sourceAdjustedLodFactor))
     );
 
     if (targetLOD !== this.currentLOD_) {
--- a/packages/core/src/core/chunk_manager.ts
+++ b/packages/core/src/core/chunk_manager.ts
@@ -157,6 +157,13 @@ export class ChunkManagerSource {
   }
 
   private setLOD(lodFactor: number): void {
+    // `scale` here is the x-width of an image pixel in virtual units at LOD 0.
+    // So (ignoring the bias term) subtracting `lodFactor` from `Math.log2(scale)`
+    // is effectively `Math.log2(virtualUnitsPerScreenPixel / xScale)`.
+    // That is, `adjustedLodFactor = Math.log2(imagePixelsPerScreenPixel)`;
+    // or in other words, how many image pixels (LOD 0) fit in a screen pixel.
+    // Use of log2 here and in ChunkManager relies on the assumption that
+    // each LOD is downsampled by a factor of 2 in X and Y.
     const sourceAdjustment =
       this.lodBias_ - Math.log2(this.dimensions_.x.lods[0].scale);
     const sourceAdjustedLodFactor = sourceAdjustment - lodFactor;


packages/core/src/core/chunk_manager.ts
```diff
@@ -28,6 +28,9 @@ export class ChunkManagerSource {
   private readonly sliceCoords_: SliceCoordinates;
   private readonly dimensions_: SourceDimensionMap;
   private currentLOD_: number = 0;
+  // TODO: make LOD bias configurable per-source or per-layer
+  // positive values nudge towards coarser resolution (higher LOD number)
+  private lodBias_: number = 0.5;
```
andy-sweet, line 33: This default value effectively turns the `floor` into a `round` right? That seems like a reasonable enough default, if we're not always flooring to satisfy the density requirements. I'm still warming up to the idea of this bias term as something we'd make users explicitly think about.

packages/core/src/core/chunk_manager.ts
```diff
@@ -154,10 +157,13 @@ export class ChunkManagerSource {
   }
 
   private setLOD(lodFactor: number): void {
+    const sourceAdjustment =
+      this.lodBias_ - Math.log2(this.dimensions_.x.lods[0].scale);
```
andy-sweet, line 168: Ignoring the bias term, I think the sourceAdjustedLodFactor is the log of the ratio between the units per screen pixel in the viewport and the units per image pixel (at LOD0). Is that correct? Given this calculation is separated from the other `lodFactor`, I wonder if adding some explanatory comments might be worth it here.

Another thought is to hold off on the log in `ChunkManager` and pass through `virtualUnitsPerScreenPixel`, then explicitly compute the log of the ratio here, which is easier for me to reason about (though I guess a little more compute work if there are lots of sources).

andy-sweet (approved):
This looks like an improvement to me.

In line with the comments in the PR description about only using the width, I wonder if we can also relax the constraint around factors of 2 when picking the LOD in the future, which could also help to improve readability here. Though this PR was at least a nice logarithm refresher.

packages/core/src/core/chunk_manager.ts
aganders3: LOD factors can also be negative but without thinking too hard about it that sounds correct to me.

I'm also not totally convinced about exposing this directly to users, hence not adding an API for it here. Neuroglancer provides the little resolution widget that I don't find particularly easy to use, but I would imagine adding low/high LOD limits and a bias term pretty intuitive.

packages/core/src/core/chunk_manager.ts
aganders3: Yes, you are correct. I can add a comment about it though.

Initially I was calculating it as `log2(imagePixelsPerScreenPixel)` in `ChunkManager`. This was quite clear, but I thought accessing `source.dimensions.x.lods[0].scale` felt like a leaky abstraction.

To your point I think this is likely to change significantly if we start using both directions and/or relaxing the factors-of-2 requirements (which we may need/want for orthogonal slices).

aganders3: > Another thought is to hold off on the log in ChunkManager and pass through virtualUnitsPerScreenPixel, then explicitly compute the log of the ratio here, which is easier for me to reason about (though I guess a little more compute work if there are lots of sources).

I kind of like this - is it more compute? It might be less work since it could combine the two log2 calls into one (and a division). In any case I'm sure it's fast enough.

aganders3: I'm going to leave it as-is with the comment because I think this will be revisited for multiple-viewports in the near future, but I'd like it to be fixed in the mean time.

PR #431, opened 2025-09-09
Add queue for cancellable limited concurrent fetches

shlomnissan:
### Summary

This PR adds a chunk-aware fetch queue that:
- Limits concurrent requests
- Aborts in-flight requests if chunks become unnecessary
- Respects priority
- Keeps chunk state transitions consistent

### Notes

- This implementation doesn't respect sequencing within the same priority. I will need to address this when implementing loading order.

### Tests & Checks

- All existing examples are working as expected
- Added test suite for the chunk queue
- Manual verification:
  - Use the console log, filter `ChunkQueue` logs to verify cancellations
  - Change `MAX_CONCURRENT` in the `ChunkQueue` + throttling to verify concurrent connection limit
  - See how chunks are loaded top left in a linear progression

### DIFF
--- a/packages/core/src/core/chunk_manager.ts
+++ b/packages/core/src/core/chunk_manager.ts
@@ -11,6 +11,7 @@ import { Box3 } from "../math/box3";
 import { almostEqual } from "../utilities/almost_equal";
 import { Logger } from "../utilities/logger";
 import { OrthographicCamera } from "../objects/cameras/orthographic_camera";
+import { ChunkQueue } from "@/utilities/chunk_queue";
 
 // Number of chunks to extend beyond the visible bounds in each direction (x/y/z)
 // These additional chunks are prefetched to improve responsiveness when panning.
@@ -149,8 +150,8 @@ export class ChunkManagerSource {
     return this.currentLOD_;
   }
 
-  public loadChunkData(chunk: Chunk) {
-    return this.loader_.loadChunkData(chunk, this.sliceCoords_);
+  public loadChunkData(chunk: Chunk, signal: AbortSignal) {
+    return this.loader_.loadChunkData(chunk, this.sliceCoords_, signal);
   }
 
   private setLOD(lodFactor: number): void {
@@ -348,11 +349,9 @@ export class ChunkManagerSource {
   }
 }
 
-type BucketItem = { source: ChunkManagerSource; chunk: Chunk };
-type Buckets = BucketItem[][];
-
 export class ChunkManager {
   private readonly sources_ = new Map<ChunkSource, ChunkManagerSource>();
+  private readonly queue_ = new ChunkQueue();
 
   public async addSource(source: ChunkSource, sliceCoords: SliceCoordinates) {
     let existing = this.sources_.get(source);
@@ -377,29 +376,16 @@ export class ChunkManager {
     const virtualUnitsPerScreenPixel = virtualWidth / bufferWidth;
     const lodFactor = Math.log2(1 / virtualUnitsPerScreenPixel);
 
-    const buckets: Buckets = [[], [], [], []];
-
     for (const [_, source] of this.sources_) {
       source.update(lodFactor, viewBounds2D);
       for (const chunk of source.chunks) {
-        if (chunk.state !== "queued" || chunk.priority === null) continue;
-        buckets[chunk.priority].push({ source, chunk });
-      }
-    }
-
-    for (const bucket of buckets) {
-      for (const { source, chunk } of bucket) {
-        if (chunk.state !== "queued") continue; // guard
-        chunk.state = "loading";
-        source
-          .loadChunkData(chunk)
-          .then(() => {
-            if (chunk.state === "loading") chunk.state = "loaded";
-          })
-          .catch((err) => {
-            Logger.error("ChunkManager", String(err));
-            if (chunk.state === "loading") chunk.state = "unloaded";
-          });
+        if (chunk.priority === null) {
+          this.queue_.cancel(chunk);
+        } else if (chunk.state === "queued") {
+          this.queue_.enqueue(chunk, (signal) =>
+            source.loadChunkData(chunk, signal)
+          );
+        }
       }
     }
   }
--- a/packages/core/src/core/chunk_manager.ts
+++ b/packages/core/src/core/chunk_manager.ts
@@ -11,7 +11,7 @@ import { Box3 } from "../math/box3";
 import { almostEqual } from "../utilities/almost_equal";
 import { Logger } from "../utilities/logger";
 import { OrthographicCamera } from "../objects/cameras/orthographic_camera";
-import { ChunkQueue } from "@/utilities/chunk_queue";
+import { ChunkQueue } from "../data/chunk_queue";
 
 // Number of chunks to extend beyond the visible bounds in each direction (x/y/z)
 // These additional chunks are prefetched to improve responsiveness when panning.
@@ -388,5 +388,7 @@ export class ChunkManager {
         }
       }
     }
+
+    this.queue_.flush();
   }
 }
--- a/packages/core/src/utilities/chunk_queue.ts
+++ b/packages/core/src/utilities/chunk_queue.ts
@@ -1,0 +1,80 @@
+import { Chunk } from "@/data/chunk";
+import { Logger } from "./logger";
+
+const MAX_CONCURRENT = 8;
+
+type LoaderFn = (signal: AbortSignal) => Promise<void>;
+
+type PendingItem = { chunk: Chunk; fn: LoaderFn };
+
+export class ChunkQueue {
+  private readonly pending_: PendingItem[] = [];
+  private readonly running_ = new Map<
+    Chunk,
+    { controller: AbortController; promise: Promise<void> }
+  >();
+
+  enqueue(chunk: Chunk, fn: LoaderFn) {
+    if (this.running_.has(chunk)) return;
+    if (this.pending_.some((p) => p.chunk === chunk)) return;
+
+    this.pending_.push({ chunk, fn });
+    this.pump();
+  }
+
+  cancel(chunk: Chunk) {
+    const idx = this.pending_.findIndex((p) => p.chunk === chunk);
+    if (idx >= 0) {
+      this.pending_.splice(idx, 1);
+      Logger.debug("ChunkQueue", "Cancelled pending request");
+      return;
+    }
+
+    const running = this.running_.get(chunk);
+    if (running) {
+      running.controller.abort();
+      Logger.debug("ChunkQueue", "Cancelled fetch request");
+    }
+  }
+
+  private pump() {
+    while (this.running_.size < MAX_CONCURRENT && this.pending_.length > 0) {
+      this.pending_.sort((a, b) => {
+        return (
+          (a.chunk.priority ?? Number.MAX_SAFE_INTEGER) -
+          (b.chunk.priority ?? Number.MAX_SAFE_INTEGER)
+        );
+      });
+      this.start(this.pending_.shift()!);
+    }
+  }
+
+  private start(item: PendingItem) {
+    const { chunk, fn } = item;
+    chunk.state = "loading";
+
+    const controller = new AbortController();
+    const promise = Promise.resolve()
+      .then(() => fn(controller.signal))
+      .then(
+        () => {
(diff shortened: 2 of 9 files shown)


packages/core/src/core/chunk_manager.ts
```diff
@@ -11,6 +11,7 @@ import { Box3 } from "../math/box3";
 import { almostEqual } from "../utilities/almost_equal";
 import { Logger } from "../utilities/logger";
 import { OrthographicCamera } from "../objects/cameras/orthographic_camera";
+import { ChunkQueue } from "@/utilities/chunk_queue";
```
andy-sweet: Nit: for consistency

```suggestion
import { ChunkQueue } from "../utilities/chunk_queue";
```

packages/core/src/utilities/chunk_queue.ts
```diff
@@ -0,0 +1,85 @@
+import { Chunk } from "@/data/chunk";
```
andy-sweet: Nit: for consistency

```suggestion
import { Chunk } from "../data/chunk";
```

packages/core/src/utilities/chunk_queue.ts
andy-sweet, line 1: Nit: since this is specifically a chunk queue (and not a general purpose priority queue), I'd expect this to live closer to the chunk module (e.g. in data vs. general utils).

packages/core/test/helpers.ts
```diff
@@ -0,0 +1,81 @@
...
+  const { shape, chunkIndex, scale, offset, ...rest } = overrides;
+
+  const defaultChunk: Chunk = {
+    state: "unloaded",
+    lod: 0,
+    visible: false,
+    prefetch: false,
+    priority: null,
+    shape: { x: 256, y: 256, z: 256, c: 1 },
+    rowStride: 256,
```
andy-sweet, line 22: For this helper, the row stride should probably just always equal `shape.x` for now. At the moment, a custom shape can be provided, which may then be inconsistent with this default.

packages/core/src/utilities/chunk_queue.ts
```diff
@@ -0,0 +1,85 @@
...
+      .then(
+        () => {
+          if (chunk.state === "loading") chunk.state = "loaded";
+          this.running_.delete(chunk);
+          this.pump();
+        },
+        (err) => {
+          if (chunk.state === "loading") chunk.state = "unloaded";
+          this.running_.delete(chunk);
+          this.pump();
```
andy-sweet: Does it make sense for these shared lines currently in both the resolve and reject functions, to be in a finally clause?

packages/core/src/utilities/chunk_queue.ts
```diff
@@ -0,0 +1,85 @@
...
+      Logger.debug("ChunkQueue", "Cancelled fetch request");
+    }
+  }
+
+  private pump() {
+    while (
+      this.running_.size < this.maxConcurrent_ &&
+      this.pending_.length > 0
+    ) {
+      this.pending_.sort((a, b) => {
```
andy-sweet: I think this can be moved outside the loop because shift does not change the order, start doesn't either, and this function is not async.

packages/core/src/utilities/chunk_queue.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  enqueue(chunk: Chunk, fn: LoaderFn) {
+    if (this.running_.has(chunk)) return;
+    if (this.pending_.some((p) => p.chunk === chunk)) return;
+
+    this.pending_.push({ chunk, fn });
+    this.pump();
+  }
+
+  cancel(chunk: Chunk) {
```
andy-sweet: What do you think about adding a `cancelAll` or `shutdown` method for this class? This would be useful for handling the case when Idetik is not constantly running in a web app (e.g. the user can navigate away from it, which is the case in OrganelleBox), because then we could abort all the running tasks (which wouldn't be needed anymore).

This might be better as a follow up task because we're not doing much on `Idetik.stop` (other than stopping the animation loop), and don't have explicit disposal/shutdown at the runtime level.

packages/core/test/chunk_queue.test.ts
```diff
@@ -0,0 +1,123 @@
...
+
+    queue.enqueue(chunk, ctl.loader);
+    await waitFor(() => chunk.state === "loading");
+
+    ctl.resolve();
+    await waitFor(() => chunk.state === "loaded");
+  });
+
+  test("starts up to max concurrent tasks and queues the rest", async () => {
+    const queue = new ChunkQueue();
```
andy-sweet: To make this test more readable, the max concurrent tasks should be passed to the constructor here rather than relying on the default which is defined elsewhere.

```suggestion
    const queue = new ChunkQueue(8);
```

Or you could use a smaller value than 8 here.

packages/core/src/utilities/chunk_queue.ts
```diff
@@ -0,0 +1,85 @@
...
+          (b.chunk.priority ?? Number.MAX_SAFE_INTEGER)
+        );
+      });
+      this.start(this.pending_.shift()!);
+    }
+  }
+
+  private start(item: PendingItem) {
+    const { chunk, fn } = item;
+    chunk.state = "loading";
```
andy-sweet: I think I better understand your comment about chunk loading states being hard to follow right now since they are mutated here and in the chunk manager source.

andy-sweet (approved):
Looks great and works well. All comments are optional.

andy-sweet:
> This implementation doesn't respect sequencing within the same priority. I will need to address this when implementing loading order.

If I disable the browser cache and run the chunk streaming example, I'm seeing that the visible LOD1 chunks are loaded/rendered before the LOD2 visible chunks, which I would not expect even with the caveat above (since LOD2 chunks are always the highest priority). This is different to main. What am I missing?

andy-sweet:
> If I disable the browser cache and run the chunk streaming example, I'm seeing that the visible LOD1 chunks are loaded/rendered before the LOD2 visible chunks, which I would not expect even with the caveat above (since LOD2 chunks are always the highest priority). This is different to main. What am I missing?

I think this is because `source.chunks` is sorted by increasing LOD (compared to `getChunks` which puts the fallback chunks first), which is used when iteratively enqueuing the chunk tasks one by one. Some ideas: batch enqueue multiple tasks at once (so they are sorted before they are started), use a sorted version of the chunks.

packages/core/src/core/chunk_manager.ts
```diff
@@ -377,29 +376,16 @@ export class ChunkManager {
...
-            if (chunk.state === "loading") chunk.state = "loaded";
-          })
-          .catch((err) => {
-            Logger.error("ChunkManager", String(err));
-            if (chunk.state === "loading") chunk.state = "unloaded";
-          });
+        if (chunk.priority === null) {
+          this.queue_.cancel(chunk);
+        } else if (chunk.state === "queued") {
+          this.queue_.enqueue(chunk, (signal) =>
```
andy-sweet, line 385: As noted in [my comment](https://github.com/chanzuckerberg/idetik/pull/431#issuecomment-3275350459), I think this does not respect chunk priority across all the chunks, since the order of `source.chunks` determines which tasks get started first.

I think this should probably be fixed in this PR.

andy-sweet (changes requested):
Sorry, I do have one blocking request related to chunk priority.

aganders3:
@andy-sweet provided a much more thorough review, but I just want to say the in-flight cancellation here is awesome! That's going to help a lot.

packages/core/src/utilities/chunk_queue.ts
shlomnissan: I moved `chunk_queue.ts` to the data directory.

packages/core/src/utilities/chunk_queue.ts
shlomnissan: Sure, we can move them.

packages/core/test/helpers.ts
shlomnissan: Great call! I omitted `rowStride` from the chunk overrides and I'm setting it manually based on the shape.

packages/core/src/utilities/chunk_queue.ts
shlomnissan: I originally passed a `ChunkManagerSource` ID into the queue and exposed a `cancelAll(sourceId)`, but I removed it since we are refactoring these abstractions. I agree this is better as a follow-up, we should first align on disposal/shutdown semantics across the runtime, then add source-scoped or runtime-scoped cancellation.

packages/core/test/chunk_queue.test.ts
shlomnissan: I'll stick to 8 since a lower value doesn't change the logic and it will be nice to test under default conditions.

packages/core/src/core/chunk_manager.ts
shlomnissan: Good catch, I missed this initially. The issue is “start-on-enqueue” as `source.chunks` is iterated, early LOD1 items grab free slots before LOD2 is enqueued, so priority isn’t honored globally.

Instead of starting inside `enqueue()`, I’ll decouple start from enqueue by adding a `flush()` method and have the chunk manager call it once all chunks are enqueued per frame. This lets the queue sort the full pending set. I’ll add a test for this behavior.

The trade-off is callers must remember to call `flush()` but I think it's fine / we can add an optional auto-flush later if needed.

shlomnissan: This test `respects priority ordering when starting` should now verify prioritization properly.

packages/core/test/chunk_queue.test.ts
```diff
@@ -0,0 +1,122 @@
...
+    queue.enqueue(pendingChunk, pendingCtl.loader);
+
+    queue.flush();
+    await waitFor(() => blockerChunk.state === "loading");
+
+    queue.cancel(pendingChunk);
+
+    blockerCtl.resolve();
+    await waitFor(() => blockerChunk.state === "loaded");
+    await Promise.resolve(); // microtask tick
```
andy-sweet: The existing comment here is useful for explaining what this line does, but it would be more useful to explain why it's needed for this test.

I think that the conditions asserted below will be true if this line is not here.

Instead I think this line is intended to demonstrate that even if there's an extra tick, these conditions are still true.

Would be good to clarify that inline.

Another thought is that I might expect some public way to understand more about the state of the chunk queue, which could actually give you something to await (either a condition or a promise). For example, the number of running and/or pending tasks. Or maybe `flush` could return a `Promise` or array of `Promises` that could be awaited. Then you wouldn't need this or an explanation for it. I prefer this approach as being less mystical and implementation-dependent.

andy-sweet (approved):
I have a remaining issue with one of the tests, but it's not blocking and I'll leave it up to you to decide how to handle it.

packages/core/test/chunk_queue.test.ts
shlomnissan: > Instead I think this line is intended to demonstrate that even if there's an extra tick

Yeah, it's there to prove the invariant holds even if the queue schedules additional.

I like the idea of exposing information about the chunk queue state. Something like the following can simplify the tests:

```ts
get pendingCount() { return this.pending_.length; }
get runningCount() { return this.running_.size; }
```

I might rewrite the tests tomorrow morning with these two accessors in place before merging.

shlomnissan: I added accessors and a helper

```ts
test("respects priority ordering when starting", async () => {
  const queue = new ChunkQueue();

  const started: number[] = [];
  const ctl0 = makeControllableLoader(() => started.push(0));
  const ctl1 = makeControllableLoader(() => started.push(1));
  const ctl5 = makeControllableLoader(() => started.push(5));

  queue.enqueue(makeChunk({ state: "queued", priority: 5 }), ctl5.loader);
  queue.enqueue(makeChunk({ state: "queued", priority: 1 }), ctl1.loader);
  queue.enqueue(makeChunk({ state: "queued", priority: 0 }), ctl0.loader);

  queue.flush();
  await expectCounts(queue, { pending: 0, running: 3 });
  expect(started).toEqual([0, 1, 5]);

  ctl0.resolve();
  ctl1.resolve();
  ctl5.resolve();
  await expectCounts(queue, { pending: 0, running: 0 });
});
```

I also refactored all existing tests and added a new one: `backfills by priority when slots free up`

PR #432, opened 2025-09-09
Support time in chunk manager and chunked image layer

### LINKED ISSUE #453: Support time dimension in the chunk manager
The goal of this task is to support time in the chunk manager, such that we could pass in a slice coordinates that can change in time and load/render appropriately. The chunk manager demo should be updated to allow basic control in the time dimension as for the z-dimension.

andy-sweet:
### Summary
This change provides basic support for the time dimension in the chunk manager, with no prefetching.

The biggest change here is that `ChunkManagerSource.chunks_` is now a 2D array instead of 1D. The outer dimension represents time, so that we can efficiently update only the chunks at the current or previous time point.

The chunk manager source only works under some constraints about the time dimension in the chunk source, and will throw on construction if those are not met.

### Related Issue
Closes #453 

### Tests & Checks
I manually tested my changes in the examples that use the chunk manager.


packages/core/src/core/chunk_manager.ts
```diff
@@ -70,41 +74,44 @@ export class ChunkManagerSource {
...
+                  z: chunkDepth,
+                  c: channels,
+                },
+                rowStride: chunkWidth,
+                rowAlignmentBytes: 1,
+                chunkIndex: { x, y, z, t },
+                scale: {
+                  x: xLod.scale,
+                  y: yLod.scale,
+                  z: zLod?.scale ?? 1,
```
andy-sweet, line 97: Note that I did not add t to scale, offset, or shape because they are unused elsewhere.

packages/core/src/core/chunk_manager.ts
```diff
@@ -129,13 +136,23 @@ export class ChunkManagerSource {
...
       this.viewBounds2DChanged(viewBounds2D) ||
-      this.zBoundsChanged(zBounds)
+      this.zBoundsChanged(zBounds) ||
+      this.lastTCoord_ !== this.sliceCoords_.t
     ) {
-      this.updateChunkVisibility(viewBounds2D);
+      updatedChunks = this.updateChunkVisibility(viewBounds2D);
     }
+
+    this.lastViewBounds2D_ = new Box2(
```
andy-sweet: Always assigning values here is to fix a bug that was present before this change, which is that the `||` was previously preventing updates when needed. As we also check t here, this bug becomes more likely to occur.

andy-sweet: Note that I'm not sure if the defensive clone is actually needed here, but I kept it for consistency.

packages/core/src/core/chunk_manager.ts
```diff
@@ -410,8 +459,8 @@ export class ChunkManager {
     const lodFactor = Math.log2(1 / virtualUnitsPerScreenPixel);
 
     for (const [_, source] of this.sources_) {
-      source.update(lodFactor, viewBounds2D);
-      for (const chunk of source.chunks) {
+      const updatedChunks = source.update(lodFactor, viewBounds2D);
```
andy-sweet: Returning the chunks that were updated (e.g. those at the previous and current time point) was the simplest solution I could think of to avoid iterating over all chunks (at all time points) here. Open to other ideas.

This pattern could be useful to avoid unnecessary checks in other dimensions, which could be useful as we see more zarr v3 datasets with more/smaller chunks.

packages/core/src/core/chunk_manager.ts
```diff
@@ -43,9 +45,11 @@ export class ChunkManagerSource {
     this.sliceCoords_ = sliceCoords;
 
     this.validateXYScaleRatios();
+    this.validateTimeDimension();
 
     // generate chunks for each LOD without loading data
-    this.chunks_ = [];
+    const chunksT = this.dimensions_.t?.lods[0].size ?? 1;
```
shlomnissan: This line is fine since `t` is already validated, so using `lods[0]` to derive `chunksT` is consistent, but it's not obvious until you read `validateTimeDimension`. It might be clearer to merge validation and initialization into a single step. For example, replacing `validateTimeDimension` with something like `timeLength()`. This way all assumptions about the `t` are defined in one place, and invariants are verified at the same time we compute the usable length.

packages/core/src/core/chunk_manager.ts
```diff
@@ -70,41 +74,44 @@ export class ChunkManagerSource {
...
-                y: yLod.scale,
-                z: zLod?.scale ?? 1,
-              },
-              offset: {
-                x: xOffset,
-                y: yOffset,
-                z: zOffset,
-              },
-            });
+            for (let t = 0; t < chunksT; ++t) {
```
shlomnissan: Consider flipping the loop order so that `t` is the outermost dimension instead of the innermost. It doesn’t change the total number of iterations, but it makes the control flow clearer (we build one full `this.chunks_[t]` slice at a time) and avoids repeatedly indexing into `this.chunks_[t]` deep inside nested loops which improves readability and keeps all chunk construction for a given timepoint localized.

packages/core/src/core/chunk_manager.ts
shlomnissan: We already have a defensive clone in the `Box2` constructor so I think we should remove it.

packages/core/examples/chunk_streaming/chunk_info_overlay.ts
```diff
@@ -23,7 +23,7 @@ export class ChunkInfoOverlay {
       return;
     }
 
-    const allChunks = chunkManagerSource.chunks;
+    const allChunks = chunkManagerSource.chunks.flat();
```
shlomnissan: I think this means we are iterating through all chunks in every frame. Iteration is over `2ms` per frame on my computer. Maybe we should return chunks per time point and make it clear in the info.

packages/core/src/core/chunk_manager.ts
```diff
@@ -146,7 +163,7 @@ export class ChunkManagerSource {
     return this.dimensions_;
   }
 
-  public get chunks(): Chunk[] {
+  public get chunks(): Chunk[][] {
```
shlomnissan: Given that we want to prevent explicit iteration over all the chunks, should we return chunks for the current time point only? This can potentially simplify some other changes.

packages/core/src/core/chunk_manager.ts
```diff
@@ -129,13 +136,23 @@ export class ChunkManagerSource {
...
+    let updatedChunks: Chunk[] = [];
 
     if (
       this.viewBounds2DChanged(viewBounds2D) ||
-      this.zBoundsChanged(zBounds)
+      this.zBoundsChanged(zBounds) ||
+      this.lastTCoord_ !== this.sliceCoords_.t
     ) {
-      this.updateChunkVisibility(viewBounds2D);
+      updatedChunks = this.updateChunkVisibility(viewBounds2D);
```
shlomnissan: If we are going to return chunks from the `update()` function, we should probably rename it. My expectation from an `update()` function is updating some internal state.

packages/core/src/core/chunk_manager.ts
```diff
@@ -203,7 +221,21 @@ export class ChunkManagerSource {
...
+      this.lastTCoord_ !== this.sliceCoords_.t
+    ) {
+      const lastTimeChunks = this.chunks_[this.lastTCoord_];
+      for (const chunk of lastTimeChunks) {
+        chunk.visible = false;
+        chunk.prefetch = false;
+        this.disposeChunk(chunk);
+      }
+      updatedChunks.push(...lastTimeChunks);
+    }
```
shlomnissan: I don't think this is misplaced, but I feel like making `updateChunkVisibility` local to a single time point might improve readability.

```ts
// ChunkManagerSource::update

if (this.lastTCoord_ !== this.sliceCoords_.t) {
  this.disposePreviousTimePointChunks();
}

if (this.viewBounds2DChanged(viewBounds2D) || this.zBoundsChanged(zBounds)) {
  this.updateChunkVisibility(viewBounds2D);
}
```
If we stick to returning the updated chunks, I think extracting this code to a well-named function will be an improvement.

packages/core/src/core/chunk_manager.ts
shlomnissan: I don't mind the current pattern, but I think we can improve it slightly by storing references of updated chunks in the source which can be rebuilt per frame:

```ts
source.update(lodFactor, viewBounds2D);
for (const change of source.updatedChunks) {
  // ...
}
```
This feels more consistent with what I would expect the scope of `update()` to be.

If you prefer not to stick to the current approach we should at least rename `update()`.

That said, if you are looking for alternatives: 

1. Extend the initial suggestion and store disposed/queued chunk references in the chunk manager source and access them via getters, avoiding additional iteration on the chunk manager.

```ts
for (const [_, source] of this.sources_) {
  source.queuedChunks.forEach((c) => {
    // enqueue
  });
  source.disposedChunks.forEach((c) => {
    // cancel
  });
}
```

2. We can store previously loaded chunk references in on the chunk manager.

```ts
this.previouslyQueued.forEach((c) => {
  if (c.priority === null) {
    // cancel
  }
});

for (const [_, source] of this.sources_) {
  // ...
}
```

shlomnissan (changes requested):
The only blocking concern I have is iterating through all the chunks in the chunks info overlay. Otherwise, this looks and works great!

I would love to see a fix for the white screen state in between timepoints, in this PR if it's small enough or as a follow up, to make this truly "demoable" on its own.

packages/core/src/core/chunk_manager.ts
shlomnissan: While I consider this change "improving readability" (as indexing is cheap) I do think it's important.

packages/core/src/core/chunk_manager.ts
andy-sweet: Agreed. I returned `{size: this.dimensions_.t?.lods[0].size ?? 1}` from `validateTimeDimension` since it's not clear to me that a name like `timeLength` or `timeSize` would do any validation or error.

packages/core/src/core/chunk_manager.ts
andy-sweet: I thought about this too. At one point it actually made things less readable, but now that this PR makes more simplifying assumptions about "t", that's not the case anymore, so I'll change it now.

packages/core/src/core/chunk_manager.ts
andy-sweet: Done. I simplified this further to just use `viewBounds2D.clone()`.

packages/core/examples/chunk_streaming/chunk_info_overlay.ts
andy-sweet: Interesting. I had primarily thought of this overlay as a debugging tool, so wasn't too concerned about performance here.

Thinking a bit more, disabling this avoids rendering it, but the calculations still happen. So at a minimum, we should probably not do anything if we're not rendering (e.g. if we turn this off with the checkbox control).

Since this includes prefetching information, I feel like only calculating information for the current time point may become limiting very shortly (i.e. in the next PR). But I'd be OK adding a method or renaming the `chunks` property in this PR to clarify.

packages/core/src/core/chunk_manager.ts
andy-sweet: > I don't mind the current pattern, but I think we can improve it slightly by storing references of updated chunks in the source which can be rebuilt per frame:

I find it hard to justify adding state to a class when a return value would suffice. So I'm going to avoid this suggestion.

But I like the suggestion of `queuedChunks` and `disposedChunks` more because separating those out gives us more semantic clarity, but feels slightly out of place in this change.

packages/core/src/core/chunk_manager.ts
andy-sweet: I like this, especially as time disposal will get more complicated in the near future (i.e. with prefetching in time). Done.

packages/core/src/core/chunk_manager.ts
andy-sweet: Agreed that I'd expect an update function to update some internal state, which it does still do.

I'd usually be pleasantly surprised if an update function return information that I could use to understand what was updated.

But I'll try to think of a better name here.

packages/core/src/core/chunk_manager.ts
andy-sweet: That feels too limiting for some upcoming work, but I'd be OK doing it for this PR and working what to do later.

packages/core/examples/chunk_streaming/chunk_info_overlay.ts
shlomnissan: It is purely a debugging tool, but I don't expect debugging tools to have a significant impact on performance, or if they do, we need to make sure that it's clear (maybe make sure it can't be enabled in production).
 
I think ensuring no additional computation when disabled is fine for this PR, we can come up with a more robust solution in the future. If we stick to the current structure, we may want to have accessors that provides all the loaded, queued, disposed chunks, and then all we need is exposing the length property of the chunks array.

andy-sweet: Agreed that minimal performance impact would be good. I've implemented all the quick fixes to make this at least as performant as `main`, with the biggest impact being the introduction of `ChunkManagerSource.getChunksAtCurrentTime`.

My thoughts are that these kind of chunk stats are incredibly useful and the best time to keep track of them is in the chunk manager module itself, where we can make efficient and informed updates. That's definitely out of scope for this PR though.

packages/core/src/core/chunk_manager.ts
andy-sweet: I added a new method to do this: `ChunkManagerSource.getChunksAtCurrentTime`. I left `get chunks` alone for now, though nothing is using it in this repo now, so we could consider removing it.

andy-sweet:
> The only blocking concern I have is iterating through all the chunks in the chunks info overlay. Otherwise, this looks and works great!

Should be fixed now. It only iterates through chunks at the current time, which is fine for this PR. I also made a couple of other related optimizations to that overlay.

> I would love to see a fix for the white screen state in between timepoints, in this PR if it's small enough or as a follow up, to make this truly "demoable" on its own.

I can't think of a quick fix and I think this is demoable enough for random scrubbing through time. With some of the upcoming PRs, the issue of white space and/or flickering may be improved. But this is generally a slightly tricky problem, since you don't want to show stale data for too long, but also don't want to change data too frequently.

andy-sweet:
> > I would love to see a fix for the white screen state in between timepoints, in this PR if it's small enough or as a follow up, to make this truly "demoable" on its own.
> 
> I can't think of a quick fix and I think this is demoable enough for random scrubbing through time. With some of the upcoming PRs, the issue of white space and/or flickering may be improved. But this is generally a slightly tricky problem, since you don't want to show stale data for too long, but also don't want to change data too frequently.

I thought a bit more about this, but still can't think of a quick fix that would fit well in this PR. Some possible bigger fixes more suitable for a separate PR.

1. Store fallback/backup renderable objects in `ChunkedImageLayer` that continue to display while waiting for enough chunks to arrive for a more recent time point.
2. Similar to (1), but implemented in the `ChunkManagerSource` closer to the LOD fallback logic. `ChunkManagerSource.getChunks` would now also return chunks at times that are recent enough to be deemed visible.

Both approaches are similar in that they relax the notion of what is visible or should be presented based on the time point. They likely both require some amount of configuration (e.g. how many frames or how much time is recent enough).

packages/core/src/core/chunk_manager.ts
```diff
@@ -43,68 +45,73 @@ export class ChunkManagerSource {
     this.sliceCoords_ = sliceCoords;
 
     this.validateXYScaleRatios();
+    const { size: chunksT } = this.validateTimeDimension();
```
shlomnissan: I think `validateTimeDimension` is a bad name because it hides with this function actually does: returning a derived time value the caller actually uses. If I only saw the name, I'd assume it performs side-effects (throws errors if invalid) but doesn't return anything meaningful. Alternatives to consider:
`resolveTimeLength()`
`getAndValidateTimeDimension()`
`validateAndGetTimeSize()`

packages/core/src/core/chunk_manager.ts
```diff
@@ -147,7 +165,7 @@ export class ChunkManagerSource {
   }
 
   public get chunks(): Chunk[] {
-    return this.chunks_;
+    return this.chunks_.flat();
```
shlomnissan: Why flatten it here? That seems to reduce the flexibility of this function. I’d prefer not to expose the entire chunk set to callers. My concern is that we’ll forget about it, and later someone will fetch the full set and iterate through everything without realizing the performance cost. I get that we need to keep it for now until we can query chunks by state, but we should at least preserve the current structure—so if someone does fetch it, they can index on a particular timepoint.

shlomnissan (approved):
I left a couple of non-blocking comments :+1:

packages/core/src/core/chunk_manager.ts
shlomnissan: - `updateAndCollectChangedChunks`
- `computeChunkChangesForView`

shlomnissan:
Rename options for `ChunkManagerSource.update()`
- `computeChunkChangesForView`
- `updateAndCollectChangedChunks`

packages/core/src/core/chunk_manager.ts
andy-sweet: Agreed that this shouldn't be exposed, so I'm just going to delete it.

packages/core/src/core/chunk_manager.ts
andy-sweet: I'm used to validators returning a validated and potentially coerced value (e.g. as with Pydantic), but I'm fine with changing the name to `getAndValidateTimeDimension`. The others seem more misleading than the current one since they either don't mention validation or imply validating something else.

shlomnissan: > I'm used to validators returning a validated and potentially coerced value

I'm not familiar with Pydantic, but when I see a validator that returns a normalized value, it typically expects that you pass it the value to validate/normalize (common pattern in string validation libraries)

packages/core/src/core/chunk_manager.ts
shlomnissan: FWIW I understand the value of at least showing the total number of chunks in the overlay. I think it's a good way to illustrate how much work Idetik is doing. So while I think it shouldn't be exposed, I'm not sure my concern outweighs the benefit, but it's your decision. I think a middle ground can be exposing the length property of the chunks array.

andy-sweet:
Sorry this isn't merged yet.

Before merging, I wanted to think a bit more about how we could

1) efficiently calculate chunk stats across multiple time points because this will be very useful (maybe even necessary for follow ups)
2) actually implement prefetching chunks across time, which could affect some of the changes here.

I did not quite finish that today.

Please let me know if not merging this is blocking or could block you, and I'll bring this in.

andy-sweet:
Decided to merge this after addressing PR comments, since all follow up work is going to be based on this anyway. And these changes should have minimal performance overhead (i.e. only initialization and disposing of chunks at the previous time point).

PR #434, opened 2025-09-10
Multiple Viewports: Viewport-scoped events

aganders3:
### Summary
This adds the concept of an `EventProvider` for the `EventDispatcher` instead of expecting all events to come directly from the canvas element. This allows for events to be scoped to a specific viewport, for example to allow each viewport to have separate camera controls.

I also added an optional `onEvent` to the `Overlay` type, so overlays can also respond to events. For example this would allow a global overlay to update depending on the which viewport has the mouse over it. We may also want to enable viewport-specific overlays at some point.

The `Viewport` class implements `EventProvider`, adding a `processEvent` method that:
* augments the event with viewport-specific information (source viewport, clipPos, and worldPos if the event has clientPos)
* dispatches the event first to viewport-specific listeners (layers and camera controls)
* dispatches the event to any global event listeners (overlays)

I also added support for `mouseenter` and `mouseleave` events. This lets the global event listeners to keep track of which viewport was last hovered, clicked, etc.

This is step three of the implementation plan in the [Multiple Viewports Tech Spec](https://docs.google.com/document/d/1OwGL7KW34hxFAENyOm2WiQ71E58XGu9Y7jKkJlyRx84/edit?tab=t.0).

### Related Issue
N/A

### Tests & Checks
Tested current examples work without modification
Added some tests for the event provider interface

### DIFF
--- a/packages/core/src/idetik.ts
+++ b/packages/core/src/idetik.ts
@@ -17,6 +17,7 @@ import {
 
 type Overlay = {
   update(idetik: Idetik, timestamp?: DOMHighResTimeStamp): void;
+  onEvent?(event: EventContext): void;
 };
 
 type IdetikParams = {
@@ -87,22 +88,15 @@ export class Idetik {
     if (params.showStats) this.stats_ = createStats();
 
     this.events = new EventDispatcher(canvas);
-    this.events.addEventListener((event: EventContext) => {
-      if (
-        event.event instanceof PointerEvent ||
-        event.event instanceof WheelEvent
-      ) {
-        const { clientX, clientY } = event.event;
-        const client = vec2.fromValues(clientX, clientY);
-        event.clipPos = this.clientToClip(client, 0);
-        event.worldPos = this.camera.clipToWorld(event.clipPos);
+    for (const viewport of this.viewports_) {
+      this.events.addViewport(viewport);
+    }
+
+    for (const overlay of this.overlays) {
+      if (overlay.onEvent) {
+        this.events.addEventListener(overlay.onEvent.bind(overlay));
       }
-      for (const layer of this.layerManager.layers) {
-        layer.onEvent(event);
-        if (event.propagationStopped) return;
-      }
-      this.viewports_[0].cameraControls?.onEvent(event);
-    });
+    }
   }
 
   public get width() {
--- a/packages/core/src/idetik.ts
+++ b/packages/core/src/idetik.ts
@@ -1,11 +1,7 @@
 import { Camera } from "./objects/cameras/camera";
 import { Layer } from "./core/layer";
 import { LayerManager } from "./core/layer_manager";
-import {
-  EventContext,
-  EventDispatcher,
-  CanvasEventProvider,
-} from "./core/event_dispatcher";
+import { EventDispatcher } from "./core/event_dispatcher";
 import { WebGLRenderer } from "./renderers/webgl_renderer";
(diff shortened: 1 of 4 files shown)


aganders3:
Tagging @phoenixAja for review somewhat arbitrarily, but we did discuss the viewports strategy early on.

I think this is at least ready for a fresh set of eyes, but I'll probably give it another once-over myself tonight as well if you want to wait for tomorrow (or later).

packages/core/src/idetik.ts
```diff
@@ -86,23 +91,23 @@ export class Idetik {
...
+    // add canvas provider if no viewport uses the canvas as its element
+    // this handles multi-viewport cases where canvas background might be exposed
+    // i.e. when viewports don't cover the entire canvas
+    if (!this.viewports_.some((v) => v.element === canvas)) {
+      this.events.addProvider(new CanvasEventProvider(canvas));
+    }
+
+    for (const overlay of this.overlays) {
+      if (overlay.onEvent) {
+        this.events.addEventListener(overlay.onEvent.bind(overlay));
```
andy-sweet: `Idetik.overlays` is public, readonly, but not a ReadonlyArray, so this might not cover all cases you intend (e.g. if an overlay is added later it won't observe events).

A few options:

1. Remove events from overlays for now (given the size of this PR, I'd probably go for this).
2. Make overlays a ReadonlyArray (also fine with me).
3. Create an overlay manager that listens for new events (don't do this!).

packages/core/src/idetik.ts
```diff
@@ -86,23 +91,23 @@ export class Idetik {
...
-        if (event.propagationStopped) return;
+    this.events = new EventDispatcher();
+    for (const viewport of this.viewports_) {
+      this.events.addProvider(viewport);
+    }
+
+    // add canvas provider if no viewport uses the canvas as its element
+    // this handles multi-viewport cases where canvas background might be exposed
+    // i.e. when viewports don't cover the entire canvas
+    if (!this.viewports_.some((v) => v.element === canvas)) {
```
andy-sweet: This could be my Friday afternoon brain speaking, but currently, `this.viewports_` is always an array of length one where the `v.element === canvas`, right?

If so, I'd prefer to just remove the loop above and the condition and just call the line here.

And then update it when there are multiple viewports, and when both cases can be tested.

andy-sweet (approved):
Maybe this my Friday brain speaking again, but this feels complicated. But I also don't have any better suggestions right now. I'm happy to approve to unblock follow-on work which may give this more context. Or maybe I should re-read the tech spec to help.

packages/core/src/idetik.ts
aganders3: I think it makes sense to remove them here. If we make the event dispatcher public maybe we don't have to automatically add them at all, which removes the need for the `onEvent` interface as well.

aganders3:
Yeah I agree this is a little complicated. Initially it felt more complicated to try to test/detect ourselves which viewport the event occurred in, but now I'm not so sure.

Do you think it's simpler to keep the DOM listeners on the canvas only and let viewports test the event position (if available) against their bounding box to decide what to do with the event?

edit: That may make it harder to do things like `mouseenter` and `mouseleave` for a viewport, but maybe that's fine and I'm overthinking it.

edit2: It would also require making the layout divs for the viewports non-interactive, which feels like an annoying requirement as well.

aganders3:
Okay I'm going to close this in favor of the much simpler #451

PR #437, opened 2025-09-11
feat: add OmeZarrChunkedImageViewer with external z-navigation support

phoenixAja:
###  Summary

  Adds a new OmeZarrChunkedImageViewer component that provides external z-navigation control for ChunkedImageLayer. 

###  Changes Made

  - Created OmeZarrChunkedImageViewer: Dedicated viewer component for ChunkedImageLayer
    - Exposes updateZSlice(zValue: number) callback via onLayerCreated prop
    - Direct sliceCoords manipulation for fluid z-navigation
    - Simplified interface focused solely on ChunkedImageLayer
  -  Created shared helper functions for both viewers
  - Miscellaneous fixes for bugs found while adding chunkedImageLayer to embrella

###  Technical Implementation

  - Returns shared sliceCoords object reference via callback for external manipulation
  - Uses existing ChunkedImageLayer streaming and LOD management capabilities
  - Maintains React component lifecycle patterns

###  Use Case

 Enables applications to implement custom z-navigation UI (sliders, keyboard controls, etc.) while leveraging ChunkedImageLayer's performance benefits.

###  Testing
  - Component builds successfully
  - External z-navigation callback functionality tested


packages/core/src/utilities/almost_equal.ts
```diff
@@ -1,3 +1,3 @@
-export function almostEqual(a: number, b: number, epsilon = 1e-6): boolean {
+export function almostEqual(a: number, b: number, epsilon = 0.02): boolean {
```
phoenixAja: I was getting this error when trying to use ChunkedImageLayer with embrella data:
  Invalid downsampling factor between levels 1 → 2: expected (2× in X and Y), but got (1.99×, 1.99×)
  
  so i adjusted this function to be more leniant, let me know anyone has concerns

aganders3: I'm not sure if it would help your specific case, but I was toying with changing this at some point to match the [Python `isClose` function](https://docs.python.org/3/library/math.html#math.isclose) behavior.

Using a relative tolerance should also be less sensitive to whether the comparison is`almostEqual(size[n-1] / size[n], 2.0)` vs. `almostEqual(size[n-1], 2.0 * size[n])`.

packages/react/src/components/viewers/OmeZarrChunkedImageViewer/OmeZarrChunkedImageViewer.tsx
```diff
@@ -0,0 +1,184 @@
...
+import { ChannelControlsList } from "../OmeZarrImageViewer/components/ChannelControlsList";
+import { ScaleBar } from "../OmeZarrImageViewer/components/ScaleBar/ScaleBar";
+
+export interface OmeZarrChunkedImageViewerProps {
+  sourceUrl?: string;
+  sourceLocalDirectory?: {
+    directory: FileSystemDirectoryHandle;
+    path?: `/${string}`;
+  };
+  region: Region;
```
andy-sweet: Optional.

Could we just use `SliceCoordinates` here? That was designed to be more explicit and then you could skip the conversion later.

We'd probably need to export it from the core library and it might change, but I think that's true of a lot of things in the core library.

packages/react/src/components/viewers/OmeZarrImageViewer/components/ScaleBar/ScaleBar.tsx
```diff
@@ -107,7 +107,10 @@ class ScaleBarOverlay {
       const lineWidthWorld = scientificFloor(containerWidthWorld);
       const lineProportion = lineWidthWorld.value / containerWidthWorld;
       lineDiv.style.width = `${lineProportion * 100}%`;
-      const numDecimalPlaces = Math.max(0, -lineWidthWorld.exponent);
+      const numDecimalPlaces = Math.max(
+        0,
+        Math.min(100, -lineWidthWorld.exponent)
+      ); // Clamp between 0-100 (toFixed limit)
```
andy-sweet: Did you run into 100 with real data? Maybe when the unit given in the metadata means the numerical scale is a very small value?

andy-sweet (commented):
Generally looks good, and no blocking feedback. Should we create a component example for this, so it's easy to quickly show off and test out in the package itself? I'd also be OK with replacing the existing example, which should probably be moved anyway.

packages/react/src/components/viewers/OmeZarrImageViewer/components/ScaleBar/ScaleBar.tsx
phoenixAja: it wasn't 100 exactly, i was just clamping to a fairly large value. I updated this to use scientiifc notation intstead for more than 6 decimal places and removed clampling

phoenixAja:
> Generally looks good, and no blocking feedback. Should we create a component example for this, so it's easy to quickly show off and test out in the package itself? I'd also be OK with replacing the existing example, which should probably be moved anyway.

I might do this in a follow up PR so i can also add a component for external Z navigation if that's alright?

packages/react/src/components/viewers/OmeZarrImageViewer/components/ScaleBar/ScaleBar.tsx
```diff
@@ -107,8 +107,19 @@ class ScaleBarOverlay {
...
-      textDiv.textContent = `${lineWidthWorld.value.toFixed(numDecimalPlaces)} ${this.unit_}`;
+      let displayText: string;
+      if (lineWidthWorld.exponent < -6) {
+        // Use scientific notation for very small numbers
+        displayText = `${lineWidthWorld.value.toExponential(2)} ${this.unit_}`;
+      } else {
+        // For reasonable scales, limit decimal places to at most 6
+        const numDecimalPlaces = Math.max(
+          0,
+          Math.min(6, -lineWidthWorld.exponent)
```
andy-sweet: Given the if condition above, this will always return `-lineWidthWorld.exponent` right? If so, I'd just remove it for automatic consistency and less code to read/understand.

packages/react/src/components/viewers/OmeZarrImageViewer/components/ScaleBar/ScaleBar.tsx
```diff
@@ -107,8 +107,19 @@ class ScaleBarOverlay {
       const lineWidthWorld = scientificFloor(containerWidthWorld);
       const lineProportion = lineWidthWorld.value / containerWidthWorld;
       lineDiv.style.width = `${lineProportion * 100}%`;
-      const numDecimalPlaces = Math.max(0, -lineWidthWorld.exponent);
-      textDiv.textContent = `${lineWidthWorld.value.toFixed(numDecimalPlaces)} ${this.unit_}`;
+      let displayText: string;
+      if (lineWidthWorld.exponent < -6) {
```
andy-sweet: I'd even consider dropping this to -3. That's about my limit for fixed notation anyway.

packages/react/src/components/viewers/OmeZarrChunkedImageViewer/OmeZarrChunkedImageViewer.tsx
```diff
@@ -0,0 +1,180 @@
...
+  const [unit, setUnit] = useState<string>();
+  const [loading, setLoading] = useState(true);
+  const [extraControlProps, setExtraControlProps] = useState<
+    ExtraControlProps[]
+  >([]);
+  const sourceRef = useRef<OmeZarrImageSource | null>(null);
+  const imageLayerRef = useRef<ChunkedImageLayer | null>(null);
+  const sliceCoordsRef = useRef<SliceCoordinates | null>(null);
+
+  // #region Initialization
```
andy-sweet: No longer relevant?

```suggestion
```

packages/react/src/components/viewers/OmeZarrChunkedImageViewer/OmeZarrChunkedImageViewer.tsx
```diff
@@ -0,0 +1,180 @@
...
+      const loadChannelMetadataPromise = loadChannelMetadata(
+        source,
+        fallbackContrastLimits
+      );
+      const loadImageMetadataPromise = loadImageMetadata(source);
+      const { channelProps, extraControlProps } =
+        await loadChannelMetadataPromise;
+      const { xUnit, yCoordRange, xCoordRange } =
+        await loadImageMetadataPromise;
+      if (sourceRef.current !== source) {
```
andy-sweet, line 94: Optional (because this is the same as in the other component).

Could and should we handle this early return earlier (i.e. right after checking for a null source)?

packages/react/src/components/viewers/OmeZarrChunkedImageViewer/OmeZarrChunkedImageViewer.tsx
```diff
@@ -0,0 +1,180 @@
...
+        sliceCoordinates,
+        channelProps
+      );
+      imageLayerRef.current = layer;
+      sliceCoordsRef.current = sliceCoords;
+
+      // Create updateZSlice function for external control
+      const updateZSlice = (zValue: number) => {
+        if (sliceCoords) {
+          sliceCoords.z = zValue;
```
andy-sweet, line 110: One thing to be mindful of here is that the z-coordinate respects the translation and scale of that dimension in the source data. That means these may not be integers.

`ImageSeriesLayer` has `setIndex` (for integer indices) and `setPosition` (like this, for coordinates) to support both.

Any component that uses `updateZSlice` will need to take that into account (i.e. by loading and using the source dimension metadata).

packages/core/src/utilities/almost_equal.ts
andy-sweet: Given this is a general purpose utility, 0.02 doesn't seem like a reasonable default. I'd probably prefer some multiple of machine precision for `number`, or like Ashley mentioned a different comparison with relative tolerance.

But I think it would be fine to pass through 0.02 from chunk_manager, and that's what you should do instead.

andy-sweet (changes requested):
Generally looks good, and the rationale to create a new simpler component definitely makes sense to me. The only blocking feedback is to pass through 0.02 to `almostEqual` instead of making it the default.

andy-sweet:
Also, please update the PR description based on any new changes.

packages/react/src/components/viewers/OmeZarrImageViewer/components/ScaleBar/ScaleBar.tsx
phoenixAja: i agree

packages/react/src/components/viewers/OmeZarrImageViewer/components/ScaleBar/ScaleBar.tsx
phoenixAja: yep

packages/core/src/utilities/almost_equal.ts
phoenixAja: i was thinking of doing ashleys' suggestion in another PR as i was aiming to keep this one small-ish, i removed the 0.02 default

packages/react/src/components/viewers/OmeZarrChunkedImageViewer/OmeZarrChunkedImageViewer.tsx
phoenixAja: thanks for the callout

packages/react/src/components/viewers/OmeZarrChunkedImageViewer/OmeZarrChunkedImageViewer.tsx
phoenixAja: yep! added an early return

andy-sweet: This is still optional, but I don't see any changes from the last review. I think that's OK (see first comment), I'm just a bit confused by the existing component. Just wanted to make sure you didn't miss a commit/change.

andy-sweet (approved):
Looks good now.

PR #444, opened 2025-09-12
Load chunks center-out within each priority

shlomnissan:
### Summary
This PR adds an `orderKey` to `Chunk`, computed each frame from the view
center to load chunks center-out within each priority.

### Tests & Checks

- Added an additional test for verifying sorting by `orderKey`.
- All existing examples are working as expected.
- Manual verification (see video below):
  - Change chunk queue capacity to 1
  - Order the view in LOD1 so it's centered
  - Zoom in from the center
  - Observe chunks in the middle are loaded first

https://github.com/user-attachments/assets/979f4ec8-adc0-485f-9c1f-e69c0a168b96

### DIFF
--- a/packages/core/src/core/chunk_manager.ts
+++ b/packages/core/src/core/chunk_manager.ts
@@ -12,7 +12,7 @@ import { almostEqual } from "../utilities/almost_equal";
 import { Logger } from "../utilities/logger";
 import { OrthographicCamera } from "../objects/cameras/orthographic_camera";
 import { ChunkQueue } from "../data/chunk_queue";
-import { clamp } from "@/utilities/clamp";
+import { clamp } from "../utilities/clamp";
 
 // Number of chunks to extend beyond the visible bounds in each direction (x/y/z)
 // These additional chunks are prefetched to improve responsiveness when panning.
@@ -201,7 +201,7 @@ export class ChunkManagerSource {
     );
 
     const paddedBounds = this.getPaddedBounds(viewBounds3D);
-    const centerIndicies = this.getCenterIndices(viewBounds2D);
+    const centerIndices = this.getCenterIndices(viewBounds2D);
 
     for (const chunk of this.chunks_) {
       const isVisible = this.isChunkWithinBounds(chunk, viewBounds3D);
@@ -230,8 +230,8 @@ export class ChunkManagerSource {
 
       if (chunk.priority !== null) {
         chunk.orderKey = Math.max(
-          Math.abs(chunk.chunkIndex.x - centerIndicies[chunk.lod].x),
-          Math.abs(chunk.chunkIndex.y - centerIndicies[chunk.lod].y)
+          Math.abs(chunk.chunkIndex.x - centerIndices[chunk.lod].x),
+          Math.abs(chunk.chunkIndex.y - centerIndices[chunk.lod].y)
         );
       }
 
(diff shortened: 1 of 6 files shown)


packages/core/src/core/chunk_manager.ts
```diff
@@ -199,6 +201,8 @@ export class ChunkManagerSource {
     );
 
     const paddedBounds = this.getPaddedBounds(viewBounds3D);
+    const centerIndicies = this.getCenterIndices(viewBounds2D);
```
andy-sweet: Typo
```suggestion
    const centerIndices = this.getCenterIndices(viewBounds2D);
```

packages/core/src/core/chunk_manager.ts
```diff
@@ -221,6 +225,14 @@ export class ChunkManagerSource {
         chunk.state = "queued";
       } else if (chunk.priority === null && chunk.state === "queued") {
         chunk.state = "unloaded";
+        chunk.orderKey = null;
+      }
+
+      if (chunk.priority !== null) {
+        chunk.orderKey = Math.max(
+          Math.abs(chunk.chunkIndex.x - centerIndicies[chunk.lod].x),
+          Math.abs(chunk.chunkIndex.y - centerIndicies[chunk.lod].y)
```
andy-sweet: Typo
```suggestion
          Math.abs(chunk.chunkIndex.x - centerIndices[chunk.lod].x),
          Math.abs(chunk.chunkIndex.y - centerIndices[chunk.lod].y)
```

packages/core/src/core/chunk_manager.ts
```diff
@@ -12,6 +12,7 @@ import { almostEqual } from "../utilities/almost_equal";
 import { Logger } from "../utilities/logger";
 import { OrthographicCamera } from "../objects/cameras/orthographic_camera";
 import { ChunkQueue } from "../data/chunk_queue";
+import { clamp } from "@/utilities/clamp";
```
andy-sweet:
```suggestion
import { clamp } from "../utilities/clamp";
```

packages/core/src/core/chunk_manager.ts
```diff
@@ -360,6 +373,39 @@ export class ChunkManagerSource {
...
+      const yLod = this.dimensions_.y.lods[lod];
+
+      const chunkWorldSizeX = xLod.chunkSize * xLod.scale;
+      const chunkWorldSizeY = yLod.chunkSize * yLod.scale;
+
+      const chunksX = Math.ceil(xLod.size / xLod.chunkSize);
+      const chunksY = Math.ceil(yLod.size / yLod.chunkSize);
+
+      const cx = (bounds.min[0] + bounds.max[0]) / 2;
+      const cy = (bounds.min[1] + bounds.max[1]) / 2;
```
andy-sweet: These calculations can be moved out of the loop.

andy-sweet (approved):
Looks great!

An alternative option might be store the x,y center of each chunk and then just use the euclidean distance between that and the viewport center. It would remove the need for `getCenterIndices` which is a little complicated and also give priority out with more precision (e.g. sometimes i can get chunks that are mostly outside the FOV load before ones that are all inside).

packages/core/src/core/chunk_manager.ts
shlomnissan: Good catch!

shlomnissan:
> An alternative option might be store the x,y center of each chunk and then just use the euclidean distance between that and the viewport center.

@andy-sweet That's a better option across the board! I updated the PR accordingly.

andy-sweet (approved):
Looks great!

PR #445, opened 2025-09-12
Use preventDefault to prevent page scroll on zoom

### LINKED ISSUE #440: Zooming with a mouse wheel on the IdetikCanvas also causes the page to scroll
When an IdetikCanvas is embedded in an application that has a vertical scrollbar, scrolling with a mouse wheel will cause the canvas to zoom and the page to scroll. We almost certainly want to prevent the vertical scroll from occurring in this case.

To reproduce, see a recent branch deployment on the OrganelleBox portal: https://organelle-box-portal-git-v2-zarr-czbiohub.vercel.app/gene/2025_03_31_A549_ACTB

andy-sweet:
### Summary
When the Idetik canvas is within a page that has a vertical scroll bar, the mouse wheel event can both zoom Idetik and scroll the page, which is jarring and almost definitely undesirable. This change prevents the default page scroll action.

### Related Issue
Closes #440 

### Tests & Checks
I temporarily manually modified navigation.css to use `150%` height for the body, which adds a vertical scrollbar. I then used the HCS example as the canvas does not cover the whole page in that case, so I can zoom in/out with the mouse wheel on the canvas, and scroll the page up/down on the left side.

### DIFF
--- a/packages/core/src/objects/cameras/controls.ts
+++ b/packages/core/src/objects/cameras/controls.ts
@@ -39,6 +39,9 @@ export class PanZoomControls implements CameraControls {
     if (!event.worldPos || !event.clipPos) return;
     const e = event.event as WheelEvent;
 
+    // Prevent the page from scrolling, the default action for wheel events.
+    e.preventDefault();
+
     const posBeforeZoom = vec3.clone(event.worldPos);
     const zoomFactor = e.deltaY < 0 ? 1.05 : 0.95;
 


andy-sweet:
@aganders3 : assigned you because you wrote these controls originally, though they have changed quite a bit due to the idetik event handling.

aganders3 (approved):
Thanks! This fix looks good to me.

My one thought is that it might be better for our event handlers to avoid handling or modifying the actual DOM event and primarily consume the `EventContext` we provide. Then this type of change would be made globally in the `EventDispatcher` or the like. Perhaps in the `EventContext` constructor for example:
https://github.com/chanzuckerberg/idetik/blob/ac9e39b95da0f604c535d581ecbbc86c2521e23f/packages/core/src/core/event_dispatcher.ts#L24

PR #447, opened 2025-09-12
Multiple Viewports: Move LOD calculation to Viewport

aganders3:
### Summary
This moves LOD calculation and world view bounds logic from `ChunkManager` to `Viewport` class. `ChunkManager.update()` and `ChunkManagerSource.update()` both now take a `Viewport` parameter instead of separate camera and buffer width parameters. This also consolidates the `OrthographicCamera`-specific type checking into the `Viewport`.

This improves separation of concerns by moving viewport-specific calculations into the `Viewport` class, making the `ChunkManager` API and main render loop cleaner. This also allows us to support different LODs in different viewports. The POC implementation just relied on the `ChunkManager` state switching rapidly as it rendered each viewport.

This is step four of the implementation plan in the [Multiple Viewports Tech Spec](https://docs.google.com/document/d/1OwGL7KW34hxFAENyOm2WiQ71E58XGu9Y7jKkJlyRx84/edit?tab=t.0).

### Related Issue
Closes N/A

### Tests & Checks
Tested existing examples work without changes

### DIFF
--- a/packages/core/src/core/chunk_manager.ts
+++ b/packages/core/src/core/chunk_manager.ts
@@ -129,7 +129,7 @@ export class ChunkManagerSource {
   public update(viewport: Viewport) {
     const lodFactor = viewport.getLodFactor();
     this.setLOD(lodFactor);
-    const viewBounds2D = viewport.getWorldViewRect();
+    const viewBounds2D = viewport.camera.getWorldViewRect();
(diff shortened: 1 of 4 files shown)


aganders3:
This is getting stale, so I'm going to close it for now and revisit as part of our Q4 work.

PR #450, opened 2025-09-12
`Viewport.getBox()` may return wrong value, causing rendering errors

### LINKED ISSUE #449: Viewport overlap with canvas calculation is wrong
On a page with a single viewport (viewport === canvas), I am seeing the wrong image and spammed logs with:

```
logger.ts:127 [2025-09-12T20:10:44.085Z][WARN][WebGLRenderer] Viewport main is entirely outside canvas bounds, skipping render
```

This is because `Viewport.cachedViewportBoundingBox_` is invalidated for _size_ changes (via the `ResizeObserver` and `devicePixelRatio` observer) but not for _position_ changes (e.g. scrolling).

aganders3:
### Summary
This simply removes `Viewport.cachedViewportBoundingBox_` for now. If this becomes a performance issue, we can revisit adding this cached value back and listening for any layout changes that might affect size _or_ position.

### Related Issue
Closes #449

### Tests & Checks
Tested on a page where the canvas does not take up the full page (can move without resizing).

### DIFF
--- a/packages/core/src/core/viewport.ts
+++ b/packages/core/src/core/viewport.ts
@@ -22,8 +22,6 @@ export class Viewport {
   public readonly layerManager: LayerManager;
   public cameraControls?: CameraControls;
 
-  private cachedViewportBox_: Box2 | null = null;
-
   constructor(config: ViewportConfig, layerManager: LayerManager) {
     this.id = config.id || config.element.id || generateUUID();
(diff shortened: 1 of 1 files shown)


andy-sweet (approved):
Not too concerned about performance right now given everything else happening in the animation loop.

PR #451, opened 2025-09-12
Add an `EventDispatcher` to each `Viewport` for viewport-scoped events

aganders3:
### Summary
This uses the existing `EventDispatcher` with very little modification, but adds one to each viewport (still currently only one).

### Related Issue
Obviates #434

### Tests & Checks
Current examples still work, including those that depend on events (value picking and controls)


aganders3:
@andy-sweet thanks for the feedback on #434. I think this ~~reigns~~ reins it in quite a bit, so I'd appreciate your feedback on this take.

edit: 🤦

andy-sweet (approved):
This is very simple, so approved. Happy to re-review something more complex when needed.

PR #461, opened 2025-09-23
Prevent flickering in the chunked image layer when changing the time point

### LINKED ISSUE #452: Support efficient time scrubbing and playback for chunk streaming
This is a tracking issue that tracks the work needed for efficient time scrubbing and playback in Idetik, and for the dynamic cell atlas Idetik demo we plan to give by the end of Q3.

Resources
- DCA Q3 [Scope and objectives doc](https://docs.google.com/document/d/1y2-WRHWUWo4hQKNTS-5VvbG_fV-1aaJURPxWVG6CKo8/edit?tab=t.0)

andy-sweet:
### Summary
This introduces some logic in the `ChunkedImageLayer` to show chunks at a previously loaded time point until all the visible chunks at the lowest LOD are loaded. This prevents those previously loaded chunks being immediately hidden, which can cause flickering in some cases of scrubbing through time.

This also exposes the most recently presented time coordinate in the `ChunkedImageLayer`, which is used in an overlay in the chunk streaming example.

### Related Issue
Closes #452 

### Tests & Checks
I manually tested the chunk streaming example.

### DIFF
--- a/packages/core/examples/chunk_streaming/main.ts
+++ b/packages/core/examples/chunk_streaming/main.ts
@@ -95,12 +95,21 @@ gui
   .add(controls.sliceCoords, "t", tRange.min, tRange.max, t.scale)
   .name("T-point");
 
-gui
+const overlaysFolder = gui.addFolder("Overlays");
+
+overlaysFolder
+  .add(controls, "showTimePointOverlay")
+  .name("Show time point overlay")
+  .onChange((show: boolean) => {
+    timePointDiv.style.display = show ? "block" : "none";
+  });
+
+overlaysFolder
   .add(controls, "showWireframes")
   .name("Show tile wireframes")
   .onChange((show: boolean) => (imageLayer.debugMode = show));
 
-gui
+overlaysFolder
   .add(controls, "showChunkInfoOverlay")
   .name("Show chunk information overlay")
   .onChange((show: boolean) => {
--- a/packages/core/examples/chunk_streaming/main.ts
+++ b/packages/core/examples/chunk_streaming/main.ts
@@ -56,12 +56,20 @@ const chunkInfoOverlay = new ChunkInfoOverlay({
   imageLayer: imageLayer,
 });
 
+const timePointDiv = document.querySelector<HTMLDivElement>("#time-point")!;
+const timePointOverlay = {
+  update(_idetik: Idetik, _timestamp?: DOMHighResTimeStamp) {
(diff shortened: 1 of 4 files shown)


packages/core/examples/chunk_streaming/main.ts
```diff
@@ -56,19 +56,28 @@ const chunkInfoOverlay = new ChunkInfoOverlay({
...
-  overlays: [chunkInfoOverlay],
+  overlays: [chunkInfoOverlay, timePointOverlay],
   showStats: true,
 }).start();
 
 const controls = {
   sliceCoords,
   showWireframes: true,
   showChunkInfoOverlay: true,
+  showTimePointOverlay: true,
```
aganders3, line 80: You need to add the following for this to show up in the lil-gui controls:

```
gui
  .add(controls, "showTimePointOverlay")
  .name("Show time point overlay")
  .onChange((show: boolean) => {
    timePointDiv.style.display = show ? "block" : "none";
  });
```

That said, I'd be fine to just make this non-configurable as the time label itself is smaller than an additional toggle.

packages/core/src/layers/chunked_image_layer.ts
```diff
@@ -38,6 +38,10 @@ export class ChunkedImageLayer extends Layer implements ChannelsEnabled {
   private zPrevPointWorld_?: number;
   private debugMode_ = false;
 
+  private static readonly STALE_PRESENTATION_MS_ = 1000;
```
aganders3, line 41: nit: I bumped this to 2000 which feels better on my connection

aganders3 (approved):
I think eventually I'd like to see an async API for updating the `sliceCoords`, but I'm happy to put that off for later. Some configuration may also eventually make sense. For now this is a meaningful improvement in UX for me, and the code is pretty small/clear.

packages/core/examples/chunk_streaming/main.ts
andy-sweet: Added. I put the visibility toggles in a folder, but that also made the controls bigger and I'm maybe regretting it...

packages/core/src/layers/chunked_image_layer.ts
andy-sweet: I imagine our connections to the server are not too different. I chose this value instead of 2000 because it hits the stale case sometimes (even when the browser cache is hit).

PR #462, opened 2025-09-23
Add playback control to chunk streaming examples

### LINKED ISSUE #455: Add a playback mode to the chunk streaming demo
The goal of this task to add controls that allow the time dimension in the chunk streaming demo to be played back at some fixed frame-rate (e.g. a play/pause button). Ideally there should also be some visual indication of what time-points have been pre-loaded.

We should try to pick some fixed constants that enable smooth playback at some respectable frame-rate in some environment, but we should not attempt any specialized optimization within the chunk manager to achieve this.

andy-sweet:
### Summary
This adds a slider to control the playback rate over the z and t dimensions in the examples that use chunk streaming.

The playback is driven by the example front-end (i.e. using `setInterval` to drive the slider position) rather than Idetik backend. Idetik tries to keep up as best it can, but there is no explicit prefetching over time yet.

For the main chunk streaming example with the very slow default playback settings of 2 time-points per second and relying on the browser cache, my machine/connection struggles to keep up (mostly due to decompression). There are simple hard-coded ways to improve this (e.g. bump the LOD-bias to only load the lowest-res LOD), but I have not committed those yet.

### Related Issue
Closes #455

### Tests & Checks
I manually tested the chunk streaming example.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -29,7 +29,7 @@
         "rollup-plugin-dts": "^6.1.1",
         "typescript": "^5.2.2",
         "typescript-eslint": "^8.1.0",
-        "vite": "^6.3.4",
+        "vite": "^6.3.6",
         "vite-plugin-eslint": "^1.8.1",
         "vite-plugin-glsl": "^1.3.0",
         "vitest": "^3.0.9"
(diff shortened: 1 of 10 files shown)


andy-sweet:
I made this a draft because it's somewhat rough, and I'm not yet convinced the example should be driving the playback. However, I will be using this as an initial driver to provide some related optimizations (e.g. pre-fetching) and wanted to share out early.

andy-sweet:
I think this is ready for review now because playback works well-enough for z in both examples. Our upcoming optimizations will hopefully make t work a little more smoothly for the main chunk streaming example.

I also refactored a lil-gui utility function to create a dimension slider with optional playback.

aganders3:
Haha the center-out prioritization gives kind of a [star wipe](https://www.youtube.com/watch?v=72bUheqRE5o) effect during playback.

aganders3 (approved):
This looks good and works well for me. I like that the code is contained to the examples, because I agree I'm not sure the UI should drive playback.

One idea is to add another gui item to set the playback interval. This would be useful for qualitatively assessing performance. Unfortunately the controls start to get a bit crowded at that point.

andy-sweet:
> One idea is to add another gui item to set the playback interval.

one simple way of doing this without using more vertical space is to replace the play/pause checkbox with a playback rate slider (where 0 means pause)

> Unfortunately the controls start to get a bit crowded at that point.

i did briefly try to make a horizontal layout with lil-gui (with a checkbox and slider), but it was more complex than I hoped, so I avoided it for now

PR #463, opened 2025-09-24
Dispatch `getChunk` to WebWorker pool

### LINKED ISSUE #456: Use WebWorkers to move chunk decompression off the main thread
Zarr chunks are typically stored with lossless compression. This is handled transparently for us by Zarrita + numcodecs, but decompression may have a significant cost. In a browser, we are typically working in a single-threaded context so CPU-bound tasks (like decompression) can block the animation loop, reducing framerates or affecting interactivity within our visualizations.

We can consider two ways to reduce impact of this:
* Reduce the amount of duplicate work by caching _decompressed_ chunks
* Move the CPU-intensive decompression step off the main thread to avoid blocking

In the browser, moving tasks off the main thread requires WebWorkers. This effort primarily focuses on the WebWorker approach, though some amount of caching may be a natural byproduct of the implementation.

We plan to spend effort in the last two weeks of Q3 investigating this. The following tasks will be used to track progress. Outputs will be posted in this issue, and I will post corresponding updates in #eng-visualization.

**Week 1: Exploration & Scoping** - Week of Sep. 14, 2025
- [x] Measure current decompression blocking times
- [x] Identify worst-case scenarios (which codecs/chunk sizes)
- [x] Back-of-envelope calculations on streaming throughput, network vs. CPU bottlenecks
- [x] Research to understand WebWorkers, best practices and memory management (namely `SharedArrayBuffer` vs. `ArrayBuffer`) 
- [x] Develop an implementation plan for Week 2

**Week 2: Implementation** - Week of Sep. 21, 2025
- [ ] Implement core worker module with minimal disruption to existing code (tentatively aiming for a drop-in replacement for `zarr.getChunk`)
- [ ] Basic performance tests and comparison, confirm chunk decompression is not blocking the main thread
- [ ] Profiling to determine memory impact

aganders3:
### Summary
This introduces a new `getChunk` replacement that dispatches the fetch + (more importantly) chunk decompression to a pool of WebWorkers. The goal was to make the replacement function close to a drop-in to minimize changes to existing code. This is now working for both the chunk streaming (zebrafish) and OME-Zarr v0.5 (marmoset neurons) examples. I think it's ready for another set of eyes.

There are some slightly complicated aspects here, mostly around what can types of objects can be transferred or copied to a WebWorker:
* use `ArrayBuffer` for zero-copy data transfer from the WebWorker back to the main thread (see [MDN Transferable Objects](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Transferable_objects))
* need to re-open Zarr arrays separately in the WebWorker, so there is a need to serialize all the required parameters - for this I added a `ZarrArrayParams` that supports the types of arrays we currently support.
* you can't pass through an `AbortController`/`AbortSignal` directly, so this PR sets up a listener for the signal on the main thread, and sends a `cancel` message to the worker pool, which creates/maintains its own separate `AbortController` for each task.

Eventually we may want to make this more generic and integrate it with the `ChunkQueue` to run any (or all) chunk-fetching operations in a worker pool. I need to think a bit more about that design. An intermediate step might be to make the worker _pool_ more generic, but keep the zarr-specific workers. None of this feels urgent until we start working with other types/sources of chunked data.

### Related Issue
Closes #456

### Tests & Checks
Tested with existing examples - framerate on chunk_streaming example seems much improved when zooming/panning (loading new chunks)

I also tested by temporarily adding two features:
1. an option to force `getChunk` on the main thread instead of the worker
2. a blocking sleep to simulate additional CPU bottleneck

With these changes it was very obvious the WebWorker was effective at keeping the main thread unblocked.

Here are some videos comparing the frame rate of the chunk streaming example on `main` vs. this branch. For this test I kept the browser cache enabled, because it exacerbates the problem when decompression is not naturally spaced out by network transfers.

### `main`

https://github.com/user-attachments/assets/1b5a2bbf-662b-4e73-9e74-1291b6a3edbb

### `getChunk` in WebWorkers (this branch)

https://github.com/user-attachments/assets/a8231cf5-6877-45b5-882d-585104e13b55


packages/core/src/data/chunk.ts
```diff
@@ -13,6 +13,7 @@ const chunkDataTypes = [
   Float32Array,
 ] as const;
 export type ChunkData = InstanceType<(typeof chunkDataTypes)[number]>;
+export type ChunkDataConstructor = (typeof chunkDataTypes)[number];
```
andy-sweet, line 15: Nit: reuse types.

```suggestion
export type ChunkDataConstructor = (typeof chunkDataTypes)[number];
export type ChunkData = InstanceType<ChunkDataConstructor>;
```

packages/core/src/data/zarr/open.ts
```diff
@@ -52,3 +73,64 @@ export async function openArray(
...
+  switch (params.type) {
+    case "fetch": {
+      rootLocation = new Location(
+        new FetchStore(params.url, params.fetchOptions)
+      );
+      break;
+    }
+    case "filesystem": {
+      if (typeof FileSystemDirectoryHandle === "undefined") {
+        throw new Error("FileSystem API not available in this context");
```
andy-sweet: From MDN, it looks like the filesystem API should be supported in web workers. What case is this handling?

packages/core/src/data/ome_zarr/worker_pool.ts
```diff
@@ -0,0 +1,282 @@
...
+      throw new Error(
+        `Unexpected chunk data type: ${typeof chunk.data}. Expected TypedArray.`
+      );
+    }
+
+    return {
+      data: chunk.data,
+      shape: chunk.shape,
+      stride: chunk.stride,
+      dtype: chunk.data.constructor.name,
```
andy-sweet: What is this used for? Isn't it just always available from `chunk.data.constructor.name`?

andy-sweet (commented):
I only started looking through this, but thought I'd leave the few comments I had.

Trying this out, performance seems great. But I think there may be an ordering issue between the `ChunkQueue` and the new worker pool. The effect on my machine is that currentLOD chunks sometimes load before fallbackLOD chunks, which is undesirable. As the simplest fix, how bad would making a pool of size 1 be? That would still prevent the main thread being blocked, but obvious lacks truely parallel decompression.

aganders3:
> how bad would making a pool of size 1 be? That would still prevent the main thread being blocked, but obvious lacks truely parallel decompression.

I'll go back to this and maybe can shrink the PR substantially. I agree it should achieve the main goal of unblocking the main thread. I think an early version I had like this had a bug, and I pursued the pool when I misunderstood the problem. I cixed the problem but kept the worker pool. I'll try this tomorrow!

packages/core/src/data/ome_zarr/worker_pool.ts
aganders3: Oh good catch, we need to pass it back from the worker but it's not necessary here.

packages/core/src/data/zarr/open.ts
aganders3: I'm confused and don't fully understand it yet. I saw that on MDN as well, but then I read that it couldn't be serialized/passed between threads, and I can't figure out how to instantiate it from a path. My only experience is getting the handle form [`window.showDirectoryPicker()`](https://developer.mozilla.org/en-US/docs/Web/API/Window/showDirectoryPicker).

What I haven't done is actually tested passing the `FileSystemDirectoryHandle` to the worker, maybe it will just work.

aganders3:
Here is a branch that removes the pool in favor of managing a single worker. It's a smaller change overall from `main`, but obviously a smaller change from this branch's current state would be to just hard-code the pool size to 1.
https://github.com/chanzuckerberg/idetik/compare/aganders3/getchunk-webworker-nopool

aganders3:
> The effect on my machine is that `currentLOD` chunks sometimes load before `fallbackLOD` chunks, which is undesirable.

I haven't thought this through too much, but here's an alternative fix for the symptom that preserves the parallelism (which does have a performance benefit). I _think_ the prioritization is still correct in terms of kicking off the `getChunk` requests, it's just that with parallel workers a `currentLOD` chunk may finish first.

```typescript
  public getChunks(): Chunk[] {
    const currentTimeChunks = this.getChunksAtCurrentTime();
    const lowResChunks = currentTimeChunks.filter(
      (chunk) =>
        chunk.lod === this.lowestResLOD_ &&
        chunk.visible &&
        chunk.state === "loaded"
    );

    // If we're at the lowest resolution LOD, we're done
    if (this.currentLOD_ === this.lowestResLOD_) {
      return lowResChunks;
    }

    // Only include current LOD chunks if all visible lowest LOD chunks are loaded
    const currentLODChunks = currentTimeChunks.filter(
      (chunk) =>
        this.allVisibleLowestLODLoaded() &&
        chunk.lod === this.currentLOD_ &&
        chunk.visible &&
        chunk.state === "loaded"
    );

    return [...lowResChunks, ...currentLODChunks];
  }

```

packages/core/src/data/zarr/open.ts
aganders3: Okay I think it just works™!

andy-sweet:
> it's just that with parallel workers a currentLOD chunk may finish first.

After some more thought I think that's even possible on `main` because the fetches are occurring concurrently. But maybe the parallel decompression increases the chance of it occurring on this branch? Anyway, I don't think it's blocking and I agree that if we really care about controlling the presentation, we should do it later (either in getChunks or in the layer).

aganders3:
> use ArrayBuffer for zero-copy data transfer from the WebWorker back to the main thread - this requires passing through the typed array's type (constructor) and buffer attribute separately so it can be re-typed in the main thread

I might be wrong about how this works and I'm going to see if we can skip this "reconstruction" step.

edit: That's great, https://github.com/chanzuckerberg/idetik/pull/463/commits/814441e3d6a8e5e99c45d9affa8a753c9351a99a makes this a lot simpler!

packages/core/src/data/ome_zarr/image_loader.ts
```diff
@@ -40,19 +42,23 @@ export class PromiseQueue<T> {
...
 // https://ngff.openmicroscopy.org/0.5/#image-layout
 export class OmeZarrImageLoader {
   private readonly metadata_: OmeZarrImage["ome"]["multiscales"][number];
   private readonly arrays_: ReadonlyArray<zarr.Array<zarr.DataType, Readable>>;
+  private readonly arrayParams_: ReadonlyArray<ZarrArrayParams>;
   private readonly loaderAttributes_: ReadonlyArray<LoaderAttributes>;
   private readonly dimensions_: SourceDimensionMap;
 
   constructor(props: OmeZarrImageLoaderProps) {
+    ensureWorkerPool();
```
andy-sweet: In the spirit of making as few changes as possible here, I might call this in `getChunk` instead. It will be called many times, but returns immediately.

packages/core/src/data/ome_zarr/image_loader.ts
```diff
@@ -88,14 +94,12 @@ export class OmeZarrImageLoader {
...
 
     const array = this.arrays_[chunk.lod];
-    const subarray = await array.getChunk(chunkCoords, { signal });
+    const arrayParams = this.arrayParams_[chunk.lod];
+    const subarray = await getChunk(array, arrayParams, chunkCoords, {
+      signal,
+    });
 
     const data = subarray.data;
-    if (!isChunkData(data)) {
```
andy-sweet, line 94: I'd consider leaving this here instead of moving it into `getChunk`, then have that function return the `zarr.Chunk` instance that `array.getChunk` returns, to make this more of a drop-in replacement.

packages/core/src/data/zarr/open.ts
```diff
@@ -52,3 +74,62 @@ export async function openArray(
...
+      arrayPath,
+      zarrVersion,
+      url: (location.store as FetchStore).url.toString(),
+    };
+  } else if (location.store instanceof WebFileSystemStore) {
+    return {
+      type: "filesystem",
+      arrayPath,
+      zarrVersion,
+      directoryHandle: location.store.directoryHandle,
```
andy-sweet, line 127: It's my first time thinking about web workers, but is there a simple reason about why this is passable to the worker but something like the `zarr.Location` or even `zarr.Array` is not?

packages/core/src/data/ome_zarr/worker_pool.ts
```diff
@@ -0,0 +1,295 @@
...
+  workerId: number;
+};
+
+const DEFAULT_WORKER_COUNT = Math.min(navigator.hardwareConcurrency || 4, 8);
+let workerPool: WorkerInstance[] = [];
+let messageId = 0;
+let workerId = 0;
+const pendingMessages = new Map<number, PendingGetChunkRequest>();
+const canceledMessages = new Set<number>();
+let poolConfigured = false;
```
andy-sweet: I'd probably just use `workerPool.length > 0` as a good enough proxy for this.

packages/core/src/data/ome_zarr/worker_pool.ts
```diff
@@ -0,0 +1,295 @@
...
+  options?: { signal?: AbortSignal }
+): Promise<{
+  data: ChunkData;
+  shape: number[];
+  stride: number[];
+}> {
+  return new Promise((resolve, reject) => {
+    const workerInstance = getAvailableWorker();
+    if (!workerInstance) {
+      reject(new Error("WebWorker pool not available"));
```
andy-sweet: To clarify this could happen in the chunk streaming example if the number of workers is less than the maximum number of concurrent tasks in the `ChunkQueue` right? In which case we fall back to fetching on the main thread?

packages/core/src/data/ome_zarr/worker_pool.ts
```diff
@@ -0,0 +1,295 @@
...
+};
+
+type WorkerInstance = {
+  worker: Worker;
+  busy: boolean;
+  pendingCount: number;
+  workerId: number;
+};
+
+const DEFAULT_WORKER_COUNT = Math.min(navigator.hardwareConcurrency || 4, 8);
```
andy-sweet: Nit: can hardwareConcurrency be 0 or undefined? MDN and my IDE type hint says it's a `number`, so I wonder if the `||` is really needed here.

packages/core/src/data/ome_zarr/worker_pool.ts
```diff
@@ -0,0 +1,295 @@
...
+  }) => void;
+  reject: (error: Error) => void;
+  abortListener?: () => void;
+  abortSignal?: AbortSignal;
+  workerId: number;
+};
+
+type WorkerInstance = {
+  worker: Worker;
+  busy: boolean;
```
andy-sweet: This seems to be assigned a lot, but I can't find it being used. Ideally, I'd prefer not to store this as extra state because it can be derived from `pendingCount`.

packages/core/src/data/ome_zarr/worker_pool.ts
```diff
@@ -0,0 +1,295 @@
...
+  worker.addEventListener("messageerror", (error) =>
+    handleWorkerError(error, worker)
+  );
+
+  return worker;
+}
+
+function getAvailableWorker(): WorkerInstance | undefined {
+  if (workerPool.length === 0) return undefined;
+  return workerPool.sort((a, b) => a.pendingCount - b.pendingCount)[0];
```
andy-sweet: Nit: I haven't convinced myself that I actually like relying on the fact that typescript returns `undefined` for OOB array indexing, but it is shorter!

```suggestion
  return workerPool.sort((a, b) => a.pendingCount - b.pendingCount)[0];
```

packages/core/src/data/ome_zarr/worker_pool.ts
```diff
@@ -0,0 +1,295 @@
...
+export function terminateWorkerPool(): void {
+  for (const workerInstance of workerPool) {
+    workerInstance.worker.terminate();
+  }
+  workerPool = [];
+  pendingMessages.clear();
+  poolConfigured = false;
+}
+
+export function ensureWorkerPool(): void {
```
andy-sweet: In the future I might prefer the worker pool be a class, to allow for easier testability.

But this doesn't have any tests yet, so this point is more or less moot right now though, and just adds extra indentation.

packages/core/src/data/ome_zarr/worker_kernel.ts
```diff
@@ -0,0 +1,144 @@
...
+      stride: number[];
+    }
+  | {
+      success: false;
+      type: ZarrWorkerMessageType;
+      error: string;
+    }
+);
+
+const arrayCache = new Map<string, zarr.Array<zarr.DataType, Readable>>();
```
andy-sweet, line 37: I think this should be a limited size. My concern here is when an application loads multiple multi-scale images over some session, this could grow to be quite large (e.g. 100s or 1000s). My concern is less about the memory that uses, and more about something like open file handles if that's relevant to the web file system store (I'm not sure that it is).

packages/core/src/data/ome_zarr/worker_kernel.ts
```diff
@@ -0,0 +1,144 @@
...
+      throw new Error("Operation was canceled");
+    }
+    throw new Error(
+      `Failed to get chunk at index ${JSON.stringify(index)}: ${error instanceof Error ? error.message : String(error)}`
+    );
+  } finally {
+    activeRequests.delete(id);
+  }
+
+  if (!isChunkData(chunk.data)) {
```
andy-sweet: Isn't already checked in `getChunk`?

packages/core/src/data/ome_zarr/worker_pool.ts
```diff
@@ -0,0 +1,295 @@
...
+type WorkerInstance = {
+  worker: Worker;
+  busy: boolean;
+  pendingCount: number;
+  workerId: number;
+};
+
+const DEFAULT_WORKER_COUNT = Math.min(navigator.hardwareConcurrency || 4, 8);
+let workerPool: WorkerInstance[] = [];
+let messageId = 0;
```
andy-sweet, line 23: Nit: I'd probably expect this one to be a UUID or similar. Granted, I have a hard time imagining some instance of Idetik running long enough and doing enough fetches that this eventually increments to infinity - maybe we could just leave this in as a signal of usage success?

andy-sweet (approved):
I'm mostly approving this due to the improved performance of usage, which means this moves us closer to getting respectable performance for the DCA data.

However, this is my first time thinking about web workers, so there's a lot I don't fully understand here.

I also left quite a lot of comments, most of which I think are worth addressing.

packages/core/src/data/zarr/open.ts
aganders3: I am not an expert here by any stretch, but my understanding is that it uses the "structured clone algorithm" which [supports these types](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm#supported_types).

The main limitation of something like `zarr.Location` or `zarr.Array` is that it cannot serialize `Function` objects, so it's limited to "only plain objects (e.g., from object literals)" for user-defined objects.

packages/core/src/data/ome_zarr/worker_pool.ts
aganders3: No, I don't think so. `getAvailableWorker()` should perhaps be renamed or even refactored - it should always return a worker if the pool is configured, it just tries to find the "least busy" worker.

packages/core/src/data/ome_zarr/worker_pool.ts
aganders3: I'll take shorter but may reconsider this function. Maybe it should just throw so it doesn't have to return an optional?

packages/core/src/data/ome_zarr/worker_pool.ts
aganders3: Yes, I considered this and I think it's a good idea. I'd probably change it to be a class when it can be generalized to be less specifically a zarr worker pool and could take the worker kernel as an input.

packages/core/src/data/ome_zarr/worker_kernel.ts
aganders3: We at least need to verify here (in the worker) that `chunk.data` is a transferable type, because `zarr.getChunk` can return `unknown[]`.

I was able to clean this up after realizing `zarr.Chunk` is a POJO. Now it checks here only that it's a `TypedArray` or `DataView`, and then keep the check for `isChunkData` where it was before (in the loader).

packages/core/src/data/zarr/open.ts
aganders3: Thanks for prompting me to rethink this. I was able to remove some overly verbose typing and just use `zarr.Chunk` in more places.

packages/core/src/data/ome_zarr/worker_pool.ts
aganders3: Yeah I'm going to leave it unless it starts causing problems. I think it's nicer to see these more readable values for debugging in logs, etc.

aganders3:
Okay I think I got all your suggestions implemented, and it's better and smaller!

I'm going to merge but of course still happy to make changes as they come up.

PR #464, opened 2025-09-24
feat: Restructure Examples and Add OmeZarrChunkedImageViewer Example

phoenixAja:
###  Summary

This PR restructures the React package examples directory to follow a consistent pattern and adds a comprehensive examples navigation page. The changes improve examples organization and user experience
  while maintaining backward compatibility. Also added a new example for OmeZarrChunkedImageViewer

###  Key Changes

  - Standardized all examples to follow the same index.html → Main.tsx → App.tsx pattern
  - Made examples fully self-contained with their own providers and dependencies
  - Created complete example structure for OmeZarrChunkedImageViewer
  - Added comprehensive examples/index.html with styled navigation grid
  - Created clickable example cards with descriptions and feature tags

###  Documentation Updates
  - Updated README.md with documentation for adding new examples

###  Testing Notes

  - All examples load correctly at their respective URLs
  - Navigation works from the index page
  - Both viewer components function as expected

To test directly: 
- run `npm run components` from project root
- navigate to http://localhost:5173/
- make sure you are connected to biohub VPN

  This PR improves the developer experience when exploring React package capabilities while maintaining clean codebase structure and providing better discoverability of available examples.


packages/react/examples/ome-zarr-chunked-image-viewer/App.tsx
```diff
@@ -0,0 +1,109 @@
...
+import { StyledEngineProvider, ThemeProvider } from "@mui/material/styles";
+import { Theme, InputSlider } from "@czi-sds/components";
+import CssBaseline from "@mui/material/CssBaseline";
+import useMediaQuery from "@mui/material/useMediaQuery";
+import { SliceCoordinates, ChunkedImageLayer } from "@idetik/core-prerelease";
+import { IdetikProvider, OmeZarrChunkedImageViewer } from "../../src";
+import { useCallback, useRef, useState } from "react";
+
+const sourceUrl =
+  "https://czii-onsite.czbiohub.org/krios1.processing/aretomo3/25jul30a/run002/vol003/Position_1_Vol.zarr";
```
andy-sweet, line 6: This requires biohub VPN access. That might be OK, but I wonder if we should just use a tomogram from the public cryoet data portal instead?

packages/react/src/components/viewers/OmeZarrImageViewer/OmeZarrImageViewer.tsx
```diff
@@ -412,7 +412,8 @@ async function loadImageMetadata(
 
   let initialZ: number;
   const zRegion = region.find(
-    (d) => d.dimension.toUpperCase() === seriesDimensionName.toUpperCase()
+    (d: Region[0]) =>
```
andy-sweet: Nit: there's nothing special about the first dimension.
```suggestion
    (d: Region[number]) =>
```

packages/react/README.md
```diff
@@ -163,13 +163,125 @@ npm install
 npm run build
 
 # Start development server
-npm run dev
+npm run components
```
andy-sweet: Nit: given the location of this README is packages/react and there are no special instructions here, I'd expect the command to be run from the same location, so revert.

```suggestion
npm run dev
```

packages/react/examples/ome-zarr-chunked-image-viewer/App.tsx
```diff
@@ -0,0 +1,109 @@
...
+  t: 0,
+  z: 296,
+};
+
+function ChunkedImageViewerDemo() {
+  const [sliceCoordinates, setSliceCoordinates] = useState<SliceCoordinates>(
+    initialSliceCoordinates
+  );
+  const updateZSliceRef = useRef<((zValue: number) => void) | null>(null);
+  const zMax = 591; // 0-indexed max for 592 slices
```
andy-sweet: Given the current definition of `SliceCoordinates`, this and the slider step should take into account the scale of the OME-Zarr. If the offset/translation is non-zero that will also affect this and the min.

packages/react/examples/ome-zarr-chunked-image-viewer/App.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+  const layerCreatedTime = useRef<number | undefined>(undefined);
+
+  console.log("Rendering App with sourceUrl:", sourceUrl);
+  const handleLayerCreated = useCallback(
+    (_layer?: ChunkedImageLayer, updateZSlice?: (zValue: number) => void) => {
+      layerCreatedTime.current = performance.now();
+      console.log(`Layer created at ${layerCreatedTime.current}`);
```
andy-sweet: I'm seeing this whenever the slider position moves. After enough slider position changes, I run into some memory allocation problems (maybe because there are lots of chunk manager sources?).

packages/react/examples/ome-zarr-image-viewer/Main.tsx
```diff
@@ -0,0 +1,13 @@
+import React from "react";
+import { createRoot } from "react-dom/client";
+import App from "./App";
+import "../../src/input.css";
+
+const domNode = document.getElementById("app")!;
+const root = createRoot(domNode);
+
+root.render(
```
andy-sweet, line 11: Compared to the previous main example/demo for this case, this is missing a lot of style. I wonder if there's a way all the demos could share the style providers (e.g. in the chunked image demo's App.tsx) - for example by reusing AppWithProviders that's currently under `src`.

packages/react/examples/ome-zarr-image-viewer/App.tsx
```diff
@@ -0,0 +1,97 @@
+import { Region } from "@idetik/core-prerelease";
```
andy-sweet, line 1: I think this is mostly copied from `src/App.tsx` (similar for `Main.tsx`). I think we should remove those files in `src` now that we have more structured examples.

andy-sweet (changes requested):
The new examples landing page is a great improvement from `main`. 

The main blocking feedback here is that the chunked image viewer demo seems to be broken (at least after I use a public dataset and update the z-slices). Maybe some of those changes had unintended effects though and I should just use the biohub vpn instead...

packages/react/examples/ome-zarr-chunked-image-viewer/App.tsx
phoenixAja: i'll take that out!

packages/react/examples/ome-zarr-chunked-image-viewer/App.tsx
phoenixAja: yeah i wanted this example though so we could debug what was going on with this image on the embrella site

andy-sweet: i'm ok to keep it for this PR (at least once I get biohub vpn access), but we'll probably want to switch it later

packages/react/src/components/AppWithProviders.tsx
```diff
@@ -3,12 +3,16 @@ import { StyledEngineProvider, ThemeProvider } from "@mui/material/styles";
...
-import App from "./App";
 import { IdetikProvider } from "./providers/IdetikProvider";
 
+interface AppWithProvidersProps {
+  children?: ReactNode;
+}
+
 // Create a wrapper component to handle providers
-export default function AppWithProviders() {
+export default function AppWithProviders({ children }: AppWithProvidersProps) {
```
andy-sweet, line 15: I wonder if this should also be in the `examples` directory instead of being something we export and index here. But happy to leave for a future discussion.

andy-sweet (commented):
I can't approve this because I can't run the new example yet.

I also noticed that the existing three examples do not expand to take up all the vertical space they can, though I'm not sure why. That is an undesirable change from main that should be fixed before merging.

Otherwise, no major comments!

packages/react/examples/ome-zarr-chunked-image-viewer/App.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+const sourceUrl =
+  "https://czii-onsite.czbiohub.org/krios1.processing/aretomo3/25jul30a/run002/vol003/Position_1_Vol.zarr";
+const initialSliceCoordinates: SliceCoordinates = {
+  t: 0,
+  z: 296,
+};
+
+function ChunkedImageViewerDemo() {
+  const [sliceCoordinates, setSliceCoordinates] = useState<SliceCoordinates>(
```
aganders3: Using state for `sliceCoordinates` and passing it as a prop to `OmeZarrChunkedImageViewer` causes it to re-initialize the layer whenever it changes. This resets the zoom level, etc.

aganders3 (changes requested):
Overall this is really nice and it's cool to have this gallery for the React examples. I haven't reviewed the whole thing, but found this bug when testing the `OmeZarrChunkedImageViewer`.

packages/react/examples/ome-zarr-chunked-image-viewer/App.tsx
phoenixAja: thanks!

phoenixAja: i also had to move fallbackContrastLimits and viewerClassNames as consts at the top of the file because they were also causing re-initialization! Fixed now

aganders3:
>To test directly:
> * `run npm run components` from project root

note for anyone reading this: you also still need to `npm run build` to build core first

aganders3:
This is great - thanks! I'm able to reproduce the problem Utz demonstrated (and worse!) with this. Now I can try to debug that.
Screenshot 2025-09-26 at 7 06 06 AM

I'll try to review the rest of this more thoroughly today.

packages/react/index.html
```diff
@@ -17,14 +85,68 @@
...
+            <span class="tag">Multi-channel</span>
+            <span class="tag">ImageViewer</span>
+          </div>
+        </a>
+
+        <a href="/examples/local-files/" class="example">
+          <h2>Local Files</h2>
+          <p>
+            Load local Zarr files using the file picker API. Demonstrates how to
+            load and visualize local OME-ZARR datasets using the File System
```
andy-sweet: Nit

```suggestion
            load and visualize local OME-Zarr datasets using the File System
```

packages/react/src/components/viewers/OmeZarrChunkedImageViewer/OmeZarrChunkedImageViewer.tsx
```diff
@@ -131,14 +131,7 @@ export function OmeZarrChunkedImageViewer({
...
     // eslint-disable-next-line react-hooks/exhaustive-deps -- Only props that trigger reinitialize
-  }, [
-    sourceUrl,
-    sliceCoordinates,
-    fallbackContrastLimits,
-    directory,
-    path,
-    runtime,
-  ]);
+  }, [sourceUrl, fallbackContrastLimits, directory, path, runtime]);
```
aganders3: Can `sliceCoordinates` be added back here to satisfy the linter? In general usage it should be mutated, not re-created.

packages/react/README.md
```diff
@@ -166,10 +166,122 @@ npm run build
...
+        ├── Main.tsx
+        └── App.tsx
+```
+
+#### Development Tips
+
+- Use relative imports to reference the React components: `import { Component } from "../../src"`
+- Import base styles with: `import "../../src/input.css"`
+- Examples automatically hot-reload during development
+- Each example runs independently and can have its own dependencies/setup
```
aganders3, line 284:
```suggestion
- Each example runs independently and can have its own dependencies/setup
- Changes to `@idetik/core` may _not_ hot-reload, and require a rebuild (`npm run build`) of that workspace
```

packages/react/examples/local-files/Main.tsx
```diff
@@ -15,7 +15,7 @@ const root = createRoot(domNode);
 
 root.render(
   <React.StrictMode>
-    <div className={cns("font-sds-body")}>
+    <div className={cns("font-sds-body", "h-screen")}>
```
aganders3: Should this use `AppWithProviders`?

packages/react/examples/organelle-box-time/Main.tsx
```diff
@@ -15,7 +15,7 @@ const root = createRoot(domNode);
 
 root.render(
   <React.StrictMode>
-    <div className={cns("font-sds-body")}>
+    <div className={cns("font-sds-body", "h-screen")}>
```
aganders3: Should this use `AppWithProviders`?

aganders3 (commented):
This generally looks good, but I think we should hold off for a decision on #465.

Sorry I broke this!

packages/react/examples/local-files/Main.tsx
phoenixAja: yes, fixed, ty!

packages/react/src/components/viewers/OmeZarrChunkedImageViewer/OmeZarrChunkedImageViewer.tsx
phoenixAja: That's a bit of a tricky situation. sliceCoordinates is a state variable that creates a new object on every slider movement, so adding it to the dependency array would recreate the layer constantly. I added the eslint-disable comment as a workaround for now.

The issue is that ChunkedImageLayer needs the initial sliceCoordinates for setup, but then it only looks for in-place mutations to sliceCoords_.z to trigger re-slicing. 

Unless i'm not thinking of a correct fix off the top of my head

packages/react/src/components/viewers/OmeZarrChunkedImageViewer/OmeZarrChunkedImageViewer.tsx
```diff
@@ -98,20 +105,16 @@ export function OmeZarrChunkedImageViewer({
...
       sliceCoordsRef.current = sliceCoords;
+      dimensionMapRef.current = loader.getSourceDimensionMap();
 
-      // Create updateZSlice function for external control
-      const updateZSlice = (zValue: number) => {
-        if (sliceCoords) {
-          sliceCoords.z = zValue;
-        }
-      };
+      setZMaxIndex?.(dimensionMapRef.current.z?.lods[0]?.size);
```
andy-sweet: My mistake - sorry!

```suggestion
      const zSize = dimensionMapRef.current.z?.lods[0]?.size;
      setZMaxIndex?.(zSize !== undefined ? zSize - 1 : undefined);
```

andy-sweet (approved):
This works well for me now. There's a bug from my commit (missed a -1 on the size -> max index) and another nit.

aganders3 (approved):

PR #466, opened 2025-09-27
Fix react example style and simplify z-index

andy-sweet:
This does two things to hopefully improve https://github.com/chanzuckerberg/idetik/pull/464

1. Fix the style issues with the react example.
This fix works by including the examples directory in the tailwind config, and then making some other changes so that the examples are relatively consistent.

2. Simplify the handling the z-coordinate.
This works by allowing the app to use an integer z-index, and forcing the viewer to do the conversion. I think this conversion from LOD0-index to physical value should really be handled by Idetik core, in which case it could be moved off this component.


phoenixAja (approved):
ty!! Looks great!!

PR #467, opened 2025-09-29
Add simple prefetching over time 

### LINKED ISSUE #454: Add prefetching in time dimension to chunk manager
The goal of this issue to add any type of prefetching across the time dimension in the chunk manager.

This is likely motivated by the ability to enable a simple playback mode, so fetching some number of frames ahead of the current time coordinate in a sliding window style might be sufficient here. We also need to pick some relatively high priority for time so that it happens before spatial prefetching.

andy-sweet:
### Summary
This adds some prefetching over time to the chunk manager that simply fetches the spatially visible lowest-resolution chunks for the next N time points.

N is a constant currently set to 10. I also changed the spatial prefetch constant to 0 to temporally disable that in favor of temporal prefetching. In the future, this should both be part of a larger user-specified configuration for the source or chunk manager. 

By default temporally prefetched chunks are prioritized below visible chunks at the current time point, but above background (non-visible) fallback LOD chunks. The temporally prefetched chunks can be bumped in priority by setting `ChunkManagerSource.prioritizePrefetchTime`, which prioritizes them above visible chunks at the current LOD.

### Related Issue
Closes #454

### Tests & Checks
I manually tested the chunk streaming examples.

### DIFF
--- a/packages/core/examples/lil_gui_utils.ts
+++ b/packages/core/examples/lil_gui_utils.ts
@@ -11,6 +11,7 @@ type DimensionSliderProps = {
   playback?: {
     maxRateHz?: number;
     stride?: number;
+    onRateChange?: (rateHz: number) => void;
   };
 };
 
@@ -35,7 +36,10 @@ export function addDimensionSlider(props: DimensionSliderProps) {
     const maxRateHz = props.playback.maxRateHz ?? 30;
     props.gui
       .add(playbackController, "rateHz", 0, maxRateHz, 1)
-      .name(`${props.dimensionName}-playback rate (Hz)`);
+      .name(`${props.dimensionName}-playback rate (Hz)`)
+      .onChange((rateHz: number) => {
+        props.playback?.onRateChange?.(rateHz);
+      });
   }
 }
 
--- a/packages/core/examples/lil_gui_utils.ts
+++ b/packages/core/examples/lil_gui_utils.ts
@@ -9,7 +9,7 @@ type DimensionSliderProps = {
   maxValue: number;
   stepValue: number;
   playback?: {
-    intervalMs: number;
+    maxRateHz?: number;
     stride?: number;
   };
 };
@@ -28,27 +28,19 @@ export function addDimensionSlider(props: DimensionSliderProps) {
   if (props.playback) {
     const playbackController = new PlaybackController({
       controller,
-      intervalMs: props.playback.intervalMs,
       start: props.minValue,
       stop: props.maxValue,
       step: props.stepValue * (props.playback.stride ?? 1),
     });
+    const maxRateHz = props.playback.maxRateHz ?? 30;
     props.gui
-      .add({ play: false }, "play")
-      .name(`Play ${props.dimensionName}`)
-      .onChange((play: boolean) => {
-        if (play) {
-          playbackController.play();
-        } else {
-          playbackController.pause();
-        }
-      });
+      .add(playbackController, "rateHz", 0, maxRateHz, 1)
+      .name(`${props.dimensionName}-playback rate (Hz)`);
   }
 }
 
 type PlaybackControllerProps = {
   controller: Controller;
-  intervalMs: number;
   start: number;
   stop: number;
   step: number;
@@ -59,25 +51,37 @@ class PlaybackController {
   private readonly start_: number;
   private readonly stop_: number;
   private readonly step_: number;
-  private readonly intervalMs_: number;
+  private rateHz_: number;
   private intervalId_?: number;
 
   constructor(props: PlaybackControllerProps) {
     this.controller_ = props.controller;
-    this.intervalMs_ = props.intervalMs;
+    this.rateHz_ = 0;
     this.start_ = props.start;
     this.stop_ = props.stop;
     this.step_ = props.step;
   }
 
-  public play() {
-    if (this.intervalId_ !== undefined) return;
-    this.intervalId_ = window.setInterval(() => {
-      this.incrementTime();
-    }, this.intervalMs_);
+  public get rateHz(): number {
+    return this.rateHz_;
   }
 
-  public pause() {
+  public set rateHz(rateHz: number) {
+    if (rateHz < 0) {
+      throw new Error(`Rate must be non-negative: ${rateHz}`);
+    }
+    if (rateHz !== this.rateHz_) {
+      this.pause();
+      this.rateHz_ = rateHz;
+    }
+    if (this.rateHz_ === 0) return;
+    const intervalMs = 1000 / rateHz;
+    this.intervalId_ = window.setInterval(() => {
+      this.incrementTime();
+    }, intervalMs);
+  }
+
+  private pause() {
     if (this.intervalId_ === undefined) return;
     window.clearInterval(this.intervalId_);
     this.intervalId_ = undefined;
--- a/packages/core/examples/lil_gui_utils.ts
+++ b/packages/core/examples/lil_gui_utils.ts
@@ -1,0 +1,94 @@
+import { SliceCoordinates } from "@/index";
+import { Controller, GUI } from "lil-gui";
+
+type DimensionSliderProps = {
+  gui: GUI;
+  sliceCoords: SliceCoordinates;
+  dimensionName: "z" | "t";
+  minValue: number;
+  maxValue: number;
+  stepValue: number;
+  playback?: {
+    intervalMs: number;
+    stride?: number;
+  };
+};
+
+export function addDimensionSlider(props: DimensionSliderProps) {
+  const controller = props.gui
+    .add(
+      props.sliceCoords,
+      props.dimensionName,
+      props.minValue,
+      props.maxValue,
+      props.stepValue
+    )
+    .name(`${props.dimensionName}-coord`);
+
+  if (props.playback) {
+    const playbackController = new PlaybackController({
+      controller,
+      intervalMs: props.playback.intervalMs,
+      start: props.minValue,
+      stop: props.maxValue,
+      step: props.stepValue * (props.playback.stride ?? 1),
+    });
+    props.gui
+      .add({ play: false }, "play")
+      .name(`Play ${props.dimensionName}`)
+      .onChange((play: boolean) => {
+        if (play) {
+          playbackController.play();
+        } else {
+          playbackController.pause();
+        }
+      });
+  }
+}
+
+type PlaybackControllerProps = {
+  controller: Controller;
+  intervalMs: number;
+  start: number;
+  stop: number;
+  step: number;
+};
+
+class PlaybackController {
+  private readonly controller_: Controller;
+  private readonly start_: number;
+  private readonly stop_: number;
+  private readonly step_: number;
+  private readonly intervalMs_: number;
+  private intervalId_?: number;
+
+  constructor(props: PlaybackControllerProps) {
+    this.controller_ = props.controller;
+    this.intervalMs_ = props.intervalMs;
+    this.start_ = props.start;
+    this.stop_ = props.stop;
+    this.step_ = props.step;
+  }
+
+  public play() {
+    if (this.intervalId_ !== undefined) return;
+    this.intervalId_ = window.setInterval(() => {
+      this.incrementTime();
+    }, this.intervalMs_);
+  }
+
+  public pause() {
+    if (this.intervalId_ === undefined) return;
+    window.clearInterval(this.intervalId_);
+    this.intervalId_ = undefined;
+  }
+
+  private incrementTime = () => {
+    const newValue = this.controller_.getValue() + this.step_;
+    if (newValue <= this.stop_) {
+      this.controller_.setValue(newValue);
+    } else {
+      this.controller_.setValue(this.start_);
+    }
+  };
+}
--- a/packages/core/examples/lil_gui_utils.ts
+++ b/packages/core/examples/lil_gui_utils.ts
@@ -26,20 +26,19 @@ export function addDimensionSlider(props: DimensionSliderProps) {
     )
     .name(`${props.dimensionName}-coord`);
 
-  const playback = props.playback;
-  if (playback) {
+  if (props.playback) {
     const playbackController = new PlaybackController({
       controller,
       start: props.minValue,
       stop: props.maxValue,
-      step: props.stepValue * (playback.stride ?? 1),
+      step: props.stepValue * (props.playback.stride ?? 1),
     });
-    const maxRateHz = playback.maxRateHz ?? 30;
+    const maxRateHz = props.playback.maxRateHz ?? 30;
     props.gui
       .add(playbackController, "rateHz", 0, maxRateHz, 1)
       .name(`${props.dimensionName}-playback rate (Hz)`)
       .onChange((rateHz: number) => {
-        playback.onRateChange?.(rateHz);
+        props.playback?.onRateChange?.(rateHz);
       });
   }
 }
--- a/packages/core/examples/lil_gui_utils.ts
+++ b/packages/core/examples/lil_gui_utils.ts
@@ -26,19 +26,20 @@ export function addDimensionSlider(props: DimensionSliderProps) {
     )
     .name(`${props.dimensionName}-coord`);
 
-  if (props.playback) {
+  const playback = props.playback;
+  if (playback) {
     const playbackController = new PlaybackController({
       controller,
       start: props.minValue,
       stop: props.maxValue,
-      step: props.stepValue * (props.playback.stride ?? 1),
+      step: props.stepValue * (playback.stride ?? 1),
     });
-    const maxRateHz = props.playback.maxRateHz ?? 30;
+    const maxRateHz = playback.maxRateHz ?? 30;
     props.gui
       .add(playbackController, "rateHz", 0, maxRateHz, 1)
       .name(`${props.dimensionName}-playback rate (Hz)`)
       .onChange((rateHz: number) => {
-        props.playback?.onRateChange?.(rateHz);
+        playback.onRateChange?.(rateHz);
       });
   }
 }
--- a/packages/core/src/core/chunk_manager.ts
+++ b/packages/core/src/core/chunk_manager.ts
@@ -333,7 +333,6 @@ export class ChunkManagerSource {
       if (delta >= 0 && delta <= PREFETCH_TIME_POINTS) continue;
       const chunks = this.chunks_[t];
       for (const chunk of chunks) {
-        chunk.visible = false;
         this.disposeChunk(chunk);
         disposedChunks.push(chunk);
       }
--- a/packages/core/src/core/chunk_manager.ts
+++ b/packages/core/src/core/chunk_manager.ts
@@ -36,6 +36,7 @@ export class ChunkManagerSource {
   private lastViewBounds2D_: Box2 | null = null;
   private lastZBounds_?: [number, number];
   private lastTCoord_?: number;
+  private fetchedTCoords_: Set<number> = new Set();
 
   constructor(loader: ChunkLoader, sliceCoords: SliceCoordinates) {
     this.loader_ = loader;
@@ -137,6 +138,12 @@ export class ChunkManagerSource {
     return this.chunks_[this.sliceCoords_.t ?? 0];
   }
 
+  public allVisibleLowestLODLoaded(): boolean {
+    return this.getChunksAtCurrentTime()
+      .filter((c) => c.visible && c.lod === this.lowestResLOD_)
+      .every((c) => c.state === "loaded");
+  }
+
   public updateAndCollectChunkChanges(lodFactor: number, viewBounds2D: Box2) {
     this.setLOD(lodFactor);
     const zBounds = this.getZBounds();
@@ -223,6 +230,7 @@ export class ChunkManagerSource {
     const updatedChunks = this.updatePreviousTimeChunks();
 
     const currentTimeChunks = this.chunks_[this.sliceCoords_.t ?? 0];
+    this.fetchedTCoords_.add(this.sliceCoords_.t ?? 0);
     for (const chunk of currentTimeChunks) {
       const isVisible = this.isChunkWithinBounds(chunk, viewBounds3D);
       const eligibleForPrefetch =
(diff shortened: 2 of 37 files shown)


andy-sweet:
@aganders3 : assigned to review for continuity
@shlomnissan : assigned to catch up on the (modest) progress we have for temporal playback

packages/core/src/core/chunk_manager.ts
```diff
@@ -267,27 +277,66 @@ export class ChunkManagerSource {
...
     chunk.data = undefined;
     chunk.state = "unloaded";
     chunk.priority = null;
     chunk.orderKey = null;
-    Logger.debug("ChunkManagerSource", `Disposing chunk in LOD ${chunk.lod}`);
+    chunk.prefetch = false;
+    Logger.debug(
+      "ChunkManagerSource",
+      `Disposing chunk ${JSON.stringify(chunk.chunkIndex)} in LOD ${chunk.lod}`
+    );
```
aganders3, line 382: Can/should this log include the timepoint? Or else perhaps log the timepoints in `disposeStaleTimeChunks`.

packages/core/src/core/chunk_manager.ts
```diff
@@ -267,27 +277,66 @@ export class ChunkManagerSource {
...
+    for (let t = this.sliceCoords_.t + 1; t < tEnd; ++t) {
+      for (const chunk of this.chunks_[t]) {
+        if (chunk.state !== "unloaded") continue;
+        const isLowestLOD = chunk.lod === this.lowestResLOD_;
+        const isVisible = this.isChunkWithinBounds(chunk, viewBounds3D);
+        if (isLowestLOD && isVisible) {
+          chunk.prefetch = true;
+          chunk.priority = this.prioritizePrefetchTime
+            ? PRI_PREFETCH_TIME_HIGH
+            : PRI_PREFETCH_TIME_LOW;
```
aganders3: What do you think of this moving into `computePriority` by giving that a new parameter like `isCurrentTimepoint`?

packages/core/src/core/chunk_manager.ts
```diff
@@ -267,27 +277,66 @@ export class ChunkManagerSource {
...
+          chunk.orderKey = t - this.sliceCoords_.t;
```
aganders3: This loses the spatial prioritization. That may be fine, or would it work to multiply this by `orderKeyByDistance`?

packages/core/src/core/chunk_manager.ts
```diff
@@ -267,27 +277,66 @@ export class ChunkManagerSource {
...
+          chunk.state = "queued";
```
aganders3: For non-time-prefetch chunks this is first checked to make sure it's transitioning from `unloaded` -> `queued` and not `loaded` -> `queued`. I'm not sure that's possible here, but I'm losing track of the state management a bit.

packages/core/src/core/chunk_manager.ts
```diff
@@ -15,12 +15,17 @@ import { ChunkQueue } from "../data/chunk_queue";
 
 // Number of chunks to extend beyond the visible bounds in each direction (x/y/z)
 // These additional chunks are prefetched to improve responsiveness when panning.
-const PREFETCH_PADDING_CHUNKS = 1;
+const PREFETCH_PADDING_CHUNKS = 0;
+
+// Fetch some number of timepoints ahead of current time.
+const PREFETCH_TEMPORAL_CHUNKS = 10;
```
aganders3: This is making an assumption of 1 timepoint/chunk, right?

edit: or maybe that's always true for "chunks" in this context, if not necessarily in the source. That is, we're already kind of using our own "logical chunks" that don't quite match the source.

packages/core/src/core/chunk_manager.ts
```diff
@@ -267,27 +277,66 @@ export class ChunkManagerSource {
...
+    return disposedChunks;
   }
 
   private disposeChunk(chunk: Chunk) {
     chunk.data = undefined;
     chunk.state = "unloaded";
     chunk.priority = null;
     chunk.orderKey = null;
-    Logger.debug("ChunkManagerSource", `Disposing chunk in LOD ${chunk.lod}`);
+    chunk.prefetch = false;
```
aganders3, line 378: Why does setting `chunk.prefetch = false` get moved in here, but `chunk.visible = false` is still set by the caller?

packages/core/src/core/chunk_manager.ts
```diff
@@ -37,6 +42,9 @@ export class ChunkManagerSource {
   private lastZBounds_?: [number, number];
   private lastTCoord_?: number;
 
+  public prioritizePrefetchTime: boolean = false;
```
aganders3, line 52: Is this sort of a proxy for what we were calling "playback mode" before?

packages/core/src/core/chunk_manager.ts
```diff
@@ -226,9 +234,11 @@ export class ChunkManagerSource {
     const center = vec2.create();
     vec2.lerp(center, viewBounds2D.min, viewBounds2D.max, 0.5);
 
-    const updatedChunks = this.updatePreviousTimeChunks();
+    const updatedChunks: Chunk[] = [];
+    updatedChunks.push(...this.disposeStaleTimeChunks());
 
     const currentTimeChunks = this.chunks_[this.sliceCoords_.t ?? 0];
+    this.fetchedTCoords_.add(this.sliceCoords_.t ?? 0);
```
aganders3: nit: in `prefetchTimeChunks` you set this _after_ iterating the chunks for each frame. The only way I can think it might make a difference is if at some point the loop gains a break or some concurrency or something.

aganders3 (approved):
This is working for me, and I think achieves our goal of doing some kind of prefetching in time. I have a number of nitpick comments and questions, but I don't see anything here as blocking.

I'd like us to look at moving `sliceCoords` _off_ the `ChunkManagerSource` before going much farther. It's probably not a big problem (and may be a feature) for time, but I think the coupling on Z is blocking other slice orientations. That's a sizeable though and I'm not sure we're all on the same page about it. Maybe there's some alternative I'm not thinking of.

packages/core/src/core/chunk_manager.ts
andy-sweet: Given the current validated constraints, it does via `chunkIndex.t`.

packages/core/src/core/chunk_manager.ts
andy-sweet: I think this does not need to be added here at all because `prefetch` is only used internally as a local variable in `computePriority`. However, it is used externally in the chunk stats overlay, so I set it here out of precaution because I think that a chunk being disposed should not have `prefetch === true`. I think I could just not add it here though.

Not that at one point one this branch, I removed `Chunk.prefetch` entirely to avoid thinking about it. But that change (if desired) should probably happen independently of this.

packages/core/src/core/chunk_manager.ts
andy-sweet: I'd probably leave it as is for now because these priorities are intentionally a little different from the ones for the current time point (since there is no fallback here yet).

packages/core/src/core/chunk_manager.ts
andy-sweet: Good point.

I think we'd want a metric that first prioritizes time, then space, so given the current constraints we could add some normalized version of `orderKeyByDistance` in [0, 1) to the temporal distance to achieve this.

packages/core/src/core/chunk_manager.ts
andy-sweet: I make a similar, but simpler check even earlier here (look up), which is that I only get here if the chunk is not queued/loading/loaded (i.e. it's not unloaded). I think that's sufficient here.

But agreed that state management is getting harder and harder to track.

At one point in this branch, I did try to remove/simplify some of the chunk state, then split the chunks into different stores. That introduced a lot of new changes that aren't entirely related to the changes here, so felt it was better to do that in a separate change.

packages/core/src/core/chunk_manager.ts
andy-sweet: I see your point.

I do think `fetchedTCoords_` is a little fragile to future changes, but it's also quite simple and hopefully effective for the current implementation.

I'll consider moving this or the other so that they're somewhat consistent.

packages/core/src/core/chunk_manager.ts
andy-sweet: Good point. We currently assuming one timepoint per chunk, with a time-scale of 1 too, so this is in chunks, time points, and duration too!

I think my intent was more like time-points or frames here and that's probably the easiest thing for users to think about if/when we expose this, so I'd be down to rename to `PREFETCH_TIME_POINTS` if that helps?

In general, being more flexible about time points/scaling needs a decent amount of implementation work.

packages/core/src/core/chunk_manager.ts
andy-sweet: Yes, pretty much. 

The example front-end is driving the actual playback right now, so there isn't any other playback logic in the chunk manager (though there might be in the future). This was a simplest way I could think of to make the example front-end to also dynamically change the prioritization based on whether it was playing back.

andy-sweet:
> I'd like us to look at moving sliceCoords off the ChunkManagerSource before going much farther. It's probably not a big problem (and may be a feature) for time, but I think the coupling on Z is blocking other slice orientations. That's a sizeable though and I'm not sure we're all on the same page about it. Maybe there's some alternative I'm not thinking of.

Agreed. I'd love to find a way to prioritize coming up with some solution soon!

packages/core/examples/lil_gui_utils.ts
```diff
@@ -25,17 +26,21 @@ export function addDimensionSlider(props: DimensionSliderProps) {
     )
     .name(`${props.dimensionName}-coord`);
 
-  if (props.playback) {
+  const playback = props.playback;
```
shlomnissan: I actually prefer using `props.playback` since it makes the source of the value explicit.

packages/core/examples/lil_gui_utils.ts
```diff
@@ -25,17 +26,21 @@ export function addDimensionSlider(props: DimensionSliderProps) {
...
+  if (playback) {
     const playbackController = new PlaybackController({
```
shlomnissan, line 30: - I think the playback mechanism should be separated from `lil-gui`. A lightweight playback engine that only works with `getValue`/`setValue` callbacks and stays UI-agnostic would be easier to reuse across different interfaces and simpler to test.
- I think `requestAnimationFrame` is preferred because it syncs updates with the browser’s refresh cycle, avoiding the time drift and inefficiencies of `setInterval`.

packages/core/src/core/chunk_manager.ts
shlomnissan: I think the name `PREFETCH_TEMPORAL_CHUNKS` is a bit misleading. It’s actually the number of timepoints to prefetch, not the number of chunks. Consider renaming to `PREFETCH_TIMEPOINTS` for clarity.

packages/core/src/core/chunk_manager.ts
```diff
@@ -15,12 +15,17 @@ import { ChunkQueue } from "../data/chunk_queue";
...
 
 const PRI_FALLBACK_VISIBLE = 0;
-const PRI_VISIBLE_CURRENT = 1;
-const PRI_FALLBACK_BACKGROUND = 2;
-const PRI_PREFETCH = 3;
+const PRI_PREFETCH_TIME_HIGH = 1;
+const PRI_VISIBLE_CURRENT = 2;
+const PRI_PREFETCH_TIME_LOW = 3;
+const PRI_FALLBACK_BACKGROUND = 4;
+const PRI_PREFETCH_SPACE = 5;
```
shlomnissan, line 35: Consider adding short inline comments explaining each priority constant. For example, `PRI_PREFETCH_TIME_HIGH` wasn’t immediately clear (I initially assumed it meant high-resolution timepoints), and `PRI_PREFETCH_SPACE` was a bit confusing since it refers to spatial chunks marked for prefetching but not currently visible. A one line note for each would make the intent clearer until we implement prefetch policies.

packages/core/src/core/chunk_manager.ts
```diff
@@ -37,6 +42,9 @@ export class ChunkManagerSource {
   private lastZBounds_?: [number, number];
   private lastTCoord_?: number;
 
+  public prioritizePrefetchTime: boolean = false;
+  private fetchedTCoords_: Set<number> = new Set();
```
shlomnissan: The term “fetch” made me think these timepoints were already fetched and loaded (because JS fetch API), but it looks like this set is tracking which timepoints have been queued or touched. A clearer name like `trackedTCoords_` or `activeTCoords_` might be better?

packages/core/src/core/chunk_manager.ts
```diff
@@ -226,9 +234,11 @@ export class ChunkManagerSource {
     const center = vec2.create();
     vec2.lerp(center, viewBounds2D.min, viewBounds2D.max, 0.5);
 
-    const updatedChunks = this.updatePreviousTimeChunks();
+    const updatedChunks: Chunk[] = [];
```
shlomnissan: I know this isn’t from the current change, but the name `updatedChunks` is a bit confusing. Since it also includes disposed chunks, I always scratch my head when I see it. Maybe something like `changedChunks` or `modifiedChunks` would better reflect that it collects both updated and disposed chunks?

packages/core/src/core/chunk_manager.ts
```diff
@@ -267,27 +277,66 @@ export class ChunkManagerSource {
       }
     }
     updatedChunks.push(...currentTimeChunks);
```
shlomnissan: - The name `currentTimeChunks` could be confusing since we also support purely 3D datasets without time, maybe a more neutral name like `activeChunks` or `chunksAtCurrentCoord` would be clearer.
- `updateAndCollectChunkChangesForCurrentLod` is doing a lot right now. It might be worth refactoring by extracting the code that computes the current timepoint’s chunks into a helper.

packages/core/src/core/chunk_manager.ts
```diff
@@ -267,27 +277,66 @@ export class ChunkManagerSource {
...
 
-  private updatePreviousTimeChunks(): Chunk[] {
-    if (this.lastTCoord_ === undefined) return [];
-    if (this.lastTCoord_ === this.sliceCoords_.t) return [];
-    const lastTimeChunks = Array(...this.chunks_[this.lastTCoord_]);
-    for (const chunk of lastTimeChunks) {
-      chunk.visible = false;
-      chunk.prefetch = false;
-      this.disposeChunk(chunk);
+  private prefetchTimeChunks(viewBounds3D: Box3): Chunk[] {
```
shlomnissan: Consider renaming `prefetchTimeChunks` since it doesn’t actually prefetch data, it just marks or queues the chunks for prefetch. Maybe `markTimeChunksForPrefetch` might make the intent clearer?

packages/core/src/core/chunk_manager.ts
```diff
@@ -267,27 +277,66 @@ export class ChunkManagerSource {
...
+    if (this.sliceCoords_.t === undefined) return [];
```
shlomnissan: I’d extract this guard from `prefetchTimeChunks` and move it to the caller, so the function only handles the queuing logic itself. That would make the contract clearer. Same goes for `disposeStaleTimeChunks()`.

packages/core/src/core/chunk_manager.ts
```diff
@@ -267,27 +277,66 @@ export class ChunkManagerSource {
...
+    const tEnd = Math.min(
+      this.chunks_.length,
+      this.sliceCoords_.t + PREFETCH_TEMPORAL_CHUNKS + 1
```
shlomnissan: This is a bit misleading. Instead of adding the +1 here and then looping with `< tEnd`, it might be clearer to drop the +1 and update the loop to use `<= tEnd`. That way the intent "prefetch N future timepoints inclusive" is more obvious.

packages/core/src/core/chunk_manager.ts
```diff
@@ -267,27 +277,66 @@ export class ChunkManagerSource {
...
+    );
+    const updatedChunks: Chunk[] = [];
+    for (let t = this.sliceCoords_.t + 1; t < tEnd; ++t) {
+      for (const chunk of this.chunks_[t]) {
+        if (chunk.state !== "unloaded") continue;
+        const isLowestLOD = chunk.lod === this.lowestResLOD_;
+        const isVisible = this.isChunkWithinBounds(chunk, viewBounds3D);
```
shlomnissan: This variable name is a bit misleading. A chunk in a future timepoint can't really be “visible” yet. Since this check is really about whether the chunk intersects the current spatial bounds. `inViewBounds` might be better?

packages/core/src/core/chunk_manager.ts
```diff
@@ -267,27 +277,66 @@ export class ChunkManagerSource {
...
+
+  private disposeStaleTimeChunks(): Chunk[] {
+    if (this.sliceCoords_.t === undefined) return [];
+    const disposedChunks: Chunk[] = [];
+    for (const t of this.fetchedTCoords_) {
+      const delta = t - this.sliceCoords_.t;
+      if (delta >= 0 && delta <= PREFETCH_TEMPORAL_CHUNKS) continue;
+      const chunks = this.chunks_[t];
+      for (const chunk of chunks) {
+        chunk.visible = false;
```
shlomnissan: I'm a little concerned that `chunk.visible = false` here conflates spatial visibility with temporal activeness. It might be clearer to decouple those concepts so “visible” only means within current timepoint.

shlomnissan (approved):
Looks great and works great (especially in Oregon 😂)! Most of my comments are just nits. A couple of things I’m a little bit worried about: the potential confusion around visible semantics (spatial vs. temporal) and some naming that could be clearer (`PREFETCH_TEMPORAL_CHUNKS`, `prefetchTimeChunks`, etc.) but nothing blocking.

packages/core/src/core/chunk_manager.ts
shlomnissan: Ha I didn't see your response to the original comment 😅

andy-sweet: Renamed!

packages/core/src/core/chunk_manager.ts
andy-sweet: Done.

packages/core/src/core/chunk_manager.ts
andy-sweet: Agreed that naming could be better here. I considered `activeTCoords_`, but it's a little vague.

I went with `tCoordsWithQueuedChunks` as the most descriptive and accurate I could think of for now.

packages/core/src/core/chunk_manager.ts
andy-sweet: I think of those as synonyms, but if it's confusing to you, I'm happy to change to `modifiedChunks`.

packages/core/src/core/chunk_manager.ts
andy-sweet: `inViewBounds` is better. Done.

packages/core/src/core/chunk_manager.ts
andy-sweet: Renamed.

packages/core/src/core/chunk_manager.ts
andy-sweet: Done. Both functions take `currentTime: number` as an argument now to remove `undefined` from `this.sliceCoords_.t` once it's been checked.

packages/core/examples/lil_gui_utils.ts
andy-sweet: Agreed that this is not a good long term home for playback.

However, given this PR is about prefetching and not playback, I won't make the change here. I created a [follow-up issue](https://github.com/chanzuckerberg/idetik/issues/469) where we can discuss some options.

packages/core/examples/lil_gui_utils.ts
andy-sweet: Done.

packages/core/src/core/chunk_manager.ts
andy-sweet: Also note that a visible chunk may be disposed (e.g. if it is not in the current or fallback LOD).

packages/core/src/core/chunk_manager.ts
andy-sweet: Good point. Given that `visible` previously meant spatial visibility, I'm going to keep that definition. Temporal visibility is currently handled differently (e.g. by indexing `this.chunks_`) and it's also easy to infer if a chunk is the current one (assuming you have `SliceCoordinates`).

Therefore, I think I can just remove this modification because the main update updates all the chunks at the current time point (including their spatial visibility) and `getChunks` only returns chunks at the current time point.

packages/core/src/core/chunk_manager.ts
andy-sweet: I'm going to leave this for now, but I think we should consider removing `Chunks.prefetch` as state after this PR is merged.

packages/core/src/core/chunk_manager.ts
andy-sweet: Done.

packages/core/src/core/chunk_manager.ts
andy-sweet: > The name currentTimeChunks could be confusing since we also support purely 3D datasets without time, maybe a more neutral name like activeChunks or chunksAtCurrentCoord would be clearer.

The renames seem vague or less clear, so I'm going to avoid them.

Even when the dataset doesn't have time, the chunk manager source gives it a unitary time dimension to simplify logic. It handles z similarly (e.g. when a dataset doesn't have a z-dimension) and I think the consistency holds well here.

andy-sweet: > updateAndCollectChunkChangesForCurrentLod is doing a lot right now. It might be worth refactoring by extracting the code that computes the current timepoint’s chunks into a helper.

Done.

packages/core/src/core/chunk_manager.ts
andy-sweet: Given the current logic for the current time point, I'm going to leave this for now. May be worth moving or changing in future refactors.

packages/core/src/core/chunk_manager.ts
andy-sweet: I added something as described above, but it maybe adds more complexity than it's worth. I'll think about whether to keep this over lunch. Otherwise, I'll revert it.

PR #468, opened 2025-09-29
Prepare for latest npm package versions

andy-sweet:
### Summary
This change is the result of running `npm version major` in both the core and react packages.

I decided to bump the major version because this is a pre-release. I'm not sure if there are breaking changes in either package and very few apps depend on either of these packages, but this offers some protection for them.

### Tests & Checks
I ran `npm install` to check that nothing changes after these bumps.
Then I ran `npm run build:all` to check that both packages build.
Then I ran `npm run examples` to check that the core examples work.
Then I ran `npm run components` to check that the react example works.

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@idetik/react-prerelease",
-  "version": "1.1.1",
+  "version": "2.0.0",
   "type": "module",
   "scripts": {
     "dev": "vite",
@@ -52,7 +52,7 @@
     "idetik"
   ],
   "devDependencies": {
-    "@idetik/core-prerelease": "^2.0.0",
+    "@idetik/core-prerelease": "^3.0.0",
     "@testing-library/react": "^16.3.0",
     "@types/classnames": "^2.3.0",
     "autoprefixer": "^10.4.20",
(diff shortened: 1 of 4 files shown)


andy-sweet:
@aganders3 : assigning you because you reviewed previous package releases.

Note that I have not run `npm publish` yet for either package. I plan to do that after this PR is approved.

After this is merged, I'll create a tag `core-prerelease-v3.0.0` so we have some simple way of tracking versions in the repo before we split things into separate core and react repos.

packages/react/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@idetik/react-prerelease",
-  "version": "1.1.1",
+  "version": "2.0.0",
```
aganders3, line 3: I thought perhaps the major version of `core` and `react` would stay in sync, but this is giving me pause. I guess maybe it's better if they're specifically _not_ related?

aganders3 (approved):
Looks good!

packages/react/package.json
andy-sweet: Well, they weren't in sync before this, so I didn't change anything. Given they are both pre-releases that should be deprecated once we have something that's actually stable, I guess it doesn't matter right now.

Equally, I guess we could make them the same while they live in the same repo. I'll leave that for the future though, given we might split before another publish anyway.

andy-sweet:
These are now published on npm. I also added a tag to the squashed merge commit with the name `prerelease-3.0.0`.

PR #470, opened 2025-09-30
Trim padding from edge chunks in OME-Zarr image loader

### LINKED ISSUE #428: Remove or hide padding on edge chunks
Fetching constant-size chunks based on the source data results in padding on the edge chunks (right and bottom). This is generally not a big issue with positive-valued images as the padding will match the renderer background (black). The padding is quite noticeable on the CryoET images because the images are not positive-valued, so with normal contrast limits the padding appears as a neutral gray tone.

Removing the padding could improve memory usage, but may be more complicated. It's probably enough to figure out a way to just discard the padding fragments for now, if that's easier.

aganders3:
### Summary

This PR adds special chunk shape handling for edge chunks in OME-Zarr datasets. Previously, edge chunks (chunks at the boundaries of an image that don't align with full chunk dimensions) would receive padded data from zarrita, which resulted in (sometimes large) areas of zero-valued padding on the bottom/right of the image (viewing a z slice in standard orientation).

#### Changes
  - `ChunkManager`: Calculate actual chunk dimensions based on remaining image size for edge chunks
  - `OmeZarrImageLoader`: Detect when received chunk data is larger than expected and remove padding in-place by copying only the valid data rows
  
This does not handle similar issues chunking in `c` or `t`. It seems unlikely that channels would be chunked in such a way, though I'm not sure about time. This can also be handled later since the chunk manager already makes some assumptions about these dimensions.

https://github.com/user-attachments/assets/539e7bae-3094-4169-930b-f7c6061d35fc

### Related Issue
Closes #428

### Tests & Checks
Tested with chunk streaming and OME-Zarr v0.5 examples
Tested with `OmeZarrChunkedImageViewer` react example

### DIFF
--- a/packages/core/src/core/chunk_manager.ts
+++ b/packages/core/src/core/chunk_manager.ts
@@ -68,9 +68,21 @@ export class ChunkManagerSource {
 
         for (let x = 0; x < chunksX; ++x) {
           const xOffset = xLod.translation + x * xLod.chunkSize * xLod.scale;
+          const actualChunkWidth = Math.min(
+            chunkWidth,
+            xLod.size - x * chunkWidth
+          );
           for (let y = 0; y < chunksY; ++y) {
+            const actualChunkHeight = Math.min(
+              chunkHeight,
+              yLod.size - y * chunkHeight
+            );
             const yOffset = yLod.translation + y * yLod.chunkSize * yLod.scale;
             for (let z = 0; z < chunksZ; ++z) {
+              const actualChunkDepth = Math.min(
+                chunkDepth,
+                (zLod?.size ?? 1) - z * chunkDepth
+              );
               const zOffset =
                 zLod !== undefined
                   ? zLod.translation + z * chunkDepth * zLod.scale
@@ -83,12 +95,12 @@ export class ChunkManagerSource {
                 priority: null,
                 orderKey: null,
                 shape: {
-                  x: chunkWidth,
-                  y: chunkHeight,
-                  z: chunkDepth,
+                  x: actualChunkWidth,
+                  y: actualChunkHeight,
+                  z: actualChunkDepth,
                   c: channels,
                 },
-                rowStride: chunkWidth,
+                rowStride: actualChunkWidth,
                 rowAlignmentBytes: 1,
                 chunkIndex: { x, y, z, t },
                 scale: {
--- a/packages/core/src/core/chunk_manager.ts
+++ b/packages/core/src/core/chunk_manager.ts
@@ -68,21 +68,10 @@ export class ChunkManagerSource {
 
         for (let x = 0; x < chunksX; ++x) {
           const xOffset = xLod.translation + x * xLod.chunkSize * xLod.scale;
-          const actualChunkWidth = Math.min(
-            chunkWidth,
-            xLod.size - x * chunkWidth
-          );
+          const rowStride = Math.min(chunkWidth, xLod.size - x * chunkWidth);
           for (let y = 0; y < chunksY; ++y) {
-            const actualChunkHeight = Math.min(
-              chunkHeight,
-              yLod.size - y * chunkHeight
-            );
             const yOffset = yLod.translation + y * yLod.chunkSize * yLod.scale;
             for (let z = 0; z < chunksZ; ++z) {
-              const actualChunkDepth = Math.min(
-                chunkDepth,
-                (zLod?.size ?? 1) - z * chunkDepth
-              );
               const zOffset =
                 zLod !== undefined
                   ? zLod.translation + z * chunkDepth * zLod.scale
@@ -95,12 +84,12 @@ export class ChunkManagerSource {
                 priority: null,
                 orderKey: null,
                 shape: {
-                  x: actualChunkWidth,
-                  y: actualChunkHeight,
-                  z: actualChunkDepth,
+                  x: Math.min(chunkWidth, xLod.size - x * chunkWidth),
+                  y: Math.min(chunkWidth, xLod.size - y * chunkWidth),
+                  z: Math.min(chunkDepth, (zLod?.size ?? 1) - z * chunkDepth),
                   c: channels,
                 },
-                rowStride: actualChunkWidth,
+                rowStride,
                 rowAlignmentBytes: 1,
                 chunkIndex: { x, y, z, t },
                 scale: {
--- a/packages/core/src/data/ome_zarr/image_loader.ts
+++ b/packages/core/src/data/ome_zarr/image_loader.ts
@@ -19,7 +19,6 @@ import { Image as OmeZarrImage } from "./0.5/image";
 import { Readable } from "@zarrita/storage";
 import { ZarrArrayParams } from "../zarr/open";
 import { getChunk } from "./worker_pool";
-import { Logger } from "../../utilities/logger";
 
 // Implements the interface required for getting array chunks in zarrita:
 // https://github.com/manzt/zarrita.js/blob/c15c1a14e42a83516972368ac962ebdf56a6dcdb/packages/indexing/src/types.ts#L52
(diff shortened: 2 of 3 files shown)


packages/core/src/data/ome_zarr/image_loader.ts
```diff
@@ -98,23 +99,68 @@ export class OmeZarrImageLoader {
       signal,
     });
 
-    const data = subarray.data;
-    if (!isChunkData(data)) {
+    this.assignChunkData(chunk, subarray);
+  }
+
+  private assignChunkData(chunk: Chunk, subarray: zarr.Chunk<zarr.DataType>) {
+    const wholeChunk = subarray.data;
```
phoenixAja: subarry feels like a very ambiguous variable name, can it be something more descriptive? Unless that's the actual terminology people use.

packages/core/src/data/ome_zarr/image_loader.ts
```diff
@@ -98,23 +99,68 @@ export class OmeZarrImageLoader {
...
+    }
+
+    if (
+      subarray.shape[this.dimensions_.x.index] === chunk.shape.x &&
+      subarray.shape[this.dimensions_.y.index] === chunk.shape.y &&
+      (!this.dimensions_.z ||
+        subarray.shape[this.dimensions_.z.index] === chunk.shape.z)
+    ) {
+      chunk.data = wholeChunk;
+      chunk.rowStride = subarray.stride[this.dimensions_.y.index];
```
phoenixAja: should this be `this.dimensions_.x.index` since it's `rowStride`?

packages/core/src/data/ome_zarr/image_loader.ts
```diff
@@ -98,23 +99,68 @@ export class OmeZarrImageLoader {
...
+      (this.dimensions_.z &&
+        subarray.shape[this.dimensions_.z.index] < chunk.shape.z)
+    ) {
       throw new Error(
-        `Subarray has an unsupported data type, data=${data.constructor.name}`
+        `Received incompatible shape for chunkIndex ${JSON.stringify(chunk.chunkIndex)} at LOD ${chunk.lod}: ` +
+          `expected shape: ${JSON.stringify(chunk.shape)}, received shape: ${subarray.shape} (too small)`
+      );
+    } else {
+      Logger.debug(
```
phoenixAja: if you're sliding though z i wonder if this has potential to spam the logs?

phoenixAja: oh i see - when you scroll through z it doesn't load/ trim new chunks. It only does that on the initial imageLayer creation?

phoenixAja:
this works well for me in the react example!

packages/core/src/core/chunk_manager.ts
```diff
@@ -68,9 +68,21 @@ export class ChunkManagerSource {
 
         for (let x = 0; x < chunksX; ++x) {
           const xOffset = xLod.translation + x * xLod.chunkSize * xLod.scale;
+          const actualChunkWidth = Math.min(
+            chunkWidth,
+            xLod.size - x * chunkWidth
+          );
```
shlomnissan: You can inline these calculations to reduce cognitive load

```ts
const rowStride = Math.min(chunkWidth, xLod.size - x * chunkWidth);

// ...

chunksAtT.push({
  state: "unloaded",
  lod,
  visible: false,
  prefetch: false,
  priority: null,
  orderKey: null,
  shape: {
    x: Math.min(chunkWidth, xLod.size - x * chunkWidth),
    y: Math.min(chunkWidth, xLod.size - y * chunkWidth),
    z: Math.min(chunkDepth, (zLod?.size ?? 1) - z * chunkDepth),
    c: channels,
  },
  rowStride,
  rowAlignmentBytes: 1,
  chunkIndex: { x, y, z, t },
  // ...
}
```

packages/core/src/data/ome_zarr/image_loader.ts
```diff
@@ -98,23 +99,68 @@ export class OmeZarrImageLoader {
...
+      for (let z = 0; z < chunk.shape.z; z++) {
+        const zStart = z * zStride;
+        for (let y = 0; y < chunk.shape.y; y++) {
+          const srcStart = zStart + y * yStride;
+          const srcEnd = srcStart + chunk.shape.x;
+          wholeChunk.set(wholeChunk.subarray(srcStart, srcEnd), offset);
+          offset += chunk.shape.x;
+        }
+      }
+      chunk.data = wholeChunk.subarray(0, offset);
```
andy-sweet: Given that `chunk.data` is eventually assigned to a view of the original chunk (using subarray), doesn't that mean that this PR will not save any memory? That seems to be confirmed with the stats.js memory usage (which hovers ~700MB for me before and after this PR).

If so, the simplest change to the existing PR is to use `slice` instead here. However, I found the logic above slightly confusing, and I wonder if just preallocating the actual compact chunk data above (and filling that in) is worth it instead given that `slice` will do that anyway.

packages/core/src/data/ome_zarr/image_loader.ts
```diff
@@ -98,23 +99,68 @@ export class OmeZarrImageLoader {
...
+    const wholeChunk = subarray.data;
+
+    if (!isChunkData(wholeChunk)) {
+      throw new Error(
+        `Subarray has an unsupported data type, data=${wholeChunk.constructor.name}`
+      );
+    }
+
+    if (
+      subarray.shape[this.dimensions_.x.index] === chunk.shape.x &&
```
andy-sweet: Nit: these if conditions are hard to read. I think pulling out some definitions like `subarrayShapeX`, etc. would help.

packages/core/src/data/ome_zarr/image_loader.ts
```diff
@@ -98,23 +99,68 @@ export class OmeZarrImageLoader {
       signal,
     });
 
-    const data = subarray.data;
-    if (!isChunkData(data)) {
+    this.assignChunkData(chunk, subarray);
+  }
+
+  private assignChunkData(chunk: Chunk, subarray: zarr.Chunk<zarr.DataType>) {
```
shlomnissan: I’d suggest renaming this to `assignChunkDataIfValid`, or pulling the shape-validation/error case out of the function. Alternatively, you could move the validation into a separate helper. In any case, handling the error outside would simplify this function and avoid bundling it into an else-if chain.

packages/core/src/data/ome_zarr/image_loader.ts
shlomnissan: I’d skip this assignment and  just use `subarray.data` directly. That makes it clear you’re working with the original buffer (no copy) and avoids any ambiguity about views vs. copies.

packages/core/src/data/ome_zarr/image_loader.ts
```diff
@@ -98,23 +99,68 @@ export class OmeZarrImageLoader {
...
+    const wholeChunk = subarray.data;
+
+    if (!isChunkData(wholeChunk)) {
+      throw new Error(
+        `Subarray has an unsupported data type, data=${wholeChunk.constructor.name}`
+      );
+    }
```
shlomnissan: I would pull this error case out as well.

packages/core/src/data/ome_zarr/image_loader.ts
shlomnissan: I think `subarray.stride[this.dimensions_.y.index]` is correct here. Using the chunk’s width would only work if the data is guaranteed to be tightly packed. But it's a bit confusing because we set `rowStride = actualWidth` when we generate chunks, and again in the else branch below.

packages/core/src/data/ome_zarr/image_loader.ts
aganders3: I guess I figured `wholeChunk` would be garbage collected and the buffer would reduce to that held by `chunk.data`, but I'm probably misunderstanding how views work!

I had a version that preallocated a new compact array, but was trying to avoid the extra allocation. That's probably not a big deal so better to do for readability alone. Allocating the new array dynamically to match the type was also a little ugly, but maybe I can find a cleaner alternative.

packages/core/src/data/ome_zarr/image_loader.ts
aganders3: Not just on creation, but any time chunks are (re)loaded. So sliding through Z it would happen when you cross a Z chunk boundary at the current and/or fallback LOD. I'm happy to remove this log though.

packages/core/src/data/ome_zarr/image_loader.ts
aganders3: I don't think so (edit: responding to Phoenix). `rowStride` might be a bit confusing of a name, but it's basically the same as the row width, which is the y-stride. The x-stride is typically 1 (fastest moving dimension).

packages/core/src/data/ome_zarr/image_loader.ts
aganders3: Yeah I was just reusing the name but I think it should be changed. I think `subarray` has been in there for a while and made sense before a lot of changes.

packages/core/src/data/ome_zarr/image_loader.ts
aganders3: Nice - viewing the whole zebrahub FOV at LOD1 that brings it down to ~500MB for me. Thanks!

packages/core/src/data/ome_zarr/image_loader.ts
```diff
@@ -98,23 +99,68 @@ export class OmeZarrImageLoader {
...
+        const zStart = z * zStride;
+        for (let y = 0; y < chunk.shape.y; y++) {
+          const srcStart = zStart + y * yStride;
+          const srcEnd = srcStart + chunk.shape.x;
+          wholeChunk.set(wholeChunk.subarray(srcStart, srcEnd), offset);
+          offset += chunk.shape.x;
+        }
+      }
+      chunk.data = wholeChunk.subarray(0, offset);
+      chunk.rowStride = chunk.shape.x;
```
shlomnissan: Is this necessary? We already set `rowStride = actualChunkWidth`.

shlomnissan (commented):
Overall this works well. I left a couple of small nits, but the only blocking item is the memory allocation issue Andy raised. I’d also suggest making `assignChunkData` a bit clearer.

packages/core/src/data/ome_zarr/image_loader.ts
shlomnissan: If you move validation and error cases outside this function, you could end up with something like:


```ts

const data = subarray.data as ChunkData;

const xIdx = this.dimensions_.x.index;
const yIdx = this.dimensions_.y.index;
const zIdx = this.dimensions_.z?.index;

const chunkWidth = chunk.shape.x;
const chunkHeight = chunk.shape.y;
const chunkDepth = this.dimensions_.z ? chunk.shape.z : 1;

const shapeWidth = subarray.shape[xIdx];
const shapeHeight = subarray.shape[yIdx];
const shapeDepth = this.dimensions_.z ? subarray.shape[zIdx!] : 1;

// handle base case (exact shape)
if (
  shapeWidth === chunkWidth &&
  shapeHeight === chunkHeight &&
  !this.dimensions_.z || shapeDepth === chunkDepth
) {
  chunk.data = data;
  chunk.rowStride = subarray.stride[yIdx];
  return;
}

// handle boreder chunks here
```

packages/core/src/data/ome_zarr/image_loader.ts
aganders3: I pulled this out and agree it's better, but it does then require a type assertion in this function.

packages/core/src/data/ome_zarr/image_loader.ts
aganders3: I moved the size checking and type validation out and refactored this to be `trimChunkPadding` which I think is clearer about the intent.

packages/core/src/core/chunk_manager.ts
```diff
@@ -105,12 +84,12 @@ export class ChunkManagerSource {
                 priority: null,
                 orderKey: null,
                 shape: {
-                  x: chunkWidth,
-                  y: chunkHeight,
-                  z: chunkDepth,
+                  x: Math.min(chunkWidth, xLod.size - x * chunkWidth),
```
andy-sweet, line 87: Nit: this has an identical definition to rowStride, so it's inconsistent to define that in the outer loop vs. defining this here. I preferred just having `actualChunkWidth` defined in the outer loop and then showing `rowStride: actualChunkWidth` because that is an indication of our expectation that the chunk data is compact. This is optional because I think you had that originally, so don't want you to chase your own tail here - it's not a big deal.

packages/core/src/data/ome_zarr/image_loader.ts
```diff
@@ -94,27 +95,81 @@ export class OmeZarrImageLoader {
...
     chunk.rowAlignmentBytes = rowAlignment;
-    chunk.rowStride = subarray.stride[this.dimensions_.y.index];
-    chunk.data = data;
+  }
+
+  private trimChunkPadding(
+    chunk: Chunk,
+    receivedChunk: zarr.Chunk<zarr.DataType>
+  ): ChunkData {
+    // type assertion is safe because we check isChunkData(receivedChunk.data) in the caller
```
andy-sweet: Optional: you only use `receivedChunk.data` and `receivedChunk.stride`, so instead of having this comment and the static cast (which should almost always be avoided), why not just pass `receivedChunkData: ChunkData` and `receivedChunkStride: number[]` as the arguments to this function?

packages/core/src/data/ome_zarr/image_loader.ts
```diff
@@ -94,27 +95,81 @@ export class OmeZarrImageLoader {
...
+    const receivedChunkData = receivedChunk.data as ChunkData;
+    const compactSize = chunk.shape.x * chunk.shape.y * chunk.shape.z;
+    const compactData = receivedChunkData.slice(0, compactSize);
```
andy-sweet: I guess you're doing this because there's no compact way of creating a new TypedArray of the same type as receivedChunkData, right? Ideally, I'd prefer this to be something closer to `new ...` to indicate all this data will be explicitly set, which also avoids some reads/writes here too.

andy-sweet (approved):
Looks good and works well for me. All my comments are optional.

packages/core/src/core/chunk_manager.ts
aganders3: I'm ambivalent about defining inline vs. the `actualChunkWidth` variable, but part of that is just not being satisfied with the naming. I'm curious how this might change if we move to "fixed-size logical chunks" as well.

Regarding `rowStride` - with how we're currently handling data I think we could likely remove this and just use the chunk shape in place of strides, assuming the data are densely packed and ordered as we expect.

All this to say I'm going to leave it as-is for now. 😬

packages/core/src/data/ome_zarr/image_loader.ts
aganders3: Good point - I also didn't like this but found it better than the TypeScript gymnastics I came up with initially. With some more searching I think this is better:

```
const compactData = new (receivedChunkData.constructor as ChunkDataConstructor)(compactSize);
```

packages/core/src/core/chunk_manager.ts
andy-sweet: > Regarding rowStride - with how we're currently handling data I think we could likely remove this and just use the chunk shape in place of strides, 

I'm assuming zarrita chunks have a stride for some reason (e.g. some readers return non-compact data, or differently ordered data, or could do in the future). Agreed that we're handling this inconsistently across Idetik right now. I think validating then using compact-ness internally is fine, but the current state of assuming is not.

PR #471, opened 2025-09-30
Prepare for npm prerelease version 4

andy-sweet:
### Summary
This change was created by running `npm version major` on both the core and react packages. I decided to bump the react version twice, so that the pre-release version numbers for both packages are in sync for now.

I also removed most of the content from the READMEs since some of it is already inaccurate, and I'd rather not encourage people to use Idetik if they stumble across it!

### Tests & Checks
I tested this with the organelle-box portal, by building locally on this branch, updating the yalc packages locally, and running a local dev build of organelle-box.


andy-sweet:
@aganders3 : assigning you because you reviewed the recent v3 prerelease that was a slightly sad failure. I also thought you would appreciate the anti-zerover these packages are progressing in ;)

This version accepts the larger bundle size to keep the pre-release history simple (i.e. commits on `main` for now), in the hope that the organelle-box portal won't actually download the worker chunks if it doesn't need them. I know almost nothing about next, so am unsure how to confirm that locally. If the larger bundle size is unacceptable and we can't prioritize a quick fix with another related release, then I'll branch from main to create yet another release (maybe a 2.1 or 3.1?).

This time, I actually tested this build with the organelle-box (at least via yalc).

aganders3:
> I also removed most of the content from the READMEs since some of it is already inaccurate, and I'd rather not encourage people to use Idetik if they stumble across it!

I'm good with this, but we should have (up-to-date!) info somewhere at least for ourselves!

aganders3 (approved):
Looks good, thanks!

I don't have an up-to-date dev env for the OrganelleBox portal locally, so didn't test myself. For what it's worth I'm still looking at options for reducing the bundle size with the WebWorker in place, but haven't found a fully working solution yet.

PR #473, opened 2025-10-03
Stop observing events when runtime stops

andy-sweet:
### Summary
This stops observing events and size changes when `Idetik.stop` is called. It also refactors the size observation logic into a utility class to improve the readability of the Idetik runtime. I also adding some added some extra logs to help catch double starts or unnecessary stops.

### Tests & Checks
I manually tested the changes with the core examples and the react components.


phoenixAja (approved):
i don't see any blockers

aganders3 (approved):
LGTM too
I yalc'd this and tested with [my branch in VCP](https://github.com/chanzuckerberg/virtual-cells-platform/compare/aganders3/update-idetik) and the HITL works as I'd expect locally. Will be good to get more testing there from @phoenixAja who is more familiar with the app.

PR #474, opened 2025-10-03
Add event dispatcher connect/disconnect in runtime start/stop

aganders3:
### Summary
I attempted to update Idetik in the VCP project in preparation of the CryoLens release. Unfortunately this broke the HITL app, which is now tied to the same Idetik dependency version. The previous version did not include #394, and the HITL app behavior changed without long-lived Idetik runtime(s).

I was able to work around some of the issues there, but duplicate event observers are still interfering. I was able to test a local build with these changes and it fixes some of the issues.

### DIFF
--- a/packages/core/src/core/event_dispatcher.ts
+++ b/packages/core/src/core/event_dispatcher.ts
@@ -52,7 +52,7 @@ export class EventDispatcher {
 
   public connect() {
     if (this.isConnected_) {
-      Logger.debug(
+      Logger.warn(
         "EventDispatcher",
         "Attempted to connect already connected event dispatcher",
         `element id: ${this.element_.id}`
--- a/packages/core/src/core/event_dispatcher.ts
+++ b/packages/core/src/core/event_dispatcher.ts
@@ -39,15 +39,47 @@ type Listener = (event: EventContext) => void;
 
 export class EventDispatcher {
   private readonly listeners_: Listener[] = [];
+  private readonly element_: HTMLElement;
+  private isConnected_ = false;
 
   constructor(element: HTMLElement) {
-    eventTypes.forEach((type) => {
-      element.addEventListener(type, this.handleEvent, { passive: false });
-    });
+    this.element_ = element;
   }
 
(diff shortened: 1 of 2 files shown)


packages/core/src/core/event_dispatcher.ts
```diff
@@ -39,17 +39,49 @@ type Listener = (event: EventContext) => void;
...
+    this.element_ = element;
   }
 
   public addEventListener(listener: Listener) {
     this.listeners_.push(listener);
   }
 
+  public connect() {
+    if (this.isConnected_) {
+      Logger.debug(
```
andy-sweet: I'd be tempted to bump this up to a warning, so that we see it in things that depend on an npm build of this library, since i think this is a misuse. The other case is less dangerous, so is fine as a debug.

andy-sweet (approved):
Looks good!

packages/core/src/core/event_dispatcher.ts
```diff
@@ -39,17 +39,49 @@ type Listener = (event: EventContext) => void;
...
+        "EventDispatcher",
+        "Attempted to disconnect already disconnected event dispatcher",
+        `element id: ${this.element_.id}`
+      );
+      return;
+    }
+    this.isConnected_ = false;
+    eventTypes.forEach((type) => {
+      this.element_.removeEventListener(type, this.handleEvent);
```
andy-sweet, line 81: MDN suggests that the value of `options` in remove does need to match `add`, but also that browsers are inconsistent about how they handle this: https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener#matching_event_listeners_for_removal

I don't think there's any harm in adding `{passive: false}` here, or maybe creating a constant `options` somewhere to ensure this is always consistent.

aganders3: Actually I get a TypeScript error if I do that. I guess this makes sense, because those docs also say:
> the only option `removeEventListener()` checks is the `capture`/`useCapture` flag

PR #477, opened 2025-10-06
Document current npm pre-release process 

andy-sweet:
### Summary
This change updates the main README to include basic instructions for our current manual release process for the pre-release packages on NPM.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,3 +30,27 @@ A layer-based library for interactive visualization of large datasets.
    `npm run test -- --run`.
 
 4. See [package.json](package.json) for other commands available.
+
+## Release
+
+Currently, we maintain Idetik prerelease packages on npm: [@idetik/core-prerelease](https://www.npmjs.com/package/@idetik/core-prerelease?activeTab=readme) and [@idetik/react-prerelease](https://www.npmjs.com/package/@idetik/react-prerelease).
+
+Our release process is currently manual and we currently bump the major version every time to easily avoid unintentional downstream breakage.
+
+To release the core package run the following the commands
+
+```shell
+git switch -c your-name/prerelease-X-Y-Z
+cd packages/core
+npm version major
+npm build
+npm login
+npm publish
+```
+
+where `X`, `Y`, and `Z` are the respective major, minor, and patch numbers of the release number.
+
+
+Then create a PR, merge it, and add a tag of the form `prerelease-X.Y.Z` to the commit on `main`.
+
+Repeat the process for `packages/react`.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -30,3 +30,55 @@ A layer-based library for interactive visualization of large datasets.
...
+
+Our release process is currently manual and we currently bump the major version every time to easily avoid unintentional downstream breakage.
+
+To release the core package run the following the commands
+
+```shell
+git switch -c your-name/prerelease-X-Y-Z
+cd packages/core
+npm version major
+npm build
```
aganders3:
```suggestion
npm run build
```

README.md
```diff
@@ -30,3 +30,55 @@ A layer-based library for interactive visualization of large datasets.
...
+```
+
+where `X`, `Y`, and `Z` are the respective major, minor, and patch numbers of the release number.
+
+Then repeat for `packages/react`.
+
+```shell
+cd ../react
+npm version major
+npm build
```
aganders3:
```suggestion
npm run build
```

README.md
```diff
@@ -30,3 +30,55 @@ A layer-based library for interactive visualization of large datasets.
...
+After both package versions have been bumped, create a PR, get it approved, and merge it to `main`.
+
+Next, checkout the corresponding commit on `main` and run the following commands for 
+
+```shell
+npm login
+cd ../core
+npm publish
+cd ../react
+npm publish
```
aganders3: Do the packages need to be rebuilt with `main` checked out? There is a script in each `package.json` so this can be done with `npm run pub`:
```
"pub": "npm run build; npm publish --access=public"
```

aganders3 (approved):
Looks good with some minor comments. This was super helpful for me right now as I am trying to make a release!

aganders3:
One more thought: perhaps we should come back and comment on the PR that it's been released, with a link to npm? I'm sure this can be automated one day.

README.md
andy-sweet: > Do the packages need to be rebuilt with main checked out?

Probably not. But out of sanity in general, I like to build against the exact commit that's tagged/associated with the commit (in case that git commit is used somehow).

`npm run pub` sounds like a good idea though

andy-sweet: Oh I see, you want to run `npm run pub` before merging. I'd like to avoid that right now in case someone makes a mistake that needs to be corrected before the PR is approved/merged. But I still like the protection of `npm run pub`, so I'll use that instead.

andy-sweet:
> One more thought: perhaps we should come back and comment on the PR that it's been released, with a link to npm? I'm sure this can be automated one day.

I prefer using the commit on `main` as being tightly associated with the release. Currently this is through a tag. Agreed this can be automated later.

README.md
aganders3: No I think we're on the same page - I meant to run `npm run pub` after merging and checking out `main`. I agree we want the exact commit.

PR #478, opened 2025-10-06
Prepare for npm prerelease version 5

aganders3:
### Summary
This was prepared according to the guide added in #477 

### Tests & Checks
This release was tested locally in VCP (CryoLens and HITL) using `yalc`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,5 +1,5 @@
 {
-  "name": "idetik",
+  "name": "idetik-release",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
@@ -8201,7 +8201,7 @@
     },
     "packages/core": {
       "name": "@idetik/core-prerelease",
-      "version": "4.0.0",
+      "version": "5.0.0",
       "license": "MIT",
       "dependencies": {
         "gl-matrix": "^3.4.3",
@@ -8215,7 +8215,7 @@
     },
     "packages/react": {
       "name": "@idetik/react-prerelease",
-      "version": "4.0.0",
+      "version": "5.0.0",
       "license": "MIT",
       "dependencies": {
         "@base-ui-components/react": "^1.0.0-beta.2",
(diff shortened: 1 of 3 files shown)


package-lock.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "idetik",
+  "name": "idetik-release",
```
andy-sweet: I still don't know where this name comes from, but I also don't care too much!

When I run `npm install` from the top, this changes back to `idetik`.

package-lock.json
```diff
@@ -8245,6 +8245,18 @@
         "react-dom": "^18.3.1"
       }
     },
+    "packages/react/node_modules/@idetik/core-prerelease": {
```
andy-sweet: I don't think this should be added as part of this PR, but I did add it by mistake a number of times. It's probably related to the ordering of commands - sorry if I messed that up in #477!

andy-sweet: I think an `npm install` is needed somewhere (at the top level, in core, in react), but strangely that doesn't fix things here.

package-lock.json
andy-sweet: I also forgot this. We need to update the required version of `core-prerelease` for `react-prerelease` as part of the release process. That also explains the below: https://github.com/chanzuckerberg/idetik/pull/478/files#r2407439872

andy-sweet: I'm not sure what command would handle this automatically (since core is not yet on npm), so I just manually edited these.

package-lock.json
aganders3: Oh sorry, it comes from the directory name and I made a new worktree for this. I'll revert that.

package-lock.json
andy-sweet, line 8243:
```suggestion
        "@idetik/core-prerelease": "^5.0.0",
```

package-lock.json
aganders3: Maybe this is from running `npm install` after updating the version in `core` but before updating the dependency pin in `react`?

packages/react/package.json
andy-sweet, line 64: This is the peerDependency, not the devDependency

```suggestion
    "@idetik/core-prerelease": "^5.0.0",
```

aganders3: omg sorry I really thought I got both

PR #489, opened 2025-10-08
Relax time translation and scale constraint in chunk manager

andy-sweet:
### Summary
This change relaxes the constraint of a zero translation and unitary scale in the time dimension in the chunk manager. It also adds an example from the public OME-Zarr dataset that has a non-unitary scale.

While relaxing this constraint is generally useful for future work, I prioritized this because many of the DCA datasets have a time scale.

### Tests & Checks
I manually tested the new example.
I manually tested the existing examples that use chunk streaming.


phoenixAja (approved):
works for me locally, but changing the t coord is fairly laggy, but i don't think that's from the changes you made?

andy-sweet:
> works for me locally, but changing the t coord is fairly laggy, but i don't think that's from the changes you made?

Do you mean the existing examples are laggier than on main? That would be surprising given there's very little extra compute going on here.

Or that the new example is laggy (e.g. compared to existing examples)? I think the new example's data is hosted in the UK, so the latency from the West coast might be pretty significant. I can achieve a modest playback of about 2Hz depending on my field of view (e.g. if all the high res chunks are loaded, things get slow).

phoenixAja:
> > works for me locally, but changing the t coord is fairly laggy, but i don't think that's from the changes you made?
> 
> Do you mean the existing examples are laggier than on main? That would be surprising given there's very little extra compute going on here.
> 
> Or that the new example is laggy (e.g. compared to existing examples)? I think the new example's data is hosted in the UK, so the latency from the West coast might be pretty significant. I can achieve a modest playback of about 2Hz depending on my field of view (e.g. if all the high res chunks are loaded, things get slow).

just this new example, sorry should have clarified. The latency makes total sense though if it's hosted in the UK

PR #490, opened 2025-10-08
Use clamp_to_edge instead of repeat as default texture wrap mode

andy-sweet:
### Summary
This uses "clamp_to_edge" instead of "repeat" as the default wrap mode for textures, which prevents tile rendering artifacts seen in the OME-Zarr v0.5 example.

This change also allows us to explicitly enable anti-aliasing, which was previously explicitly disabled in #302 to fix a similar tile rendering artifact seen in other/older examples.

### Tests & Checks
I manually tested all the core examples.
I manually tested all the React examples.

### DIFF
--- a/packages/core/src/objects/textures/texture.ts
+++ b/packages/core/src/objects/textures/texture.ts
@@ -83,9 +83,9 @@ export abstract class Texture extends Node {
   public mipmapLevels = 1;
   public unpackAlignment: TextureUnpackRowAlignment = 4;
   public unpackRowLength = 0;
-  public wrapR: TextureWrapMode = "repeat";
-  public wrapS: TextureWrapMode = "repeat";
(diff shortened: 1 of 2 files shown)


andy-sweet:
@aganders3 : I assigned you because you thought of this and mentioned you might create a PR. As that hasn't happened yet, I thought I'd get this in.

I considered keeping the texture repeat (since that is not uncommon for rendering in other settings), and explicitly setting "clamp_to_edge" for the data textures. But decided it's probably not worth the extra complexity right now.

aganders3 (approved):
Nice - thanks for getting to what I didn't!

I'm actually not sure which (R, S, or T) is the through-plane, which I think is used for channels in the 2D texture array, but we may not need to change the mode for that. It makes sense to keep it consistent though and I think for our purposes this is the right default for now.

PR #492, opened 2025-10-10
Support chunk streaming of multiple channels

### LINKED ISSUE #372: ChunkedImageLayer does not support multi-channel images

### LINKED ISSUE #406: Avoid second slice in point value picking with chunked image layer
With this change, we create a copy of the sliced data:
https://github.com/chanzuckerberg/idetik/pull/404#discussion_r2301645176

Is slicing twice necessary? Can we use the underlying texture data via the renderable object?

This is a non-blocking comment because it's not directly related to the change, but if there's a way to improve this, let's add a comment or create an issue before moving on.

_Originally posted by @shlomnissan in https://github.com/chanzuckerberg/idetik/pull/403#discussion_r2301702188_

andy-sweet:
### Summary
This adds support for multiple channels in the chunked image layer.

Support is limited to streaming one or all channels. Presentation waits until all channels are ready for a particular LOD. A relatively simple future change should be able to add support for a subset of channels, but this change is already too big.

The approach relies on introducing a new helper type `SlicedChunk2D`, which

- logically groups the source's chunks into an array of channels;
- stores a data buffer for a slice in Z over all channels;
- and provides a method to update the z-slice.

While this introduces more types and some amount of redirection, I think it simplifies and encapsulates some of the associated logic that was previously spread around `ChunkedImageLayer`.

The other key change is that `ChunkedImageLayer` keeps track of partially loaded images (in `loadedChunks_`) and fully loaded images (in `visibleImages_`). This effectively waits for all channels to be loaded for a particular tile until the image is created and presented.

### Related Issue
Closes #372
Closes #406 

### Tests & Checks
I manually tested the React examples.


andy-sweet:
This is a draft and not ready for review yet. I pushed it up to motivate a smaller part I split from it in #493

packages/core/examples/image2d_from_omezarr4d_hcs/main.ts
aganders3, line 1: Good idea to update this existing example 👍

andy-sweet:
This is now ready for review, but it's also pretty large, mostly due to the introduction of a new helper class (`SlicedChunk2D`). I think this might be best reviewed as a group to assess the general approach, and if/when there's consensus that this is good enough, then we can split it in two (one for `SlicedChunk2D` with tests, and one for the integration).

andy-sweet:
I marked this as a draft to indicate that I'm not actively looking for reviews. Of course, any comments or ideas are still welcome.

andy-sweet:
Closing this for now because it is not intended to be actively reviewed, and because I'd like to propose a better long term approach (one channel per layer, blending layers).

PR #493, opened 2025-10-10
Basic support for channel dimension in chunk manager source

andy-sweet:
### Summary
This adds some support for the channel dimension in the chunk manager source. The support requires that the there must be one channel per chunk, as well as some other sensible constraints for the channel dimension that are validated on construction.

This approach requires minimal changes within the chunk manager source, and no changes in any dependencies of it. The main downside of the approach is that incurs a performance penalty in `updateChunksAtTimeIndex` that is linear in the number of channels. This could be avoided by making changes to certain data structures (e.g. storing the channel index in a new dimension in `ChunkManagerSource.chunks_`, changing the `Chunk` type), but these would likely have a relatively larger impact on dependencies.

The motivation for this change is twofold.

1.  Split some of the larger work around support multi-channel chunk streaming: https://github.com/chanzuckerberg/idetik/pull/492
2. Unblock work attempting to split the chunk manager source as part of larger redesign/refactor work.

### Tests & Checks
I manually tested the core examples that use chunk streaming.


andy-sweet:
@aganders3 : I assigned you because you mentioned this may unblock some of your work on splitting `ChunkManagerSource`. Other comments/reviews are welcome though!

packages/core/src/core/chunk_manager_source.ts
```diff
@@ -96,33 +95,35 @@ export class ChunkManagerSource {
...
+                  lod,
+                  visible: false,
+                  prefetch: false,
+                  priority: null,
+                  orderKey: null,
+                  shape: {
+                    x: Math.min(chunkWidth, xLod.size - x * chunkWidth),
+                    y: Math.min(chunkHeight, yLod.size - y * chunkHeight),
+                    z: Math.min(chunkDepth, (zLod?.size ?? 1) - z * chunkDepth),
+                    c: 1,
```
aganders3, line 110: I think (but am not certain) we should just remove this and maintain that internal chunks are single channel. I'm pretty sure this would also require adding a top-level `channel` property on `Chunk`, though.

This removes the constraint of channel chunk size == 1. The downside is that we'd need to extract the channel in the `image_loader`, which right now would result in redundant work (decompression) to get multiple channels from the same chunk (but it's off the main thread at least).

packages/core/src/core/chunk_manager_source.ts
aganders3, line 37: I think it's okay to leave for now, but `Chunk[t][c][spatial]` would eventually allow for some efficiency improvements.

packages/core/src/core/chunk_manager_source.ts
```diff
@@ -203,7 +204,7 @@ export class ChunkManagerSource {
   }
 
   public loadChunkData(chunk: Chunk, signal: AbortSignal) {
-    return this.loader_.loadChunkData(chunk, this.sliceCoords_, signal);
+    return this.loader_.loadChunkData(chunk, signal);
```
aganders3, line 207: I love this change!

packages/core/src/core/chunk_manager_source.ts
```diff
@@ -378,6 +389,28 @@ export class ChunkManagerSource {
     return prefetchedChunks;
   }
 
+  private isChunkChannelInSlice(chunk: Chunk): boolean {
+    return (
+      this.sliceCoords_.c === undefined ||
+      this.sliceCoords_.c === chunk.chunkIndex.c
+    );
+  }
```
aganders3, line 397: I'm trying to avoid additional usage of `this.sliceCoords_` since I think it should be removed, but this can be updated when that happens.

aganders3 (approved):
Thanks! This works well for me.

> Unblock work attempting to split the chunk manager source as part of larger redesign/refactor work.

I think more specifically it unblocks removing `sliceCoords_` as state on `ChunkManagerSource`, which could (maybe should?) be done before splitting that class.

packages/core/src/core/chunk_manager_source.ts
aganders3: I haven't looked at https://github.com/chanzuckerberg/idetik/pull/492 yet to see how it might work in that context.

packages/core/src/core/chunk_manager_source.ts
aganders3: Ah, you called this out in the PR description. Non-blocking but I'd be okay with doing it here too.

packages/core/src/core/chunk_manager_source.ts
andy-sweet: I'm assuming it or some version of it will either be passed through or stored on some view of the chunk store though?

Also, I'm not totally convinced that `c` belongs on `SliceCoordinates` because it's a little different to `z` and `t`. It's always a discrete value (or maybe multiple discrete values in the future), and it's unlikely that it should change often (unless we wanted to couple it with the visibility of channels).

packages/core/src/core/chunk_manager_source.ts
andy-sweet: Agreed. It may add some complexity and performance penalties elsewhere though.

packages/core/src/core/chunk_manager_source.ts
andy-sweet: > I think (but am not certain) we should just remove this and maintain that internal chunks are single channel.

Do you mean removing `Chunk.shape.c` as a property in this PR? I think that's viable once we've removed the old loading path, but I'd rather just leave that path alone for now.

aganders3: Oh of course, I keep forgetting the `Chunk` type is shared in both loading paths.

Issue #487, opened 2025-10-07
Configurable Policy


shlomnissan:
Idetik is a library for building high-performance, web-based image viewers. Because datasets vary widely in size, dimensionality, and access patterns, the ability to tune viewer behavior is a major value add. This first iteration introduces a configurable “policy” system to replace existing hard-coded values in the chunk manager and queue.

### Goal

Deliver a flexible configuration foundation that replaces hard-coded constants, allows users to tune viewer behavior, and prepares Idetik for adaptive policies (e.g., memory- or network-aware scheduling) without adding complexity in v1.

### Configuration Objects

Policies are now divided into two configuration objects to match Idetik’s architecture:
- RuntimePolicyConfig: global runtime settings shared by all sources (network, memory).
- SourcePolicyConfig: per-image-source configuration (prefetching, priorities, and LOD behavior).

This separation ensures that each data source can tune its own behavior while the runtime maintains consistent global resource constraints.

#### RuntimePolicyConfig

```ts
export type RuntimePolicyConfig = {
  memory: { cpu: number };
  network: {
    maxConcurrent: number;
    retryCount?: number;
    timeoutMs?: number 
  };
  profile?: string;
};
```

#### SourcePolicyConfig

```ts
type PriorityCategory =
  | "fallbackVisible"
  | "prefetchTimeHigh"
  | "visibleCurrent"
  | "prefetchTimeLow"
  | "fallbackBackground"
  | "prefetchSpace";

export type SourcePolicyConfig = {
  profile?: string; // optional display/debug name
  prefetch: { x: number; y: number; z: number; t: number };
  priorityOrder: PriorityCategory[];
  // LOD configuration (new)
  lod: {
    min?: number;
    max?: number;
    bias?: number;  // positive favors coarser (lower-res) LODs
  };
};
```

This configuration defines prefetch behavior, memory budget, network settings, and reorderable priority categories. It will replace constants such as `MAX_CONCURRENT`, `PREFETCH_PADDING_CHUNKS`, `PREFETCH_TIME_POINTS`, and all `PRI_*` values.

### Memory Limit

A new soft memory limit (`memory.cpu`) will be introduced:
- The system will track total CPU memory used by loaded chunks.
- If a new allocation would exceed the defined limit, Idetik will log an informational warning rather than enforcing eviction.
- This allows users to refine their configuration (e.g., lower time prefetch) without breaking rendering.

### Profiles

[...]

PR #494, opened 2025-10-10
Add image source loading policies

shlomnissan:
### Summary

This PR introduces an image source policy object and related helpers used to
configure loading behavior for a given `ChunkManagerSource`. It replaces the
previous hard-coded constants with a structured policy definition. The only new
feature added in this change is support for specifying an LOD range, allowing
users to skip certain LOD levels for a given image.

### Related Issue
The next task for #487 is adding a config object to the runtime that handles memory and network properties.

### Tests & Checks

- All existing checks have passed.
- Added basic unit tests.

### DIFF
--- a/packages/core/examples/ca_wave_dynamics/main.ts
+++ b/packages/core/examples/ca_wave_dynamics/main.ts
@@ -5,6 +5,8 @@ import {
   OmeZarrImageSource,
   OrthographicCamera,
   Color,
+  createExplorationPolicy,
+  createPlaybackPolicy,
 } from "@";
 import { PanZoomControls } from "@/objects/cameras/controls";
 import { addDimensionSlider } from "../lil_gui_utils";
@@ -43,7 +45,12 @@ const channelProps: ChannelProps[] = [
 ];
 
 const camera = new OrthographicCamera(left, right, top, bottom);
-const imageLayer = new ChunkedImageLayer({ source, sliceCoords, channelProps });
+const imageLayer = new ChunkedImageLayer({
+  source,
+  sliceCoords,
+  policy: createExplorationPolicy(),
+  channelProps,
+});
 imageLayer.debugMode = true;
 
 const timePointDiv = document.querySelector<HTMLDivElement>("#time-point")!;
@@ -81,7 +88,8 @@ addDimensionSlider({
     onRateChange: (rateHz: number) => {
       const source = imageLayer.chunkManagerSource;
       if (source) {
-        source.prioritizePrefetchTime = rateHz > 0;
+        source.imageSourcePolicy =
+          rateHz > 0 ? createPlaybackPolicy() : createExplorationPolicy();
       }
     },
   },
--- a/packages/core/examples/ca_wave_dynamics/main.ts
+++ b/packages/core/examples/ca_wave_dynamics/main.ts
@@ -86,11 +86,8 @@ addDimensionSlider({
   stepValue: t.scale,
   playback: {
     onRateChange: (rateHz: number) => {
-      const source = imageLayer.chunkManagerSource;
-      if (source) {
-        source.imageSourcePolicy =
-          rateHz > 0 ? createPlaybackPolicy() : createExplorationPolicy();
-      }
+      imageLayer.imageSourcePolicy =
+        rateHz > 0 ? createPlaybackPolicy() : createExplorationPolicy();
     },
   },
 });
--- a/packages/core/src/core/chunk_manager_source.ts
+++ b/packages/core/src/core/chunk_manager_source.ts
@@ -13,21 +13,28 @@ import { almostEqual } from "../utilities/almost_equal";
 import { Logger } from "../utilities/logger";
 import { clamp } from "../utilities/clamp";
 
+/*
+Unique symbol used as a capability token to allow internal modules to update
+the image source policy. Only code that imports this symbol can call
+setImageSourcePolicy; all other callers will be rejected. Acts like a "friend"
+access key, preventing accidental external mutation.
+*/
+export const INTERNAL_POLICY_KEY = Symbol("INTERNAL_POLICY_KEY");
+
 export class ChunkManagerSource {
   private readonly chunks_: Chunk[][];
   private readonly loader_;
   private readonly lowestResLOD_: number;
   private readonly sliceCoords_: SliceCoordinates;
   private readonly dimensions_: SourceDimensionMap;
+  private readonly tIndicesWithQueuedChunks_: Set<number> = new Set();
+  private readonly sourceMaxSquareDistance2D_: number;
   private policy_: ImageSourcePolicy;
   private policyChanged_ = false;
   private currentLOD_: number = 0;
   private lastViewBounds2D_: Box2 | null = null;
   private lastZBounds_?: [number, number];
   private lastTCoord_?: number;
-
-  private tIndicesWithQueuedChunks_: Set<number> = new Set();
-  private sourceMaxSquareDistance2D_: number;
 
   constructor(
     loader: ChunkLoader,
@@ -194,14 +201,14 @@ export class ChunkManagerSource {
     return this.currentLOD_;
   }
 
-  public get imageSourcePolicy(): Readonly<ImageSourcePolicy> {
-    return this.policy_;
-  }
+  public setImageSourcePolicy(newPolicy: ImageSourcePolicy, key: symbol) {
+    if (key !== INTERNAL_POLICY_KEY) {
+      throw new Error("Unauthorized policy mutation");
+    }
 
-  public set imageSourcePolicy(policy: ImageSourcePolicy) {
-    if (this.policy_ !== policy) {
+    if (this.policy_ !== newPolicy) {
+      this.policy_ = newPolicy;
       this.policyChanged_ = true;
-      this.policy_ = policy;
 
       Logger.info(
         "ChunkManagerSource",
@@ -235,13 +242,8 @@ export class ChunkManagerSource {
       Math.min(this.lowestResLOD_, this.policy_.lod.max)
     );
 
-    const target = Math.max(minPolicyLOD, Math.min(maxPolicyLOD, desiredLOD));
-
+    const target = clamp(desiredLOD, minPolicyLOD, maxPolicyLOD);
     if (target !== this.currentLOD_) {
-      Logger.debug(
-        "ChunkManagerSource",
-        `LOD changed from ${this.currentLOD_} to ${target}`
-      );
       this.currentLOD_ = target;
     }
   }
@@ -428,10 +430,6 @@ export class ChunkManagerSource {
     chunk.priority = null;
     chunk.orderKey = null;
     chunk.prefetch = false;
-    Logger.debug(
-      "ChunkManagerSource",
-      `Disposing chunk ${JSON.stringify(chunk.chunkIndex)} in LOD ${chunk.lod}`
-    );
   }
 
   private computePriority(
--- a/packages/core/src/core/chunk_manager_source.ts
+++ b/packages/core/src/core/chunk_manager_source.ts
@@ -196,13 +196,15 @@ export class ChunkManagerSource {
   }
 
   public set imageSourcePolicy(policy: ImageSourcePolicy) {
-    this.policy_ = policy;
+    if (this.policy_ != policy) {
+      this.policy_ = policy;
 
-    Logger.info(
-      "ChunkManagerSource",
-      "Using image source policy:",
-      this.policy_.profile
-    );
+      Logger.info(
+        "ChunkManagerSource",
+        "Using image source policy:",
+        this.policy_.profile
+      );
+    }
   }
 
   public loadChunkData(chunk: Chunk, signal: AbortSignal) {
--- a/packages/core/src/core/chunk_manager_source.ts
+++ b/packages/core/src/core/chunk_manager_source.ts
@@ -5,6 +5,7 @@ import {
   SliceCoordinates,
   coordToIndex,
 } from "../data/chunk";
+import { ImageSourcePolicy } from "./image_source_policy";
 import { ReadonlyVec2, vec2, vec3 } from "gl-matrix";
 import { Box2 } from "../math/box2";
 import { Box3 } from "../math/box3";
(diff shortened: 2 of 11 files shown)


packages/core/examples/ca_wave_dynamics/main.ts
```diff
@@ -81,7 +88,8 @@ addDimensionSlider({
     onRateChange: (rateHz: number) => {
       const source = imageLayer.chunkManagerSource;
       if (source) {
-        source.prioritizePrefetchTime = rateHz > 0;
+        source.imageSourcePolicy =
+          rateHz > 0 ? createPlaybackPolicy() : createExplorationPolicy();
```
shlomnissan: I'm not sure we need to replace the policy here. Using `createPlaybackPolicy()` makes more sense to me but I wanted to make it clear that we can update policies after source creation.

I have some questions about exposing the `source`. It might be better to add a mutator for `imageSourcePolicy` in the layer and have it delegate to the image source.

packages/core/src/core/chunk_manager_source.ts
```diff
@@ -5,65 +5,53 @@ import {
...
+      this.policy_.profile
+    );
+
     this.validateXYScaleRatios();
     const { size: chunksT } = this.getAndValidateTimeDimension();
     const { size: chunksC } = this.getAndValidateChannelDimension();
 
-    const xLod0 = this.dimensions.x.lods[0];
-    const yLod0 = this.dimensions.y.lods[0];
+    const xLod0 = this.dimensions_.x.lods[0];
```
shlomnissan, line 61: I think this was an existing bug?

andy-sweet: There is a `dimensions` property getter that aliases `dimensions_`, but I prefer this.

packages/core/src/core/chunk_manager_source.ts
```diff
@@ -203,33 +191,54 @@ export class ChunkManagerSource {
...
       0,
-      Math.min(maxLOD, Math.floor(sourceAdjustedLodFactor))
+      Math.min(this.lowestResLOD_, this.policy_.lod.min)
     );
+    const maxPolicyLOD = Math.max(
+      minPolicyLOD,
+      Math.min(this.lowestResLOD_, this.policy_.lod.max)
+    );
+
+    const target = Math.max(minPolicyLOD, Math.min(maxPolicyLOD, desiredLOD));
```
andy-sweet: Maybe, I'm misreading exactly what's going on here, but I would expect something as simple as the following.

```suggestion
    // Intersect dataset bounds with policy bounds.
    const target = clamp(desiredLOD, this.policy_.lod.min, this.policy_.lod.max);
```

This assumes that both `minPolicyLOD` and `maxPolicyLOD` are in `[0, this.lowestResLOD_]`, but that should probably be validated somewhere if it already isn't.

packages/core/src/core/chunk_manager_source.ts
```diff
@@ -203,33 +191,54 @@ export class ChunkManagerSource {
     return this.currentLOD_;
   }
 
+  public get imageSourcePolicy(): Readonly<ImageSourcePolicy> {
+    return this.policy_;
+  }
+
+  public set imageSourcePolicy(policy: ImageSourcePolicy) {
+    if (this.policy_ != policy) {
+      this.policy_ = policy;
```
andy-sweet: It might be worth validating `policy.min` and `policy.max` to check they are in `[0, this.lowestResLOD_]`.

packages/core/src/core/image_source_policy.ts
```diff
@@ -0,0 +1,150 @@
...
+      "fallbackBackground",
+      "prefetchSpace",
+    ],
+  };
+  return createImageSourcePolicy(mergeConfig(base, overrides));
+}
+
+function validatePolicyConfig(config: ImageSourcePolicyConfig) {
+  for (const [k, v] of Object.entries(config.prefetch)) {
+    if (v === null) continue; // z/t may be omitted
```
andy-sweet:
```suggestion
    if (v === undefined) continue; // z/t may be omitted
```

packages/core/src/core/image_source_policy.ts
```diff
@@ -0,0 +1,150 @@
...
+    if (v < 0) {
+      throw new Error(`prefetch.${k} must be a non-negative number`);
+    }
+  }
+
+  const lod = config.lod;
+  if (lod?.min != null && lod?.max != null && lod.min > lod.max) {
```
andy-sweet:
```suggestion
  if (lod?.min != undefined && lod?.max != undefined && lod.min > lod.max) {
```

packages/core/src/core/image_source_policy.ts
```diff
@@ -0,0 +1,150 @@
...
+  };
+  return createImageSourcePolicy(mergeConfig(base, overrides));
+}
+
+export function createPlaybackPolicy(
+  overrides: Partial<ImageSourcePolicyConfig> = {}
+): ImageSourcePolicy {
+  const base: ImageSourcePolicyConfig = {
+    profile: "playback",
+    prefetch: { x: 0, y: 0, z: 0, t: 20 },
```
andy-sweet, line 106: Not necessarily against this change, but just want to confirm that you doubled the number of chunks prefetched over time compared to the constant on `main`.

packages/core/test/image_source_policy.test.ts
```diff
@@ -0,0 +1,106 @@
...
+  expect(p.prefetch).toEqual({ x: 0, y: 0, z: 0, t: 20 });
+  expect(Object.keys(p.priorityMap)).toHaveLength(5);
+  expect(p.priorityMap).toEqual({
+    fallbackVisible: 0,
+    prefetchTime: 1,
+    visibleCurrent: 2,
+    fallbackBackground: 3,
+    prefetchSpace: 4,
+  });
+});
```
andy-sweet, line 74: These tests mostly cover default values, which I tend not to value too much because they often just mirror a value in the implementation and need to be updated any time those change.

If you find value in these I think it's fine to keep them, but I think that tests using non-default values are probably more important (e.g. for the priority map) and should be added.

packages/core/src/core/image_source_policy.ts
```diff
@@ -0,0 +1,150 @@
...
+
+  return Object.freeze(resolved);
+}
+
+export function createExplorationPolicy(
+  overrides: Partial<ImageSourcePolicyConfig> = {}
+): ImageSourcePolicy {
+  const base: ImageSourcePolicyConfig = {
+    profile: "exploration",
+    prefetch: { x: 1, y: 1, z: 1, t: 0 },
```
andy-sweet, line 89: Just to confirm that the default values here represent a change from `main`. I think those changes are fine, but wanted to check those are intended.

packages/core/src/core/image_source_policy.ts
```diff
@@ -0,0 +1,150 @@
...
+    priorityOrder: [
+      "fallbackVisible",
+      "visibleCurrent",
+      "prefetchSpace",
```
andy-sweet, line 93: Again, a confirmation of the change from `main` - this bumps prefetchSpace ahead of prefetchTime and fallbackBackground.

packages/core/src/core/chunk_manager_source.ts
```diff
@@ -285,7 +294,7 @@ export class ChunkManagerSource {
     const disposedChunks: Chunk[] = [];
     for (const t of this.tIndicesWithQueuedChunks_) {
       const delta = t - currentTimeIndex;
-      if (delta >= 0 && delta <= PREFETCH_TIME_POINTS) continue;
+      if (delta >= 0 && delta <= this.policy_.prefetch.t) continue;
```
andy-sweet, line 303: As this method and `markTimeChunksForPrefetch` now depend on the value of `this.policy_.prefetch.t`, we must check for changes to `this.policy_.prefetch.t` (or given its immutability, equivalently `this.policy_`) in `updateAndCollectChunkChanges` to ensure that `updateAndCollectChunkChangesForCurrentLod` is called when the policy changes even though the xytz bounds do not.

packages/react/src/components/viewers/OmeZarrChunkedImageViewer/OmeZarrChunkedImageViewer.tsx
```diff
@@ -189,6 +190,7 @@ function createLayer(
   const layer = new ChunkedImageLayer({
     source,
     sliceCoords,
+    policy: createExplorationPolicy(),
```
andy-sweet: Doesn't need to happen in this change, but I might also expect a preconfigured policy that does no prefetching, which might be a sensible default here (and elsewhere) as it does the least preemptive work/memory usage.

packages/core/examples/ca_wave_dynamics/main.ts
andy-sweet: > I have some questions about exposing the source

Some need for this may go away once we incorporate playback internally.

> It might be better to add a mutator for imageSourcePolicy in the layer and have it delegate to the image source.

That might be a good idea in this PR anyway, since the `ChunkedImageLayer.policy_` can currently be different from the `ChunkManagerSource.policy_`, which could lead to bugs in the future.

andy-sweet (changes requested):
Generally looks great!

My only blocking feedback is regarding the need to check for changes to `ChunkManagerSource.policy_` to ensure appropriate disposal and prefetching, or something equivalent to that (e.g. calling the dispose and prefetch methods after setting).

packages/core/src/core/chunk_manager_source.ts
shlomnissan: Thanks for spotting this issue!

packages/core/examples/ca_wave_dynamics/main.ts
shlomnissan: I initially wanted to do that, but decided against it since the source remains externally accessible, the same potential bugs would still exist. I was actually looking for something like a C++ friend function to prevent that, and ended up using an internal exported symbol as a workaround.

packages/core/src/core/chunk_manager_source.ts
shlomnissan: The default for `this.policy_.lod.max` is `Number.MAX_SAFE_INTEGER`, which effectively means “no upper limit.” The extra structure computes the intersection between the dataset’s legal LOD range and the user’s policy window, while also repairing bad inputs. `min > max` validation already happens when the policy is created, but I’m being a bit more defensive here so the chunk manager doesn’t rely on assumptions made upstream.

That said, I can simplify the final line by using `clamp()` for target. I’ll make that change.

packages/core/src/core/chunk_manager_source.ts
shlomnissan: See previous comment.

packages/core/src/core/image_source_policy.ts
shlomnissan: Yeah, when I tested playback in the office this value gave better results, but it depends on many factors, so I’m not sure there’s a single meaningful value to set here.

packages/core/src/core/image_source_policy.ts
shlomnissan: Yep, that was intentional. The values in `main` weren’t really tuned for any specific preset. I’d expect prefetching in all three spatial dimensions to be the default for "exploration".

packages/core/src/core/image_source_policy.ts
shlomnissan: Same reasoning as the previous comment.

packages/core/test/image_source_policy.test.ts
shlomnissan: I actually think default value tests are worth keeping because they lock in the public contract. Defaults define how the system behaves when users do nothing, and that’s often where subtle regressions slip in. If someone later tweaks a bias, range, or priority order thinking it’s harmless, these tests will catch it. Types don’t protect that.

I agree that we can add more robust non-default tests. I'll try to think about some before requesting review.

packages/react/src/components/viewers/OmeZarrChunkedImageViewer/OmeZarrChunkedImageViewer.tsx
shlomnissan: That's a good idea! I added a `no-prefetch` policy and updated this call.

packages/core/src/core/chunk_manager_source.ts
```diff
@@ -5,65 +5,55 @@ import {
...
 export class ChunkManagerSource {
-  // First dimension of the array is time, the others cover LOD, x, y, z, c.
   private readonly chunks_: Chunk[][];
   private readonly loader_;
   private readonly lowestResLOD_: number;
   private readonly sliceCoords_: SliceCoordinates;
   private readonly dimensions_: SourceDimensionMap;
+  private readonly tIndicesWithQueuedChunks_: Set<number> = new Set();
+  private readonly sourceMaxSquareDistance2D_: number;
+  private policy_: Readonly<ImageSourcePolicy>;
```
andy-sweet: Nit: ImageSourcePolicy is already Readonly

```suggestion
  private policy_: ImageSourcePolicy;
```

andy-sweet (approved):
This looks good to me now. If you add more tests to handle non-default policy values, great! But this can also go on in as is.

packages/core/src/core/chunk_manager_source.ts
```diff
@@ -5,65 +5,55 @@ import {
...
-const PRI_PREFETCH_TIME_HIGH = 1;
-// Visible chunks at the current LOD.
-const PRI_VISIBLE_CURRENT = 2;
-// Non-prioritized prefetch chunks not at the current time.
-const PRI_PREFETCH_TIME_LOW = 3;
-// Non-visible chunks at the fallback LOD.
-const PRI_FALLBACK_BACKGROUND = 4;
-// Prefetch non-visible chunks at the current time.
-const PRI_PREFETCH_SPACE = 5;
+export const INTERNAL_POLICY_KEY = Symbol("INTERNAL_POLICY_KEY");
```
andy-sweet, line 22: It took a little time to understand what this was for - it's a way to ensure that only internal modules can use `ChunkManagerSource.setImageSourcePolicy`, right?

I think an explanatory comment here to help explain that would help other readers realize that more quickly. Alternatively, you could add a more descriptive error message in `ChunkManagerSource.setImageSourcePolicy` itself.

packages/core/src/core/image_source_policy.ts
```diff
@@ -0,0 +1,167 @@
...
+      "prefetchTime",
+    ],
+  };
+  return createImageSourcePolicy(mergeConfig(base, overrides));
+}
+
+function validatePolicyConfig(config: ImageSourcePolicyConfig) {
+  for (const [k, v] of Object.entries(config.prefetch)) {
+    if (v === undefined) continue; // z/t may be omitted
+    if (v < 0) {
```
andy-sweet, line 138: This can come in a later change (because it requires other work), but I would expect to be able to define asymmetric prefetches.

Right now, this is done implicitly for time, which only prefetches in the future. But I would expect to be able to configure a policy to prefetch backwards in time too.

Whereas, space prefetches are implicitly symmetric right now (i.e. they prefetch in negative and positive directions from the current bounds).

PR #495, opened 2025-10-14
(SMALL) Move geometry instancing into renderables

shlomnissan:
### Summary

This PR moves plane geometry instancing from layers into renderable objects.
I noticed this while reviewing code, and it seemed to me that it breaks the
renderable abstraction. IMO the only type of renderable that should receive
geometry spec is a triangulated mesh (including instanced and multi-resolution).

### Tests & Checks

- All checks have passed


andy-sweet (approved):
Works for me!

PR #496, opened 2025-10-17
Add box geometry generator

shlomnissan:
### Summary

This PR adds a box geometry generator module. I'm currently using it to
test orbit camera controls, and we may also need it for volume rendering
proxy geometry. The implementation is mostly ported from [three.js](https://github.com/mrdoob/three.js/blob/dev/src/geometries/BoxGeometry.js).

### Tests & Checks
- Visual verification.
- All checks have passed.
- I can’t think of any meaningful tests that don't feel like overkill.

https://github.com/user-attachments/assets/9ae22e37-15a3-4067-9880-a01ea790bc06

### DIFF
--- a/packages/core/src/objects/geometry/box_geometry.ts
+++ b/packages/core/src/objects/geometry/box_geometry.ts
@@ -1,0 +1,126 @@
+import { Geometry } from "../../core/geometry";
+
+type Axis = "x" | "y" | "z";
+type Vec3Obj = Record<Axis, number>;
+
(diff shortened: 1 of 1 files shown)


aganders3 (approved):
This seems fine to me. A lot of this could be shared with `PlaneGeometry`. There are some subtle and not-so-subtle differences but I also don't expect to need to update either frequently, so I don't mind the repetition.

edit: I know it's "hello world" level but after looking at 2D for so long it's neat to see the video of the cube rotating!

Issue #399, opened 2025-08-22
`LayerManager.add` hides async behavior


aganders3:
`LayerManager.add` as not marked async but it calls `layer.onAttached`, which is async:
https://github.com/chanzuckerberg/idetik/blob/c3b24c5d7582e4b5ce589330bcf55b40035e39d3/packages/core/src/core/layer_manager.ts#L33-L39

I noticed this when working on sharing a `ChunkManagerSource` between multiple layers - the `ChunkManager` keeps sources in a map to provide an existing `ChunkManagerSource` if available, but the hidden async behavior was causing a race so each layer would get a new, separate `ChunkManagerSource` instance.
https://github.com/chanzuckerberg/idetik/blob/c3b24c5d7582e4b5ce589330bcf55b40035e39d3/packages/core/src/core/chunk_manager.ts#L331-L340

andy-sweet:
Maybe could have been caught by related lint checks described in #392?

aganders3:
`layer.onAttached()` is async because it calls `chunkManager.addSource()`, which is async because it fetches source metadata over the network.

The best solution I can think of here is to make initialization lazy in the layer's `update()` instead. This is natural because layers already have a concept of `state` we can use to support the async behavior. A layer will need to track a promise internally, returning early from `update()` until its `ChunkManagerSource` is ready.

edit: actually I'm going to think about this a little more, there might be a simpler solution by just storing promises in `ChunkManager.sources_`:
```
  public addSource(source: ChunkSource): Promise<ChunkManagerSource> {
    let existing = this.sources_.get(source);
    if (!existing) {
      existing = source.open().then(loader => new ChunkManagerSource(loader, loader.getAttributes()));
      this.sources_.set(source, existing);
    }
    return existing;
  }
```

aganders3:
#497 provides an effective workaround to the issue this is currently causing, but perhaps we just want to make `onAttached` sync and require layers to handle any async initialization through their `state`.

PR #497, opened 2025-10-17
Updates to layer attachment lifecycle

aganders3:
### Summary
This fixes a race condition where multiple layers simultaneously requesting the same data source can create duplicate ChunkManagerSource instances (see #399).

This also adds an additional `onDetached` method for layers to implement cleanup on removal, and uses it to clear some state from the `ChunkedImageLayer`. Finally, this adds a guard to the `ChunkedImageLayer` to prevent unsupported attachment to multiple `LayerManager`s (in effect, simultaneous attachment to multiple viewports).

### Related Issue
related to #399 but may not close - the method still hides async behavior, this just fixes the immediate problem it was causing
[Doc with thoughts on Idetik layers](https://docs.google.com/document/d/1TSdHHSB0fq5dbbmeWIMSalwF-fZHWndTYcIZy2mhrIM/edit?usp=sharing)

### Tests & Checks
All existing tests pass
Manually verified race condition fix by checking that concurrent layer additions return the same `ChunkManagerSource` promise

### DIFF
--- a/packages/core/src/core/layer.ts
+++ b/packages/core/src/core/layer.ts
@@ -3,7 +3,6 @@ import { RenderableObject } from "./renderable_object";
 import { clamp } from "../utilities/clamp";
 import { Logger } from "../utilities/logger";
 import { EventContext } from "./event_dispatcher";
-import { Viewport } from "./viewport";
 
 export type LayerState = "initialized" | "loading" | "ready";
 export type blendMode = "normal" | "additive" | "subtractive" | "multiply";
@@ -18,10 +17,6 @@ export interface LayerOptions {
   opacity?: number;
   blendMode?: blendMode;
 }
-
-export type RenderContext = {
-  viewport: Viewport;
-};
 
 export abstract class Layer {
   public abstract readonly type: string;
@@ -64,7 +59,7 @@ export abstract class Layer {
     this.opacity_ = clamp(value, 0.0, 1.0);
   }
 
-  public abstract update(context?: RenderContext): void;
+  public abstract update(): void;
 
   public onEvent(_: EventContext): void {}
 
--- a/packages/core/src/core/layer.ts
+++ b/packages/core/src/core/layer.ts
@@ -3,6 +3,7 @@ import { RenderableObject } from "./renderable_object";
 import { clamp } from "../utilities/clamp";
 import { Logger } from "../utilities/logger";
 import { EventContext } from "./event_dispatcher";
+import { Viewport } from "./viewport";
 
 export type LayerState = "initialized" | "loading" | "ready";
 export type blendMode = "normal" | "additive" | "subtractive" | "multiply";
@@ -17,6 +18,10 @@ export interface LayerOptions {
   opacity?: number;
   blendMode?: blendMode;
 }
+
+export type RenderContext = {
+  viewport: Viewport;
+};
 
 export abstract class Layer {
   public abstract readonly type: string;
@@ -59,7 +64,7 @@ export abstract class Layer {
     this.opacity_ = clamp(value, 0.0, 1.0);
   }
 
-  public abstract update(): void;
+  public abstract update(context?: RenderContext): void;
 
   public onEvent(_: EventContext): void {}
 
@@ -68,6 +73,8 @@ export abstract class Layer {
   // manager, but for now, we allow optional overrides to avoid requiring
   // placeholder implementations.
   public async onAttached(_context: IdetikContext) {}
+
+  public onDetached(): void {}
 
   public get objects() {
     return this.objects_;
--- a/packages/core/src/layers/chunked_image_layer.ts
+++ b/packages/core/src/layers/chunked_image_layer.ts
@@ -73,11 +73,24 @@ export class ChunkedImageLayer extends Layer implements ChannelsEnabled {
   }
 
   public async onAttached(context: IdetikContext) {
+    if (this.chunkManagerSource_) {
+      throw new Error(
+        `ChunkedImageLayer cannot be attached to multiple viewports. ` +
+          `Layer is already attached. ` +
+          `Create separate layer instances for each viewport.`
+      );
+    }
     this.chunkManagerSource_ = await context.chunkManager.addSource(
       this.source_,
       this.sliceCoords_,
       this.policy_
     );
+  }
+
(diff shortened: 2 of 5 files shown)


packages/core/src/layers/chunked_image_layer.ts
```diff
@@ -73,13 +73,26 @@ export class ChunkedImageLayer extends Layer implements ChannelsEnabled {
...
     this.chunkManagerSource_ = await context.chunkManager.addSource(
       this.source_,
       this.sliceCoords_,
       this.policy_
     );
   }
 
+  public onDetached(): void {
+    this.chunkManagerSource_ = undefined;
+    this.visibleChunks_.clear();
```
andy-sweet: I think I would expect this to release the image renderables in the pool too because the way I understand `onDetached` is that the layer instance might be reused somewhere else or later, so the pool of renderables might also be useful to keep around.

The easiest way to do that is to define a helper method that could be called here and in updateChunks.

packages/core/src/layers/chunked_image_layer.ts
```diff
@@ -73,13 +73,26 @@ export class ChunkedImageLayer extends Layer implements ChannelsEnabled {
   }
 
   public async onAttached(context: IdetikContext) {
+    if (this.chunkManagerSource_) {
+      throw new Error(
+        `ChunkedImageLayer cannot be attached to multiple viewports. ` +
```
andy-sweet: This error message talks about viewports, but the PR description calls out that onAttached/onDetached is related to the static runtime context and not the dynamic viewport context, so this wording feels out of place and potentially confusing.

packages/core/src/core/layer.ts
```diff
@@ -59,7 +64,7 @@ export abstract class Layer {
     this.opacity_ = clamp(value, 0.0, 1.0);
   }
 
-  public abstract update(): void;
+  public abstract update(context?: RenderContext): void;
```
andy-sweet: As mentioned in the PR, this is added but never used. The changes related to this addition are so small, I don't think there any value in adding them to this PR. However, adding them in this PR mixes two separate concepts (again as described in the PR), does have some downside.

Instead this change should be added at the same time it is used and/or tested.

andy-sweet (changes requested):
This might be borderline nitty, but I don't like that this PR explicitly contains two separate changes (as called out in the PR description) and then also does not use or test one of the changes, which itself is very small.

packages/core/src/core/layer.ts
aganders3: This is fair - I included it here to prompt discussion about whether it's needed at all, but could have indicated that in the PR description. The alternatives I can think of are:

* thread the `viewport` instance through the `LayerManager` where it calls `layer.onAttached`. I find it a bit clunky for the `LayerManager` (child) to need a reference to its parent (`Viewport`), but maybe this is fine because we currently have a 1:1 parent:child relationship of `Viewport`:`LayerManager` and I don't really expect that to change.

* get rid of `onAttached` lifecycle and expect layers to lazily initialize everything they need through the `RenderContext` passed to their `update()` (which could include `ChunkManager` - currently the only thing passed in the runtime context).

andy-sweet: I don't love the first option. The second option sounds fine to me with the potential benefit of removing the special idetik runtime context logic, but there might be some value in that separate (as in the PR description), so probably warrants a discussion outside of a comment reply here.

aganders3: I've removed the `RenderContext` from this PR.

packages/core/src/layers/chunked_image_layer.ts
aganders3: I see what you mean. I feel like the error was descriptive in what the user might be trying to achieve, but you're right it did not match with the PR description. I tried to make it more clear what the problem is, but happy to change it again if you have ideas.

packages/core/src/layers/chunked_image_layer.ts
```diff
@@ -73,13 +73,25 @@ export class ChunkedImageLayer extends Layer implements ChannelsEnabled {
   }
 
   public async onAttached(context: IdetikContext) {
+    if (this.chunkManagerSource_) {
+      throw new Error(
+        "ChunkedImageLayer is already attached. " +
+          "A layer cannot be attached to multiple LayerManagers simultaneously."
```
andy-sweet, line 79: Nit: I don't think there's a need to mention anything outside the scope of what this knows about, but I'll leave it up to you to decide based on other pending changes.

```suggestion
          "A layer cannot be attached to multiple context's simultaneously."
```

packages/core/src/layers/chunked_image_layer.ts
```diff
@@ -101,12 +113,10 @@ export class ChunkedImageLayer extends Layer implements ChannelsEnabled {
 
     const orderedByLOD = this.chunkManagerSource_.getChunks();
     const current = new Set(orderedByLOD);
-    this.visibleChunks_.forEach((image, chunk) => {
-      if (!current.has(chunk)) {
-        this.visibleChunks_.delete(chunk);
-        this.pool_.release(poolKeyForImageRenderable(chunk), image);
-      }
-    });
+    const nonVisibleChunks = Array.from(this.visibleChunks_.keys()).filter(
```
andy-sweet, line 116: Ah, it's unfortunate this requires an extra iteration over `this.visibleChunks_` compared to main. I guess that `releaseAndRemoveChunks` could take an optional `exclude: Set<Chunk>` instead of the chunks to remove, but maybe that's a little ugly.

andy-sweet (approved):
This looks good to me now!

packages/core/src/layers/chunked_image_layer.ts
aganders3: Yes, I considered that but also found it ugly. I think this code is likely to change especially once we start profiling, so I think it's acceptable for now.

I think ES2025 has a set-difference method but I think it's too new and might iterate anyway.

packages/core/src/layers/chunked_image_layer.ts
aganders3: Ugh, sorry I made your suggested change but committed it to the wrong branch so it was not merged here 🤦.

Issue #483, opened 2025-10-06
Extract React Components into a Separate Repository


phoenixAja:
#### Summary
Move shared React components into a dedicated repository to decouple UI development from the core development. Previously, co-locating components with core packages caused dependency entanglement, version lock-in, and slowed iteration on both the core and front-end layers.  

This refactor will isolate the component library, enabling faster UI iteration, cleaner dependency management, and independent release cycles.

---

#### Goals
- Create a new standalone repository for React components.
- Remove build and packaging configuration for UI components from the main repo.
- Preserve existing import paths through a transitional alias or package version bump.
- Set up CI/CD for independent publishing and versioning.

andy-sweet:
@phoenixAja : closing this as done - reopen if that's not the case.

PR #499, opened 2025-10-17
refactor: extract React package to standalone repository

phoenixAja:
### Summary
Extracts the React package (@idetik/react-prerelease) to its own repository at https://github.com/chanzuckerberg/idetik-react. This simplifies the monorepo structure and allows independent development and release cycles for the React components.

### Changes
  - Removed packages/react directory and all source files
  - Updated package.json to remove react workspace and related scripts
  - Updated GitHub Actions workflow to remove React build, lint, and test steps
  - Updated README.md with simplified release process for core package only

  ### Related Issue
[483](https://github.com/chanzuckerberg/idetik/issues/483)

### Tests & Checks
  - Verified core package build still works successfully
  - Confirmed all React package references have been removed or updated to point to new repository
  - GitHub Actions workflow updated to only build/test core package


README.md
```diff
@@ -33,54 +33,35 @@ A layer-based library for interactive visualization of large datasets.
...
-```shell
-cd ../react
-npm version major
-npm install
-npm run build
-```
-
-After both package versions have been bumped, create a PR, get it approved, and merge it to `main`.
-
-Next, checkout the corresponding commit on `main` and run the following commands for
```
aganders3, line 73: I think we want to keep this workflow, where the release is built off `main` after merging the prep PR.

README.md
```diff
@@ -33,54 +33,35 @@ A layer-based library for interactive visualization of large datasets.
 
 ## Release
 
-Currently, we maintain Idetik prerelease packages on npm: [@idetik/core-prerelease](https://www.npmjs.com/package/@idetik/core-prerelease?activeTab=readme) and [@idetik/react-prerelease](https://www.npmjs.com/package/@idetik/react-prerelease).
+Currently, we maintain the [@idetik/core-prerelease](https://www.npmjs.com/package/@idetik/core-prerelease?activeTab=readme) package on npm.
+
+**Note:** The React package ([@idetik/react-prerelease](https://www.npmjs.com/package/@idetik/react-prerelease)) has been moved to its own repository at https://github.com/chanzuckerberg/idetik-react. See that repository for React package release instructions.
```
andy-sweet: Nit: I don't think there's a big need for this because not many people currently care about the split (other than us), so I'd probably just remove it to keep this concise.
```suggestion
```

.github/workflows/lint-test-build.yml
```diff
@@ -67,4 +63,4 @@ jobs:
         run: npm ci
 
       - name: Build
-        run: npm run build:all
+        run: npm run build --workspace=@idetik/core-prerelease
```
andy-sweet: The workspace is already defined in `package.json`

```suggestion
        run: npm run build
```

package.json
```diff
@@ -1,15 +1,12 @@
...
-    "packages/react"
+    "packages/core"
   ],
   "scripts": {
     "build": "npm run build --workspace=@idetik/core-prerelease",
-    "build:all": "npm run build --workspaces",
     "build:examples": "npm run build:examples --workspace=@idetik/core-prerelease",
     "examples": "npm run dev --workspace=@idetik/core-prerelease",
-    "components": "npm run dev --workspace=@idetik/react-prerelease",
     "format:all": "npm run format --workspaces"
```
andy-sweet: Nit: this would save me a few keystrokes.

```suggestion
    "format": "npm run format --workspaces=@idetik/core-prerelease"
```

README.md
andy-sweet: Agreed.

andy-sweet (changes requested):
The only blocking feedback I have here is to run `npm install` and commit the related changes to `package-lock.json`.

Other comments are nits.

I think there's some more cleanup that could occur in some of the other config files that no longer need to refer to react things, but that can happen later.

I don't feel too strongly about keeping workspaces, but I'd lean towards removing them if we're not using them because it avoids duplication of script definitions in `package.json`. Again, that can be done after this change is merged.

README.md
phoenixAja: made this more clear. Thanks!

aganders3:
@phoenixAja I hope you don't mind I pushed the changes to `package-lock.json` that @andy-sweet requested.

Otherwise this LGTM now!

edit: actually I think the steps for publishing are still not quite correct. `npm run pub` should be run after merging to `main`.

package-lock.json
```diff
@@ -1,12 +1,11 @@
 {
-  "name": "idetik",
+  "name": "idetik-review",
```
andy-sweet: Nit: this doesn't really matter, but the old name of `idetik` seems reasonable given the name of the repo. I think this probably gets generated based on the name of your directory given there is no name in package.json.

```suggestion
  "name": "idetik",
```

andy-sweet (approved):
Looks good to me now. Agreed that the order of release instructions has shifted a little bit, but given our release process is relatively loose right now, that's not blocking.

package-lock.json
aganders3: Sorry that's my fault for using a worktree again. Good catch!

phoenixAja:
> Looks good to me now. Agreed that the order of release instructions has shifted a little bit, but given our release process is relatively loose right now, that's not blocking.

fixed this, sorry head was in too many different places yesterday, it's synced now with what's on main (just removing react update)

PR #500, opened 2025-10-19
Add orbit controls for perspective camera

shlomnissan:
### Summary

This PR adds orbit controls for perspective cameras, using spherical coordinates
and common mappings: right mouse button to orbit, middle (or shift + left) mouse
to pan, and scroll to zoom. This feature was prioritized to accelerate 3D volume
rendering development.

### Notes

- Test this feature on the temporary `shlomnissan/feature/volume-layer-2d1f` branch. Open the "Volume Rendering" example from the list to verify the controls.
- Input event updates aren’t particularly stable, they fire inconsistently and can make the controls feel jittery. We can achieve smoother motion by updating the camera each frame in the main loop and scaling movements by the frame’s delta time. I’ll likely post a follow-up once this is merged.
- This local implementation also reflects what I had in mind for 2D controls. I’m not a fan of having zoom and pan methods baked into the camera base class. These actions belong in the control layer, not the camera itself.
- The Spherical implementation follows a common convention used in game development, where phi controls horizontal rotation around the Y-axis and theta controls vertical rotation relative to the equator. These axis directions and angle orientations are ultimately arbitrary.

### Tests & Checks

- All checks have passed.
- Visual verification (see video below).
- Added basic unit tests for the spherical coordinates class.

https://github.com/user-attachments/assets/f31a1ff4-0e73-43df-bf68-4a414a97d69d

### DIFF
--- a/packages/core/src/core/transforms.ts
+++ b/packages/core/src/core/transforms.ts
@@ -1,4 +1,6 @@
-import { mat4, vec3, quat } from "gl-matrix";
+import { mat4, mat3, vec3, quat } from "gl-matrix";
+
+const WORLD_UP = vec3.fromValues(0, 1, 0);
 
 export class TrsTransform {
   private dirty_ = true;
@@ -42,6 +44,21 @@ export class TrsTransform {
 
   public setScale(vec: vec3) {
     vec3.copy(this.scale_, vec);
+    this.dirty_ = true;
+  }
+
+  public targetTo(target: vec3) {
+    // Avoid zero-length forward vector
+    if (vec3.equals(this.translation_, target)) {
+      target = vec3.clone(target);
+      target[2] += 1e-3;
+    }
+
+    const m = mat4.targetTo(mat4.create(), this.translation_, target, WORLD_UP);
+    const rotation = mat3.fromMat4(mat3.create(), m);
+    quat.fromMat3(this.rotation_, rotation);
+    quat.normalize(this.rotation_, this.rotation_);
+
     this.dirty_ = true;
   }
 
--- a/packages/core/src/core/transforms.ts
+++ b/packages/core/src/core/transforms.ts
@@ -1,5 +1,6 @@
-import { mat4, mat3, vec3, quat } from "gl-matrix";
+import { mat4, mat3, vec3, quat, glMatrix } from "gl-matrix";
 
+// +Y is world up to match gl-matrix
 const WORLD_UP = vec3.fromValues(0, 1, 0);
 
 export class TrsTransform {
@@ -48,10 +49,11 @@ export class TrsTransform {
   }
 
   public targetTo(target: vec3) {
-    // Avoid zero-length forward vector
+    // Prevent zero-length forward vector by nudging
+    // target slightly along +Z
     if (vec3.equals(this.translation_, target)) {
       target = vec3.clone(target);
-      target[2] += 1e-3;
+      target[2] += glMatrix.EPSILON;
     }
 
     const m = mat4.targetTo(mat4.create(), this.translation_, target, WORLD_UP);
--- a/packages/core/src/math/spherical.ts
+++ b/packages/core/src/math/spherical.ts
@@ -1,4 +1,3 @@
-import { clamp } from "@/utilities/clamp";
 import { vec3 } from "gl-matrix";
 
 export class Spherical {
@@ -10,12 +9,6 @@ export class Spherical {
     this.radius = radius;
     this.phi = phi;
     this.theta = theta;
-  }
-
-  public makeSafe() {
-    const ε = 1e-3;
-    const limit = Math.PI / 2 - ε;
-    this.theta = clamp(this.theta, -limit, limit);
   }
 
   public toVec3() {
--- a/packages/core/src/math/spherical.ts
+++ b/packages/core/src/math/spherical.ts
@@ -1,0 +1,29 @@
+import { clamp } from "@/utilities/clamp";
+import { vec3 } from "gl-matrix";
+
+export class Spherical {
+  public radius;
+  public phi;
+  public theta;
+
+  constructor(radius: number, phi: number, theta: number) {
+    this.radius = radius;
+    this.phi = phi;
+    this.theta = theta;
+  }
+
+  public makeSafe() {
+    const ε = 1e-3;
+    const limit = Math.PI / 2 - ε;
+    this.theta = clamp(this.theta, -limit, limit);
+  }
+
+  public toVec3() {
+    const c = Math.cos(this.theta);
+    return vec3.fromValues(
+      this.radius * Math.sin(this.phi) * c,
+      -this.radius * Math.sin(this.theta),
+      this.radius * Math.cos(this.phi) * c
+    );
+  }
+}
--- a/packages/core/src/objects/cameras/orbit_controls.ts
+++ b/packages/core/src/objects/cameras/orbit_controls.ts
@@ -1,9 +1,10 @@
 import { CameraControls } from "./controls";
 import { EventContext } from "../../core/event_dispatcher";
 import { PerspectiveCamera } from "./perspective_camera";
-import { Spherical } from "@/math/spherical";
+import { Spherical } from "../../math/spherical";
 
-import { vec3 } from "gl-matrix";
+import { glMatrix, vec3 } from "gl-matrix";
+import { clamp } from "../../utilities/clamp";
 
 const MOUSE_BUTTON_NONE = -1;
 const MOUSE_BUTTON_LEFT = 0;
@@ -15,14 +16,14 @@ const ZOOM_SPEED = 0.0009;
 
 export class OrbitControls implements CameraControls {
   private readonly camera_: PerspectiveCamera;
-  private readonly spherical_: Spherical;
+  private readonly sphericalPos_: Spherical;
   private readonly target_ = vec3.create();
 
   private currMouseButton_ = MOUSE_BUTTON_NONE;
 
   constructor(camera: PerspectiveCamera, radius = 1, yaw = 0, pitch = 0) {
     this.camera_ = camera;
-    this.spherical_ = new Spherical(radius, yaw, pitch);
+    this.sphericalPos_ = new Spherical(radius, yaw, pitch);
     this.update();
   }
 
@@ -58,11 +59,10 @@ export class OrbitControls implements CameraControls {
     const dx = e.movementX ?? 0;
     const dy = e.movementY ?? 0;
 
-    const m = e.shiftKey;
-    const b = this.currMouseButton_;
-
-    const doOrbit = b === MOUSE_BUTTON_LEFT && !m;
-    const doPan = (b === MOUSE_BUTTON_LEFT && m) || b === MOUSE_BUTTON_MIDDLE;
+    const doOrbit = this.currMouseButton_ === MOUSE_BUTTON_LEFT && !e.shiftKey;
+    const doPan =
+      (this.currMouseButton_ === MOUSE_BUTTON_LEFT && e.shiftKey) ||
+      this.currMouseButton_ === MOUSE_BUTTON_MIDDLE;
 
     if (doOrbit) this.orbit(dx, dy);
     if (doPan) this.pan(dx, dy);
@@ -71,7 +71,6 @@ export class OrbitControls implements CameraControls {
   }
 
   private onWheel(event: EventContext) {
-    if (!event.worldPos || !event.clipPos) return;
     const e = event.event as WheelEvent;
     e.preventDefault(); // prevent the page from scrolling
 
@@ -89,14 +88,20 @@ export class OrbitControls implements CameraControls {
   }
 
   private orbit(dx: number, dy: number) {
-    this.spherical_.phi -= dx * ORBIT_SPEED;
-    this.spherical_.theta += dy * ORBIT_SPEED;
-    this.spherical_.makeSafe();
+    this.sphericalPos_.phi -= dx * ORBIT_SPEED;
+    this.sphericalPos_.theta += dy * ORBIT_SPEED;
+
+    // Prevent the camera from reaching the poles (±π/2), where
+    // the view direction would flip and orbit controls would invert.
+    // We clamp the elevation angle slightly before ±π/2 to maintain
+    // stable rotation and avoid gimbal-like artifacts.
+    const limit = Math.PI / 2 - glMatrix.EPSILON;
+    this.sphericalPos_.theta = clamp(this.sphericalPos_.theta, -limit, limit);
   }
 
   private pan(dx: number, dy: number) {
     // Scale pan speed by distance so movement feels consistent
-    const speed = this.spherical_.radius * PAN_SPEED;
+    const speed = this.sphericalPos_.radius * PAN_SPEED;
     const delta = vec3.create();
 
     vec3.scaleAndAdd(delta, delta, this.camera_.right, dx);
@@ -109,11 +114,18 @@ export class OrbitControls implements CameraControls {
   private zoom(dy: number) {
     // Exponential for smooth, distance-independent zoom
     const scale = Math.exp(dy * ZOOM_SPEED);
-    this.spherical_.radius = Math.max(0.01, this.spherical_.radius * scale);
+    this.sphericalPos_.radius = Math.max(
+      0.01,
+      this.sphericalPos_.radius * scale
+    );
   }
 
   private update() {
-    const p = vec3.add(vec3.create(), this.target_, this.spherical_.toVec3());
+    const p = vec3.add(
+      vec3.create(),
+      this.target_,
+      this.sphericalPos_.toVec3()
+    );
     this.camera_.transform.setTranslation(p);
     this.camera_.transform.targetTo(this.target_);
   }
--- a/packages/core/src/objects/cameras/orbit_controls.ts
+++ b/packages/core/src/objects/cameras/orbit_controls.ts
@@ -1,0 +1,120 @@
+import { CameraControls } from "./controls";
+import { EventContext } from "../../core/event_dispatcher";
+import { PerspectiveCamera } from "./perspective_camera";
+import { Spherical } from "@/math/spherical";
+
+import { vec3 } from "gl-matrix";
+
+const MOUSE_BUTTON_NONE = -1;
+const MOUSE_BUTTON_LEFT = 0;
+const MOUSE_BUTTON_MIDDLE = 1;
+
+const ORBIT_SPEED = 0.009;
+const PAN_SPEED = 0.001;
+const ZOOM_SPEED = 0.0009;
+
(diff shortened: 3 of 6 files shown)


packages/core/src/math/spherical.ts
```diff
@@ -0,0 +1,29 @@
...
+    this.theta = theta;
+  }
+
+  public makeSafe() {
+    const ε = 1e-3;
+    const limit = Math.PI / 2 - ε;
+    this.theta = clamp(this.theta, -limit, limit);
+  }
+
+  public toVec3() {
```
aganders3, line 14: Might be a nitpick but I'd suggest `toCartesian` to indicate what the method _does_ and let the return type indicate the type.

packages/core/src/math/spherical.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  constructor(radius: number, phi: number, theta: number) {
+    this.radius = radius;
+    this.phi = phi;
+    this.theta = theta;
+  }
+
+  public makeSafe() {
+    const ε = 1e-3;
+    const limit = Math.PI / 2 - ε;
```
aganders3: nit: I'd prefer to just use `epsilon` instead of adding non-ASCII unicode characters

aganders3: I'm also not sure `makeSafe` needs to be public. It seems like it could just be called as part of `toVec3` but I'm not great at thinking in spherical coordinates so I could be missing something. As long as you're working purely in spherical coordinates I don't think there's anything inherently unsafe about any values.

packages/core/src/objects/cameras/orbit_controls.ts
```diff
@@ -0,0 +1,120 @@
+import { CameraControls } from "./controls";
+import { EventContext } from "../../core/event_dispatcher";
+import { PerspectiveCamera } from "./perspective_camera";
+import { Spherical } from "@/math/spherical";
```
aganders3:
```suggestion
import { Spherical } from "../../math/spherical";
```

packages/core/src/objects/cameras/orbit_controls.ts
```diff
@@ -0,0 +1,120 @@
...
+const MOUSE_BUTTON_NONE = -1;
+const MOUSE_BUTTON_LEFT = 0;
+const MOUSE_BUTTON_MIDDLE = 1;
+
+const ORBIT_SPEED = 0.009;
+const PAN_SPEED = 0.001;
+const ZOOM_SPEED = 0.0009;
+
+export class OrbitControls implements CameraControls {
+  private readonly camera_: PerspectiveCamera;
```
aganders3, line 18: It shouldn't be much work to make this work with both `PerspectiveCamera` and `OrthographicCamera`. My experience is that orthographic is not uncommon for scientific viz even with a fully 3D view.

packages/core/src/math/spherical.ts
```diff
@@ -0,0 +1,29 @@
+import { clamp } from "@/utilities/clamp";
```
aganders3:
```suggestion
import { clamp } from "../utilities/clamp";
```

packages/core/src/core/transforms.ts
```diff
@@ -1,4 +1,6 @@
-import { mat4, vec3, quat } from "gl-matrix";
+import { mat4, mat3, vec3, quat } from "gl-matrix";
+
+const WORLD_UP = vec3.fromValues(0, 1, 0);
```
aganders3, line 4: I'm a little confused by this given the description of world coordinates in `webgl_renderer.ts`. From that, I think our `WORLD_UP` should actually be `(0, -1, 0)`. It also feels like we should be able to centralize this value somewhere to make sure it's renderer independent.

packages/core/test/spherical.test.ts
```diff
@@ -0,0 +1,92 @@
...
+  const theta = PI_OVER_2 + Number.EPSILON;
+  const s = new Spherical(2, phi, theta);
+
+  s.makeSafe();
+  const expected = PI_OVER_2 - Number.EPSILON;
+
+  expect(s.phi).toBe(glMatrix.toRadian(30));
+  expect(s.theta).toBeCloseTo(expected);
+  expect(s.radius).toBe(2);
+});
```
aganders3: This is testing implementation details I'm not sure are particularly high-value. If `makeSafe` is made private these can just be removed, and make sure the `toVec3` tests cover the appropriate degenerate cases.

Also I'm not sure these tests are correct as `Number.EPSILON` is much smaller than the epsilon defined in `makeSafe`. So the tests still pass if you change them to `expect(s.theta).toBeCloseTo(theta);`

packages/core/src/math/spherical.ts
```diff
@@ -0,0 +1,29 @@
...
+  public theta;
+
+  constructor(radius: number, phi: number, theta: number) {
+    this.radius = radius;
+    this.phi = phi;
+    this.theta = theta;
+  }
+
+  public makeSafe() {
+    const ε = 1e-3;
```
andy-sweet: Nit: I don't mind using non-(non-extended) ascii for variable names, but given that theta, phi, and pi don't get the full Greek treatment, I'm not sure I like making an exception for epsilon.

packages/core/src/math/spherical.ts
andy-sweet: Nit: I'd prefer to see this as a constant defined outside of the method (e.g. at the module or static class level).

packages/core/src/core/transforms.ts
```diff
@@ -45,6 +47,21 @@ export class TrsTransform {
     this.dirty_ = true;
   }
 
+  public targetTo(target: vec3) {
+    // Avoid zero-length forward vector
+    if (vec3.equals(this.translation_, target)) {
+      target = vec3.clone(target);
+      target[2] += 1e-3;
```
andy-sweet: Is the RHS value related to other small values defined elsewhere? I wonder if having a shared source of what epsilon means with respect to transforms/coordinates might help (e.g. like gl-matrix's EPSILON).

Is bumping in positive z a convention? Does it effectively bump the target forward when used with the orbit controls?

packages/core/src/objects/cameras/orbit_controls.ts
```diff
@@ -0,0 +1,120 @@
...
+  private readonly spherical_: Spherical;
+  private readonly target_ = vec3.create();
+
+  private currMouseButton_ = MOUSE_BUTTON_NONE;
+
+  constructor(camera: PerspectiveCamera, radius = 1, yaw = 0, pitch = 0) {
+    this.camera_ = camera;
+    this.spherical_ = new Spherical(radius, yaw, pitch);
```
aganders3: nit: I think `spherical_` is a bit ambiguous about what this does, so I'd probably call it `position_` or maybe `sphericalPos_`.

packages/core/test/spherical.test.ts
```diff
@@ -0,0 +1,92 @@
...
+const PI_OVER_4 = PI / 4;
+
+function expectVec3Close(v: vec3, [x, y, z]: [number, number, number]) {
+  expect(v[0]).toBeCloseTo(x);
+  expect(v[1]).toBeCloseTo(y);
+  expect(v[2]).toBeCloseTo(z);
+}
+
+test("makeSafe clamps when theta > +π/2", () => {
+  const phi = glMatrix.toRadian(30);
```
andy-sweet: Nit: Is there something special about 30 degrees in this test and in others, or is it just some non-special value? If so, I'd prefer it to be defined as a constant like PI/4. Or maybe we could just use PI/4 as a non-special value too?

I'd also be fine with PI/6 and avoiding the call to `toRadian`.

aganders3 (changes requested):
This generally looks good to me and seems to work well in the demo. Mostly my comments are nits and some questions, but I'll be happy to approve as-is if needed to unblock progress on the volume renderer.

> This local implementation also reflects what I had in mind for 2D controls. I’m not a fan of having zoom and pan methods baked into the camera base class. These actions belong in the control layer, not the camera itself.

This is fine with me but I feel it would make sense to do it here or in a fast follow-on PR to avoid confusion. Using the versions on the camera classes makes this Just Work™ with both camera types, but as a compromise we could make utility functions like `panCamera` and `zoomCamera` (or `panOrtho`/`zoomOrtho`) that would prevent some re-implementation and keep it out of the camera class.

packages/core/src/math/spherical.ts
aganders3: Sorry, one more thought on `makeSafe`: should it first normalize `theta` to `[-pi, pi]` before clamping? Likewise it might be worth normalizing `phi` to `[0, 2pi]`.

packages/core/src/math/spherical.ts
```diff
@@ -0,0 +1,29 @@
...
+  public phi;
+  public theta;
+
+  constructor(radius: number, phi: number, theta: number) {
+    this.radius = radius;
+    this.phi = phi;
+    this.theta = theta;
+  }
+
+  public makeSafe() {
```
andy-sweet: Given the name in this context, I think of this as a way to manually restrict the domain of theta for spherical coordinates in Idetik (e.g. to guarantee uniqueness of the mapping between spherical and Cartesian coordinates).

I think that's fine, but it feels incomplete because 

- it's only for theta and not phi;
- and it doesn't happen on every change to theta (i.e. this has to explicitly called).

If this is intended to be a special case for orbit controls, then maybe the clamping logic for theta belongs there? Otherwise, this feels a bit like spherical coordinates for orbit controls.

packages/core/src/objects/cameras/orbit_controls.ts
andy-sweet: Nit:

```suggestion
import { Spherical } from "../../math/spherical";
```

packages/core/src/objects/cameras/orbit_controls.ts
```diff
@@ -0,0 +1,120 @@
...
+const ZOOM_SPEED = 0.0009;
+
+export class OrbitControls implements CameraControls {
+  private readonly camera_: PerspectiveCamera;
+  private readonly spherical_: Spherical;
+  private readonly target_ = vec3.create();
+
+  private currMouseButton_ = MOUSE_BUTTON_NONE;
+
+  constructor(camera: PerspectiveCamera, radius = 1, yaw = 0, pitch = 0) {
```
andy-sweet, line 24: I'm fine with position arguments for spherical coordinates, given it follows some convention and that might be called frequently, but I'd prefer a props style structure here so we effectively have keyword arguments to improve readability when constructing this.

shlomnissan:
> Mostly my comments are nits and some questions

@aganders3 I think we should reserve “Request changes” for cases where something would break or block downstream work. Otherwise, comments like these are great to capture but shouldn’t block progress. Of course, we should still address them before merging.

> that would prevent some re-implementation and keep it out of the camera class

@aganders3 I don’t particularly mind the re-implementation. Modules like camera controls should stay self-contained so they can evolve independently and keep their own abstractions clear.

packages/core/src/objects/cameras/orbit_controls.ts
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  private onPointerMove(event: EventContext) {
+    if (this.currMouseButton_ == MOUSE_BUTTON_NONE) return;
+
+    const e = event.event as PointerEvent;
+    const dx = e.movementX ?? 0;
+    const dy = e.movementY ?? 0;
+
+    const m = e.shiftKey;
```
andy-sweet: Nit: I guess this m is for modifier? It took a few seconds to infer/guess that, so maybe `shift` might be better here (or just use e.shiftKey)?

packages/core/src/math/spherical.ts
shlomnissan: I used `glMatrix.EPSILON` instead.

andy-sweet (approved):
Generally looks pretty clean to me and the example in the related branch works well enough. All my comments are optional - either nits or points of confusion for me.

packages/core/src/math/spherical.ts
shlomnissan: Why? `limit` is only used within `makeSafe`, so I’d prefer to keep it scoped locally. It’s specific to the normalization logic for spherical coordinates, not something I expect to reuse elsewhere within or outside this module.

packages/core/src/core/transforms.ts
shlomnissan: - I updated this to use `glMatrix.EPSILON`
- The small bump just ensures the forward vector isn’t zero-length when the target coincides with the camera position. I picked Z arbitrarily since it’s typically the forward axis in our coordinate system, but the direction doesn’t materially affect the result. I'll update the comment so its clearer.

packages/core/test/spherical.test.ts
shlomnissan: I used PI/6 as non-special value.

packages/core/src/math/spherical.ts
andy-sweet: The main reason was that it was unclear whether the connection to epsilon was intended be longer range than the scope here (i.e. is 1e-3 supposed to exactly the same or some factor of another epsilon elsewhere). I think you might have redefined that epsilon to more explicitly use something else, in which case that reason goes away.

The other reason is to only perform the two calculations here once, but that's pretty negligible and some interpreter optimization probably already does that, so very optional.

aganders3 (approved):
> I think we should reserve “Request changes” for cases where something would break or block downstream work. Otherwise, comments like these are great to capture but shouldn’t block progress. Of course, we should still address them before merging.

Sure - none of this breaks or blocks work from what I see.

packages/core/src/math/spherical.ts
shlomnissan: I’d keep it as `toVec3`. The name emphasizes the return type rather than the coordinate system. Since spherical → cartesian is already implied by the class context `toVec3` stays concise and consistent with typical math APIs.

packages/core/src/math/spherical.ts
shlomnissan: - I updated to use `glMatrix.EPSILON`.
- I’d keep `makeSafe` public and avoid calling it from `toVec3`. `toVec3` should remain a pure conversion. Mutating internal angles there would be unexpected. This pattern is common (see [three.js implementation](https://github.com/mrdoob/three.js/blob/dev/src/math/Spherical.js#L84)) note that three.js uses slightly different conventions and normalizes phi.
- `theta` naturally lies in [-pi/2, pi/2], so wrapping isn’t required for correctness. Normalizing `phi` isn’t necessary either. It’s harmless, but if we were to do it, I’d prefer handling it externally where wrapping ties more directly to user input, rather than every time we clamp.

packages/core/src/objects/cameras/orbit_controls.ts
shlomnissan: I agree. I might make this a follow-up tasks since it does requires modifications zooming and panning and I would like to test this separately.

packages/core/src/objects/cameras/orbit_controls.ts
shlomnissan: I kept it short because I didn't want the formatter breaking lines. I think it's clear from context but I'm happy to change it if you feel like it hinders clarity.

packages/core/src/core/transforms.ts
shlomnissan: The `WORLD_UP` here is meant to follow the conventional right-handed coordinate system with +Y up, which keeps math utilities (like `targetTo`) and general transforms consistent with `gl-matrix`. The confusion between world and clip space is exactly why I didn’t centralize this constant, but I can add a comment to clarify that distinction.

packages/core/test/spherical.test.ts
shlomnissan: I think guard-rail tests are useful since it’s easy to regress that logic especially when tweaking math. Since I decided to keep `makeSafe` public, I think keeping this test is reasonable. And good catch on `Number.EPSILON`. I can update it to use `glMatrix.EPSILON` to match the implementation so the intent is clearer.

packages/core/src/objects/cameras/orbit_controls.ts
andy-sweet: Even `mod` could help here if it passes the format checks. I'm not used to seeing `m` as a single letter variable name in code.

Very optional though.

packages/core/src/math/spherical.ts
andy-sweet: > As long as you're working purely in spherical coordinates I don't think there's anything inherently unsafe about any values.

I'm jumping on this comment, but it was also unclear to me what safe/unsafe or correct/incorrect values are in this context since I couldn't naturally think of any divide-by-zero or similar issues.

aganders3: > `theta` naturally lies in [-pi/2, pi/2]

This is only true if dealing with small increments and ensuring `makeSafe` is called after each increment, so it depends on additional knowledge of how this class is supposed to be used. This does feel like it becomes closer to what @andy-sweet described in another comment as "spherical coordinates for orbit controls".

For example as a user I would not initially think to wrap the angles myself when constructing `Spherical` from some other calculation. If the input is some large theta it will just clamp to pi/2, which I would not expect.

shlomnissan: I understand. `makeSafe` is a common name in math libraries for spherical or polar coordinates, but I agree that this function isn’t really about “safety.” It’s more about moving the coordinates toward a canonical form, though the implementation only does that partially. Since my preference for `makeSafe` is mostly familiarity rather than clarity, I’ll rename it to `constrainElevation` for now, unless you can think of a clearer name.

packages/core/src/objects/cameras/orbit_controls.ts
shlomnissan: Renamed to `sphericalPos_`.

packages/core/src/math/spherical.ts
shlomnissan: Okay, I'm convinced. Ignore: https://github.com/chanzuckerberg/idetik/pull/500#discussion_r2445662401

I'm going to move clamping to orbit controls. If we need to constrain elevation elsewhere, we can bring it back.

packages/core/src/objects/cameras/orbit_controls.ts
shlomnissan: I inlined both `m` and `b` and updated the formatting.

packages/core/test/spherical.test.ts
shlomnissan: Now that I inlined `makeSafe` all related tests are deleted.

packages/core/src/math/spherical.ts
shlomnissan: I still think a generic spherical coordinates class should include a function to bring its angles into canonical form, but I agree that subtracting `eps` is specific to a particular use case. My bias here probably comes from Three.js’s implementation, even though this version wasn’t directly ported from it

PR #504, opened 2025-10-21
Make Idetik context required

andy-sweet:
### Summary
This change makes the Idetik runtime context required. This was previously blocked by the React component not using the runtime, but this is no longer the case. The only update needed is in the viewport test, which now uses a test `LayerManager` that is provided with the Idetik context.

### Tests & Checks
Any related failures should be caught at compile time. I also tested the React examples just in case there is non-compiled JS usage somewhere.


shlomnissan (approved):
LGTM 👍

Issue #486, opened 2025-10-07
Add viewport support with shared data sources


aganders3:
Currently, the chunk management system has viewport-specific state (LOD calculation, chunk visibility) tightly coupled with data source management (chunk pool, data loading) and layer state (slice coordinates). This architecture has limitations where:

1. **Multiple viewports need different LODs for the same data source** - The current `ChunkManagerSource` stores a single LOD value, but different viewports viewing the same data at different zoom levels need independent LOD calculations.

2. **Memory is wasted on duplicate chunks** - `ChunkManagerSource` is tightly coupled to a specific layer by its shared `sliceCoors` attribute, though it is only stored in the `ChunkManager` keyed by the source. This makes it impossible to share across layers with different slices, thus it's not currently possible to:
    1. share a _source_ across layers with different slices/views
    2. share _chunks_ (and therefore memory) across layers with different slices/views

3. **Chunk visibility is stored per-chunk, not per-viewport** - The current `Chunk` type has boolean `visible` and `prefetch` flags, but these should be viewport-specific (or be the result of an aggregation across viewports) when multiple viewports possibly render the same chunks.

4. **LOD calculation happens in `ChunkManagerSource`** - LOD should be viewport-driven since it depends on the camera's zoom level and viewport dimensions, both of which are owned by the viewport object.

#### Related links
[Multiple Viewports Tech Spec](https://docs.google.com/document/d/1OwGL7KW34hxFAENyOm2WiQ71E58XGu9Y7jKkJlyRx84/edit?tab=t.0#heading=h.iybqw6c78lgi)

aganders3:
`ChunkManagerSource` is currently tightly coupled to both the `Viewport` (camera, element dimensions), and the slice/orientation. These are separate but related problems, therefore I think a solution should be designed simultaneously for this issue and #481 (enabling axis-aligned slicing in different orientations).

The current implementation of `ChunkManagerSource` has barriers to:
1. Sharing data across viewports
2. Displaying different slice orientations (i.e. not just XY planes with slicing in Z)

These are clearly related when we want to display different slice orientations in different viewports, but I think the relationship goes beyond that because some parts of `ChunkManagerSource` rely on Z-slice and single-view assumptions simultaneously.

### 1. Barriers to Sharing Chunk Data Across Viewports

There's no separation between "chunk data storage" and "chunk view state", making it impossible to have multiple views of the same underlying data.

[...]

PR #506, opened 2025-10-22
Introduce multiple-viewport API (breaking changes)

aganders3:
### Summary

This introduces an API for instantiating a multi-viewport runtime. The bulk of the important changes are in `idetik.ts`. The rest of the changes (primarily in the examples) are pretty mechanical, but useful for seeing what the updated API looks like.

In the interest of keeping the configuration explicit, and validation simple, this is a _breaking change_, hence all the updated examples. In summary:
* When instantiating the runtime, you _must_ specify the `viewports` array
* For single-viewport applications, simply specify a single viewport in the array
* Each viewport in the config requires an HTML element that defines its bounds, if you omit this it will default to the base `canvas` element
  * In general this is what you want for a single-viewport app
  * Viewports cannot share elements, so for a multi-viewport case you must specify unique elements for all viewports
  * You may omit the explicit element (again, defaulting to the canvas) for _one_ viewport in a multi-viewport definition
* The following properties/methods of the `Idetik` object ("runtime") have been removed, and must be accessed via the viewport instead:
  * `camera`
  * `cameraControls`
  * `clientToClip`
  * `events`
  * `layerManager`
* You can access a viewport through the runtime object with:
  * `viewports` property, which returns a readonly array of `Viewport` objects
  * `getViewport(id: string): Viewport | undefined` if you have a viewport ID
  * In the common case of a single-viewport app, use `runtime.viewports[0]` to get the viewport
  
I also included a comprehensive JSDoc comment for the `Idetik` constructor to document how it works. Hopefully this is helpful as we look into generating API docs.

Finally, this includes a new example demonstrating multiple viewports in `packages/core/examples/multi_viewport`. Because the `ChunkManager` and associated machinery is not compatible with multiple viewports, I used a `VolumeLayer` and `PerspectiveCamera` in the second viewport as a temporary workaround. This example can serve as a testing rig as we add multi-viewport capability to the chunking infrastructure.
Screenshot 2025-10-22 at 10 21 33 AM

### Related Issue
#486 (does not close)

### Tests & Checks
Tested existing examples and updated automated tests
Included a new example demonstrating multiple viewports, with the limitation that only one can use the `ChunkManager`

N.B. I did not update the React code here in anticipation of #499 merging soon. We will need to update the react library accordingly when this is published. Tests on this branch will probably fail until incorporating the react package removal.

### DIFF
--- a/packages/core/examples/image2d_from_omezarr4d_hcs/main.ts
+++ b/packages/core/examples/image2d_from_omezarr4d_hcs/main.ts
@@ -34,13 +34,12 @@ wellPaths.forEach((path) => {
 });
 
 const camera = new OrthographicCamera(0, 840, 0, 360);
-const controls = new PanZoomControls(camera);
 const app = new Idetik({
   canvas: document.querySelector<HTMLCanvasElement>("canvas")!,
   viewports: [
     {
       camera,
-      cameraControls: controls,
+      cameraControls: new PanZoomControls(camera),
     },
   ],
 }).start();
--- a/packages/core/examples/image2d_from_omezarr4d_hcs/main.ts
+++ b/packages/core/examples/image2d_from_omezarr4d_hcs/main.ts
@@ -37,9 +37,15 @@ const camera = new OrthographicCamera(0, 840, 0, 360);
 const controls = new PanZoomControls(camera);
 const app = new Idetik({
   canvas: document.querySelector<HTMLCanvasElement>("canvas")!,
-  camera,
-  cameraControls: controls,
+  viewports: [
+    {
+      camera,
+      cameraControls: controls,
+    },
+  ],
 }).start();
+
+const viewport = app.viewports[0];
 
 const region: Region = [
   { dimension: "T", index: { type: "point", value: 0 } },
@@ -53,7 +59,7 @@ const imageSelector = document.querySelector("#image") as HTMLSelectElement;
 
 const onImageChange = async () => {
   console.debug("onImageChange: ", imageSelector.value);
-  app.layerManager.removeAll();
+  viewport.layerManager.removeAll();
   const imageUrl =
     plateUrl + "/" + wellSelector.value + "/" + imageSelector.value;
   const source = new OmeZarrImageSource(imageUrl);
@@ -80,7 +86,7 @@ const onImageChange = async () => {
     ],
     lod: 0,
   });
-  app.layerManager.add(newLayer);
+  viewport.layerManager.add(newLayer);
   newLayer.addStateChangeCallback((state) => {
     if (state === "ready" && newLayer.extent) {
       camera.setFrame(0, newLayer.extent.x, 0, newLayer.extent.y);
@@ -91,7 +97,7 @@ const onImageChange = async () => {
 
 const onWellChange = async () => {
   console.debug("onWellChange: ", wellSelector.value);
-  app.layerManager.removeAll();
+  viewport.layerManager.removeAll();
   const path = wellSelector.value;
   const well = await loadOmeZarrWell(plateUrl, path);
   console.debug("well", well);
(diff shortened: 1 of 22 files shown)


packages/core/examples/image2d_from_omezarr4d_hcs/main.ts
```diff
@@ -37,10 +37,16 @@ const camera = new OrthographicCamera(0, 840, 0, 360);
 const controls = new PanZoomControls(camera);
 const app = new Idetik({
   canvas: document.querySelector<HTMLCanvasElement>("canvas")!,
-  camera,
-  cameraControls: controls,
+  viewports: [
+    {
+      camera,
+      cameraControls: controls,
```
andy-sweet: Nit: I think it's also fine to inline the construction of controls here as in some of the other examples.

packages/core/src/core/viewport.ts
```diff
@@ -24,7 +25,10 @@ export class Viewport {
   public readonly events: EventDispatcher;
   public cameraControls?: CameraControls;
 
-  constructor(config: ViewportConfig, layerManager: LayerManager) {
+  constructor(
+    config: ViewportConfig & { element: HTMLElement },
```
andy-sweet: Optional: it might be worth defining this type with a different name since it is used in more than one place and I had to think about it a little when seeing it for the first time. Then you could also add the layer manager and distinguish between a user provided config and the arguments needed for construction.

Maybe something like `type ViewportProps = ViewportConfig & { element: HTMLElement, layerManager: LayerManager }`.

You could also clarify the name as `props` or `params` instead of config to help clarify (since this is more than a config).

packages/core/src/idetik.ts
```diff
@@ -44,42 +36,72 @@ export class Idetik {
   private readonly stats_?: Stats;
   private readonly sizeObserver_: PixelSizeObserver;
 
+  /**
```
andy-sweet, line 39: I still feel like it's a little early to add big doc comments to code, but I also don't mind it.

packages/core/src/idetik.ts
```diff
@@ -44,42 +36,72 @@ export class Idetik {
...
+   *   Each viewport renders with its own camera, layers, and controls.
+   *   The `element` property is optional and defaults to the canvas if not provided.
+   *   Elements must be unique across viewports.
+   *   The `id` property is optional but useful for referencing specific viewports later.
+   * @param params.overlays - Optional array of overlay objects that update each frame (e.g., for HUD elements)
+   * @param params.showStats - Optional flag to display performance statistics
+   *
+   * @example
+   * // Single viewport (element defaults to canvas)
+   *  const camera = new OrthographicCamera(0, 1024, 0, 1024);
```
```suggestion
   * const camera = new OrthographicCamera(0, 1024, 0, 1024);
```

packages/core/test/viewport.test.ts
```diff
@@ -19,6 +20,10 @@ function createTestCamera(): OrthographicCamera {
   return new OrthographicCamera(-1, 1, -1, 1);
 }
 
+function createTestContext(): IdetikContext {
+  return { chunkManager: new ChunkManager() };
+}
+
 function createTestLayerManager(): LayerManager {
   const chunkManager = new ChunkManager();
   return new LayerManager({ chunkManager });
```
andy-sweet:
```suggestion
return new LayerManager(createTestContext());
```

andy-sweet (approved):
This looks big, but isn't really. All my comments are nits/optionals. New example is great!

packages/core/src/idetik.ts
aganders3: Yeah it's more of a trial than anything as I'm totally unfamiliar with JSDoc. We still need to look into related tooling, etc. I just figured if any function should be documented it's this one.

PR #511, opened 2025-10-27
Allow OME-Zarr version to be specified on construction of image source

andy-sweet:
### Summary
This change allows the OME-Zarr version to be explicitly specified when constructing the OME-Zarr source. While we can rely on the current guesswork within zarrita to eventually successfully open different OME-Zarr versions, doing so currently causes failed requests which triggers multiple errors to be logged. I've also seen outright failures (e.g. on some new OrganelleBox visualization data), though that may represent an issue with those data instead of the current guesswork logic.

This is a breaking change because given the overloaded constructor, the version must come second.

### Tests & Checks
I manually tested the updated example where the OME-Zarr version is now specified.


andy-sweet:
I'm tempted to change the way these sources are constructed by using one props object and/or defining factory methods/functions.

Before marking this ready for review, I should probably update our metadata reading functions to also allow a version (OME-Zarr or Zarr) to be specified too.

andy-sweet:
@aganders3 : I assigned you for review, because I remember you making some comments about the versioning here before. The current solution is a little rough (partly because the OmeZarrImageSource constructor overloading but also because the metadata utilities are a little silly), so I'd appreciate any better ideas you had here.

aganders3 (approved):
I think this is fine. My only ideas would be bigger changes, so I'd just stick with what you have here as it's minimal and clear.

This repeated pattern is unfortunate. A simple "fix" would be to allow `omeZarrToZarrVersion` to take/return `undefined`, but I think that becomes more confusing, not less.
```
  const zarrVersion = version ? omeZarrToZarrVersion(version) : undefined;
  const group = await openGroup(location, zarrVersion);
```

Looking at the kind of odd assortment of utilities here it makes me feel like a lot of this belongs outside Idetik, but obviously that's not a change for today. It would be nice to perhaps separate some of the logic for guessing the version from opening a group/array, but that may not be entirely possible. This prompted me to look into how zarrita guesses the version and it's kind of weird.

andy-sweet:
> A simple "fix" would be to allow omeZarrToZarrVersion to take/return undefined, but I think that becomes more confusing, not less.

I actually don't mind this. I might bring it in here.

andy-sweet:
> Looking at the kind of odd assortment of utilities here it makes me feel like a lot of this belongs outside Idetik, but obviously that's not a change for today.

I think some of them belong outside idetik (e.g. hcs metadata), but some of them should be better internalized (e.g. omero visualization metadata).

andy-sweet:
> This prompted me to look into how zarrita guesses the version and it's kind of weird.

Yes, this was another motivation to open this. And why I only want to rely on that as a last resort.

PR #512, opened 2025-10-27
Export ImageSourcePolicy type

### LINKED ISSUE #510: Explicitly export ImageSourcePolicy
We currently export `ImageSourcePolicyConfig` and the policy utility functions, but not the `ImageSourcePolicy` type itself.

We wanted to allow passing a policy through in the React component, and mostly wanted to use the utility functions, but then typing the parameters required using `ReturnType<createImageSourcePolicy>` which works, but is a little awkward.

andy-sweet:
### Summary
This change exports `ImageSourcePolicy` as a type, which allows dependents to directly import and use that as a type instead of relying on things like `ReturnType<typeof createImageSourcePolicy>`.

### Related Issue
Closes #510

### Tests & Checks
This only exports an additional type. No manual testing needed.

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -28,6 +28,7 @@ export { OmeZarrImageSource } from "./data/ome_zarr/image_source";
 
 export type {
   PriorityCategory,
+  ImageSourcePolicy,
   ImageSourcePolicyConfig,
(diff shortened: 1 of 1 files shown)


shlomnissan (approved):
:+1:

PR #513, opened 2025-10-27
Add bounding boxes to renderable objects

shlomnissan:
### Summary

This PR adds bounding boxes to renderable objects. These will be used for
frustum-culling debugging or any renderer-level culling we may need, though the main
path is more likely to generate boxes directly from chunks. The computation is lazy,
so it introduces no extra perf cost, but we can roll some of this work back if it's no
longer useful beyond debugging.

### Tests & Checks
- All checks have passed
- Added basic tests for new `Box3` functionality

### DIFF
--- a/packages/core/src/core/geometry.ts
+++ b/packages/core/src/core/geometry.ts
@@ -1,4 +1,6 @@
 import { Node } from "./node";
+import { Box3 } from "../math/box3";
+import { vec3 } from "gl-matrix";
 
 export type Primitive = "triangles" | "points" | "lines";
 
@@ -34,6 +36,7 @@ type GeometryAttribute = {
 };
 
 export class Geometry extends Node {
+  private boundingBox_: Box3 | null = null;
   protected primitive_: Primitive;
   protected attributes_: GeometryAttribute[];
   protected vertexData_: Float32Array;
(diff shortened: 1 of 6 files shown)


packages/core/src/core/geometry.ts
```diff
@@ -53,18 +56,25 @@ export class Geometry extends Node {
 
   public addAttribute(attr: GeometryAttribute) {
     this.attributes_.push(attr);
+    this.boundingBox_ = null;
+  }
+
+  public getAttribute(type: GeometryAttributeType) {
```
andy-sweet, line 62: Do you anticipate this being used externally? Otherwise, I'd prefer this to be private and even inlined to its current single usage.

I'm generally a bit hazy about how we handle geometry attributes, but even `addAttribute` is a unexpected, as it seems bound to `vertexData_` and so the attributes feel like they should be fixed rather than mutable. Definitely outside the scope of this PR though!

packages/core/src/math/box3.ts
```diff
@@ -38,4 +38,37 @@ export class Box3 {
...
+    if (p[1] < this.min[1]) this.min[1] = p[1];
+    if (p[2] < this.min[2]) this.min[2] = p[2];
+    if (p[0] > this.max[0]) this.max[0] = p[0];
+    if (p[1] > this.max[1]) this.max[1] = p[1];
+    if (p[2] > this.max[2]) this.max[2] = p[2];
+  }
+
+  public applyTransform(matrix: mat4) {
+    const { min, max } = this;
+    const corners: vec3[] = [
```
andy-sweet, line 53: This seems fine for now, but if we expect this to be called frequently outside of debugging, I wonder if we should consider an oriented box instead of an axis-aligned one. Or some other way of optimizing this path.

andy-sweet (approved):
Good to me for short term debugging use. For longer term / critical path use, there may be some optimizations worth pursuing, but can consider those when needed.

packages/core/src/core/geometry.ts
shlomnissan: That's a really good point!!! It probably makes more sense to pass a set of immutable attributes when the geometry is initialized or its underlying data is updated. I'm going to inline `getAttribute` for now.

packages/core/src/math/box3.ts
shlomnissan: AABB is faster for frustum culling checks which is the main use-case for now and the corner expansion method is common in rendering engines for box primitives.

I do expect it to be called frequently in the frustum culling PR. We can optimize this path by caching the bounding box at the renderable object and only computing it if the renderable's transform is dirty.

I'll measure and consider the tradeoffs in the next PR.

PR #514, opened 2025-10-28
[WIP] Refactoring to share chunk data across multiple viewports

aganders3:
### Summary
This PR introduces the ChunkStore/ChunkStoreView architecture to enable efficient chunk sharing across multiple viewports. Previously, `ChunkManagerSource` conflated two responsibilities: managing the global chunk data store and determining viewport-specific chunk visibility. This design prevented resource sharing when multiple viewports rendered the same data source from different perspectives, leading to memory leaks and duplicate chunk loading.

The implementation splits `ChunkManagerSource` into two complementary classes:
  - `ChunkStore`: Manages the canonical data store and global state for all chunks for a single source
  - `ChunkStoreView`: Represents a layer/viewport-specific view into that store, calculating LOD, visibility, and prefetch requirements based on its camera and slice position

Multiple layers (including in separate viewports) can now share the same `ChunkStore` while each maintains its own `ChunkStoreView`. The `ChunkStore` aggregates chunk requirements across all views: chunks are kept loaded if any view needs them, and disposed when no views reference them. Priority for loading is determined by taking the minimum priority across all views.

`ChunkStoreView` maintains an efficient sparse `Map<Chunk, ChunkViewState>` rather than tracking all chunks. View state aggregation happens synchronously during the update cycle, with `forgetChunk()` called inline to clean up no-longer-visible chunks within a view.

### Related Issue
For detailed architecture diagrams and class relationships, see the [multiple viewports tech spec](https://docs.google.com/document/d/1OwGL7KW34hxFAENyOm2WiQ71E58XGu9Y7jKkJlyRx84/edit?tab=t.0).

This is the final PR of the multi-viewport implementation plan (see tech spec):
  - #407: WebGL infrastructure
  - #419: Viewport class
  - #451: Viewport events
  - #506: Idetik API integration
  - #514 (this PR): `ChunkManagerSource` refactor

### Tests & Checks
  - Existing chunk management tests updated to use new ChunkStore/ChunkStoreView API
  - All tests pass
  - Existing examples work without modification
  - Added a second `ChunkedImageLayer` to the multi-viewport example and confirmed it behaves as expected, with chunk memory sharing confirmed (according to stats.js)
  - Validated that chunks are properly disposed when no views need them (memory goes back down)


packages/core/src/core/chunk_store.ts
```diff
@@ -0,0 +1,266 @@
...
+  public removeView(view: ChunkStoreView): void {
+    const affectedChunks = Array.from(view.chunkViewStates.keys());
+    this.views_.delete(view);
+
+    for (const chunk of affectedChunks) {
+      this.aggregateChunkViewStates(chunk);
+    }
+  }
+
+  public get views(): ReadonlySet<ChunkStoreView> {
```
andy-sweet: This isn't used externally. Do you expect it to be in the future or could it be removed?

aganders3: I think I had this public for use in an overlay, but it can probably be removed at least for now.

aganders3: (same with `ChunkStoreView.viewport` I think)

packages/core/src/core/chunk_store_view.ts
```diff
@@ -0,0 +1,451 @@
...
+    const currentTimeChunks = this.store_.getChunksAtTime(currentTimeIndex);
+    const currentLODChunks = currentTimeChunks.filter(
+      (chunk) =>
+        chunk.lod === this.currentLOD_ &&
+        this.chunkViewStates_.get(chunk)?.visible === true &&
+        chunk.state === "loaded"
+    );
+
+    // If we're at the lowest resolution LOD, only return current LOD chunks
+    const lowestResLOD = this.store_.getLowestResLOD();
```
andy-sweet, line 71: This reflects the logic on main.
But as the policy allows a specification of the min and max LOD, I wonder if this would be a good place to reflect that so that each view could have an effective minimum LOD instead of sharing one across the whole store that is always the same for a source/store. That would fix a maybe-bug (at the least, undesirable behavior) on main even in the single layer case.

packages/core/src/core/chunk_store_view.ts
```diff
@@ -0,0 +1,451 @@
...
+
+    const orthoCamera = camera as OrthographicCamera;
+    const viewBounds2D = orthoCamera.getWorldViewRect();
+    const virtualWidth = Math.abs(viewBounds2D.max[0] - viewBounds2D.min[0]);
+    const canvasElement = this.viewport_.element as HTMLCanvasElement;
+    const bufferWidth = this.viewport_
+      .getBoxRelativeTo(canvasElement)
+      .toRect().width;
+    const virtualUnitsPerScreenPixel = virtualWidth / bufferWidth;
+    const lodFactor = Math.log2(1 / virtualUnitsPerScreenPixel);
```
andy-sweet: Some large part of this feels like it should be a method on the viewport because it's using the viewport's camera and element to perform these calculations.

This may also allow multiple layers/views within the same viewport to reuse the same calculation and may also allow skipping calculations more easily (currently each layer calls this on every tick once the store/view is available).

packages/core/src/core/chunk_store_view.ts
```diff
@@ -0,0 +1,451 @@
...
+      this.lastZBounds_ = zBounds;
+      this.lastTCoord_ = sliceCoords.t;
+    }
+  }
+
+  public get currentLOD(): number {
+    return this.currentLOD_;
+  }
+
+  public forgetChunk(chunk: Chunk): void {
```
andy-sweet: Nit: I'd prefer a name like `maybeForgetChunk` since the forgetting is conditional.

packages/core/src/layers/chunked_image_layer.ts
```diff
@@ -38,7 +36,8 @@ export class ChunkedImageLayer extends Layer implements ChannelsEnabled {
   private readonly channelChangeCallbacks_: (() => void)[] = [];
   private policy_: ImageSourcePolicy;
   private channelProps_?: ChannelProps[];
-  private chunkManagerSource_?: ChunkManagerSource;
+  private chunkStore_?: ChunkStore;
```
andy-sweet: I had not expected the layer would need to store both the ChunkStore and ChunkStoreView, but I see why it is needed (because of the current idetik context attach/detach, and more specifically because of the need to call `ChunkStore.removeView`).

Actually, there are many other uses of `ChunkStore` here too, though maybe some of those could be moved to `ChunkStoreView`.

packages/core/src/layers/chunked_image_layer.ts
```diff
@@ -73,50 +72,62 @@ export class ChunkedImageLayer extends Layer implements ChannelsEnabled {
...
-      this.sliceCoords_,
-      this.policy_
-    );
+    this.chunkStore_ = await context.chunkManager.addSource(this.source_);
   }
 
   public onDetached(): void {
-    this.chunkManagerSource_ = undefined;
-    this.releaseAndRemoveChunks(this.visibleChunks_.keys());
-    this.clearObjects();
```
andy-sweet, line 92: Is this an accidental removal? Or is there a reason these would not be needed anymore?

packages/core/src/idetik.ts
```diff
@@ -158,13 +157,10 @@ export class Idetik {
       this.updateSize();
     }
 
+    // Update chunk manager before rendering to process all views
```
andy-sweet, line 160: I haven't thought this through yet, but doesn't this represent a change to the order of the main idetik loop compared to main because now the chunk manager updates the chunk queue before the layer/view is updated with the current value of sliceCoords, LOD, and camera to determine what should be loaded?

This might not have a huge practical impact (because the next tick will rectify it), but feels slightly off to me (i.e. we're using stale information to make a determination about what should be loaded or cancelled).

packages/core/src/layers/chunked_image_layer.ts
aganders3: Oh thanks - it was accidental and I even thought I had caught it before! I'll add it back.

packages/core/src/idetik.ts
aganders3: I think you're right this is a small change compared to main and doesn't quite sit well with me either, but I need to think about it. My first instinct is to make the `ChunkManager.update` more event-driven, but I'm worried that brings other complications.

packages/core/src/layers/chunked_image_layer.ts
aganders3: I agree and would ideally avoid it. I need to think if there's a better way to create/remove the view.

I think you're right the other uses could move to `ChunkStoreView`, though. In fact this might represent a bug because `allVisibleLowestLODLoaded` will wait for chunks to load across any views before returning true.

packages/core/src/core/chunk_store_view.ts
aganders3: I think you're probably right, and moving the initial `lodFactor` calculation at least should be quite simple. I'm curious how this might change with updates for 3D LOD and visibility, especially if a viewport mixes 2D and 3D objects.

packages/core/src/idetik.ts
andy-sweet: A very simple "fix" would be to move the chunk manager update after rendering the layers. That's still a change from main, but allows the view to make updates to the chunk view state before the chunk manager does anything with that information (e.g. enqueues chunks to be loaded, cancels pending or running loads), which is my main concern here (i.e. we should only load what we need).

packages/core/src/layers/chunked_image_layer.ts
```diff
@@ -73,45 +72,56 @@ export class ChunkedImageLayer extends Layer implements ChannelsEnabled {
...
+      !this.chunkStore_.allVisibleLowestLODLoaded(currentTimeIndex) &&
       !this.isPresentationStale()
     ) {
       return;
     }
     this.lastPresentationTimeStamp_ = performance.now();
     this.lastPresentationTimeCoord_ = this.sliceCoords_.t;
 
-    const orderedByLOD = this.chunkManagerSource_.getChunks();
+    const orderedByLOD = this.chunkStoreView_.getChunksToRender(
```
andy-sweet, line 116: I like the small name change from `getChunks` to `getChunksToRender` - it helps a lot with communicating expectations of that method!

packages/core/src/layers/chunked_image_layer.ts
```diff
@@ -157,6 +167,16 @@ export class ChunkedImageLayer extends Layer implements ChannelsEnabled {
     this.zPrevPointWorld_ = zPointWorld;
   }
 
+  private releaseAndRemoveChunks(chunks: Iterable<Chunk>): void {
```
andy-sweet: nit: maybe move this back to its original location to reduce the diff

packages/core/src/core/chunk_store.ts
```diff
@@ -0,0 +1,262 @@
...
+  private readonly dimensions_: SourceDimensionMap;
+  private readonly views_: Set<ChunkStoreView> = new Set();
+
+  constructor(loader: ChunkLoader) {
+    this.loader_ = loader;
+    this.dimensions_ = this.loader_.getSourceDimensionMap();
+    this.lowestResLOD_ = this.dimensions_.numLods - 1;
+
+    this.validateXYScaleRatios();
+    const { size: chunksT } = this.getAndValidateTimeDimension();
```
andy-sweet, line 23: On main, there's a similar validation for the channel dimension. Did you deliberately remove that here? Or is this PR just a little behind main?

aganders3: It's just behind main because I've been working on various iterations for a while and rebasing.
Thanks for calling these out!

packages/core/src/core/chunk_manager.ts
```diff
@@ -25,41 +19,22 @@ export class ChunkManager {
...
-      );
-    }
-
-    const viewBounds2D = camera.getWorldViewRect();
-    const virtualWidth = Math.abs(viewBounds2D.max[0] - viewBounds2D.min[0]);
-    const virtualUnitsPerScreenPixel = virtualWidth / bufferWidth;
-    const lodFactor = Math.log2(1 / virtualUnitsPerScreenPixel);
-
+  public update() {
     for (const [_, source] of this.sources_) {
```
andy-sweet: Nit: the unpacked argument should be store, not source. I would also expect `ChunkManager.sources_` to be named `ChunkManager.stores_` (i.e. with the value determining the name and not the key), but that is ambiguous on main (because of ChunkSource and ChunkManagerSource), so I don't feel as strongly about that.

packages/core/src/core/chunk_manager.ts
```diff
@@ -1,71 +1,45 @@
...
+import { ChunkSource } from "../data/chunk";
 import { ChunkQueue } from "../data/chunk_queue";
-import { ChunkManagerSource } from "./chunk_manager_source";
-import { ImageSourcePolicy } from "./image_source_policy";
+import { ChunkStore } from "./chunk_store";
 
 export class ChunkManager {
-  private readonly sources_ = new Map<ChunkSource, ChunkManagerSource>();
+  private readonly stores_ = new Map<ChunkSource, ChunkStore>();
   private readonly pendingSources_ = new Map<
```
andy-sweet: nit: you're going to hate me, but as you changed `sources_` to `stores_` this should be `pendingStores_` for consistency...

packages/core/src/core/chunk_store_view.ts
```diff
@@ -0,0 +1,451 @@
...
+    const bufferWidth = this.viewport_
+      .getBoxRelativeTo(canvasElement)
+      .toRect().width;
+    const virtualUnitsPerScreenPixel = virtualWidth / bufferWidth;
+    const lodFactor = Math.log2(1 / virtualUnitsPerScreenPixel);
+
+    this.setLOD(lodFactor);
+
+    const zBounds = this.getZBounds(sliceCoords);
+    const changed =
```
andy-sweet, line 109: Shouldn't `changed` also take into account if the current LOD changed? This is consistent with main, so no need to make the change here, but just commenting so I don't forget.

packages/core/src/core/chunk_store_view.ts
andy-sweet: Another related thought is that `ChunkStoreView.viewport_` is only used in this function, which makes me think it should be passed through as a parameter here instead of being a class property.

packages/core/src/core/chunk_store_view.ts
```diff
@@ -0,0 +1,451 @@
...
+        viewBoundsCenter2D
+      );
+    }
+  }
+
+  private updateStaleTimeChunks(currentTimeIndex: number): void {
+    for (const t of this.tIndicesWithQueuedChunks_) {
+      const delta = t - currentTimeIndex;
+      if (delta >= 0 && delta <= this.policy_.prefetch.t) continue;
+      this.tIndicesWithQueuedChunks_.delete(t);
```
andy-sweet: Looking at main, I would have thought that you'd need to call `this.chunkViewStates_.set(chunk, {...})` here. But maybe I missed how the new logic works to ensure cancellation and/or disposal in this case.

packages/core/src/layers/chunked_image_layer.ts
aganders3: I refactored a bit in [edac65e](https://github.com/chanzuckerberg/idetik/pull/514/commits/edac65e30321971b40d04e43b6d7d2fb530bdedd) such that now the only need for `ChunkStore` is to add/remove the view.

I think this is also more correct wrt `allVisibleLowestLODLoaded` as mentioned, and would fit well with your comment above to use the policy to constrain LOD per-view.

packages/core/src/core/chunk_store_view.ts
andy-sweet: Oh right, this is where `markUnused` comes in.

packages/core/src/core/chunk_store.ts
```diff
@@ -0,0 +1,293 @@
...
+      for (const [chunk, _viewState] of view.chunkViewStates) {
+        affectedChunks.add(chunk);
+      }
+    }
+
+    for (const chunk of affectedChunks) {
+      this.aggregateChunkViewStates(chunk);
+    }
+
+    return Array.from(affectedChunks);
```
andy-sweet: nit: chunk manager is the only caller of this function and I don't think it cares about the order the chunks, so you could just return the set here (which i think is in the same order anyway?).

packages/core/src/core/chunk_store_view.ts
aganders3: > it should be passed through as a parameter here

I think I like this. I was feeling a bit off about viewport being a property but sliceCoords not. I think this also clarifies that the `ChunkStoreView` is more per-layer than per-viewport.

packages/core/src/core/chunk_store.ts
```diff
@@ -0,0 +1,282 @@
...
+
+  public getLowestResLOD(): number {
+    return this.lowestResLOD_;
+  }
+
+  public loadChunkData(chunk: Chunk, signal: AbortSignal) {
+    return this.loader_.loadChunkData(chunk, signal);
+  }
+
+  public disposeChunk(chunk: Chunk) {
```
andy-sweet, line 115: This is only used internally, so should be private?

packages/core/src/core/chunk_store_view.ts
andy-sweet: I want to confirm my understanding here since this logic was not clear on main, and may be even less clear on this branch.

The layer calls `ChunkStoreView.updateChunkViewStates`, which will mark any previously prefetched/queued chunk views that are now temporally stale with priority null (via `markUnused`).

The chunk manager calls `ChunkStore.updateAndCollectChunkChanges` which then calls `ChunkStore.aggregateChunkViewStates` with each chunk view that each chunk store view knows about (including the now temporally stale ones).

`ChunkStore.aggregateChunkViewState` calls `ChunkStoreView.maybeForgetChunk` with unused chunks (including the now temporally stale ones). It finally calls `ChunkStore.disposeChunk` on any loaded chunk with null priority (maybe including some now temporally stale ones).

packages/core/src/core/chunk_store_view.ts
```diff
@@ -0,0 +1,463 @@
...
+        viewBounds3D,
+        viewBoundsCenter2D
+      );
+    }
+  }
+
+  private updateStaleTimeChunks(currentTimeIndex: number): void {
+    for (const t of this.tIndicesWithQueuedChunks_) {
+      const delta = t - currentTimeIndex;
+      if (delta >= 0 && delta <= this.policy_.prefetch.t) continue;
```
andy-sweet: Since this function does not change the chunk views, I think it has no effect other than mutating `tIndicesWithQueuedChunks`, which I think does nothing since it's only read here.

I think that means we could remove `tIndicesWithQueuedChunks`, which is great!

But I'm trying to understand how we now capture the logic of not disposing/forgetting loaded chunks at time points that are still in prefetch range. Ohhh - because `markTimeChunksForPrefetch` is called which updates the chunk views before anything is disposed.

So let's remove `tIndicesWithQueuedChunks` and this method? That's a nice improvement from main.

andy-sweet (commented):
I'm still working through this. My thoughts so far are mixed, but leaning towards positive. I think some changes here improve the clarity of loading over main, while handling more complex usage.

But other changes make some of the flow hard to follow and introduce more dependencies between classes that can be hard to follow and might lead to unexpected behavior. In particular, the dependency cycle between `ChunkStore` and `ChunkStoreView` seems like a code smell. However, I don't have suggested alternatives yet, so don't take this feedback too strongly.

packages/core/src/core/chunk_manager.ts
aganders3: Haha well at least now I also caught that the closure here should be `initializeStore` before pushing this fix.

aganders3:
> In particular, the dependency cycle between ChunkStore and ChunkStoreView seems like a code smell.

Thanks - this part also stuck out to me, especially as I took care to avoid it in the `Viewport`/`LayerManager` relationship.

I do have one other idea now inspired by your comments here, which is to give more responsibility to the `ChunkManager`. Specifically, make it a factory for `ChunkStoreView` such that the layer does not need to keep a reference to a `ChunkStore`, and a `ChunkStore` would not need to keep track of its own `ChunkStoreView` objects. `ChunkManager` would then manage both `ChunkStore` and `ChunkStoreView` instances. This is possible if we make `viewport` a parameter for `ChunkViewStore.updateChunkViewStates(viewport, sliceCoords)` as you suggested.

I will probably try this today to see if I'm missing anything, or see if it is somehow worse.

aganders3:
@andy-sweet PTAL at the changes in [8607e1b](https://github.com/chanzuckerberg/idetik/pull/514/commits/8607e1bc7b9f032b521bf3c76a6f8e6f9ecd6d2b) when you get a chance. I think it's an improvement!

With these changes:
  - `ChunkStore` is closer to a pure data container
  - `ChunkStoreView` computes what's needed for an individual view
  - `ChunkManager` coordinates stores and aggregates across views

packages/core/src/layers/chunked_image_layer.ts
andy-sweet: I like the new changes.

However, I think the only need for the `ChunkStore` reference in the layer is for `ChunkInfoOverlay`. Adding/removing the view is done with the `ChunkSource` and `ChunkManager`.

As you also removed the circular dependency between store and view, I think it would fine to expose `ChunkStoreView.store` as a public property. That way you don't need to explicitly store the store here.

packages/core/src/core/chunk_manager.ts
```diff
@@ -1,76 +1,148 @@
...
+  private readonly pendingStores_ = new Map<ChunkSource, Promise<ChunkStore>>();
+  private readonly views_ = new Map<ChunkStore, ChunkStoreView[]>();
   private readonly queue_ = new ChunkQueue();
 
-  public async addSource(
-    source: ChunkSource,
-    sliceCoords: SliceCoordinates,
-    policy: ImageSourcePolicy
-  ) {
+  public async addSource(source: ChunkSource): Promise<ChunkStore> {
```
andy-sweet: This can and probably should be private now.

packages/core/src/core/chunk_store_view.ts
aganders3: Nice catch - I ported this over pretty mechanically so it's good to remove!

packages/core/src/core/chunk_manager.ts
```diff
@@ -1,76 +1,148 @@
...
+    source: ChunkSource,
+    policy: ImageSourcePolicy
+  ): Promise<ChunkStoreView> {
+    const store = await this.addSource(source);
+    const view = new ChunkStoreView(store, policy);
+    this.views_.set(store, (this.views_.get(store) ?? []).concat(view));
+    return view;
+  }
+
+  public removeView(source: ChunkSource, view: ChunkStoreView): void {
```
andy-sweet: A little minor, but do you need to pass source here?

I know you can and it makes the implementation of this function easier. But I don't think it would too hard to find the store, then source associated with the view, in which case you simplify the function signature nicely.

andy-sweet:
> @andy-sweet PTAL at the changes in [8607e1b](https://github.com/chanzuckerberg/idetik/pull/514/commits/8607e1bc7b9f032b521bf3c76a6f8e6f9ecd6d2b) when you get a chance. I think it's an improvement!
> 
> With these changes:
> 
> * `ChunkStore` is closer to a pure data container
> * `ChunkStoreView` computes what's needed for an individual view
> * `ChunkManager` coordinates stores and aggregates across views

Agreed this is a general improvement and this change set looks more like what I had imagined, but with all the nasty details worked out. I especially like that `ChunkManager` acts a coordinator now because that ties that logic closer to chunk state updates and the chunk queue.

I left a few comments, which are probably worth addressing.

But I'm going to take a break from this to wait for other high level feedback.

aganders3:
Closing in favor of rebased #531

PR #515, opened 2025-10-28
WIP: develop style branch for volume rendering changes

seankmartin:
### Summary

This is to help for an overall idea of where volume rendering is at, PRs would come in smaller feature branches.


shlomnissan (changes requested):
This PR looks good overall. I'm only blocking because I think the cull face mode belongs on the renderable object and not the layer. It’s not a big deal to keep it on the layer for now if there’s a specific reason for that choice?

shlomnissan:
Sorry this review was for #526.

PR #516, opened 2025-10-28
feat: add Texture3D class and use in webgl renderer

seankmartin:
### Summary

These changes allow the creation of a 3D texture object as a subclass of `Texture`, and implements checks in the webgl renderer to use the 3D texture. The 3D texture handling currently shares the same path in the code as the texture 2d array in the renderer (except for getting the texture type).

### Tests & Checks

Manually tested changes from an example in the [draft volume rendering branch](https://github.com/chanzuckerberg/idetik/pull/515/files#diff-066e85ab63c16922810b3437fea8bddff3b60147dda5b1443d6811b983e5e394)

This is using the example dataset for the chunk streaming
rendering3d

### DIFF
--- a/packages/core/src/objects/textures/texture_3d.ts
+++ b/packages/core/src/objects/textures/texture_3d.ts
@@ -1,0 +1,95 @@
+import {
+  DataTextureTypedArray,
+  Texture,
+  bufferToDataType,
+} from "./texture";
+import { Chunk, ChunkData } from "../../data/chunk";
+
+export class Texture3D extends Texture {
+  private data_: DataTextureTypedArray;
+  private readonly width_: number;
+  private readonly height_: number;
+  private readonly depth_: number;
+
+  constructor(
+    data: DataTextureTypedArray,
+    width: number,
+    height: number,
+    depth: number
+  ) {
+    super();
+    this.dataFormat = "scalar";
+    this.dataType = bufferToDataType(data);
+
+    this.data_ = data;
+    this.width_ = width;
+    this.height_ = height;
+    this.depth_ = depth;
+  }
+
+  public set data(data: DataTextureTypedArray) {
+    this.data_ = data;
+    this.needsUpdate = true;
+  }
+
+  public get type() {
+    return "Texture3D";
+  }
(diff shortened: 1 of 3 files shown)


packages/core/src/objects/textures/texture_3d.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  public get height() {
+    return this.height_;
+  }
+
+  public get depth() {
+    return this.depth_;
+  }
+
+  public updateWithChunk(chunk: Chunk, data?: ChunkData) {
```
andy-sweet: Recommended

I think you could simplify the signatures of this method and `createWithChunk` to just take a chunk. I believe the optional `data` argument is only needed in the `ChunkedImageLayer` to handle slicing in Z on the CPU, which should not be needed for this 3D texture or an initial pass of volume rendering (as I understand it).

If it is needed later in some form, we can add it back in.

packages/core/src/renderers/webgl_textures.ts
andy-sweet: Blocking

I think this needs to be updated to handle `TEXTURE_3D` as well. I don't mind if you do that inline or define a helper function like `isTextureStorage3D` or `textureHasDepth` (better names are welcome!)

andy-sweet (changes requested):
Generally looks good, but I am requesting changes since I think `WebGLTextures.computeStorageBytes` needs to take into account the depth of 3D textures.

packages/core/src/objects/textures/texture_3d.ts
seankmartin: Thanks Andy, sounds good!

packages/core/src/renderers/webgl_textures.ts
seankmartin: Thank you for spotting this! Introduced `isTextureStorage3D` helper

andy-sweet (approved):
Looks good to me now! You may want to wait for @shlomnissan to review.

shlomnissan (approved):
LGTM :+1:

PR #518, opened 2025-10-28
Remove row stride from chunk and texture

### LINKED ISSUE #517: Remove row stride from chunk
Currently, each chunk has the `rowStride` property. This was added because zarrita chunks have a stride property.

Chunk data is currently 4D, so `rowStride` is insufficient to capture the general stride information that zarrita provides.

We are also inconsistent about using `rowStride` (e.g. when slicing through z).

As we add support for 3D rendering, we would need a similar property for height, which complicates things.

Therefore, we have decided to remove `rowStride` from `Chunk` and make any related simplifications to assume the data is compact.

andy-sweet:
### Summary
This change removes the `Chunk.rowStride` and the `Texture.unpackRowLength` properties. Instead, it validates that each zarr chunk is compact (i.e. the shape exactly specifies the stride) after loading.

### Related Issue
Closes #517

### Tests & Checks

### DIFF
--- a/packages/core/src/data/ome_zarr/image_loader.ts
+++ b/packages/core/src/data/ome_zarr/image_loader.ts
@@ -92,7 +92,7 @@ export class OmeZarrImageLoader {
       );
     }
 
-    validateCompactChunk(receivedChunk);
+    validateTightlyPackedChunk(receivedChunk);
 
     const receivedShape = {
       x: receivedChunk.shape[this.dimensions_.x.index],
@@ -195,7 +195,7 @@ export class OmeZarrImageLoader {
       );
     }
 
-    validateCompactChunk(subarray);
+    validateTightlyPackedChunk(subarray);
 
     if (subarray.shape.length !== 2 && subarray.shape.length !== 3) {
       throw new Error(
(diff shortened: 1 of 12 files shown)


andy-sweet:
@shlomnissan : assigned to you because this was partially motivated by volume rendering.

I considered also removing the unpack row alignment here too because we validate that the chunk is compact (so no need to ever skip bytes when unpacking in that case), but texture data does not have to come from chunks so I left it alone for now.

packages/core/src/data/ome_zarr/image_loader.ts
```diff
@@ -92,6 +92,8 @@ export class OmeZarrImageLoader {
       );
     }
 
+    validateCompactChunk(receivedChunk);
```
shlomnissan: Nit: `validateTightlyPackedChunk` might read a bit clearer. That could just be my familiarity but “tightly packed” tells me what the function checks.

shlomnissan (approved):
Looks good :+1:

packages/core/src/data/ome_zarr/image_loader.ts
andy-sweet: Ah, that is a more common term, so I'll make the change.

PR #519, opened 2025-10-29
Remove rowStride property from test chunk omissions

andy-sweet:
### Summary
This is a very tiny miss from PR #518. Since this change removes a property that is omitted in a type, it is not strictly needed, but helps to keep things as clean and concise as possible.

### Tests & Checks
This is test only code, so automated checks are sufficient.


shlomnissan (approved):
:+1:

PR #520, opened 2025-10-29
Prepare for prerelease version 7

andy-sweet:


andy-sweet:
@phoenixAja : assigned you for review as some recent changes to core would be useful for idetik react (e.g. exact specification of ome-zarr version).

phoenixAja (approved):

PR #521, opened 2025-10-29
Add frustum culling to the renderer

shlomnissan:
### Summary

This PR introduces a `Frustum` type that can be constructed from a projection
matrix and integrates frustum culling into the rendering pipeline. The
integration was primarily done to verify correctness, but I think it makes
sense to keep it as part of the pipeline.

### Note

In the next PR, I plan to integrate the `Frustum` into the `ChunkManagerSource`, replacing the current 2D view-bounds intersection logic. This will move us closer to a unified 2D/3D system that supports chunked images, orthoslices, and volume rendering.

### Tests & Checks

- All the checks have passed
- Added unit tests for `Plane` and `Frustum`
- Visual verification
  - I added a "Rendered Objects" count to the chunk info overlay
  - I manually verified the perspective camera case in 3D (see video below)

https://github.com/user-attachments/assets/02c859c1-30fd-4209-b6b6-b14ffa2959b3

### DIFF
--- a/packages/core/src/math/frustum.ts
+++ b/packages/core/src/math/frustum.ts
@@ -1,0 +1,75 @@
+import { Plane } from "./plane";
+import { Box3 } from "./box3";
+import { mat4, vec3 } from "gl-matrix";
+
+export class Frustum {
+  private readonly planes_: [Plane, Plane, Plane, Plane, Plane, Plane];
+
+  constructor(m: mat4) {
+    this.planes_ = [
+      new Plane(vec3.create(), 0),
+      new Plane(vec3.create(), 0),
+      new Plane(vec3.create(), 0),
+      new Plane(vec3.create(), 0),
+      new Plane(vec3.create(), 0),
+      new Plane(vec3.create(), 0),
+    ];
+    this.setWithViewProjection(m);
+  }
+
+  // Uses the fast plane-extraction algorithm described in
+  // Gribb & Hartmann (1997): https://tinyurl.com/5x5htcwm
+  public setWithViewProjection(m: mat4) {
+    const n = vec3.create();
+
+    // Left: row3 + row0
+    this.planes_[0].set(
+      vec3.set(n, m[3] + m[0], m[7] + m[4], m[11] + m[8]),
+      m[15] + m[12]
+    );
+
+    // Right: row3 - row0
+    this.planes_[1].set(
+      vec3.set(n, m[3] - m[0], m[7] - m[4], m[11] - m[8]),
+      m[15] - m[12]
+    );
+
+    // Top: row3 - row1
+    this.planes_[2].set(
+      vec3.set(n, m[3] - m[1], m[7] - m[5], m[11] - m[9]),
+      m[15] - m[13]
+    );
+
+    // Bottom: row3 + row1
+    this.planes_[3].set(
+      vec3.set(n, m[3] + m[1], m[7] + m[5], m[11] + m[9]),
+      m[15] + m[13]
+    );
+
+    // Near: row3 + row2
+    this.planes_[4].set(
+      vec3.set(n, m[3] + m[2], m[7] + m[6], m[11] + m[10]),
+      m[15] + m[14]
+    );
+
+    // Far: row3 - row2
+    this.planes_[5].set(
+      vec3.set(n, m[3] - m[2], m[7] - m[6], m[11] - m[10]),
+      m[15] - m[14]
+    );
+
+    for (const p of this.planes_) p.normalize();
+  }
+
+  public intersectsWithBox3(box: Box3) {
+    const v = vec3.create();
+    for (const plane of this.planes_) {
+      const n = plane.normal;
+      v[0] = n[0] > 0 ? box.max[0] : box.min[0];
+      v[1] = n[1] > 0 ? box.max[1] : box.min[1];
+      v[2] = n[2] > 0 ? box.max[2] : box.min[2];
+      if (plane.signedDistanceToPoint(v) < 0) return false;
+    }
+    return true;
+  }
+}
--- a/packages/core/src/math/plane.ts
+++ b/packages/core/src/math/plane.ts
@@ -1,0 +1,31 @@
+import { vec3 } from "gl-matrix";
+
+export class Plane {
+  public normal: vec3;
+  public signedDistance: number;
+
+  constructor(normal: vec3 = vec3.fromValues(0, 1, 0), distance = 0) {
+    this.normal = vec3.clone(normal);
+    this.signedDistance = distance;
+  }
+
+  public set(normal: vec3, distance: number) {
+    this.normal = vec3.clone(normal);
+    this.signedDistance = distance;
+  }
+
+  public signedDistanceToPoint(point: vec3) {
+    // Algebraic convention ax + by + cz + d = 0
+    // Negative values mean the point lies opposite the plane's normal
+    return vec3.dot(this.normal, point) + this.signedDistance;
+  }
+
+  public normalize() {
+    const len = vec3.length(this.normal);
+    if (len > 0) {
+      const inv = 1 / len;
+      vec3.scale(this.normal, this.normal, inv);
+      this.signedDistance *= inv;
+    }
+  }
+}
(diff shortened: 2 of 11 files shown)


packages/core/src/objects/cameras/camera.ts
```diff
@@ -34,6 +35,12 @@ export abstract class Camera extends RenderableObject {
     return vec3.fromValues(m[4], m[5], m[6]);
   }
 
+  get frustum() {
+    return new Frustum(
```
shlomnissan, line 39: It would be more efficient to cache the frustum rather than recreating it each time `get frustum()` is called but it requires a reliable way to track transform updates beyond a simple “dirty” flag. A common pattern in scene graphs is to maintain a transform version counter, which could also enable caching bounding boxes at the renderable level. This feels outside the scope of this PR so I’ll address it in a follow-up

packages/core/src/renderers/webgl_renderer.ts
```diff
@@ -102,9 +106,17 @@ export class WebGLRenderer extends Renderer {
     return this.textures_.textureInfo;
   }
 
-  private renderLayer(layer: Layer, camera: Camera) {
+  private renderLayer(layer: Layer, camera: Camera, frustum: Frustum) {
     this.state_.setBlendingMode(layer.transparent ? layer.blendMode : "none");
-    layer.objects.forEach((_, i) => this.renderObject(layer, i, camera));
+
+    let renderedObjects = 0;
```
andy-sweet: Blocker

As defined, `WebGLRenderer.renderedObjects` will be the number of objects rendered in the last layer. I'm guessing it should be the total number of render objects over all layers, in which case you should probably just set `this.renderedObjects_` to 0 before any layers are rendered and increment the property directly.

packages/core/src/math/plane.ts
```diff
@@ -0,0 +1,31 @@
+import { vec3 } from "gl-matrix";
+
+export class Plane {
+  public normal: vec3;
+  public distance: number;
```
andy-sweet: Optional

I interpret this as meaning the distance from the origin of the coordinate space in which this plane lives. However, it looks like this value could be negative and still be meaningful (i.e. if I understood correctly, it would effectively positively offset the plane from the origin in the direction of the normal) so I wonder if `distance` is a good name since distance implies non-negativity. I wonder if something like `constant` or `offset` might be a more accurate description.

packages/core/src/math/plane.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  public distanceToPoint(point: vec3) {
+    // Algebraic convention ax + by + cz + d = 0
+    // Negative values mean the point lies opposite the plane's normal
+    return vec3.dot(this.normal, point) + this.distance;
+  }
+
+  public normalize() {
+    const len = vec3.length(this.normal);
+    if (len > 0) {
```
andy-sweet, line 25: Optional

Is normalizing a plane with a normal of length 0 something we expect to happen with normal usage? If not, at least logging a warning here might be helpful to catch misuse.

packages/core/src/math/plane.ts
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  public set(normal: vec3, distance: number) {
+    this.normal = vec3.clone(normal);
+    this.distance = distance;
+  }
+
+  public distanceToPoint(point: vec3) {
+    // Algebraic convention ax + by + cz + d = 0
+    // Negative values mean the point lies opposite the plane's normal
```
andy-sweet, line 19: Optional

By default, I think of distances as non-negative, so maybe `signedDistanceToPoint` would be a more appropriate name here.

packages/core/test/frustum.test.ts
```diff
@@ -0,0 +1,131 @@
...
+
+  const b = new Box3(
+    vec3.fromValues(0.5, -0.2, -0.2),
+    vec3.fromValues(1.0, 0.2, 0.2)
+  );
+  expect(frustum.intersectsWithBox3(b)).toBe(true);
+});
+
+test("orthographic frustum with inside box", () => {
+  const vp = mat4.ortho(mat4.create(), -1, 1, -1, 1, 1, 10);
```
andy-sweet: Optional

A test helper function that allows us to write `const vp = ortho({left: -1, right: 1, bottom: -1, top: 1, near: 1, far: 10})` might help readers reason about this test immediately without having to jump to gl-matrix docs.

packages/core/test/frustum.test.ts
```diff
@@ -0,0 +1,131 @@
...
+  const frustum = new Frustum(vp);
+
+  const b = new Box3(
+    vec3.fromValues(-0.25, -0.25, -3.25),
```
andy-sweet, line 77: Question

What's a good way to think about the z-coordinates here? The near and far planes are defined as 1 and 10, so I was imaging the ortho frustum as a 3D box from (-1, -1, 1) to (1, 1, 10). Or maybe another way to phrase this is what coordinate space is the orthographic frustum defined in?

packages/core/src/math/frustum.ts
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  public intersectsWithBox3(box: Box3) {
+    const v = vec3.create();
+    for (let i = 0; i < 6; ++i) {
+      const n = this.planes_[i].normal;
+      v[0] = n[0] > 0 ? box.max[0] : box.min[0];
+      v[1] = n[1] > 0 ? box.max[1] : box.min[1];
+      v[2] = n[2] > 0 ? box.max[2] : box.min[2];
+      if (this.planes_[i].distanceToPoint(v) < 0) return false;
```
andy-sweet: Nit: minor readability improvement since `i` is only used for indexing `this.planes_`.

```suggestion
    for (const plane of this.planes_) {
      const n = plane.normal;
      v[0] = n[0] > 0 ? box.max[0] : box.min[0];
      v[1] = n[1] > 0 ? box.max[1] : box.min[1];
      v[2] = n[2] > 0 ? box.max[2] : box.min[2];
      if (plane.distanceToPoint(v) < 0) return false;
```

andy-sweet (changes requested):
Generally looks good with thorough unit tests as I've come to expect!

Most of my comments are optional or questions out of my own ignorance.

The one blocking comment is that I think `Renderer.renderedObjects_` is miscalculated in the WebGL render pass.

packages/core/src/math/plane.ts
shlomnissan: I think I’ll leave it as-is. A zero-length normal would indicate an upstream logic error, so it’s better caught in tests or assertions rather than at runtime. Normalizing silently is consistent with how most math libraries handle this.

packages/core/test/frustum.test.ts
shlomnissan: The orthographic frustum is defined in view space. `gl-matrix` follows OpenGL right-handed convention where the camera looks down the −Z axis. The near and far values are distances from the camera so with near = 1 and far = 10 the valid Z range is [−10, −1]. That’s why the box uses negative Z coordinates.. it’s positioned in front of the camera within the frustum’s view-space bounds

https://github.com/toji/gl-matrix/blob/master/src/mat4.js#L1667

shlomnissan:
@andy-sweet I accidentally amended the original commit instead of creating a new one for the PR updates, but all changes have been addressed

packages/core/test/frustum.test.ts
andy-sweet: Thanks for clarifying. This makes sense with respect to the rest of the design since the frustum can take any view-projection matrix. In these tests it takes one for use with GL, so then it follows that the box used for the intersection test needs to account for the negation in z that follows from that.

andy-sweet (approved):
Looks good now!

andy-sweet:
My only other thought here is that updating the volume renderer example to be similar to the video in the PR description would be ideal. But not blocking at all, especially if it requires other non-trivial changes.

shlomnissan:
> My only other thought here is that updating the volume renderer example to be similar to the video in the PR description would be ideal.

@andy-sweet The example in the video uses console logs to report how many boxes are rendered per frame. This was debug code for manual verification and should not be included. We can't use the chunk info overlay to show this either because this example doesn't use the chunk manager yet. Also, I want to keep the example minimal and have @seankmartin use it as a sandbox for volume rendering work.

PR #522, opened 2025-10-31
feat: add automated release workflow with semantic-release

phoenixAja:
## Summary

Implements automated NPM releases using semantic-release and enforces
conventional commit format for PR titles. Once merged, versions are
automatically published to NPM when PRs are merged to `main`, with
version numbers determined by commit types.

  ## Changes

  - Added automated release workflow (`.github/workflows/release.yml`)
- Added PR title validation workflow
(`.github/workflows/pr-title-check.yml`)
  - Added semantic-release configuration (`.releaserc.json`)
  - Fixed TypeScript types (`@types/react`, `@czi-sds/components`)
  - Updated README with new release process

  ## How Versioning Works

  ### PR Title Format (Required)

  Examples:
  Good: feat(slider): add zoom controls
  Good: fix(viewer): resolve memory leak
  Bad: Add zoom controls (invalid)

  ### Version Bump Rules

  | Type | Bump | Example | Release? |
  |------|------|---------|----------|
  | `feat:` | Minor | 7.0.0 → 7.1.0 | Yes |
| `fix:`, `perf:`, `refactor:`, `chore:` | Patch | 7.0.0 → 7.0.1 | Yes |
  | `feat!:` or `BREAKING CHANGE:` | Major | 7.0.0 → 8.0.0 | Yes |
  | `docs:`, `style:`, `test:`, `ci:` | None | - | No |

  ## Release Flow

  1. PR with `feat(slider): add zoom controls` merged to `main`
  2. Release workflow runs automatically
  3. Analyzes commit and determines version bump
  4. Publishes to NPM, creates GitHub release, updates CHANGELOG
  5. Package is live on NPM

### DIFF
--- a/.github/workflows/pr-title-check.yml
+++ b/.github/workflows/pr-title-check.yml
@@ -1,0 +1,48 @@
+name: PR Title Check
+permissions:
+  contents: read
+  pull-requests: read
+
+on:
+  pull_request:
+    types:
+      - opened
+      - edited
+      - synchronize
+      - reopened
+
+jobs:
+  check-pr-title:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check PR title
+        uses: amannn/action-semantic-pull-request@v5
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          # Configure which types are allowed (must match semantic-release config)
+          types: |
+            feat
+            fix
+            docs
+            style
+            refactor
+            perf
+            test
+            build
+            ci
+            chore
+            revert
+          # Do not require a scope
+          requireScope: false
+          # Subject must not be empty
+          subjectPattern: ^.+$
+          subjectPatternError: |
+            The subject "{subject}" found in the pull request title "{title}"
+            doesn't match the configured pattern. Please ensure your PR title
+            follows the conventional commit format.
+          # Validate PR title after the PR body is updated
+          validateSingleCommit: false
+          # Ignore merge commits
+          ignoreLabels: |
+            ignore-semantic-pr
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,55 @@
+name: Release
+
+on:
+  push:
+    branches:
+      - main
+
+permissions:
+  contents: write
+  issues: write
+  pull-requests: write
+  id-token: write
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    if: ${{ ! contains(github.event.head_commit.message, '[skip ci]') }}
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          token: ${{ secrets.RELEASE_TOKEN }}
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
+          cache: 'npm'
+
+      - name: Install dependencies
+        run: |
+          npm ci
+          npx playwright install chromium
+
+      - name: Compile TypeScript
+        run: npm run compile
+
+      - name: Lint
+        run: |
+          npm run lint
+          npm run format-check
+
+      - name: Run tests with coverage
+        run: npm run test-with-coverage
+
+      - name: Build
+        run: npm run build
+
+      - name: Release
+        env:
+          GITHUB_TOKEN: ${{ secrets.RELEASE_TOKEN }}
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+        run: npx semantic-release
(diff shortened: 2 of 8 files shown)


andy-sweet:
I've already approved something very similar on idetik-react, so while I approved, I'm not going to actually hit the approve button because at least one other person should approved this and ideally the whole team.

I think we mentioned updating the package name from `@idetik/prerelease` to `idetik` and winding back the version number to 0.something. I guess we don't strictly need to do that before making this PR, since we can just remove the tags/versions after we've done that right?

phoenixAja:
> I've already approved something very similar on idetik-react, so while I approved, I'm not going to actually hit the approve button because at least one other person should approved this and ideally the whole team.
> 
> I think we mentioned updating the package name from `@idetik/prerelease` to `idetik` and winding back the version number to 0.something. I guess we don't strictly need to do that before making this PR, since we can just remove the tags/versions after we've done that right?

yes, i was going to rename the package and reset the tags in a follow up PR!

.github/workflows/pr-title-check.yml
```diff
@@ -0,0 +1,48 @@
...
+          subjectPattern: ^.+$
+          subjectPatternError: |
+            The subject "{subject}" found in the pull request title "{title}"
+            doesn't match the configured pattern. Please ensure your PR title
+            follows the conventional commit format.
+          # Validate PR title after the PR body is updated
+          validateSingleCommit: false
+          # Ignore merge commits
+          ignoreLabels: |
+            ignore-semantic-pr
```
aganders3, line 48: Where does this label come from? Is it something we do manually or assign somewhere?

.github/workflows/release.yml
```diff
@@ -0,0 +1,55 @@
...
+  release:
+    runs-on: ubuntu-latest
+    if: ${{ ! contains(github.event.head_commit.message, '[skip ci]') }}
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          token: ${{ secrets.RELEASE_TOKEN }}
```
aganders3, line 24: Is the release token necessary here?

.github/workflows/release.yml
```diff
@@ -0,0 +1,55 @@
...
+      - name: Compile TypeScript
+        run: npm run compile
+
+      - name: Lint
+        run: |
+          npm run lint
+          npm run format-check
+
+      - name: Run tests with coverage
+        run: npm run test-with-coverage
```
aganders3, line 46: I think there's a way to reuse these steps/checks from our other workflow, but I will be honest I don't know how to do it. This is pretty straightforward so optional if you'd like to sort that out.

README.md
```diff
@@ -35,42 +35,75 @@ A layer-based library for interactive visualization of large datasets.
...
+We use [semantic-release](https://github.com/semantic-release/semantic-release) to automatically handle versioning, changelog generation, npm publishing, and GitHub releases.
+
+#### How It Works
+
+1. **Use Conventional Commits**: When creating PRs, ensure your PR title follows the [Conventional Commits](https://www.conventionalcommits.org/) format:
+   - `feat: add new feature` → triggers a **minor** version bump (e.g., 7.0.0 → 7.1.0)
+   - `fix: resolve bug` → triggers a **patch** version bump (e.g., 7.0.0 → 7.0.1)
+   - `feat!: breaking change` or `BREAKING CHANGE:` in commit footer → triggers a **major** version bump (e.g., 7.0.0 → 8.0.0)
+   - `refactor:`, `perf:`, `chore:` → triggers a **patch** version bump
+   - `docs:`, `style:`, `test:`, `ci:`, `build:` → no release
```
aganders3, line 49: Just calling out (for myself, perhaps) that this now needs to be part of PR review to ensure we're using the correct tag. Maybe this doesn't matter so much until we hit 1.0 though?

aganders3 (approved):
I'm not really sure how to test this, but it looks good to me! I left a few questions but I don't have actionable feedback.

Is there a way to keep (major?) versions of the separate libraries (`core` and `react`) in sync? Do we even want/expect to do that?

andy-sweet:
> Is there a way to keep (major?) versions of the separate libraries (core and react) in sync? Do we even want/expect to do that?

Unsure how to do it, but I don't want or expect it. I expect that the public APIs of both packages might evolve slightly differently.

README.md
andy-sweet: Maybe we could make the action add/replace a comment with the version number change it will trigger?

aganders3:
Thanks - that makes sense when I think about it.

I think core is currently a peer-dependency of the react library, so I think we should indicate prominently the version relationship on that repo. This can be considered with other options over there, though.

.github/workflows/release.yml
phoenixAja: yes it is that's to bypass branch protection for semantic to push to main without a PR

.github/workflows/pr-title-check.yml
phoenixAja: you can add this label to your PR if you want to skip validation (for whatever reason)

README.md
phoenixAja: i can work on this as a follow up

PR #523, opened 2025-10-31
feat: rename npm package from pre-release to core

phoenixAja:
## Summary
  Renames the package from `@idetik/core-prerelease` to `@idetik/core` and resets versioning to `0.1.0`.

  ## Changes
  - Updated package name in `packages/core/package.json`
  - Reset version to `0.1.0`
  - Updated all workspace references in root `package.json`
  - Updated package references in READMEs and documentation
  - Updated GitHub workflows to use new package name

  ## Migration Notes
  The old `@idetik/core-prerelease` package will be deprecated on npm with a notice directing users to `@idetik/core`.

  Users should update their dependencies:
  ```bash
  npm uninstall @idetik/core-prerelease
  npm install @idetik/core
```

Pre-Merge Steps
  - delete existing tags

  Post-Merge Steps

  - Publish @idetik/core@0.1.0 to npm
  - Deprecate @idetik/core-prerelease on npm

### DIFF
--- a/.github/workflows/deploy-examples.yml
+++ b/.github/workflows/deploy-examples.yml
@@ -33,10 +33,10 @@ jobs:
         run: npm ci
 
       - name: Build core library
-        run: npm run build --workspace=@idetik/core-prerelease
+        run: npm run build --workspace=@idetik/core
 
       - name: Build examples for production
-        run: npm run build:examples --workspace=@idetik/core-prerelease
+        run: npm run build:examples --workspace=@idetik/core
 
       - name: Setup Pages
         uses: actions/configure-pages@v4
--- a/.github/workflows/lint-test-build.yml
+++ b/.github/workflows/lint-test-build.yml
@@ -30,17 +30,17 @@ jobs:
 
       - name: Compile TypeScript
         run: |
-          npm run compile --workspace=@idetik/core-prerelease
-          npm run build --workspace=@idetik/core-prerelease
+          npm run compile --workspace=@idetik/core
+          npm run build --workspace=@idetik/core
 
       - name: Lint
         run: |
-          npm run lint --workspace=@idetik/core-prerelease
-          npm run format-check --workspace=@idetik/core-prerelease
+          npm run lint --workspace=@idetik/core
+          npm run format-check --workspace=@idetik/core
 
       - name: Run tests with coverage
         run: |
-          npm run test-with-coverage --workspace=@idetik/core-prerelease
+          npm run test-with-coverage --workspace=@idetik/core
 
   build:
     needs: lint-and-test
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ A layer-based library for interactive visualization of large datasets.
 
 ## Release
 
-Currently, we maintain the [@idetik/core-prerelease](https://www.npmjs.com/package/@idetik/core-prerelease?activeTab=readme) package on npm.
+We maintain the [@idetik/core](https://www.npmjs.com/package/@idetik/core?activeTab=readme) package on npm.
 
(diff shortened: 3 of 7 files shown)


andy-sweet:
I think @shlomnissan preferred a package name of just `idetik`, which I also prefer at this stage - if we want to subdivided the single package in this repo into subpackages like @idetik/core, @idetik/render, @idetik/data, etc. we can do that later. 

I'm not sure what the current `@idetik/...` namespacing means in npm (if anything), so I'm not sure if that causes any issues with this decision.

But I am in general favor of this PR and will approve it once the team is happy with a package name.

aganders3:
> I'm not sure what the current `@idetik/...` namespacing means in npm (if anything), so I'm not sure if that causes any issues with this decision.

The namespacing is a real feature, and intended to prevent name collisions and ensure ownership: you have to own the user/org in the first part (thanks to @bchu1, we do!).

Since the bare `idetik` is not taken, I agree we should use it (or at least squat it).

edit to add: I'm also supportive of this change

shlomnissan (approved):
Looks good to me.

I would consider renaming to `idetik` and drop workspaces altogether. I don't think this structure adds any value right now.

andy-sweet:
> Since the bare idetik is not taken, I agree we should use it (or at least squat it).
> edit to add: I'm also supportive of this change

> I would consider renaming to idetik and drop workspaces altogether.

@phoenixAja : if you're also in agreement, then it sounds like we should rename the package in this repo to `idetik`. I think this should be done either in this PR, or before this is merged so we avoid publishing more packages on npm than we need.

> drop workspaces altogether.

I'd also be in favor of dropping workspaces for now since there is only one package in this repo now. I see this is a related, but separate concern from this PR (which is about the package name built and published to npm). I don't mind if you include that change in this PR, but I think it's cleaner and better practice to do it in a separate one (also doesn't have to be you!).

phoenixAja:
> > Since the bare idetik is not taken, I agree we should use it (or at least squat it).
> > edit to add: I'm also supportive of this change
> 
> > I would consider renaming to idetik and drop workspaces altogether.
> 
> @phoenixAja : if you're also in agreement, then it sounds like we should rename the package in this repo to `idetik`. I think this should be done either in this PR, or before this is merged so we avoid publishing more packages on npm than we need.
> 
> > drop workspaces altogether.
> 
> I'd also be in favor of dropping workspaces for now since there is only one package in this repo now. I see this is a related, but separate concern from this PR (which is about the package name built and published to npm). I don't mind if you include that change in this PR, but I think it's cleaner and better practice to do it in a separate one (also doesn't have to be you!).

agreed that i think workspaces should be dropped in a different PR. I will work on renaming

phoenixAja:
I hate to throw a wrench into things, but I actually think we should keep the package as `@idetik/core` instead of renaming it to `idetik`. Scoped packages give us org-level control, meaning only our team can publish under `@idetik/*`, while an unscoped `idetik` would live in the global namespace and be owned by a single account. Using the org scope makes all our packages clearly part of the same ecosystem and lets us manage access through org teams rather than tracking individual maintainers. With an unscoped package the original publisher remains the ultimate owner, which can create a risk if that account becomes inactive for whatever reason.

If we publish an unscoped package named `idetik`, although we’d own it now, there’s less structural protection for grouping related modules and less clarity that it’s part of our namespace. Anyone outside our org could still publish things like `idetik-utils `or `idetik-react`, which could easily confuse users.

Keeping `@idetik/core` ensures consistent naming alongside `@idetik/react` and future modules (`@idetik/cli`, `@idetik/utils`, etc.), all under our verified org scope. I feel like this is safer, cleaner, and easier to manage long-term.

PR #524, opened 2025-11-05
Fix some prefetched chunks from being disposed

andy-sweet:
### Summary
This fixes a bug that was introduced in #493 because `chunk.prefetch` is not equivalent to `eligibleForPrefetch` in the `shouldDispose` method.

That bug can be reproduced on `main` by loading the default chunk streaming example, panning the camera, and observing the chunk stats changing in a way that they shouldn't (i.e. some prefetched chunks get reloaded).

### Tests & Checks
I tested the chunk streaming example.

### DIFF
--- a/packages/core/src/core/chunk_manager_source.ts
+++ b/packages/core/src/core/chunk_manager_source.ts
@@ -357,7 +357,7 @@ export class ChunkManagerSource {
           isFallbackLOD,
           isCurrentLOD,
           isVisible,
-          chunk.prefetch,
+          eligibleForPrefetch,
           isChannelInSlice
         )
       ) {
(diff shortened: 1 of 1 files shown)


andy-sweet:
@aganders3 : assigned you for review because you also reviewed #493 . I swear I remember mentally asserting the equivalence of these things, but I guess this is why we should have tests! The logic here is still hard to follow and this change does not attempt to fix that, but I'm open to ideas.

For context, I found this while working on something else - efficient tracking of chunk stats so that I can efficiently check how many time points have been prefetched enough for presentation (related to playback buffering).

aganders3 (approved):
This looks good as a fix for now. I agree the logic is still a bit confusing here, but it's worth fixing this bug.

I'm just skimming, but my suggestion on cleaning up the logic is that I think it's trying to maintain some of these boolean flags as mutually exclusive, and they don't need to be. If they do need to be mutually exclusive, we should enumerate them instead of bookkeeping the different combinations.

PR #525, opened 2025-11-10
feat: add volume shaders for different texture data types

seankmartin:
### Summary

Primary goal here is to add support for different texture data types in volume rendering.

This adds support for passing width, height, depth, and a Texture3D to the volume renderable object. The dataType in the Texture3D is used to determine the correct shader program name to use for the volume renderable based on the type of data stored in the texture. This is using a similar pattern to how the image renderable uses the texture to determine the program name.

Three new shader programs are added, a `floatVolume`, `intVolume` and `uintVolume`. Also added is an example volume rendering fragment shader, and passing UVs from the vertex shader for volume rendering to the fragment shader for volume rendering. This volume rendering is still straight down Z of the volume, will be updated later to respect camera parameters and allow inputs to modify normalization of values.

For the example of trying these, a simple checker board style small texture has been defined in the volume layer.  https://github.com/chanzuckerberg/idetik/pull/515/files has more progress on using actual data here, but this PR is separated into two to hopefully be easier to review. 

image

### Tests & Checks

Manually tested changes by changing the data type of the texture used in the volume layer example. Also checked other pre-existing examples just in case. This changes the multiple-viewports example to look as:
image

### DIFF
--- a/packages/core/src/layers/volume_layer.ts
+++ b/packages/core/src/layers/volume_layer.ts
@@ -1,4 +1,5 @@
 import { Layer } from "../core/layer";
+import { Texture3D } from "../objects/textures/texture_3d";
 import { VolumeRenderable } from "../objects/renderable/volume_renderable";
 
 export class VolumeLayer extends Layer {
@@ -7,7 +8,10 @@ export class VolumeLayer extends Layer {
   constructor() {
     super();
 
-    const renderable = new VolumeRenderable();
+    const data = new Uint8Array([127, 0, 0, 127, 127, 0, 0, 127]);
+    const texture = new Texture3D(data, 2, 2, 2);
+    texture.unpackAlignment = 1;
+    const renderable = new VolumeRenderable(1, 1, 1, texture);
     renderable.wireframeEnabled = true;
 
     this.addObject(renderable);
--- a/packages/core/src/layers/volume_layer.ts
+++ b/packages/core/src/layers/volume_layer.ts
@@ -8,7 +8,7 @@ export class VolumeLayer extends Layer {
   constructor() {
     super();
 
-    const data = new Uint8Array([127, 0, 0, 127, 127, 0, 0, 127]);
+    const data = new Int8Array([127, 0, 0, 127, 127, 0, 0, 127]);
     const texture = new Texture3D(data, 2, 2, 2);
     texture.unpackAlignment = 1;
     const renderable = new VolumeRenderable(1, 1, 1, texture);
--- a/packages/core/src/objects/renderable/volume_renderable.ts
+++ b/packages/core/src/objects/renderable/volume_renderable.ts
@@ -27,12 +27,12 @@ function dataTypeToVolumeShader(dataType: TextureDataType): Shader {
     case "byte":
     case "int":
     case "short":
-      return "intVolumeImage";
+      return "intVolume";
     case "unsigned_short":
     case "unsigned_byte":
     case "unsigned_int":
-      return "uintVolumeImage";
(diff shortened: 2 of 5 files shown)


packages/core/src/layers/volume_layer.ts
```diff
@@ -7,7 +8,10 @@ export class VolumeLayer extends Layer {
   constructor() {
     super();
 
-    const renderable = new VolumeRenderable();
+    const data = new Int8Array([127, 0, 0, 127, 127, 0, 0, 127]);
+    const texture = new Texture3D(data, 2, 2, 2);
+    texture.unpackAlignment = 1;
```
shlomnissan, line 13: I think it makes more sense to set the alignment to tightly packed as a reasonable default? @aganders3 currently `Texture.unpackAlignment` is set to 4 (which is OpenGL's default I think) but I assume data in chunks is almost always tightly packed?

packages/core/src/objects/renderable/volume_renderable.ts
```diff
@@ -1,14 +1,38 @@
...
+    width: number,
+    height: number,
+    depth: number,
+    texture: Texture3D
+  ) {
     super();
-    this.geometry = new BoxGeometry(1, 1, 1, 1, 1, 1);
-    this.programName = "volume";
+    this.geometry = new BoxGeometry(width, height, depth, 1, 1, 1);
+    this.setTexture(0, texture);
```
shlomnissan, line 16: I feel like this API is confusing and fragile. The index number have no semantic meaning. I assumed it means texture units until poking at the implementation. Is this intentional?

@andy-sweet tagging you because I think you implemented this?

packages/core/src/renderers/shaders/index.ts
```diff
@@ -76,8 +78,18 @@ export const shaderCode: Record<Shader, ShaderCode> = {
     vertex: meshVertexShader,
     fragment: labelImage,
   },
-  volume: {
+  floatVolume: {
     vertex: volumeVertexShader,
     fragment: volumeFragmentShader,
   },
+  intVolume: {
```
shlomnissan, line 85: An incremental improvement would be to let each renderable object supply its own program attributes to a single shader program instead of defining new programs with the same shader code. We also don’t need to distinguish between vertex and fragment defines. I think using the same set for both stages should be simpler and sufficient.

shlomnissan (approved):
Looks good to me.

seankmartin:
> Looks good to me.

Thanks for the review!

packages/core/src/objects/renderable/volume_renderable.ts
andy-sweet: Correct - the semantics are that the index number corresponds to the texture unit for this renderable object's fragment shader. Agreed this is fragile and confusing. 

I implemented `setTexture` in this way to work around the awkward design of `RenderableObject.textures`, actually support multiple textures per renderable object, and also allow stale textures to be disposed when they are changed.

I'd welcome an improvement here, which would likely introduce a more explicit mapping between a specific renderable object's textures (e.g. explicit properties for each one rather than an array of textures), and then some way for a renderer to map these to texture units (which might be renderer specific).

Feel free to create an issue.

packages/core/src/layers/volume_layer.ts
aganders3: Yes I think zarr/zarrita will never provide padded arrays and we should set our default to 1 (I think you're right 4 is the WebGL default). In the loader we currently set the unpack alignment to the bytes per element, but I think this is unnecessary as well.

PR #526, opened 2025-11-10
feat: add face culling management to webgl state and allow layers to indicate culling mode

seankmartin:
### Summary

Primary goal - allow the volume rendering layer to cull back faces during rendering.
Adds to the webgl state to allow setting a cullFace mode. Primarily for culling back faces, but allows all modes (front, back, front and back) or being disabled. Follows a similar pattern to the blending mode in the webgl state abstraction.

The example of volume rendering looks a bit odd right now with this, but I'm not entirely sure why as the state abstraction seems correctly setup. To debug this, I tried creating a simpler box geometry, similar to neuroglancer style https://github.com/google/neuroglancer/blob/master/src/webgl/bounding_box.ts -- this one is actually in glsl, but it's the same idea of hardcoding all the faces, normals, and uvs in one vertex array and hardcoding the indices in another array, as opposed to the face building operation which the box_geometry uses. The back face culling seems fine with the simple geometry https://github.com/chanzuckerberg/idetik/pull/515

Since the scope of this PR is to setup how layers could indicate to the rendering wanting to use face culling, that box geometry issue could possibly be solved separately.

### Tests & Checks

Manually tested the examples

### DIFF
--- a/packages/core/src/core/layer.ts
+++ b/packages/core/src/core/layer.ts
@@ -3,7 +3,6 @@ import { RenderableObject } from "./renderable_object";
 import { clamp } from "../utilities/clamp";
 import { Logger } from "../utilities/logger";
 import { EventContext } from "./event_dispatcher";
-import { CullingMode } from "../renderers/WebGLState";
 
 export type LayerState = "initialized" | "loading" | "ready";
 export type blendMode = "normal" | "additive" | "subtractive" | "multiply";
@@ -29,7 +28,6 @@ export abstract class Layer {
   public transparent: boolean;
   private opacity_: number;
   public blendMode: blendMode;
-  private cullFaceMode_: CullingMode = "none";
 
   constructor({
     transparent = false,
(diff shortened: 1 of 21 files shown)


packages/core/src/core/layer.ts
```diff
@@ -28,6 +29,7 @@ export abstract class Layer {
   public transparent: boolean;
   private opacity_: number;
   public blendMode: blendMode;
+  private cullFaceMode_: CullingMode = "none";
```
shlomnissan: I think this should be applied per renderable object rather than per layer. It determines which triangle faces are culled which feels like an object-level attribute rather than a layer-level setting.

shlomnissan (changes requested):
This PR looks good overall. I'm only blocking because I think the cull face mode belongs on the renderable object and not the layer. It’s not a big deal to keep it on the layer for now if there’s a specific reason for that choice?

packages/core/src/core/layer.ts
seankmartin: Agreed, that makes more sense, thank you Shlomi

shlomnissan (approved):
LGTM!

PR #528, opened 2025-11-14
chore: remove semantic-release git plugin to fix release workflow

phoenixAja:
## Summary

This removes the `@semantic-release/git` and `@semantic-release/changelog` plugins from the semantic-release configuration to fix the automated release workflow.

## Problem

The release workflow was failing when `@semantic-release/git` tried to push version bump commits back to the `main` branch. This violated our branch protection rules
which require:
- All changes to be made through pull requests
- Required status checks to pass

Even with a PAT that has write permissions, the branch protection rules block direct pushes. While we could add a bypass for this, it would either:
1. Allow personal accounts to push directly to main (security risk)
2. Require setting up a GitHub App with bypass permissions (significant complexity)

## Solution

Remove the git plugin from semantic-release. The workflow will still:
- Analyze commits and determine the next version
- Publish to npm with the correct version
- Create GitHub releases with full release notes
- Create git tags for each release

The only change is that version bumps won't be committed back to the repository. The release notes that would have been in `CHANGELOG.md` will instead appear in GitHub
Releases, which is where most users look for them anyway.

## Test Plan

This can only be tested by merging to main and verifying that the release workflow completes successfully and publishes to npm.

### DIFF
--- a/.releaserc.json
+++ b/.releaserc.json
@@ -46,22 +46,10 @@
         }
       }
     ],
-    "@semantic-release/changelog",
     [
       "@semantic-release/npm",
       {
         "pkgRoot": "packages/core"
-      }
-    ],
-    [
-      "@semantic-release/git",
-      {
-        "assets": [
-          "CHANGELOG.md",
-          "packages/core/package.json",
-          "package-lock.json"
-        ],
-        "message": "chore(release): ${nextRelease.version} [skip ci]\n\n${nextRelease.notes}"
       }
     ],
     "@semantic-release/github"
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,8 +9,6 @@
       ],
       "devDependencies": {
         "@rollup/plugin-typescript": "^12.1.2",
-        "@semantic-release/changelog": "^6.0.3",
-        "@semantic-release/git": "^10.0.1",
         "@types/node": "^24.3.1",
         "@types/webgl2": "^0.0.11",
         "@typescript-eslint/eslint-plugin": "^8.1.0",
@@ -2073,25 +2071,6 @@
       "dev": true,
       "license": "MIT"
     },
-    "node_modules/@semantic-release/changelog": {
-      "version": "6.0.3",
-      "resolved": "https://registry.npmjs.org/@semantic-release/changelog/-/changelog-6.0.3.tgz",
-      "integrity": "sha512-dZuR5qByyfe3Y03TpmCvAxCyTnp7r5XwtHRf/8vD9EAn4ZWbavUX8adMtXYzE86EVh0gyLA7lm5yW4IV30XUag==",
-      "dev": true,
-      "license": "MIT",
-      "dependencies": {
-        "@semantic-release/error": "^3.0.0",
-        "aggregate-error": "^3.0.0",
-        "fs-extra": "^11.0.0",
-        "lodash": "^4.17.4"
-      },
-      "engines": {
-        "node": ">=14.17"
-      },
-      "peerDependencies": {
-        "semantic-release": ">=18.0.0"
-      }
-    },
     "node_modules/@semantic-release/commit-analyzer": {
       "version": "13.0.1",
       "resolved": "https://registry.npmjs.org/@semantic-release/commit-analyzer/-/commit-analyzer-13.0.1.tgz",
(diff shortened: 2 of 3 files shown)


andy-sweet:
> The only change is that version bumps won't be committed back to the repository. The release notes that would have been in CHANGELOG.md will instead appear in GitHub Releases, which is where most users look for them anyway.

On an initial read, this seems problematic because the version bump in package.json should be committed to the repo at some point, right? Did I misunderstand something? Note that I don't much about `CHANGELOG.md` given that it's automatically generated.

andy-sweet:
Another thought is to automatically update package.json, package-lock.json, and CHANGELOG.md in the PR that's making other changes. One potential problem is that this has to run just before squash/merging and needs to check that the version bump is consistent with `main`, to avoid version change races.

phoenixAja:
> > The only change is that version bumps won't be committed back to the repository. The release notes that would have been in CHANGELOG.md will instead appear in GitHub Releases, which is where most users look for them anyway.
> 
> On an initial read, this seems problematic because the version bump in package.json should be committed to the repo at some point, right? Did I misunderstand something? Note that I don't much about `CHANGELOG.md` given that it's automatically generated.

So actually the source of truth for versions is the git tag, not what's in package.json. With this method package.json will be out of sync with the tags, which i think is ok since users will be downloading from NPM and i think most users who want to look at versions within our repo will check the tags. We could add an action that checks that the user updated package.json version based on the conventional commit title. 

This is definitely a tradeoff between if we want strict branch protection for admins or an out of sync package.json.

andy-sweet:
> With this method package.json will be out of sync with the tags, which i think is ok since users will be downloading from NPM 

I don't mean to do this yet again within the versioning work, but I think it might be worth polling the team on this. I don't have much experience with publishing npm packages, but given our release apporach my gut instinct is that a commit on `main` should correspond to a version and that `package.json` should reflect that version as per the [npm guide on package.json](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#version)

> If you plan to publish your package, the most important things in your package.json are the name and version fields as they will be required. The name and version together form an identifier that is assumed to be completely unique. Changes to the package should come along with changes to the version. If you don't plan to publish your package, the name and version fields are optional.

While I love the release automation, I also feel like I should be follow a manual process to achieve the same results with a commit main without requiring changes to the contents of the repo. This may become important if we ever change the way we handle releases.

If we decide to follow the approach suggested by this PR, I think we should at least enable the following setting for the repo

> Enable release immutability Loading

so that the tags cannot be removed (given they are the source of truth). This is maybe a good idea regardless.

phoenixAja:
haha that's fine i don't mind discussing this here. 


>> I don't mean to do this yet again within the versioning work, but I think it might be worth polling the team on this. I don't have much experience with publishing npm packages, but given our release apporach my gut instinct is that a commit on `main` should correspond to a version and that `package.json` should reflect that version as per the [npm guide on package.json](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#version)
> 

Right, I agree and that's why i added the github plugin in the first place. The issue is that to commit a package.json change back to main we would have to add the bypass rule (which I am fine with, but this would make it easier to push to main on accident). i think we do have some other potential solutions though, namely adding an action that updated package.json in the PR or an action that blocks merging unless package.json has been bumped.

Updating package.json in our repo isn't necessary because semantic-release automatically updates it in the CI environment before publishing to npm. It determines the version by analyzing git tags (last released version) and commits (what changed), then temporarily updates package.json and publishes. Since this happens in CI, these changes aren't committed back to the repo.

> If we decide to follow the approach suggested by this PR, I think we should at least enable the following setting for the repo
> 
> > Enable release immutability Loading
> 
> so that the tags cannot be removed (given they are the source of truth). This is maybe a good idea regardless.

yes agreed! :)  thanks for finding that.

andy-sweet:
> Updating package.json in our repo isn't necessary because semantic-release automatically updates it in the CI environment before publishing to npm. 

Just to clarify, I understand it's not necessary for the practical publishing process on npm, but it's those temporary/non-committed changes to the repo's contents that I have some philosophical objection to, and might have practical implications in the future (e.g. if we change the way we automate releases).

Because this is maybe more of a philosophical objection (rather than immediately practical one), I think the best option is to ask for a second opinion to tell me I'm overthinking it - does that make sense?

phoenixAja:
> Just to clarify, I understand it's not necessary for the practical publishing process on npm, but it's those temporary/non-committed changes to the repo's contents that I have some philosophical objection to, and might have practical implications in the future (e.g. if we change the way we automate releases).
> 
right, sorry for over explaining then 😅

> Because this is maybe more of a philosophical objection (rather than immediately practical one), I think the best option is to ask for a second opinion to tell me I'm overthinking it - does that make sense?

sure! ~~but actually i think i'm going to add an action first to block merging until package.json is updated according to semantic release and then i'll poll opinions! (Just since i think that would be a nice thing to add with this approach if we go this direction)~~ nevermind this could cause race condition issues (if outside PRs are merged in after this)

andy-sweet (approved):
Based on the discussion with the team on Slack, I'm OK approving this now, ideally with the following repo setting enabled

> Enable release immutability Loading

and assuming idetik-react is reflected to mirror these settings.

This should help us share updates quickly and we can see how it goes. We might want to revisit this around public launch time.

aganders3 (approved):
+1 to what @andy-sweet mentioned above
co-signing the approach here for posterity

let's see how it goes!

PR #529, opened 2025-11-14
fix: Pass ChannelCount uniform to improve multi-channel image fragment shader

aganders3:
### Summary
I was seeing rather large FPS drops/changes with window size.

Initially I just wrote this off as the number of tiles being rendered or something and didn't really think about it, but then @shlomnissan said:
> Why would rendering FPS change based on window size and LOD?

Great question! This prompt gave me pause. Even at LOD0 and full screen it's something like 24 tiles which should be no problem. I hadn't done the profiling to see what the limitation was but I saw the same behavior in the chunk streaming example on main. I think vsync can make it seem more drastic because FPS seems to hover around integer fractions of the refresh rate (20, 30, 40, 60 FPS, etc,).

So I set out to do that profiling. `gl.finish()` did not seem to be enough to actually block until the draw was done, but I found this code which does (h/t Claude):
```
gl.flush(); // Submit commands
gl.finish(); // Wait for completion
// Read a pixel to force full pipeline flush
const dummy = new Uint8Array(4);
gl.readPixels(0, 0, 1, 1, gl.RGBA, gl.UNSIGNED_BYTE, dummy);
```
This showed rendering at full screen on my external monitor takes ~**10 ms**. Full screen on my laptop monitor is about 2x the pixels and takes close to **20 ms**. Double the pixels double the time felt like a fragment shader issue.

So....I looked again at our fragment shader. It's very simple, but there is a loop over `MAX_CHANNELS`. Hard-coding the loop to 1 channel fixes the problem completely. I think this is a case of the GPU/shader compiler having poor branch prediction and optimization. So even though we `continue` the loop it is slow! Passing a uniform for `ChannelCount` also fixes the problem and I get a very solid 120 FPS (except examples that include volume rendering, but this is not necessarily unexpected at this time).

### Related Issue
[Slack thread](https://czi-sci.slack.com/archives/C07GHGSJSQP/p1763069155880939?thread_ts=1763059524.732459&cid=C07GHGSJSQP)

### Tests & Checks
Tested FPS from stats.js and with logging timing
GPU time is now comfortably <10ms/frame for the chunk_streaming examples.

### DIFF
--- a/packages/core/src/objects/renderable/image_renderable.ts
+++ b/packages/core/src/objects/renderable/image_renderable.ts
@@ -20,6 +20,7 @@ type SingleUniformValues = {
 
 type ArrayUniformValues = {
   ImageSampler: number;
+  ChannelCount: number;
   "Visible[0]": boolean[];
   "Color[0]": number[];
   "ValueOffset[0]": number[];
--- a/packages/core/src/objects/renderable/image_renderable.ts
+++ b/packages/core/src/objects/renderable/image_renderable.ts
@@ -15,6 +15,7 @@ type SingleUniformValues = {
   Color: vec3;
   ValueOffset: number;
   ValueScale: number;
+  ChannelCount: number;
 };
 
 type ArrayUniformValues = {
@@ -76,6 +77,7 @@ export class ImageRenderable extends RenderableObject {
         Color: color.rgb,
         ValueOffset: -contrastLimits[0],
         ValueScale: 1 / (contrastLimits[1] - contrastLimits[0]),
+        ChannelCount: 1,
       };
     } else {
       // Texture2DArray case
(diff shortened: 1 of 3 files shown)


packages/core/src/objects/renderable/image_renderable.ts
andy-sweet, line 24: Not technically required for current type checks/compilation, but helps clarify this type given the current design.

```suggestion
  ChannelCount: number;
  "Visible[0]": boolean[];
```

andy-sweet (approved):
Looks good and works for me! Thanks for the debugging effort here.

packages/core/src/objects/renderable/image_renderable.ts
aganders3: Ah, I misunderstood here the point of `ArrayUniformValues` vs. `SingleUniformValues`.

Actually this is the _more_ important place to include it, so thanks.

PR #531, opened 2025-11-18
feat: refactor to share chunk data across multiple layers/viewports

aganders3:
### Summary
This is picking up where #514 left off, basically just that PR rebased on `main`.

This PR introduces a `ChunkStore`/`ChunkStoreView` architecture to enable efficient chunk sharing across multiple viewports. Currently, `ChunkManagerSource` conflates two responsibilities: managing a canonical chunk data store (for a given source) and determining viewport-specific chunk visibility. This design prevents resource sharing when multiple viewports render overlapping portions of the same data source.

The implementation splits `ChunkManagerSource` into two complementary classes:
  - `ChunkStore`: Manages the canonical data store and global state for all chunks for a single source by aggregating view state across associated `ChunkStoreView` objects
  - `ChunkStoreView`: Represents a layer/viewport-specific "view" into that store, calculating LOD, visibility, and prefetch requirements based on its camera and slice position

Multiple layers (including in separate viewports) can now share the same `ChunkStore` while each maintains its own `ChunkStoreView`. The `ChunkStore` aggregates chunk requirements across all views: chunks are kept loaded if any view needs them, and disposed when no views reference them. Priority for loading is determined by taking the minimum priority across all views.

`ChunkStoreView` maintains an efficient sparse `Map<Chunk, ChunkViewState>` rather than tracking all chunks. View state aggregation happens synchronously during the update cycle, with `maybeForgetChunk()` called inline to clean up no-longer-visible chunks within a view.

`chunkManager.update()` is now called *after* rendering, as chunk view states are updated as the runtime iterates over viewports to render. This performs the view state aggregation and ultimately chunk queueing and discarding.

### Related Issue
For architecture diagrams and class relationships, see the [multiple viewports tech spec](https://docs.google.com/document/d/1OwGL7KW34hxFAENyOm2WiQ71E58XGu9Y7jKkJlyRx84/edit?tab=t.0).

This is the final PR of the multi-viewport implementation plan (see tech spec):
  - #407: WebGL infrastructure
  - #419: Viewport class
  - #451: Viewport events
  - #506: Idetik API integration
  - #531 (this PR): `ChunkManagerSource` refactor

### Tests & Checks
  - Existing chunk management tests updated to use new `ChunkStore`/`ChunkStoreView` API
  - All tests pass
  - Existing examples work without modification
  - Added a second `ChunkedImageLayer` to the multi-viewport example and confirmed it behaves as expected, with chunk memory sharing confirmed (according to stats.js)
  - Validated that chunks are properly disposed when no views need them (memory goes back down)


packages/core/src/layers/chunked_image_layer.ts
```diff
@@ -166,8 +170,16 @@ export class ChunkedImageLayer extends Layer implements ChannelsEnabled {
...
   }
 
-  public get chunkManagerSource(): ChunkManagerSource | undefined {
-    return this.chunkManagerSource_;
+  public get chunkStore(): ChunkStore | undefined {
+    return this.chunkStoreView_?.store;
+  }
+
+  public get chunkStoreView(): ChunkStoreView | undefined {
+    return this.chunkStoreView_;
```
andy-sweet: Nit

To clarify, as with `chunkManagerSource` before, these are only exposed for `ChunkInfoOverlay`, right? And we could imagine removing them in the future?

In that case, I'd probably remove `chunkStore` since it can be accessed as `chunkStoreView.store`.

packages/core/src/idetik.ts
```diff
@@ -163,15 +162,11 @@ export class Idetik {
...
-      if (viewport.camera.type === "OrthographicCamera") {
-        this.chunkManager_.update(
-          viewport.camera as OrthographicCamera,
-          viewport.getBoxRelativeTo(this.canvas).toRect().width
-        );
-      }
       this.renderer_.render(viewport);
     }
 
+    this.chunkManager_.update();
```
andy-sweet, line 168: This change in order of render/update is worth calling out in the PR description, especially because I've temporarily forgotten why it is needed...

Intuitively (without any other context), I think I'd expect update to come first followed by render (i.e. load then present), but in practice I'm not sure it's critical.

packages/core/src/renderers/webgl_renderer.ts
```diff
@@ -88,15 +88,15 @@ export class WebGLRenderer extends Renderer {
     const frustum = viewport.camera.frustum;
 
     for (const layer of opaque) {
-      layer.update();
+      layer.update({ viewport });
```
andy-sweet: Nit: you could define the context object once outside the loops, though no idea if this has any practical effect on performance.

packages/core/src/core/chunk_manager.ts
```diff
@@ -1,76 +1,156 @@
...
-  ) {
+  ): Promise<ChunkStoreView> {
+    const store = await this.addSource(source);
+    const view = new ChunkStoreView(store, policy);
+    this.views_.set(store, (this.views_.get(store) ?? []).concat(view));
+    return view;
+  }
+
+  public removeView(view: ChunkStoreView): void {
+    // TODO: log or throw if store or view is not found
```
andy-sweet: Nit: are you planning to add this in this PR?

packages/core/src/core/chunk_manager.ts
```diff
@@ -1,76 +1,156 @@
...
+    chunk.orderKey = orderKeyForMinPriority;
+
+    if (chunk.priority !== null && chunk.state === "unloaded") {
+      chunk.state = "queued";
+    } else if (chunk.priority === null && chunk.state === "queued") {
+      chunk.state = "unloaded";
+    }
+
+    if (chunk.priority === null && chunk.state === "loaded") {
+      store.disposeChunk(chunk);
```
andy-sweet: The new simpler disposal logic seems to reintroduce the bug that following PR fixed: https://github.com/chanzuckerberg/idetik/pull/524/files

I can reproduce that by running the main chunk streaming example, panning, and observing that the number of ready chunks drops while prefetched chunks are refetched.

On main, the key is understanding the difference between `chunk.prefetch` and `eligibleForPrefetch` (which takes into account LOD and loading state). I'm not sure how best to translate that into this PR.

andy-sweet (changes requested):
No major issues here since I looked at a slightly older version of this, but I did find one bug that needs to be addressed, hence the request for changes. Even after addressing that, I think this PR should get more than just my approval before being merged - let me know if that doesn't make sense!

packages/core/src/core/chunk_manager.ts
aganders3: Thanks for the reminder - I'll take a closer look at this.

aganders3: Can you check with [913c9b4](https://github.com/chanzuckerberg/idetik/pull/531/commits/913c9b4a3c731436cba82ccca68d490246b415dc)?

I removed some outdated stuff that was accidentally included from before the loading policies were added.

I also made a change to pretty much treat `prefetch` as `eligibleForPrefetch` which I think is more intuitive. It means not trying to juggle prefetch/loaded as mutually exclusive states, but as separate properties (which they are).

packages/core/src/layers/chunked_image_layer.ts
aganders3: Yep, that's correct and a good suggestion!

packages/core/src/core/chunk_manager.ts
aganders3: Sorry I should have called this out for discussion. I'm happy to update that in this PR but wanted input on throw vs. log. Do you have any thoughts?

packages/core/src/core/chunk_manager.ts
andy-sweet: That fixes the bug. I still find it hard to reason about the prefetch property, but that was true before this PR too. I agree that this PR makes things less confusing by removing `eligibleForPrefetch`.

packages/core/src/core/chunk_manager.ts
```diff
@@ -1,76 +1,160 @@
...
+    chunk.priority = minPriority;
+    chunk.orderKey = orderKeyForMinPriority;
+
+    if (chunk.priority !== null && chunk.state === "unloaded") {
+      chunk.state = "queued";
+    } else if (chunk.priority === null && chunk.state === "queued") {
+      chunk.state = "unloaded";
+    }
+
+    if (this.shouldDispose(chunk)) {
```
andy-sweet: Optional

I'd prefer to inline `shouldDispose` since it's only used here and is not dissimilar to the ones above. For similar reasons, I think the currently public `ChunkStore.disposeChunk` should be moved here and probably inlined too.

andy-sweet: I also think this in line with my expectations of the ChunkManager managing chunks (i.e. mutating their state including their optional data), vs. the `ChunkStore` which just stores and allows loading of the chunks.

packages/core/src/core/chunk_manager.ts
andy-sweet: I don't have a strong preference. Typically, I prefer throwing, but in this case I think continuing seems safe enough, so logging works too.

andy-sweet (approved):
I'm approving this to acknowledge that I previously requested changes, but you should probably at least one other approval before merging.

packages/core/src/core/chunk_manager.ts
aganders3: Thanks - I think both suggestions are good changes. Probably an intermediate `shouldDispose` boolean is enough to make it more self-documenting.

packages/core/src/core/chunk_manager.ts
aganders3: I opted to throw since I think an error here represents some corrupted state in the ChunkManager

shlomnissan:
@aganders3 @andy-sweet Continuing the conversation about `ChunkStore` and `ChunkStoreView`:

I still think their relationship is flipped:
- Right now `ChunkManager` tracks `views_` per store and even does a reverse lookup.
- It also owns `aggregateChunkViewStates` and `updateAndCollectChunkChanges`.
These are per-store responsibilities, yet they live in the manager.

A cleaner split IMO:
- `ChunkStore` owns its `ChunkStoreView[]`. It’s the place where all view states are merged into final chunk state.
- `ChunkManager` owns the set of `ChunkStores` (one per `ChunkSource`) and the global `ChunkQueue` store lifecycle.
- `ChunkStoreView` focused on cameras, viewports, and policies. Computes per-view `ChunkViewState`. Nothing else.
- `ChunkStore` just collects those states and resolves the final flags for each chunk.

This feels like the right interim model. Thoughts?

I’ll keep reviewing/thinking about it tomorrow.

These are the open questions I have:
- How do we reprioritize when a viewport is minimized?
- What does sharing a single layer across multiple viewports look like? (separate views per layer?)
- Is the layer manager even useful now?

**I'm happy to stamp as it is so we can work on the demo and we can revisit this design later**

aganders3:
Initially I instead had `views_: ChunkStoreView[]` as a property on `ChunkStore`, which I think is what you're suggesting. This also works, but I lifted some of the responsibilities out to the ChunkManager in order to break the parent/child relationship and (direct) cycle between the `ChunkStore` and `ChunkStoreView`. However I agree it created this awkward set of maps:
```typescript
class ChunkManager {
  private readonly stores_ = new Map<ChunkSource, ChunkStore>();
  private readonly views_ = new Map<ChunkStore, ChunkStoreView[]>();
```

The other motivation was this way a layer would only need a reference to a `ChunkStoreView` instead of also holding a reference to the corresponding `ChunkStore`. Given we have `ChunkStoreView.store` as a public property, maybe this is a distinction without a difference. We can also avoid needing this reference by adding `ChunkStoreView.dispose()` to have it remove itself from `ChunkStore._views` (instead of needing `ChunkStoreView.removeView(view)`).

I think you're right though, and I'm talking myself (back) into the design you're proposing...

> * How do we reprioritize when a viewport is minimized?

I think this is still an open question, but I honestly haven't given it much thought. It does seem reasonable to distinguish between "closed" and "minimized", where a minimized viewport would continue to request data in the background but at a lower priority, but it may need to keep track of some things e.g. what size it expects to be restored to. This could also be related to disposal policies and resource management. For example I think Neuroglancer maintains some number of chunks in a "recent" state before freeing memory.

> * What does sharing a single layer across multiple viewports look like? (separate views per layer?) (from Slack: I assume the same layer will need to have separate store views?)

In the current implementation this is not supported. The simple workaround is to make another instance of the layer and it can share some, though I recognize this is a little unsatisfactory and perhaps unintuitive.

I don't really like the idea of a layer having multiple `ChunkStoreView` instances because I think it would also need to maintain separate lists of renderables for each (or its list of renderables will be constantly changing), but there might be a better solution I'm not thinking of.

I have an open question which is: what _should_ happen if a layer is in multiple viewports? I'm thinking here if a `ChunkedImageLayer` is placed in a 2D viewport (as we have) and then we also want to see it represented in the 3D view intersecting a volume rendering. In Neuroglancer for example the 3D view ("sections") is a representation of what you see in the 2D views. If the layer had a `ChunkStoreView` for each viewport, it would show the whole slice in 3D, not the crop of the viewport (what we probably want).

> * Is the layer manager even useful now?

The layer manager has had pretty small utility in general to this point, so I'm not sure. Ultimately it is a list, and with 3D views and OIT even the ordering in the list is perhaps not as important as we once thought. Maybe it could be useful as an intermediary for managing layers in multiple viewports, instead of introducing multi-viewport-aware logic in a layer itself? That's a pretty speculative idea though.

PR #532, opened 2025-11-18
feat: allow simple LOD and time coordinate selection in volume layer

seankmartin:
### Summary

This is designed to help us in discussing and getting feedback on how we should integrate the volume layer with the chunk manager. This first pass shown here is really designed to allow quickly setting up volume rendering itself, with the focus being on picking a time + lod combo, loading all chunks corresponding to that, and doing the rendering.

This was to allow making quick progress on rendering without getting too bogged down on data loading.

I'm very sorry this description is a bit brief! I'd be happy to point to specific details or chat through this together as to how to integrate this later.

Please see https://github.com/chanzuckerberg/idetik/tree/seankmartin-aranega/develop-demo-marching for the example shown in the image, that branch builds on top of this simple data loading to actually do blending, turn off depth testing, and do ray marching based on the camera position:

image

image

### DIFF
--- a/packages/core/src/layers/volume_layer.ts
+++ b/packages/core/src/layers/volume_layer.ts
@@ -139,6 +139,8 @@ export class VolumeLayer extends Layer {
       this.lastLoadedTime_ === this.sliceCoords_.t
     )
       return chunks;
+    this.clearObjects();
+    this.releaseAndRemoveChunks(this.visibleChunks_.keys());
     Logger.debug("VolumeLayer", `Loading chunks for LOD ${this.lod_}`);
     for (const chunk of chunks) {
       this.chunkManagerSource_.loadChunkData(
--- a/packages/core/src/layers/volume_layer.ts
+++ b/packages/core/src/layers/volume_layer.ts
@@ -2,7 +2,10 @@ import { Chunk, ChunkSource, SliceCoordinates } from "../data/chunk";
 import { Layer, LayerOptions } from "../core/layer";
 import { VolumeRenderable } from "../objects/renderable/volume_renderable";
 import { IdetikContext } from "../idetik";
-import { ChunkManagerSource } from "../core/chunk_manager_source";
+import {
+  ChunkManagerSource,
+  INTERNAL_POLICY_KEY,
+} from "../core/chunk_manager_source";
 import { ImageSourcePolicy } from "../core/image_source_policy";
 import { Texture3D } from "../objects/textures/texture_3d";
 import { Logger } from "../utilities/logger";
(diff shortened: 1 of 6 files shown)


packages/core/src/layers/volume_layer.ts
```diff
@@ -1,24 +1,183 @@
...
+    if (this.state !== "ready") this.setState("ready");
+  }
 
-    this.addObject(renderable);
-    this.setState("ready");
+  private releaseAndRemoveChunks(chunks: Iterable<Chunk>) {
+    for (const chunk of chunks) {
+      const volume = this.visibleChunks_.get(chunk);
+      if (volume) {
+        this.visibleChunks_.delete(chunk);
```
aganders3, line 175: I think we should use a pool here, similar to the `ChunkedImageLayer`. Otherwise the texture data associated with each new volume renderable is never cleaned up and it will cause a memory leak. The renderer only disposes textures marked "stale" by their renderable object.

This can probably come in a separate PR, but should be flagged.

packages/core/src/objects/renderable/volume_renderable.ts
```diff
@@ -13,7 +13,8 @@ export class VolumeRenderable extends RenderableObject {
   ) {
     super();
     this.geometry = new BoxGeometry(width, height, depth, 1, 1, 1);
-    this.cullFaceMode = "back";
+    // TODO temp change, works with SimpleGeometry not box geometry
+    this.cullFaceMode = "none";
```
aganders3, line 17: In the end it doesn't matter but I'm used to culling *front* faces for volume rendering (drawing on the inside of a cube).

aganders3 (commented):
Sorry, this will need some changes to rebase on #531. The good news is this will also enable sharing data between a VolumeLayer and ChunkedImageLayer.

I have a couple comments here in the meantime, but happy to take another look after rebase. In general though this approach looks good to me.

packages/core/src/objects/renderable/volume_renderable.ts
seankmartin: Yes we intended the cull the front faces, I should have been clearer with the comment my bad! We had two different geometries, a BoxGeometry which was previously introduced. And a SimpleBoxGeometry which we introduced. For some reason we were having issues with face culling on the BoxGeometry and instead of getting too bogged down in that with everything else needed for volume rendering we introduced the SimpleBoxGeometry. But to keep the diff a bit cleaner here this PR excluded the SimpleBoxGeometry -- and as such also turned off face culling

packages/core/src/layers/volume_layer.ts
seankmartin: https://github.com/chanzuckerberg/idetik/pull/553 includes that pool. If 553 is too large we can considering splitting some back out as a separate PR. I'll leave this PR open for now as a reference but will likely close it in favor of 553 or something split off from 553.

Thank you very much for the feedback here about this!

PR #534, opened 2025-11-18
feat: add AWS S3 authentication support for private data

phoenixAja:
### Summary
Adds support for loading OME-Zarr data from private S3 buckets with AWS Signature V4 authentication. This enables secure access to private datasets during local development.

### Changes

**New `S3FetchStore` class** (`packages/core/src/data/zarr/s3_fetch_store.ts`)
- Extends `FetchStore` with AWS Signature V4 authentication
- Generates fresh signatures per-request for both GET and HEAD requests
- Uses Web Crypto API for browser compatibility
- **Signing key caching**: Caches AWS signing keys per date/region combination (valid for 24 hours). Reduces crypto overhead by ~60-80% after the first request of the day
- **Localhost-only safety check**: Prevents usage outside local development environments (localhost, 127.0.0.1). Throws error on production domains to prevent accidental credential exposure

**OmeZarrImageSource updates**
(`packages/core/src/data/ome_zarr/image_source.ts`)
- Added `fromS3()` static factory method for creating authenticated S3 sources
- New `S3OmeZarrImageSourceProps` type for S3-specific configuration

**Worker thread support** (`packages/core/src/data/zarr/open.ts`)
- Added "s3" type to `ZarrArrayParams` union for serialization
- Updated `createZarrArrayParams()` to serialize S3FetchStore properties (credentials, region, etc.)
- Updated `openArrayFromParams()` to reconstruct S3FetchStore from serialized params
- Enables authenticated chunk loading in worker threads

### Usage
```typescript
const source = OmeZarrImageSource.fromS3({
  url: 's3-url',
  version: undefined,
  credentials: {
    accessKeyId: 'YOUR_ACCESS_KEY',
    secretAccessKey: 'YOUR_SECRET_KEY',
    sessionToken: 'YOUR_SESSION_TOKEN'  // optional
  },
  region: 'us-east-1'
});

await source.open();
```

Security & Limitations

- Local development only: Safety checks prevent usage outside localhost environments
- Credentials in memory: AWS credentials are stored in memory and passed to worker threads
- S3 CORS required: Bucket must be configured with CORS to allow:
  - AllowedMethods: GET, HEAD
  - AllowedHeaders: Authorization, x-amz-*, Content-Type, Range
  - ExposeHeaders: Content-Length, ETag

Testing

- Tested with private S3 datasets using temporary credentials from VCP data CLI
- Verified all existing examples work correctly


phoenixAja:
I'm wondering if this is an overly complicated solution (originally i didn't think that every path request needed it's own signature generation). Perhaps a backend proxy (downloads and hosts the zarr file) is a better solution? Let me know if you have thoughts.

andy-sweet:
> I'm wondering if this is an overly complicated solution (originally i didn't think that every path request needed it's own signature generation). Perhaps a backend proxy (downloads and hosts the zarr file) is a better solution? Let me know if you have thoughts.

My initial feeling is that this is overly complex for similar reasons - I also thought that the signature would be generated once for bucket access and could be passed through the request header in a much more straightforward way. The other reason is the potential security issue of the browser having access to these credentials at all, though they are temporary and I see your mitigation, so maybe that's not a big issue.

But I'm still going to take a deeper look at this because 

1) I'm interested to see how this actually works
2) I feel like there might be a way to simplify things in terms of the API

packages/core/src/data/ome_zarr/image_source.ts
```diff
@@ -13,19 +12,43 @@ import {
...
   readonly version?: OmeZarrVersion;
+  readonly fetchOptions?: FetchOptions;
 
   /**
    * @param url URL of Zarr root
+   * @param version OME-Zarr version
+   * @param fetchOptions Optional fetch configuration (e.g., authentication headers for private S3 data)
    */
-  constructor(url: string, version?: OmeZarrVersion);
+  constructor(
```
andy-sweet: Regardless if this PR goes in, I'm going to create an issue to improve the API of `OmeZarrImageSource` - the overloaded constructor is already awkward and if we add anything else (as in this PR), it becomes borderline unreadable.

packages/core/src/data/ome_zarr/image_source.ts
```diff
@@ -13,19 +12,43 @@ import {
...
+/** Options for customizing fetch requests, such as adding authentication headers for private S3 data */
+export type FetchOptions = {
+  /** RequestInit overrides to customize fetch behavior (e.g., custom headers for S3 authentication) */
+  overrides?: RequestInit;
+  /** Whether to use suffix requests for range queries */
+  useSuffixRequest?: boolean;
+  /** AWS credentials for S3 authentication (will generate signatures per-request) */
+  credentials?: AwsCredentials;
+  /** AWS region for S3 bucket (e.g., 'us-east-1') */
+  region?: string;
```
andy-sweet: I would not expect these to be part of generic fetch options because they are specific to S3. I think there's a way to restructure the types and functions in this PR to avoid this because the authenticated fetch store has the `credentials` and `region` properties.

andy-sweet:
In the example in the PR description

> const source = new OmeZarrImageSource(s3Url, undefined, {
  credentials: { accessKeyId, secretAccessKey, sessionToken },
  awsConfig: { region: 'us-east-1', service: 's3' }
});

is s3Url the S3 URI or the equivalent HTTPS URL?

packages/core/src/data/zarr/authenticated_fetch_store.ts
```diff
@@ -0,0 +1,272 @@
...
+
+/**
+ * A FetchStore that generates AWS Signature V4 headers for each request.
+ * This is necessary because AWS signatures are path-specific and expire quickly.
+ *
+ * SECURITY WARNING: This class is only intended for local development.
+ * Credentials are passed to worker threads and stored in memory.
+ * Do not use in production - implement a secure backend proxy instead.
+ */
+export class AuthenticatedFetchStore extends FetchStore {
```
andy-sweet: Given this is S3 specific, I would consider renaming this to `S3Store` and would probably just implement `AsyncReadable` using an instance of `FetchStore` as part of the private implementation.

aganders3:
I think this may be fine to unblock our DCA work, but I agree it seems complicated. I also don't love the restriction that users will have to host the app locally. I think this removes a lot of the benefit of using a web-based workflow to begin with. It's no easier for users than running a local proxy (though it should be more performant).

That might be unavoidable if what users need is a full-on AWS secret key, and for this method you need that to make the signatures. Even if it was a single signature it runs the same risk.

I was really hoping for something more like [cloudfront signed cookies](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-signed-cookies.html#private-content-how-signed-cookies-work) where my understanding is we could get a cookie from a small app and make subsequent https requests with a constant modification of requests to add `credentials: 'include'`:
```
fetch('https://1230-492390-48blahblah.cloudfront.net/path/to/some/zarr/chunk', {
  credentials: 'include'
})
```

packages/core/src/data/ome_zarr/metadata_loaders.ts
```diff
@@ -55,9 +59,16 @@ function removeProperty<O, P extends keyof O>(obj: O, prop: P): Omit<O, P> {
 
 export async function loadOmeZarrPlate(
   url: string,
-  version?: Version
+  version?: Version,
+  fetchOptions?: {
```
andy-sweet: Ah, I understand the need for `(Authenticated)FetchOptions` a bit more now. That being said, I'm pretty tempted to remove these utility functions from the library and move them to the single core example where they're used.

packages/core/src/data/ome_zarr/image_source.ts
phoenixAja: agree! I was debating whether this was the place to fix it (in this PR), but this new overload is making things more complicated :(

phoenixAja:
> I think this may be fine to unblock our DCA work, but I agree it seems complicated. I also don't love the restriction that users will have to host the app locally. I think this removes a lot of the benefit of using a web-based workflow to begin with. It's no easier for users than running a local proxy (though it should be more performant).
> 
> That might be unavoidable if what users need is a full-on AWS secret key, and for this method you need that to make the signatures. Even if it was a single signature it runs the same risk.
> 
> I was really hoping for something more like [cloudfront signed cookies](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-signed-cookies.html#private-content-how-signed-cookies-work) where my understanding is we could get a cookie from a small app and make subsequent https requests with a constant modification of requests to add `credentials: 'include'`:
> 
> ```
> fetch('https://1230-492390-48blahblah.cloudfront.net/path/to/some/zarr/chunk', {
>   credentials: 'include'
> })
> ```

I wish but this doesn't seem to be the case, we need to generate new credentials for each url requested (signature requires url as part of the hash). I can look into the cloudfront stuff to see if there's something that i missed potentially though, thanks for suggesting that

packages/core/src/data/ome_zarr/image_source.ts
phoenixAja: yeah, it's a bit tricky because we need these fetchOptions to initiate the AuthenticatedFetchStore :/

packages/core/src/data/ome_zarr/metadata_loaders.ts
phoenixAja: you mean the functions in this file? Or what utility functions specifically?

phoenixAja:
> In the example in the PR description
> 
> > const source = new OmeZarrImageSource(s3Url, undefined, {
> > credentials: { accessKeyId, secretAccessKey, sessionToken },
> > awsConfig: { region: 'us-east-1', service: 's3' }
> > });
> 
> is s3Url the S3 URI or the equivalent HTTPS URL?

it's the https S3 URL - so it looks like `https://${S3_CONFIG.bucket}.s3.amazonaws.com/${S3_CONFIG.key}`

packages/core/src/data/ome_zarr/metadata_loaders.ts
andy-sweet: I was thinking about `loadOmeZarrPlate` and `loadOmeZarrWell`. They are utility functions for loading the OME-Zarr HCS metadata, which is only sometimes relevant. We use them in a core example. One application that needs these metadata is organelle-box and it does not use these functions.

phoenixAja: yeah then it would be nice to extract them then, but not here ideally. It would be a good clean up task (as well as clean up the constructor code that we were talking about earlier)

andy-sweet: A simpler idea is to not add support for them when they are on a private S3 bucket in this PR. They do not currently support the file-system store, so that is consistent.

Along with other changes, I think that would allow us to simplify the public API (i.e. not export so many AWS-specific types).

packages/core/src/data/ome_zarr/image_source.ts
andy-sweet: I made a suggestion in https://github.com/chanzuckerberg/idetik/pull/539

I think some form of that should go in before this PR.

andy-sweet (changes requested):
I have not looked too closely at the implementation of the `AuthenticatedFetchStore`, but as it's well isolated I'm happy to approve those changes.

But I can't approve this PR as it is right now because

1) the `OmeZarrImageSource` constructor signature becomes too complicated
2) the `OmeZarrImageSource` constructor implementation becomes too complicated
3) we expose new types (e.g. `FetchOptions`, `AuthenticatedFetchOptions`) in the constructor and part of the main index which are mostly related to AWS credentials

I think something like #539 effectively fixes (1) and (2). And then I think (3) can be fixed pretty easily from there by following some of my inline comments.

phoenixAja:
> I have not looked too closely at the implementation of the `AuthenticatedFetchStore`, but as it's well isolated I'm happy to approve those changes.
> 
> But I can't approve this PR as it is right now because
> 
> 1. the `OmeZarrImageSource` constructor signature becomes too complicated
> 2. the `OmeZarrImageSource` constructor implementation becomes too complicated
> 3. we expose new types (e.g. `FetchOptions`, `AuthenticatedFetchOptions`) in the constructor and part of the main index which are mostly related to AWS credentials
> 
> I think something like #539 effectively fixes (1) and (2). And then I think (3) can be fixed pretty easily from there by following some of my inline comments.

so then do you want to merge #539 and then we can update this to use the new constructor and then you're good to merge this? That's fine with me, but we need to make sure everyone is ok with pushing a breaking change to main

andy-sweet:
> But I can't approve this PR as it is right now because
> 
> 1. the `OmeZarrImageSource` constructor signature becomes too complicated
> 2. the `OmeZarrImageSource` constructor implementation becomes too complicated
> 3. we expose new types (e.g. `FetchOptions`, `AuthenticatedFetchOptions`) in the constructor and part of the main index which are mostly related to AWS credentials
> 
> I think something like #539 effectively fixes (1) and (2). And then I think (3) can be fixed pretty easily from there by following some of my inline comments.

I have one last request before approving, which is to address (3). I made an attempt at that in https://github.com/chanzuckerberg/idetik/pull/542 to show you what I mean.

packages/core/src/data/ome_zarr/image_source.ts
```diff
@@ -92,6 +97,24 @@ export class OmeZarrImageSource {
     });
   }
 
+  /**
+   * Creates an OmeZarrImageSource from an S3 HTTP(S) URL.
+   *
+   * @param url URL of Zarr root
```
andy-sweet, line 103: I now realize these docs are a lie given that there's only a `props` param. That's true for the other factory methods, so I'll fix that in another PR...

packages/core/src/data/zarr/open.ts
```diff
@@ -112,12 +116,23 @@ export function createZarrArrayParams(
   arrayPath: string,
   zarrVersion: Version | undefined
 ): ZarrArrayParams {
-  if (location.store instanceof FetchStore) {
+  if (location.store instanceof S3FetchStore) {
```
andy-sweet, line 119: I forgot that we're effectively doing deserialization of the source/store here for serialization back on the web worker. Maybe that motivates a single `from` factory method with a corresponding `to`, but again not specific to this PR.

andy-sweet (approved):
Looks good now that this is separated quite nicely!

Thanks for the effort working out the S3 issues here. And for your patience on my reviews & suggestions.

andy-sweet:
Can you update the PR description to reflect the change?

PR #537, opened 2025-11-20
chore: configure dependabot to add a conventional commit prefix

aganders3:
### Summary
Every time dependabot makes a PR it fails the PR title test. I'm hoping this will fix it.

I'm not sure it will take effect for the existing PRs, so we might need to wait for the next to confirm it's working.

I also added configs to skip the title check for draft PRs just to show a bit less red in the PR list.

### DIFF
--- a/.github/depentabot.yml
+++ b/.github/depentabot.yml
@@ -5,9 +5,8 @@ updates:
     schedule:
       interval: "weekly"
     commit-message:
-      # fix will trigger a patch release
+      # see .releaserc.json in the repo root for how these prefixes map to version bumps
       prefix: "fix"
-      # chore will *also* trigger a patch release
       prefix-development: "chore"
       # includes the scope in the prefix, like this:
       # fix(deps):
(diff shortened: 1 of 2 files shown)


.github/depentabot.yml
```diff
@@ -0,0 +1,15 @@
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    commit-message:
+      # fix will trigger a patch release
+      prefix: "fix"
+      # chore does not trigger a release
```
phoenixAja: for us chore triggers a patch release (see `.releaserc.json`)

aganders3: I think that's still fine, but maybe w just use `chore` for both in that case?

aganders3: Or maybe use `build` or `ci` for dev deps?

I'm not familiar enough with JS to know what kinds of updates could represent breaking changes downstream. Like once our package is built does that mean even the runtime deps are kind of "baked in"? Would it only be peer deps that cause dependency hell?

phoenixAja (approved):
ty!

.github/depentabot.yml
phoenixAja: i think having fix and chore are both ok here (and chore makes sense)! I was just saying that for you to update the comment.

PR #539, opened 2025-11-25
feat: use factory methods instead of overloaded constructors for OmeZarrImageSource

### LINKED ISSUE #536: Improve OmeZarrImageSource creation
> Regardless if this PR goes in, I'm going to create an issue to improve the API of `OmeZarrImageSource` - the overloaded constructor is already awkward and if we add anything else (as in this PR), it becomes borderline unreadable. 

 _Originally posted by @andy-sweet in [#534](https://github.com/chanzuckerberg/idetik/pull/534/files#r2546350168)_

One simple way to improve the creation API is to use a type to define the arguments, so we can define mutually exclusive arguments explicitly. Alternatively, we may wish to consider factory functions and a private constructor.

andy-sweet:
### Summary
This changes the public way to create an `OmeZarrImageSource` from using overloaded constructors to using factory methods. This improves readability of the implementation and will likely improve the readability of external usage, especially as we add more source types (e.g. private s3 buckets in #534) and arguments.

### Related Issue
Closes #536

### Tests & Checks


andy-sweet:
@phoenixAja : assigned you for review because this came up during #534 and will conflict with the changes there.

We don't use factory methods much, but I think the approach works quite well here. Some alternatives would be to pass a union of types to a public constructor or define factory functions.

andy-sweet:
This is technically a breaking change, but I'm not sure if adding `!` to the `feat` in the commit title will cause a major version increment which is undesirable.

phoenixAja:
> This is technically a breaking change, but I'm not sure if adding ! to the feat in the commit title will cause a major version increment which is undesirable.

I also thought about factory method for this but the breaking change gave me pause (also yes adding feat! will cause a major version bump)

phoenixAja (approved):
even though this is breaking i think it's ok.. out 0 based versioning tells users that our API is stabilizing anyways.

PR #540, opened 2025-11-26
refactor: use SourceDimensionMap instead of LoaderAttributes 

### LINKED ISSUE #417: Remove `LoaderAttributes`
> `ChunkDimensionMap` stores all the information that `LoaderAttributes` in a slightly more ergonomic way for most usage (i.e. dimension first, rather than LOD-first), so we can remove the attributes here. We could consider removing it everywhere, but it is currently used in multiple applications.

_Originally posted by @andy-sweet in https://github.com/chanzuckerberg/idetik/pull/400#discussion_r2299026253_

andy-sweet:
### Summary
This removes `LoaderAttributes` in favor of using `SourceDimensionMap`. This mostly simplifies existing logic in examples, but it may complicate logic in some of the old loading paths related to `Region`.

### Related Issue
Closes #417

### Tests & Checks

### DIFF
--- a/packages/core/src/data/ome_zarr/image_loader.ts
+++ b/packages/core/src/data/ome_zarr/image_loader.ts
@@ -7,7 +7,6 @@ import {
   SourceDimension,
   SourceDimensionMap,
   isChunkData,
-  LoaderAttributes,
   ChunkData,
   ChunkDataConstructor,
 } from "../chunk";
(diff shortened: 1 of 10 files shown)


andy-sweet:
I marked this as draft because it's a breaking change and there is some risk around the old loading path. I think we should definitely bring this change in once that old loading path is removed, but until then it may not be worth it.

andy-sweet:
@aganders3 : I assigned you for review because I think I remember that you were supportive of this idea. While this is in draft to avoid accidental merging, I think it is reviewable as is. Not a high priority though.

packages/core/src/data/ome_zarr/image_loader.ts
```diff
@@ -210,18 +207,24 @@ export class OmeZarrImageLoader {
...
     };
 
-    const xOffset = calculateOffset(indices.length - 1);
-    const yOffset = calculateOffset(indices.length - 2);
+    const xLod = this.dimensions_.x.lods[lod];
+    const xIndex = indices[indices.length - 1];
+    const xOffset = calculateOffset(xIndex, xLod);
+    const yIndex = indices[indices.length - 2];
+    const yLod = this.dimensions_.y.lods[lod];
+    const yOffset = calculateOffset(yIndex, yLod);
```
aganders3, line 227:
```suggestion
    const xLod = this.dimensions_.x.lods[lod];
    const xIndex = indices[this.dimensions_.x.index];
    const xOffset = calculateOffset(xIndex, xLod);
    const yIndex = indices[this.dimensions_.y.index];
    const yLod = this.dimensions_.y.lods[lod];
    const yOffset = calculateOffset(yIndex, yLod);
```

Unless I'm still mixing up source/world spaces, I think this is effectively the same but clearer and with fewer assumptions (or centralized assumptions).

aganders3 (approved):
In general this looks great! I think it's fine to optimize for simplicity in the newer code paths, and I don't think the added complexity in the `Region`-based loading is too bad.

This is a breaking change however because it's removing the public `getAttributes`. I'm pretty sure some of the react viewer code relies on that method.

andy-sweet:
> This is a breaking change however because it's removing the public getAttributes. I'm pretty sure some of the react viewer code relies on that method.

Correct. But I have made this ready for review now, as after your feedback and another look, I think this is worth bringing in. I'm happy to make the necessary changes on the react side, especially as there is a usable older version.

andy-sweet:
@aganders3 : you approved the draft, but any objections to me merging this breaking change now it's ready for review?

aganders3:
No objections - LGTM!

PR #543, opened 2025-12-03
fix: refactor size observer for immediate resizing

shlomnissan:
### Summary

This PR fixes a glitch where the canvas would look stretched for a moment when
resizing the viewport. The `PixelSizeObserver` relied on the animation loop
calling `getAndResetChanged()` for resizing which introduced a one frame delay.
I added a callback instead of waiting for the next animation frame to mitigate it.

### Tests & Checks

I verified this visually. See before/after videos below.

#### Before

https://github.com/user-attachments/assets/53ef7b6f-1acf-4fd5-aa89-fc324ba58db5

#### After

https://github.com/user-attachments/assets/0484675e-5caa-4fe6-91ef-bbcf6dadd2f2

### DIFF
--- a/packages/core/src/idetik.ts
+++ b/packages/core/src/idetik.ts
@@ -112,7 +112,12 @@ export class Idetik {
         sizeDependents.push(viewport.element);
       }
     }
-    this.sizeObserver_ = new PixelSizeObserver(sizeDependents);
+    this.sizeObserver_ = new PixelSizeObserver(sizeDependents, () => {
+      this.updateSize();
+      for (const viewport of this.viewports_) {
+        this.renderer_.render(viewport);
+      }
+    });
   }
 
   public get renderedObjects() {
@@ -155,11 +160,6 @@ export class Idetik {
 
   private animate(timestamp?: DOMHighResTimeStamp) {
     if (this.stats_) this.stats_.begin();
-
-    // Must resize before render b/c changing canvas coordinate space clears it.
-    if (this.sizeObserver_.getAndResetChanged()) {
-      this.updateSize();
-    }
 
     for (const viewport of this.viewports_) {
       this.renderer_.render(viewport);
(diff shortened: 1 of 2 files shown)


packages/core/src/utilities/pixel_size_observer.ts
```diff
@@ -7,10 +7,11 @@ export class PixelSizeObserver {
   private resizeObserver_?: ResizeObserver;
   private mediaQuery_?: MediaQueryList;
   private onMediaQueryChange_?: () => void;
-  private changed_ = false;
+  private onResize_: () => void;
```
andy-sweet: Optional

Consider naming this `onChange` because it's called for both an element resize (via `ResizeObserver`) and a pixel density change. Naming was generally a headache here because of the overloaded meaning of size of the elements and pixel size of the canvas buffer, so very much optional.

Maybe one way to solve this is to split this back into two things and handle a little extra complexity in the runtime, but that's probably better handled in a separate PR.

packages/core/src/idetik.ts
```diff
@@ -112,7 +112,12 @@ export class Idetik {
         sizeDependents.push(viewport.element);
       }
     }
-    this.sizeObserver_ = new PixelSizeObserver(sizeDependents);
+    this.sizeObserver_ = new PixelSizeObserver(sizeDependents, () => {
+      this.updateSize();
```
andy-sweet: Optional

Consider inlining `updateSize` here since this is no longer in the main animation loop. Or moving the render calls into `updateSize`. I think the split adds more redirection which hurts readability here.

andy-sweet (approved):
Ah, I noticed this recently too, so thanks for the fix! This is an improvement in code readability too.

For context, this was introduced by @bchu1 in https://github.com/chanzuckerberg/idetik/pull/294 where there were initially some rendering/flickering issues, which I do not observe here (I think because this maintains the order of updating the size then rendering). I'm not sure why we used the flag instead of immediately rendering, but maybe we just wanted to only render in the animation loop?

Given this update, I wonder if we can simplify things further, but that can wait for another PR.

packages/core/src/idetik.ts
shlomnissan: I'll inline for now.

PR #544, opened 2025-12-03
feat: add damping to orbit controls

shlomnissan:
### Summary

This PR adds inertial damping to `OrbitControls` to create smoother camera
interactions. To support this the `CameraControls` interface has been updated to
include an `onUpdate(dt)` method and the main runtime loop now calculates and
propagates delta time (in seconds).

I'll add damping to the zoom/pan camera in another PR.

### Tests & Checks

Visual verification. See video below

https://github.com/user-attachments/assets/ac2508bf-9552-4a13-be2d-946a94322f10

### DIFF
--- a/packages/core/src/idetik.ts
+++ b/packages/core/src/idetik.ts
@@ -10,7 +10,7 @@ import {
 import { PixelSizeObserver } from "./utilities/pixel_size_observer";
 
 type Overlay = {
-  update(idetik: Idetik, dt?: number): void;
+  update(idetik: Idetik): void;
 };
 
 type IdetikParams = {
@@ -34,7 +34,7 @@ export class Idetik {
   private readonly stats_?: Stats;
   private readonly sizeObserver_: PixelSizeObserver;
   private lastAnimationId_?: number;
-  private lastTimestamp_: number = 0;
+  private lastTimestampSinceStart_: number = 0;
 
   /**
    * Creates a new Idetik visualization runtime instance.
@@ -155,7 +155,7 @@ export class Idetik {
       this.sizeObserver_.connect();
 
       this.lastAnimationId_ = requestAnimationFrame((timestamp) => {
-        this.lastTimestamp_ = timestamp;
+        this.lastTimestampSinceStart_ = timestamp;
         this.animate(timestamp);
       });
     } else {
@@ -167,8 +167,10 @@ export class Idetik {
   private animate(timestamp: DOMHighResTimeStamp) {
     if (this.stats_) this.stats_.begin();
 
-    // cap at 100ms to prevent spiraling on tab switching
-    const dt = Math.min(timestamp - this.lastTimestamp_, 100) / 1000;
+    // cap dt to prevent large time-step jumps when resuming from background tabs
+    const dt = Math.min(timestamp - this.lastTimestampSinceStart_, 100) / 1000;
+
+    this.lastTimestampSinceStart_ = timestamp;
 
     for (const viewport of this.viewports_) {
       viewport.cameraControls?.onUpdate(dt);
@@ -178,7 +180,7 @@ export class Idetik {
     this.chunkManager_.update();
 
     for (const overlay of this.overlays) {
-      overlay.update(this, dt);
+      overlay.update(this);
     }
 
     if (this.stats_) this.stats_.end();
--- a/packages/core/src/idetik.ts
+++ b/packages/core/src/idetik.ts
@@ -34,7 +34,9 @@ export class Idetik {
   private readonly stats_?: Stats;
   private readonly sizeObserver_: PixelSizeObserver;
   private lastAnimationId_?: number;
-  private lastTimestampSinceStart_: number = 0;
+
+  // this value will be set after start
+  private lastTimestamp_: DOMHighResTimeStamp = 0;
 
   /**
    * Creates a new Idetik visualization runtime instance.
@@ -155,7 +157,7 @@ export class Idetik {
       this.sizeObserver_.connect();
 
       this.lastAnimationId_ = requestAnimationFrame((timestamp) => {
-        this.lastTimestampSinceStart_ = timestamp;
+        this.lastTimestamp_ = timestamp;
         this.animate(timestamp);
       });
     } else {
@@ -168,9 +170,9 @@ export class Idetik {
     if (this.stats_) this.stats_.begin();
 
     // cap dt to prevent large time-step jumps when resuming from background tabs
-    const dt = Math.min(timestamp - this.lastTimestampSinceStart_, 100) / 1000;
+    const dt = Math.min(timestamp - this.lastTimestamp_, 100) / 1000;
 
-    this.lastTimestampSinceStart_ = timestamp;
+    this.lastTimestamp_ = timestamp;
 
     for (const viewport of this.viewports_) {
       viewport.cameraControls?.onUpdate(dt);
--- a/packages/core/src/idetik.ts
+++ b/packages/core/src/idetik.ts
@@ -10,7 +10,7 @@ import {
 import { PixelSizeObserver } from "./utilities/pixel_size_observer";
 
 type Overlay = {
-  update(idetik: Idetik, timestamp?: DOMHighResTimeStamp): void;
+  update(idetik: Idetik, dt?: number): void;
 };
 
 type IdetikParams = {
@@ -25,7 +25,6 @@ export type IdetikContext = {
 };
 
 export class Idetik {
-  private lastAnimationId_?: number;
   private readonly chunkManager_: ChunkManager;
   private readonly context_: IdetikContext;
   private readonly renderer_: WebGLRenderer;
@@ -34,6 +33,8 @@ export class Idetik {
   public readonly overlays: Overlay[];
   private readonly stats_?: Stats;
   private readonly sizeObserver_: PixelSizeObserver;
+  private lastAnimationId_?: number;
+  private lastTimestamp_: number = 0;
 
   /**
    * Creates a new Idetik visualization runtime instance.
@@ -152,24 +153,32 @@ export class Idetik {
         viewport.events.connect();
       }
       this.sizeObserver_.connect();
-      this.animate();
+
+      this.lastAnimationId_ = requestAnimationFrame((timestamp) => {
+        this.lastTimestamp_ = timestamp;
+        this.animate(timestamp);
+      });
     } else {
       Logger.warn("Idetik", "Idetik runtime already started");
     }
     return this;
   }
 
-  private animate(timestamp?: DOMHighResTimeStamp) {
+  private animate(timestamp: DOMHighResTimeStamp) {
     if (this.stats_) this.stats_.begin();
 
+    // cap at 100ms to prevent spiraling on tab switching
+    const dt = Math.min(timestamp - this.lastTimestamp_, 100) / 1000;
+
     for (const viewport of this.viewports_) {
+      viewport.cameraControls?.onUpdate(dt);
       this.renderer_.render(viewport);
     }
 
     this.chunkManager_.update();
 
     for (const overlay of this.overlays) {
-      overlay.update(this, timestamp);
+      overlay.update(this, dt);
     }
 
     if (this.stats_) this.stats_.end();
--- a/packages/core/src/objects/cameras/controls.ts
+++ b/packages/core/src/objects/cameras/controls.ts
@@ -5,6 +5,7 @@ import { EventContext } from "../../core/event_dispatcher";
 const LEFT_MOUSE_BUTTON = 0;
 
 export interface CameraControls {
+  onUpdate(dt: number): void;
   onEvent(event: EventContext): void;
 }
 
@@ -34,6 +35,8 @@ export class PanZoomControls implements CameraControls {
         break;
     }
   }
+
+  public onUpdate(_delta: number) {}
 
   private onWheel(event: EventContext) {
     if (!event.worldPos || !event.clipPos) return;
--- a/packages/core/src/objects/cameras/orbit_controls.ts
+++ b/packages/core/src/objects/cameras/orbit_controls.ts
@@ -5,7 +5,6 @@ import { Spherical } from "../../math/spherical";
 
 import { glMatrix, vec3 } from "gl-matrix";
 import { clamp } from "../../utilities/clamp";
-import { lerp } from "../../utilities/lerp";
 
 const MOUSE_BUTTON_NONE = -1;
 const MOUSE_BUTTON_LEFT = 0;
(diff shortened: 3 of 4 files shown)


packages/core/src/idetik.ts
```diff
@@ -152,24 +153,32 @@ export class Idetik {
...
     for (const viewport of this.viewports_) {
+      viewport.cameraControls?.onUpdate(dt);
       this.renderer_.render(viewport);
     }
 
     this.chunkManager_.update();
 
     for (const overlay of this.overlays) {
-      overlay.update(this, timestamp);
+      overlay.update(this, dt);
```
andy-sweet: Why did this need to change in this PR? You could still pass timestamp here, right?

I'm not immediately aware there are any overlay implementations that use the timestamp, but that means there are probably none that use the delta either, so there's no motivation for this change, which also complicates this PR.

packages/core/src/idetik.ts
```diff
@@ -34,6 +33,8 @@ export class Idetik {
   public readonly overlays: Overlay[];
   private readonly stats_?: Stats;
   private readonly sizeObserver_: PixelSizeObserver;
+  private lastAnimationId_?: number;
+  private lastTimestamp_: number = 0;
```
andy-sweet: Optional

At first I was concerned about where 0 came from as an initial value, and would have thought that Performance.timeOrigin or Performance.now() evaluated on the first render pass would be more appropriate.

But now I realize that this value is never actually used, so I think this is fine, but might warrant an inline comment to explain that this value will be clobbered after `start`.

Alternatively, you could just leave this as being possibly undefined (for the first pass), in which case the delta could be 0 as it is now. Or you could avoid this state entirely and pass both the last and current timestamp into `animate`.

packages/core/src/idetik.ts
```diff
@@ -152,24 +153,32 @@ export class Idetik {
...
       Logger.warn("Idetik", "Idetik runtime already started");
     }
     return this;
   }
 
-  private animate(timestamp?: DOMHighResTimeStamp) {
+  private animate(timestamp: DOMHighResTimeStamp) {
     if (this.stats_) this.stats_.begin();
 
+    // cap at 100ms to prevent spiraling on tab switching
```
andy-sweet: By spiraling, I assume you mean the camera? If so, then this clamp and comment better belongs in the camera update itself since there could be multiple consumers of this delta.

packages/core/src/objects/cameras/orbit_controls.ts
```diff
@@ -55,6 +67,25 @@ export class OrbitControls implements CameraControls {
     }
   }
 
+  public onUpdate(dt: number) {
```
andy-sweet, line 75: Optional

For durations of type `number`, include the unit to clarify - i.e. `deltaSeconds` or `dtSecs`.

Optional because if I were going to guess a unit, I'd guess the SI unit of seconds, but I don't like to guess - especially when ms durations are not uncommon in code.

packages/core/src/objects/cameras/controls.ts
```diff
@@ -5,6 +5,7 @@ import { EventContext } from "../../core/event_dispatcher";
 const LEFT_MOUSE_BUTTON = 0;
 
 export interface CameraControls {
+  onUpdate(dt: number): void;
```
andy-sweet, line 8: Optional

`onUpdate` makes me think this is a callback responding to an `update` function somewhere, but that's not the case. For that reason, I'd consider `update` instead.

packages/core/src/objects/cameras/orbit_controls.ts
```diff
@@ -13,6 +14,7 @@ const MOUSE_BUTTON_MIDDLE = 1;
 const ORBIT_SPEED = 0.009;
 const PAN_SPEED = 0.001;
 const ZOOM_SPEED = 0.0009;
+const DAMPING_FACTOR = 3;
```
andy-sweet: Optional

I would expect this to be included in `OrbitParams`. Optional because the default value works well enough for the 3D example.

packages/core/src/idetik.ts
```diff
@@ -152,24 +153,32 @@ export class Idetik {
...
+    const dt = Math.min(timestamp - this.lastTimestamp_, 100) / 1000;
```
andy-sweet, line 173: I might be sleepy, but when do you update `this.lastTimestamp_` after the initial call to `requestAnimationFrame`?

andy-sweet (changes requested):
Examples work well enough for me, but there's a couple of minor blocking issues for me.

1) There's no need or motivation to change the `Overlay` interface in this PR.
2) It seems like `Idetik.lastTimestamp_` is never updated after the first call to `requestAnimationFrame`.

aganders3:
I like this feature! I think this implementation happens to feel okay for the default values, but the approach breaks down for low damping and the controls become very sluggish. Instead I would expect behavior similar to the [three.js example](https://threejs.org/examples/?q=controls#misc_controls_arcball) where the control is responsive while dragging, but the camera continues moving after release and damping slows it to a stop.

To do this I would propose tracking camera velocity (angular + linear) as state instead of a target position. Then apply damping to the velocity rather than interpolating toward a target position. This is intuitive because when damping is 0 there is no "target position" (the camera would orbit forever).

As a nit I also think the PR title should change to reflect this is a "feat" not a "fix"

shlomnissan:
> There's no need or motivation to change the Overlay interface in this PR.

@andy-sweet the current timestamp parameter is dead code that's not used by the overlays anyway. I swapped it for `dt` because that's the value calculations typically need downstream. I would rather remove the time parameter entirely over keeping a timestamp and it's a small enough change that doesn't require its own PR.

I get the desire to keep the PR focused but I consider passing the timestamp around bad since it may lead to recomputing dt in a bunch of different places and this was a low-cost opportunity to make it explicit.

https://docs.google.com/document/d/1oDLs5DusTBd9ACpma0T0vMgm4DHTsMUiCCiO95fdROQ/edit?disco=AAABuJrKCNQ

shlomnissan:
> Instead I would expect behavior similar to the three.js example where the control is responsive while dragging, but the camera continues moving after release and damping slows it to a stop.

@aganders3 Ahh that's a great call! I didn't actually look at other resources. I always assumed this approach is simple an intuitive but what you're describing sounds much better.

shlomnissan:
> As a nit I also think the PR title should change to reflect this is a "feat" not a "fix"

Ha! I'm still getting used to this.

packages/core/src/objects/cameras/controls.ts
shlomnissan: I think I prefer to keep it as if for consistency with `onEvent`.

packages/core/src/objects/cameras/orbit_controls.ts
shlomnissan: I agree that it might be more precise but I feel like it's redundant. The renderers I'm familiar with use delta time in seconds as the standard simulation step. When I read `deltaTimeSeconds` I think there must be a `timestampMs` nearby and the explicit naming is a defensive choice.

packages/core/src/idetik.ts
shlomnissan: I disagree that it complicates the PR. I shared my thoughts about this here: https://github.com/chanzuckerberg/idetik/pull/544#issuecomment-3613010193

But I will remove the timestamp altogether for now.

packages/core/src/idetik.ts
shlomnissan: I renamed it to `lastTimestampSinceStart_` so it's clearer. Do you think it addresses the initial concern?

packages/core/src/idetik.ts
shlomnissan: The clamp is in the right place. It prevents large time-step when resuming from background tabs. I will update the comment so it's clear.

packages/core/src/idetik.ts
shlomnissan: Ahh I completely missed it and the safety clamp masked the issue!

shlomnissan:
@aganders3 I followed a similar implementation to three.js which implements frame-based damping. I think the implementation I have here is slightly more robust but it means that the damping factor and behavior don't map 1:1 with three.js. I also switched to friction per frame so the damping factor is scaled down.

packages/core/src/idetik.ts
andy-sweet: I personally find the new name more confusing than before (i.e. `lastTimestamp_`). A timestamp is always with respect to some epoch, and I think this name implies that epoch is the last call to `Idetik.start`, but from the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/DOMHighResTimeStamp), I don't think that's the case (unless we know that coincides with when navigation starts?).

Also, I did not notice this before, but `number` broadens the type beyond `DOMHighResTimeStamp` since the latter describes useful context and constraints about what this number represents.

So with something like 

```
private lastTimestamp_: DOMHighResTimeStamp = 0;
```

I'm happy enough, because 0 is a reasonable default in this context (i.e. we guess that no duration has passed since navigation starts, until we actually have a value).

My preferred solution is to avoid state entirely and pass both the last and current timestamp to `animate` as shown here: https://github.com/chanzuckerberg/idetik/compare/shlomnissan/fix/control-damping-bc21...andy-sweet/timestamp?expand=1

But adopting that is optional!

packages/core/src/idetik.ts
andy-sweet: > I disagree that it complicates the PR

The change to overlays has no effect on the functionality or implementation of this feature, but introducing it here increases the code diff and triggers this conversation, so it clearly complicates the PR - I think that's inarguable.

I understand you think that using a delta improves the overlay design (and removing the timestamp certainly simplifies it!). I appreciate your thoughts and would have been happy to address or review them in a separate PR. 

But they are out of place in this PR, which is why I originally requested them to be reverted. Given that the known practical effect on overlays is zero, and this has already taken up too much time, I'm fine with just letting this slide - but would appreciate avoiding these types of changes in the future.

packages/core/src/idetik.ts
```diff
@@ -152,24 +153,34 @@ export class Idetik {
...
     }
     return this;
   }
 
-  private animate(timestamp?: DOMHighResTimeStamp) {
+  private animate(timestamp: DOMHighResTimeStamp) {
     if (this.stats_) this.stats_.begin();
 
+    // cap dt to prevent large time-step jumps when resuming from background tabs
+    const dt = Math.min(timestamp - this.lastTimestampSinceStart_, 100) / 1000;
```
andy-sweet: I'm still not convinced about the clamp in general (e.g. I'm not sure I would want the clamped value when implementing playback), but as it's only being used by the camera controls, I'm fine for now.

packages/core/src/idetik.ts
```diff
@@ -34,6 +33,8 @@ export class Idetik {
   public readonly overlays: Overlay[];
   private readonly stats_?: Stats;
   private readonly sizeObserver_: PixelSizeObserver;
+  private lastAnimationId_?: number;
+  private lastTimestampSinceStart_: number = 0;
```
```suggestion
  private lastTimestampSinceStart_: DOMHighResTimeStamp = 0;
```

andy-sweet: Justified in https://github.com/chanzuckerberg/idetik/pull/544#discussion_r2590602114

andy-sweet (approved):
Good enough for me, though I'd appreciate using the more meaningful type for the timestamp and reverting the name.

packages/core/src/idetik.ts
shlomnissan: Yeh I think you're right if we look at it in playback terms. This is mostly a standard safeguard for physics to prevent the simulation from exploding when you switch tabs and come back to a large dt. In a dedicated 3D engine I'd usually use a clamped dt everywhere but in the context of Idetik maybe it's better to just clamp it locally

packages/core/src/idetik.ts
shlomnissan: You're right it certainly complicated the review given we exchanged several messages about it.

When I disagreed I was looking at it strictly from a code complexity standpoint (since the change only impacted two lines of code) but I see your point on including it distracted from the PR's main purpose.

I generally don't believe PRs need to be strictly atomic. I prefer leaving the code a bit better than I found it. But I agree this is something we should align on as a team to avoid this kind of friction in the future.

packages/core/src/idetik.ts
shlomnissan: I reverted the name and added a comment.

packages/core/src/objects/cameras/orbit_controls.ts
```diff
@@ -55,6 +70,32 @@ export class OrbitControls implements CameraControls {
     }
   }
 
+  public onUpdate(dt: number) {
+    this.currPos_.phi += this.orbitVelocity_.phi;
+    this.currPos_.theta += this.orbitVelocity_.theta;
+    this.currPos_.radius += this.zoomVelocity_ * this.currPos_.radius;
```
aganders3: Optional: could this be captured by a single `sphericalVelocity`?

packages/core/src/objects/cameras/orbit_controls.ts
```diff
@@ -55,6 +70,32 @@ export class OrbitControls implements CameraControls {
...
+    // stable rotation and avoid gimbal-like artifacts.
+    const limit = Math.PI / 2 - glMatrix.EPSILON;
+    this.currPos_.theta = clamp(this.currPos_.theta, -limit, limit);
+    this.currPos_.radius = Math.max(0.01, this.currPos_.radius);
+
+    const p = vec3.add(vec3.create(), this.currCenter_, this.currPos_.toVec3());
+    this.camera_.transform.setTranslation(p);
+    this.camera_.transform.targetTo(this.currCenter_);
+
+    const damping = Math.pow(1.0 - this.dampingFactor_, dt * 60);
```
aganders3: I think this 60 comes from a presumed/target FPS?

packages/core/src/objects/cameras/orbit_controls.ts
```diff
@@ -55,6 +70,32 @@ export class OrbitControls implements CameraControls {
...
+    this.orbitVelocity_.phi *= damping;
+    this.orbitVelocity_.theta *= damping;
+    this.zoomVelocity_ *= damping;
```
aganders3: I think you should add a threshold here to set velocities to 0.0 below some epsilon. Otherwise this may take many frames to actually reach zero. It should also save some work to return early from this method if velocity is exactly 0.0.

aganders3 (approved):
Behavior looks good to me now - thanks! It's a small change but I think a threshold to set slow velocities to true zero should be included.

packages/core/src/objects/cameras/orbit_controls.ts
shlomnissan: I considered it but decided the computation is trivial so I skipped it. But I will make the change, early return makes sense.

packages/core/src/objects/cameras/orbit_controls.ts
shlomnissan: Yes. It's still framerate independent but I'm using this value as a normalization constant. I'll extract it into a named constant.

shlomnissan:
@andy-sweet I'm going to merge but feel free to remove the clamp or revert delta time when you work on playback buffering if needed.

Issue #538, opened 2025-11-25
3D example constantly disposing chunk


andy-sweet:
When running the OME-Zarr v0.5 example with a maximized browser window on a 1080p monitor, the logger constantly logs the following messages

```
[2025-11-25T18:45:14.098Z][DEBUG][ChunkManager] Disposing chunk {"x":0,"y":0,"z":1,"c":0,"t":0} in LOD 3
logger.ts:126 [2025-11-25T18:45:14.098Z][DEBUG][ChunkManager] Disposing chunk {"x":0,"y":1,"z":1,"c":0,"t":0} in LOD 3
logger.ts:126 [2025-11-25T18:45:14.098Z][DEBUG][ChunkManager] Disposing chunk {"x":1,"y":0,"z":1,"c":0,"t":0} in LOD 3
logger.ts:126 [2025-11-25T18:45:14.098Z][DEBUG][ChunkManager] Disposing chunk {"x":1,"y":1,"z":1,"c":0,"t":0} in LOD 3
```

I think these are intended to be prefetched fallback chunks that are not visible (i.e. we start at z=0).

If that's true, I believe this bug may be incurring a decent amount of work by continuously disposing then fetching these chunks. Even if the fetch responses are cached, they still need to be decompressed.

I think this bug was introduced in #514, likely because that slightly changes the definition of what prefetch means.

andy-sweet:
One idea I had to recapture the prior definition of disposal is to add an extra condition to [`shouldDisposeChunk`](https://github.com/chanzuckerberg/idetik/blob/ca1b622b2c68fc961f7e9671cbe3d1ef3ae4c1de/packages/core/src/core/chunk_manager.ts#L167) which would check if `chunk.lod !== store.getLowestResLOD()` only disposing if that's true.

andy-sweet:
One idea that @aganders3 had is to change the definition of prefetch according to the following diff

```diff
diff --git a/packages/core/src/core/chunk_store_view.ts b/packages/core/src/core/chunk_store_view.ts
index 5cd6644f..eeb603b8 100644
--- a/packages/core/src/core/chunk_store_view.ts
+++ b/packages/core/src/core/chunk_store_view.ts
@@ -261,11 +261,14 @@ export class ChunkStoreView {
       const isCurrentLOD = chunk.lod === this.currentLOD_;
       const isFallbackLOD = chunk.lod === this.store_.getLowestResLOD();
 
-      const prefetch =
[...]

PR #545, opened 2025-12-03
refactor: Move ChunkStoreView ownership to ChunkStore

aganders3:
### Summary
This is a follow-up for #531 as discussed in https://github.com/chanzuckerberg/idetik/pull/531#issuecomment-3560334403.

The main purpose of this refactor is to transfer ownership of the `ChunkStoreView` instances to their corresponding `ChunkStore`, away from `ChunkManager`.
* Added natural `ChunkStore.views_ : ChunkStoreView[]` 
* Removed confusing mapping and reverse-lookup on `ChunkManager`
* Moved store-specific functions to `ChunkStore`:
  * `updateAndCollectChunkChanges`
  * `aggregateChunkViewStates`
* `ChunkStoreView.store` is no longer a public attribute
* Automated cleanup of `ChunkStore` when no views remain

### Related Issue
#486

### Tests & Checks
Tested all examples

### DIFF
--- a/packages/core/src/core/chunk_manager.ts
+++ b/packages/core/src/core/chunk_manager.ts
@@ -14,14 +14,7 @@ export class ChunkManager {
     policy: ImageSourcePolicy
   ): Promise<ChunkStoreView> {
     const store = await this.getOrCreateStore(source);
-    const view = new ChunkStoreView(store, policy);
-    store.addView(view);
-
-    if (store.views.length === 1) {
-      this.stores_.set(source, store);
-      this.pendingStores_.delete(source);
-    }
-
+    const view = store.createView(policy);
     return view;
   }
 
@@ -43,7 +36,12 @@ export class ChunkManager {
 
     const newPending = initializeStore();
     this.pendingStores_.set(source, newPending);
-    return newPending;
+
+    const store = await newPending;
+    this.stores_.set(source, store);
+    this.pendingStores_.delete(source);
+
+    return store;
   }
 
   public update() {
@@ -64,7 +62,7 @@ export class ChunkManager {
     this.queue_.flush();
 
     for (const [source, store] of this.stores_) {
-      if (store.views.length === 0) {
+      if (store.canDispose()) {
         this.stores_.delete(source);
       }
     }
--- a/packages/core/src/core/chunk_manager.ts
+++ b/packages/core/src/core/chunk_manager.ts
@@ -1,5 +1,4 @@
-import { Logger } from "../utilities/logger";
-import { Chunk, ChunkSource } from "../data/chunk";
+import { ChunkSource } from "../data/chunk";
 import { ChunkQueue } from "../data/chunk_queue";
 import { ChunkStore } from "./chunk_store";
 import { ChunkStoreView } from "./chunk_store_view";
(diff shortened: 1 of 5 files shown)


packages/core/src/core/chunk_manager.ts
```diff
@@ -8,80 +7,48 @@ import { ImageSourcePolicy } from "./image_source_policy";
...
-      throw new Error("Cannot remove view: store not managed by ChunkManager");
-    }
+    store.addView(view);
 
-    const index = views.indexOf(view);
-    if (index === -1) {
-      throw new Error(
-        `Cannot remove view: view not found in store's view list (source: ${source})`
-      );
+    if (store.views.length === 1) {
```
andy-sweet: Optional

It took me a while to reason about what the body of this is doing and how the condition is related. I think is handling the case of effectively transferring the pending store to an existing store, and it does so only after the first view is added.

I think it's more clear to handle this logic in `getOrCreateStore` itself by awaiting `newPending` and then executing the body of this condition after that - that's because that function is handling the checking and creation of pending/existing stores.

packages/core/src/core/chunk_manager.ts
```diff
@@ -8,80 +7,48 @@ import { ImageSourcePolicy } from "./image_source_policy";
...
-
-  public removeView(view: ChunkStoreView): void {
-    const store = view.store;
-    const source = this.getSourceForStore(store);
-
-    const views = this.views_.get(store);
-    if (!views) {
-      throw new Error("Cannot remove view: store not managed by ChunkManager");
-    }
+    store.addView(view);
```
andy-sweet: Optional

Given the new relationship between `ChunkStore` and `ChunkStoreView`, I might expect `ChunkStore.addView` (maybe with a different name) to handle creating the `ChunkStoreView` and return it, which would simplify things a little here.

andy-sweet (approved):
This simplifies a few things (e.g. the implementation `ChunkManager`) at the cost of the introducing a circular dependency between `ChunkStore` and `ChunkStoreView`. 

For me, that's fine for now, so I'm approving. All my comments are optional.

I'd also prefer that the fix for #538 were not included in this PR, since the code movement means that the diff does not express the change well. The fix works for me, which is enough for me to approve, but I did not look too closely at the new logic (and admittedly the old logic was hard for me to follow!).

packages/core/src/core/chunk_manager.ts
aganders3: This is a little confusing. Previously it was handled in `getOrCreateStore` but doing this here prevents a race to delete a store that was just created and does not yet have any views.

Perhaps another way to handle this is with a flag on `ChunkStore` to indicate it's okay to remove.

aganders3:
Regarding the fix for #538: I at least put it in its own commit (with minimal noise) here: https://github.com/chanzuckerberg/idetik/pull/545/commits/1e575f5f59e01180a7088046176e9ee676279197 but I'm happy to move that to a separate PR... I was just being lazy.

packages/core/src/core/chunk_manager.ts
andy-sweet: > doing this here prevents a race to delete a store that was just created 

Is this related to the `delete` call in `ChunkManager.update`? If so, I think I see the order of operations that could cause this, though I've never been clear on what atomicity is guaranteed around async/await in javascript (because it looks like the actual operations should be effectively synchronous since they all occur after open `source.open` which is the source of async here).

I think you could fix that by adding an extra check to `ChunkManager.update` to check for `this.pendingStores_.has(source)`, which adds a little complexity while maybe improving overall clarity.

Anyway, this is even more optional now, so I'll leave it up to you.

PR #546, opened 2025-12-04
fix: Prevent unwanted disposal of fallback chunks and add sanity check

### LINKED ISSUE #538: 3D example constantly disposing chunk

aganders3:
### Summary
This is intended to fix #538, unwanted disposal of fallback chunks (not visible or prefetch, but assigned a priority) by disposing all chunks with null priority. This also adds a logic sanity check and will throw an error if any visible or prefetch chunks have null priority. See `chunk_store.ts`: 202-219.

### Related Issue
Closes #538

### Tests & Checks
Tested all examples, in particular the OME-Zarr v0.5 example that was exhibiting the bug

### DIFF
--- a/packages/core/src/core/chunk_store.ts
+++ b/packages/core/src/core/chunk_store.ts
@@ -209,13 +209,18 @@ export class ChunkStore {
     }
 
     const shouldDisposeChunk =
-      chunk.state === "loaded" && !chunk.visible && !chunk.prefetch;
+      chunk.state === "loaded" && chunk.priority === null;
     if (shouldDisposeChunk) {
+      if (chunk.visible || chunk.prefetch) {
+        throw new Error(
+          `Chunk state inconsistency detected: priority is null but visible=${chunk.visible} or prefetch=${chunk.prefetch} ` +
+            `for chunk ${JSON.stringify(chunk.chunkIndex)} in LOD ${chunk.lod}`
+        );
+      }
+
       chunk.data = undefined;
       chunk.state = "unloaded";
-      chunk.priority = null;
       chunk.orderKey = null;
-      chunk.prefetch = false;
       Logger.debug(
         "ChunkStore",
         `Disposing chunk ${JSON.stringify(chunk.chunkIndex)} in LOD ${chunk.lod}`


packages/core/src/core/chunk_store.ts
```diff
@@ -209,13 +209,18 @@ export class ChunkStore {
     }
 
     const shouldDisposeChunk =
-      chunk.state === "loaded" && !chunk.visible && !chunk.prefetch;
+      chunk.state === "loaded" && chunk.priority === null;
     if (shouldDisposeChunk) {
```
andy-sweet, line 213: This has become small enough where inlining it improves readability.

```suggestion
if (chunk.state === "loaded" && chunk.priority === null) {
```

packages/core/src/core/chunk_store.ts
```diff
@@ -209,13 +209,18 @@ export class ChunkStore {
...
+      if (chunk.visible || chunk.prefetch) {
+        throw new Error(
```
andy-sweet, line 215: I'm fine with this, though it makes me yearn for a true dev/debug assert in javascript.

andy-sweet (approved):
Thanks for separating this out! The disposal logic is much more clear to me now!

packages/core/src/core/chunk_store.ts
andy-sweet: Oh, now I realize we have similar vars for the other conditions. Better to keep these consistent rather than change one. Ideally, I'd like to inline them all, but totally optional!

PR #547, opened 2025-12-05
feat: introduce new uniforms and fix volume jittering when zoomed out

aranega:
### Summary
This PR introduces:

1. fix of the volume chunks jittering when zooming out,
2. new uniforms for the camera and the layers that can be injected in the shader program from the base-class or sub-classes,
3. new entries in the menu to control those uniforms,
4. face culling fix,
5. volume shader cleaning.

The jittering that we could see when zooming out on the rendered volume came from the way the position was passed from the vertex shader to the fragment shader.

The uniforms that have been introduced are simply relying on the same abstract method pattern that is also used to introduce uniforms from the objects to render.

We cleaned the volume shader by also removing some allocation that was done inside the for-loop.

### Tests & Checks
I manually tested the changes by:

1. running the "volume" example
6. manually clicking on the different menus to activate/change the way the volume is rendered
7. manually zooming out until the volume is small enough and rotate the camera to verify that the jitter we hadd disappeard


seankmartin (approved):

shlomnissan:
@aranega @seankmartin Is the intention that this will eventually be merged into `main` or are you preparing another PR?

seankmartin:
@shlomnissan we'd plan for a few more changes to the shader before making a PR to main, the intention of this is some fixes that might be helpful for the demo. And also a few more controls inside the volume rendering example itself which might be useful for debugging etc

shlomnissan:
@seankmartin Ahh that makes sense. Thanks! I will review the code shortly.

packages/core/examples/volume_rendering/index.html
```diff
@@ -36,13 +36,23 @@
...
+        background-color: rgba(0, 0, 0, 0.5);
+      }
     </style>
   </head>
   <body>
     <div id="main">
       <canvas id="canvas"></canvas>
     </div>
     <div id="time-point"></div>
+    <div id="lod"></div>
```
shlomnissan, line 55: Since the LOD is fixed to the user's selection I don't think the overlay is needed. We might want an LOD overlay once it's computed dynamically but I suggest we remove this for now.

packages/core/examples/volume_rendering/main.ts
```diff
@@ -56,6 +62,8 @@ idetik.start();
 const controls = {
   showWireframes: volumeLayer.debugMode,
   showTimePointOverlay: true,
+  showLodOverlay: true,
```
shlomnissan, line 65: If we remove the overlay we should remove this control as well.

packages/core/src/core/layer.ts
```diff
@@ -129,6 +129,7 @@ export abstract class Layer {
     const cameraPos = camera.position;
     const tmpA = vec3.create();
     const tmpB = vec3.create();
+    const DISTANCE_EPSILON = 1e-5;
```
shlomnissan, line 132: We typically use `glMatrix.EPSILON`.

packages/core/src/core/layer.ts
```diff
@@ -139,8 +140,21 @@ export abstract class Layer {
 
       const da = vec3.squaredDistance(cameraPos, tmpA);
       const db = vec3.squaredDistance(cameraPos, tmpB);
+      const diff = db - da;
```
shlomnissan, line 143: - Consider renaming `tmpA` and `tmpB` to something more descriptive like `centerA` and `centerB`.
- Is there a reason this sorting function lives on the base layer class rather than the volume-specific class? Sorting by squared distance is appropriate for volume bricks but it won’t match visual rendering order for typical mesh rendering which usually relies on view-space depth. Placing it on the volume class may avoid implying that square-distance sorting is a generic ordering strategy.

packages/core/src/core/layer.ts
```diff
@@ -139,8 +140,21 @@ export abstract class Layer {
...
     });
   }
+
+  /**
+   * Get uniforms for shader program. Override in derived classes that need custom uniforms.
+   * @returns Object containing uniform name-value pairs
+   */
+  public getUniforms(): Record<string, unknown> {
+    return {}; // Default implementation returns no uniforms
+  }
```
shlomnissan, line 159: I’m not sure about this pattern. I’d prefer that “uniforms” remain a renderer-only concept. In other words, non-renderer abstractions (like layers or cameras) should expose semantic properties (e.g. `maxIntensity`) and the renderer decides how to translate those into actual uniforms when iterating over the layers.

Concretely, if `VolumeLayer` has `this.maxIntensity` the renderer can detect the layer type and set the appropriate uniforms there. That keeps GPU details in the rendering layer and avoids leaking uniform management into the volume API.

Tagging @andy-sweet and @aganders3 here for their view on this since this might be an existing pattern.

packages/core/src/layers/volume_layer.ts
```diff
@@ -53,6 +60,14 @@ export class VolumeLayer extends Layer {
     return this.sourcePolicy_;
   }
 
+  public get hitMisses(): boolean {
+    return this.hitMisses_;
+  }
+
+  public set hitMisses(showHitMiss: boolean) {
+    this.hitMisses_ = showHitMiss;
+  }
```
shlomnissan, line 69: We should avoid trivial accessors/mutators. These functions add unnecessary boilerplate without providing any real encapsulation benefit. If you need have a trivial accessor and mutator just make the member variable public. I think this applies to all the new class members added here.

packages/core/src/layers/volume_layer.ts
```diff
@@ -225,6 +280,17 @@ export class VolumeLayer extends Layer {
 
     this.updateChunks();
   }
+
+  public getUniforms(): Record<string, unknown> {
```
shlomnissan, line 283: See https://github.com/chanzuckerberg/idetik/pull/547/files#r2593990984

packages/core/src/renderers/shaders/volume_vert.glsl
```diff
@@ -7,10 +7,10 @@ uniform mat4 Projection;
 uniform mat4 ModelView;
 
 out vec2 TexCoords;
-out vec4 NormalizedPosition;
+out vec3 ModelPosition;
```
shlomnissan: When I read `ModelPosition` I assume it's the position in world space. Maybe `Position` is a better name?

packages/core/src/renderers/shaders/volume_frag.glsl
```diff
@@ -39,75 +44,86 @@ bool intersectBox(vec3 rayOrigin, vec3 rayDir, vec3 boxMin, vec3 boxMax, out flo
...
     if (!hit) {
-        discard;
+        if (ShowHitMisses) {
+            fragColor = vec4(1.0, 0.0, 0.0, 1.0);
+            return;
+        }
+        // Because we later clamp the number of samples to at least 1,
+        // and position starts at entry point,
+        // we don't have to discard here
+        //discard;
```
shlomnissan: If that's the case I think you want to explicitly initialize `tEnter` and `tExit` otherwise the value is undefined right?

packages/core/src/renderers/shaders/volume_frag.glsl
```diff
@@ -39,75 +44,86 @@ bool intersectBox(vec3 rayOrigin, vec3 rayDir, vec3 boxMin, vec3 boxMax, out flo
...
-    // Model space ranges from -BoxSize/2 to +BoxSize/2, normalize to [0, 1]
-    vec3 nearPoint = (nearPointHomogeneous.xyz / nearPointHomogeneous.w) / BoxSize + 0.5;
-    vec3 farPoint = (farPointHomogeneous.xyz / farPointHomogeneous.w) / BoxSize + 0.5;
+    // Ray direction in view space points toward camera
+    vec3 rayOriginView = viewPos.xyz;
+    vec3 rayDirView = normalize(-rayOriginView);
 
-    // Find the start and end of the ray within the volume from the near point to the far point
+    // Transform to model space for intersection test
+    mat4 inverseModelView = inverse(ModelView);
```
shlomnissan: Doing this per fragment is extremely expensive. We can
- Precompute it on the CPU and pass as a separate uniform.
- Compute in the vertex shader and pass as a varying.
- Potentially avoid entirely by passing the camera position in model?

shlomnissan (changes requested):
I didn’t review the volume rendering algorithm in detail. Just a quick scan. I would allocate time for it before we merge to main. But I did call out a few things. The only blockers for me are potentially undefined behavior in the shader and expensive inverse computation per fragment.

aranega:
@shlomnissan Thanks a lot for the feedback! We were thinking about proposing those as different independent PRs on `main` later. I'll solve the points you raised regarding the volume shader in this PR, but I'll address your other points in the various PRs that will target `main`.

packages/core/src/renderers/shaders/volume_frag.glsl
aranega: I precomputed it on the CPU and passed it to the vertex shader where I moved the computation of the ray origin and direction to ease the computation in the vertex shader.

packages/core/src/renderers/shaders/volume_frag.glsl
aranega: `tEnter` and `tExit` are computed in the `intersectBox` they are output parameter so they shouldn't be uninitialized after the execution of the function, or did you spot something problematic in the `intersectBox` function?

I  normalized the box geometry also, so the computation slightly changed, but it's mainly the same.

packages/core/src/renderers/shaders/volume_vert.glsl
aranega: I added the computation of the ray direction and origin in the vertex shader, so I removed `ModelPosition` and substituted it in the fragment shader with the 2 new variables.

aranega: We went back to consider the ray direction computation in the fragment shader as we suspect that interpolation from the vertex shader makes the final result less precise

packages/core/src/renderers/shaders/volume_frag.glsl
aranega: We went back to consider the ray direction computation in the fragment shader [see here](https://github.com/chanzuckerberg/idetik/pull/547#discussion_r2599184916)

aranega:
@shlomnissan I'm closing this PR as I'll open another one targetting `main` with your notes took into account.

PR #548, opened 2025-12-09
feat: prevent double blending of chunked image layers' LODs

### LINKED ISSUE #372: ChunkedImageLayer does not support multi-channel images

andy-sweet:
### Summary
This uses the stencil buffer to enable correct blending across multiple chunked image layers, specifically in the expected case that the layer's chunk source has multiple LODs.

https://github.com/user-attachments/assets/1041cad5-80ad-49da-a906-e89b5946f387

The renderable objects are now sorted by LOD in decreasing resolution order so that higher resolution chunks are rendered first and claim pixels in the stencil buffer, preventing fragment shader passes for lower resolution chunks for those claimed pixels.

There is likely some overhead in using the stencil test, but skipping the fragment shader passes for the lower resolution chunks will also help performance.

### Related Issue
Closes #372

### Tests & Checks
I tested the examples.
I did some basic performance assessment on the main chunk streaming example comparing this branch with `main`, using the follow strategy.
1) I observed the stats.js overlay to check frame time as I explored the data. Qualitatively, I saw no difference - it was 0-1ms in both cases at almost all times.
2) I used Chrome performance profiling when loading the initial state of the example and qualitatively compared the traces. I saw no obvious difference.


packages/core/src/core/layer.ts
```diff
@@ -80,6 +80,10 @@ export abstract class Layer {
     return this.objects_;
   }
 
+  public useStencil(): boolean {
```
andy-sweet: This is my least favorite part of the PR because it's a new instance of public state about the layer directly influencing the behavior of the renderer (rather than via the layer's renderable objects). The naming is also very tied to OpenGL.

packages/core/src/renderers/webgl_renderer.ts
```diff
@@ -111,9 +113,24 @@ export class WebGLRenderer extends Renderer {
...
 
+  private initStencil() {
+    // We use the stencil buffer to mark pixels objects have been drawn,
+    // which is used to avoid overdrawing high resolution tiles with lower
+    // resolution ones.
+    const clearValue = 0;
+    this.gl_.clearStencil(clearValue);
+    this.gl_.stencilMask(0xff);
+    this.gl_.stencilFunc(this.gl_.EQUAL, clearValue, 0xff);
+    this.gl_.stencilOp(this.gl_.KEEP, this.gl_.KEEP, this.gl_.INCR);
```
andy-sweet, line 124: I'm not sure if the second argument (dpfail - stencil test passes, depth test fails) should be `KEEP` or `INCR`. If the depth test has failed and we depth testing is always enabled, I guess we should use `INCR` here to skip unnecessary checks?

shlomnissan:
@andy-sweet the demo looks great! I will try to get to it today/tomorrow. HQ week schedule is a bit crazy.

phoenixAja (approved):
looks good! The example works for me locally

shlomnissan:
@andy-sweet Can you explain the issue this solves and why stencil was preferred over other approaches?
The solution seems to highlight a draw-order assumption so I want to understand the constraints better.

packages/core/examples/image2d_from_omezarr4d_hcs/main.ts
```diff
@@ -64,34 +62,31 @@ const onImageChange = async () => {
...
-      camera.update();
-    }
+  const layer2 = new ChunkedImageLayer({
+    source,
+    sliceCoords: { t: 0, z: initZ, c: 2 },
+    channelProps: [{ color: [1, 0, 1], contrastLimits: contrastLimits[1] }],
+    policy,
+    transparent: true,
+    blendMode: "additive",
+    opacity: 1,
```
shlomnissan: I don't think this is needed since `1` is the default value for opacity.

shlomnissan:
Reading through the code again I think I understand the goal here: for a given image layer once a higher-LOD tile covers a pixel the lower-LOD tile for that layer shouldn’t also contribute, otherwise we’re effectively double-blending that layer.

Using stencil as an “already has high LOD” mask does achieve that but it also makes correctness depend on draw order (which is already the case anyway?) and ties this behavior to global stencil state which feels a bit weird but probably fine.

I’m okay with that as a scoped solution but we should be explicit that this is culling low-LOD not really blending across multiple layers. When I first read the summary I wasn't sure about the relationship between stencil and blending.

packages/core/src/renderers/webgl_renderer.ts
```diff
@@ -111,9 +113,24 @@ export class WebGLRenderer extends Renderer {
...
+  }
+
   private renderLayer(layer: Layer, camera: Camera, frustum: Frustum) {
     this.state_.setBlendingMode(layer.transparent ? layer.blendMode : "none");
-
+    const useStencil = layer.useStencil();
```
shlomnissan: Related to your previous comment, I think a cleaner approach would be something like:

```ts
const shouldUseStencil =
  layer.type === "ChunkedImageLayer" &&
  layer.hasMultipleObjects &&
  layer.hasMultipleLODs;
```

This feels specific to this layer type so we may not need a layer-level method at all.

If we do keep a getter on the layer naming it `shouldUseStencil` or `needsStencil` would be clearer than `useStencil`. The current name reads like a command rather than a flag and it’s a bit odd for the layer to dictate renderer level behavior.

packages/core/src/renderers/webgl_renderer.ts
shlomnissan: I think what you have here is correct. We only want to increment stencil when the fragment actually survives both the stencil and depth tests and contributes to the framebuffer.

shlomnissan (approved):
LGTM!

andy-sweet:
Thanks both for the quick reviews!

> Reading through the code again I think I understand the goal here: for a given image layer once a higher-LOD tile covers a pixel the lower-LOD tile for that layer shouldn’t also contribute, otherwise we’re effectively double-blending that layer.

I'm not sure if it was my original goal, but the double blending problem was an issue when I wanted to enable blending of multiple chunked image layers that had multiple LODs.
 
> Using stencil as an “already has high LOD” mask does achieve that but it also makes correctness depend on draw order (which is already the case anyway?) and ties this behavior to global stencil state which feels a bit weird but probably fine.

I also don't love the object order dependency on what is drawn. We could consider exposing something in `RenderableObject` like draw a priority that must be explicitly set by a layer, and which could then be used by a renderer appropriately. But I didn't want to overcomplicate this change, so avoided it for now.

> I’m okay with that as a scoped optimization but we should be explicit that this is culling low-LOD not really blending across multiple layers. When I first read the summary I wasn't sure about the relationship between stencil and blending.

The goal here was not optimization (though stenciling may optimize multi-resolution drawing in some cases), rather correctly blending layers that have multiple LODs. But I appreciate the confusion, so I reworded the PR title and summary to help clarify that.

shlomnissan:
Also worth mentioning, I'm pretty sure virtual textures can solve this problem as well:
- Draw one quad for the entire slice.
- Use a page table / indirection texture in the fragment shader that tells you:
  - which tile to sample
  - which mip level to use per pixel

Then the “only one LOD per pixel” logic lives entirely in the shader logic / page table update, not in draw order or stencil.

andy-sweet:
> Also worth mentioning, I'm pretty sure virtual textures can solve this problem as well:
> 
> * Draw one quad for the entire slice.
> * Use a page table / indirection texture in the fragment shader that tells you:
>   
>   * which tile to sample
>   * which mip level to use per pixel
> 
> Then the “only one LOD per pixel” logic lives entirely in the shader logic / page table update, not in draw order or stencil.

Once we have virtual texturing, I'd be supportive of removing this change and using that to solve this issue instead.

andy-sweet:
> @andy-sweet Can you explain the issue this solves and why stencil was preferred over other approaches?

I think I addressed the first part of this in my [other comment](https://github.com/chanzuckerberg/idetik/pull/548#issuecomment-3638528797). As for other approaches, I couldn't think of anything simple that would fix the same problem this does. The only idea that seemed like it would work without artifacts was filling an offscreen buffer in a separate render pass, but that seemed to be far more complex and likely less performant too.

shlomnissan:
> Once we have virtual texturing, I'd be supportive of removing this change and using that to solve this issue instead.

FWIW I can't think of a cleaner solution than what you have here with our current architecture

packages/core/src/renderers/webgl_renderer.ts
andy-sweet: Given this is only used for this layer, I like your suggestion and will implement something like it. In the future, I can imagine this be used by other layers (e.g. a chunked label image layer), but we can worry about that then.

andy-sweet: In the end, I implemented a compromise, which is to define `Layer.hasMultpleLODs` which like the now gone `Layer.useStencil` returns `false` by default but is conditionally true for `ChunkedImageLayer`. This feels less renderer implementation specific, but also allows us to avoid specific type checking/importing in the renderer.

aganders3 (approved):
Thanks @andy-sweet - nice work!

PR #549, opened 2025-12-09
fix: improve typing of WebGL context

andy-sweet:
### Summary
This makes `WebGLRenderer.gl_` non-nullable, but assigning the return of `getContext` to a local `gl` and checking that before assigning it to `gl_`. This allows us to remove the private `WebGLRenderer.gl` property that performs a static cast.

### Related Issue
Came up when writing #548 

### Tests & Checks
I manually tested the examples.


aganders3 (approved):
Nice eye - LGTM!

PR #552, opened 2025-12-12
chore: make examples site build failure throw

andy-sweet:
### Summary
This changes the initial build step to throw instead of log and error when generating the manifest fails. This causes the check to fail rather than log the error and pass.

### DIFF
--- a/packages/core/vite-plugin-examples-nav.js
+++ b/packages/core/vite-plugin-examples-nav.js
@@ -42,14 +42,9 @@ function examplesManifestPlugin() {
       if (!isExamplesBuild) {
         return;
       }
-
-      try {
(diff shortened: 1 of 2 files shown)


packages/core/vite-plugin-examples-nav.js
andy-sweet, line 61: Not sure if we should just allow throws here too. I avoided here because in the context of CI, we don't expect live changes. And a logged error might be fine for live development.

aganders3: This makes sense to me. I think because the files will continue to be watched you'd be able to see the error and fix it without restarting the dev server.

aganders3 (approved):
Thanks!

PR #553, opened 2025-12-16
feat: Introduce volume rendering

aranega:
### Summary
This PR introduces:

* fix of the volume chunks jittering when zooming out,
* new uniforms for the layers that can be injected in the shader program,
* new entries in the menu to control new uniforms,
* face culling fix,
* volume shader cleaning,
* new blending webgl method.

We cleaned the volume shader by also removing some allocation that was done inside the for-loop, and moved some expensive computation to the CPU side.

### Tests & Checks
We manually tested the changes by:

* running the "volume" example
* manually clicking on the different menus to activate/change the way the volume is rendered
* manually zooming out until the volume is small enough and rotate the camera to verify that the jitter we had disappeard


seankmartin:
One thing to note is that a part of the changes here are related to the multi viewport example. This is because the multi viewport example was used in the demo, and we thought it might be nice to update the example to match.

If you'd prefer to keep more similar to the old multi viewport example we can look into that, but since the multi viewport example is using a volume layer it will need some changes either way.

packages/core/src/objects/geometry/simple_box_geometry.ts
aganders3, line 1: Let's make this work with `BoxGeometry` or replace that implementation with this one.

packages/core/src/core/layer.ts
```diff
@@ -120,4 +125,12 @@ export abstract class Layer {
...
     this.objects_ = [];
   }
+
+  /**
+   * Get uniforms for shader program. Override in derived classes that need custom uniforms.
+   * @returns Object containing uniform name-value pairs
+   */
+  public getUniforms(): Record<string, unknown> {
+    return {}; // Default implementation returns no uniforms
+  }
```
aganders3, line 135: This isn't a huge deal, but I agree with @shlomnissan's previous comment that we should not consider uniforms as part of the layer. Instead we can configure more ergonomic properties on the layer that will be ultimately provided as uniforms from each renderable object.

packages/core/src/layers/volume_layer.ts
```diff
@@ -1,24 +1,261 @@
...
+    }
+
+    return this.createVolume(chunk);
+  }
+
+  public async onAttached(context: IdetikContext) {
+    if (this.chunkStoreView_) {
+      throw new Error(
+        "VolumeLayer is already attached. " +
+          "A layer cannot be attached to multiple contexts simultaneously."
```
aganders3: nit: only certain layers cannot be attached to multiple contexts, where they will use input from the viewport to determine chunk visibility/LOD. The current setup for this layer (loading all chunks for a given timepoint and fixed LOD) should work fine in multiple viewports.

packages/core/src/layers/volume_layer.ts
```diff
@@ -1,24 +1,261 @@
...
+      AlphaThreshold: this.alphaThreshold,
+    };
   }
 }
+
+export function poolKeyForChunk(chunk: Chunk) {
+  return [
+    `lod${chunk.lod}`,
+    `shape${chunk.shape.x}x${chunk.shape.y}`,
+    `align${chunk.rowAlignmentBytes}`,
```
aganders3: I think not having `chunk.shape.z` in the key may be why it's possible to get texture/buffer mismatch errors.

packages/core/src/renderers/webgl_renderer.ts
```diff
@@ -132,12 +133,26 @@ export class WebGLRenderer extends Renderer {
...
       this.gl_.clear(this.gl_.STENCIL_BUFFER_BIT);
     }
+
+    // This should probably be more generic, but for now we do some checks
+    // specifically for volume rendering
+    if (isVolumeLayer(layer)) {
+      this.state_.setDepthTesting(false);
+      this.state_.setDepthMask(false);
+      layer.reorderObjects(camera, "front-to-back");
+    }
```
aganders3: Above we check `layer.objects.length > 0 && layer.hasMultipleLODs();` to turn on the stencil test, maybe something like `layer.objects.length > 0 && layer.hasOverlappingRenderables()` would be a good name for this?

Sorting can probably be handled in the layer since the layer has access to the camera in the `RenderContext` passed to its `update` method.

This is probably good enough for now though, since this will likely be revisited when we look at OIT.

packages/core/examples/multi_viewport/main.ts
```diff
@@ -24,29 +27,38 @@ const bottom = 900;
...
 const zMin = z.translate;
 const zMax = z.translate + z.scale * z.shape - z.scale;
-const zRange = { min: zMin, max: zMax };
+
+// Calculate volume center
+// const volumeCenter = [
+//   (left + right) / 2, // x center
+//   (top + bottom) / 2, // y center
+//   (zMin + zMax) / 2, // z center
+// ] as [number, number, number];
```
aganders3: Remove if unused
```suggestion
```

packages/core/examples/multi_viewport/main.ts
```diff
@@ -24,29 +27,38 @@ const bottom = 900;
...
 
 // shared source between viewports
 const source = OmeZarrImageSource.fromHttp({ url });
 
-const camera3D = new PerspectiveCamera();
-const volumeLayer = new VolumeLayer();
+// Shared timepoint across all viewports
+const sharedTimepoint = 400;
```
aganders3: nit: maybe rename `initialTimepoint` as we need the mutable object below to actually share it

packages/core/examples/volume_rendering/main.ts
```diff
@@ -1,16 +1,132 @@
...
+  c: 0,
+};
 
 const camera = new PerspectiveCamera();
+const volumeLayer = new VolumeLayer({
+  source,
+  sliceCoords,
+  policy: createExplorationPolicy(),
+});
+// volumeLayer.debugMode = true;
```
aganders3:
```suggestion
```

packages/core/src/core/chunk_store_view.ts
```diff
@@ -145,6 +189,10 @@ export class ChunkStoreView {
     return this.currentLOD_;
   }
 
+  public get lowestResLOD(): number {
+    return this.store_.getLowestResLOD();
+  }
+
```
aganders3: I'm not opposed to this, but I'm also not sure we need it here. I think @andy-sweet pointed out it would also be nice if this respected the loading policy.

packages/core/src/layers/volume_layer.ts
```diff
@@ -1,24 +1,261 @@
...
+  }
+
+  private updateChunks() {
+    if (!this.chunkStoreView_) return;
+
+    const chunksToRender = this.chunkStoreView_.getChunksToRender(
+      this.sliceCoords_
+    );
+
+    // Check if we need to update
```
aganders3:
```suggestion
```

packages/core/src/layers/volume_layer.ts
```diff
@@ -1,24 +1,261 @@
...
+
+  public update(_context?: RenderContext) {
+    if (!this.chunkStoreView_) return;
+
+    // Initialize LOD to coarsest if not set
+    if (this.lod_ === -1) {
+      this.lod_ = this.chunkStoreView_.lowestResLOD;
+    }
+
+    // Mark chunks for the desired LOD as visible
```
aganders3:
```suggestion
```

aganders3 (commented):
I've got a few comments and nits, but overall I think this is looking really good!

packages/core/src/objects/geometry/simple_box_geometry.ts
seankmartin: This was due to an issue arising with culling that lead to
image

Debugging the BoxGeometry, the front faces look fine, they are in expected CCW winding order. E.g. Here is the face with index 4 (the front z face):

```
Face 4 [
  [ -0.5,  0.5,  0.5 ], \\ top left
  [ -0.5, -0.5,  0.5 ], \\ bottom left
  [  0.5,  0.5,  0.5 ], \\ top right
  [ -0.5, -0.5,  0.5 ], \\ bottom left
  [  0.5, -0.5,  0.5 ], \\ bottom right
  [  0.5,  0.5,  0.5 ], \\ top right
]
```
So both triangles in the face are counter clockwise when viewed from the front.

But for the back Z face:
```
Face 5 [
  [  0.5,  0.5, -0.5 ], \\ top right
  [ -0.5,  0.5, -0.5 ], \\ top left
  [  0.5, -0.5, -0.5 ], \\ bottom right
  [  0.5, -0.5, -0.5 ], \\ bottom right
  [ -0.5,  0.5, -0.5 ], \\ top left
  [ -0.5, -0.5, -0.5 ], \\ bottom left
]
```
Which are also in counter clockwise order, but should actually be in clockwise order. I pushed a [commit](https://github.com/chanzuckerberg/idetik/pull/553/commits/97aca87474d7b6882795283510704f5d65ea7eb3) that removes a check but not sure if that has wider impact on the code. The check was for if the face was at the front or back to then pack the indices into the index buffer differently. But because of the udir and vdir in the function meaning that the positions of a b c and d on a face change anyway because of udir and vdir I don't think the extra wdir if check is needed to change the order of the indices. I see from the commit introducing the box geometry that it is based on https://github.com/mrdoob/three.js/blob/master/src/geometries/BoxGeometry.js and I could be missing something but I don't see the equivalent check on wdir to change the index insertion order in that either. So maybe this is the correct fix?

cc @shlomnissan in case this would break something intended in the BoxGeometry, and sorry if so!

seankmartin: I should also I think mention that I set front faces to actually be CW (in the GL state) in Idetik because I saw a y-flip in the co-ordinate space which would flip the winding unless I'm mistaken

packages/core/src/layers/volume_layer.ts
seankmartin: Nice spot thank you! Yes seems to be ok with that change

packages/core/src/renderers/webgl_renderer.ts
seankmartin: I changed this, thank you for the suggestion! Was wondering if the `renderLayer` in the renderer should just return straight away if `layer.objects.length === 0`? (essentially a `layer.isEmpty()` early return). Might save repeated checks, especially if this kind of `layer.hasX` pattern becomes more common. Obviously not a requirement, just wondering if that could help

packages/core/src/core/chunk_store_view.ts
seankmartin: Thanks Ashley, at the moment this is used in the volume layer so that the LOD in the volume layer can be set to -1 by default. And then get changed to the actual lowest res LOD as informed by the chunk store view. Is there somewhere else you'd suggest to put this? Or a different way of doing this?

And when you say respecting the loading policy, is that taking into account the LOD bias? Not sure I fully grasp what is intended there, sorry!

packages/core/src/core/layer.ts
seankmartin: Thanks both, I'll make an issue to track this as this would need a bit of moving things around to make it work that probably is better to separate into a later refactoring (though let me know if something blocking is later realised because of this)

Edit, see https://github.com/chanzuckerberg/idetik/issues/555

packages/core/src/objects/geometry/simple_box_geometry.ts
aganders3: I don't think BoxGeometry is used anywhere else, so for me pretty much any solution is fine here. I just didn't want to merge a second separate box implementation. Thanks!

packages/core/src/core/chunk_store_view.ts
aganders3: Sorry, what I mean is that I think the loading policy includes a way to restrict the available LODs, so instead of grabbing the coarsest from the source (`ChunkStore`) it would be something like `min(this.store_.getLowestResLOD(), this.policy.lods.max)` or something like that.

It's not a big deal though given this works and we will make changes here eventually to try to automatically select an LOD.

packages/core/src/renderers/webgl_renderer.ts
aganders3: That makes sense to me!

packages/core/src/renderers/shaders/volume_frag.glsl
```diff
@@ -13,17 +13,91 @@ uniform mediump usampler3D ImageSampler;
...
+        //     CameraPositionModel, RayDirModel, boundingboxMin - vec3(0.015), boundingboxMax + vec3(0.015)
+        // );
+        // tEnter = rayIntersections.x;
+        // tExit = rayIntersections.y;
+        fragColor = vec4(1.0, 0.0, 0.0, 1.0);
+        return;
     }
-    float displayValue = alpha;
-    fragColor = vec4(displayValue, displayValue, displayValue, 1.0);
+    tExit = max(tEnter, tExit);
```
aganders3: I'm confused by this. I do see the red at times if `EnableRayCorrection` is on, but the correction code in here is commented out, and the image looks okay when I turn it off.

aganders3: Oh I think maybe the only correction needed is `tExit = max(tEnter, tExit);`?

Should we just add that to `findBoxIntersectionsAlongRay`? I think we can also clamp `tEnter = max(0.0, tEnter)` to prevent rays starting behind the camera.

seankmartin: The correction inside the if is something we were exploring, it doesn't fully fix the issue anyway so I'll take it out. But yes definitely some adjustment of both tExit and tEnter are needed like is there at the moment. Agreed on the extra tEnter one you suggested just to be on the safe side!

I've a funny feeling the issue might still be around with these (it is at very specific camera angles, LODs etc), but let me make this check anyway on tEnter and I can pull a lot of this temp code here out into an issue to avoid causing confusion here.

Edit: one thing to add, the red was just because the only times we got the issue was when we had `invalidIntersection`. In other words, `invalidIntersection` didn't always cause an issue, but whenever there was one, it was at a fragment that had `invalidIntersection`.

Further edit: see https://github.com/chanzuckerberg/idetik/issues/556

packages/core/src/renderers/shaders/volume_frag.glsl
```diff
@@ -13,17 +13,91 @@ uniform mediump usampler3D ImageSampler;
 uniform mediump sampler3D ImageSampler;
 #endif
 
-in vec2 TexCoords;
+uniform highp vec3 CameraPositionModel;
+in highp vec3 PositionModel;
+
+// The bounding box in model space is normalized to -0.5 to 0.5
+vec3 boundingboxMin = vec3(-0.50);
+vec3 boundingboxMax = vec3(0.50);
```
aganders3, line 21: Sorry I have one more question! We're marching in the normalized model space so the `SampleDensity` is in samples per texture unit. Should this be adjusted for the chunk shape and/or scale (texture units per world-space unit)?

I don't think this will be visible most of the time, given chunk sizes are relatively consistent across a volume, but it might help make the parameter more consistent and intuitive across LODs or datasets.

This adjustment can probably be handled on the CPU though (especially if uniforms are per-chunk, not per-layer). I could also be missing something here!

seankmartin: Yes definitely! Wanted to make a change to this scaling anyway, you beat me to it in making the issue sorry

aganders3: No worries! If we have it documented with an issue or even with a TODO I would be happy to merge this PR without it.

packages/core/examples/multi_viewport/main.ts
```diff
@@ -57,51 +59,65 @@ new Idetik({
       id: "volume",
       element: document.querySelector<HTMLDivElement>("#viewport-left")!,
       camera: camera3D,
-      cameraControls: new OrbitControls(camera3D, { radius: 3 }),
+      cameraControls: new OrbitControls(camera3D, {
+        radius: 1500,
```
shlomnissan: @aganders3 I noticed this value in the demo branch. Why is it so large?

packages/core/examples/multi_viewport/main.ts
```diff
@@ -57,51 +59,65 @@ new Idetik({
...
+      }),
```
shlomnissan, line 80: It would be nice to add `target` to `orbitControls` to make sure the volume is centered in the viewport.

packages/core/examples/multi_viewport/main.ts
```diff
@@ -57,51 +59,65 @@ new Idetik({
...
   ],
   showStats: true,
 }).start();
 
 const gui = new GUI({ width: 300 });
 
-const topRightViewportFolder = gui.addFolder("Top Right Viewport (Slice 1)");
+const sharedTime = { t: initialTimepoint };
+volumeCoords.t = sharedTime.t;
+sliceCoords.t = sharedTime.t;
```
shlomnissan: I think there's some duplication here. Is it possible to unify `sharedTime` and `volumeCoords` so we only set this value once?

packages/core/examples/multi_viewport/main.ts
```diff
@@ -57,51 +59,65 @@ new Idetik({
...
+  minValue: 0,
+  maxValue: 800,
+  stepValue: 1,
+  playback: {
+    maxRateHz: 30,
+    stride: 1,
+    onRateChange: () => {
+      // Sync the time value to all viewports
+      volumeCoords.t = sharedTime.t;
+      sliceCoords.t = sharedTime.t;
```
shlomnissan: We continue to update these two values to keep them in sync. @andy-sweet Should we make `t` an object and pass it by reference so we don't have to update it manually?

packages/core/examples/multi_viewport/main.ts
```diff
@@ -57,51 +59,65 @@ new Idetik({
...
+// Keep all viewports synchronized by updating on every frame
+const syncTime = () => {
+  if (volumeCoords.t !== sharedTime.t) {
+    volumeCoords.t = sharedTime.t;
+  }
+  if (sliceCoords.t !== sharedTime.t) {
+    sliceCoords.t = sharedTime.t;
+  }
+  requestAnimationFrame(syncTime);
+};
```
shlomnissan: I don't think we should leak this workaround to the public API. No-op for this PR.

packages/core/examples/volume_rendering/main.ts
```diff
@@ -1,16 +1,131 @@
...
+};
 
 const camera = new PerspectiveCamera();
+const volumeLayer = new VolumeLayer({
+  source,
+  sliceCoords,
+  policy: createExplorationPolicy(),
+});
+
+// Largely copied from chunk streaming example
```
shlomnissan: I don't think this comment is needed.

packages/core/examples/volume_rendering/main.ts
```diff
@@ -1,16 +1,131 @@
...
+const t = { translate: 0.0, scale: 1.0, shape: 791 };
+const tMin = t.translate;
+const tMax = t.translate + t.scale * t.shape - t.scale;
+const tRange = { min: tMin, max: tMax };
+const timePointDiv = document.querySelector<HTMLDivElement>("#time-point")!;
+const timePointOverlay = {
+  update(_idetik: Idetik, _timestamp?: DOMHighResTimeStamp) {
```
shlomnissan: I think I removed `timestamp` from this interface. It should be removed here since it's not used.

packages/core/examples/volume_rendering/main.ts
```diff
@@ -1,16 +1,131 @@
...
+    const time = sliceCoords.t;
+    timePointDiv.textContent = `t = ${time}`;
+  },
+};
 
-new Idetik({
+const spherical = {
+  radius: 4000,
+};
+const cameraControls = new OrbitControls(camera, spherical);
```
shlomnissan: - This value keeps growing. Why is it so high? Why does it need to be higher than the viewports example?
- Same comment about [centering the volume](https://github.com/chanzuckerberg/idetik/pull/553/changes#r2632097623) in the viewport.
- You should inline this variable. It doesn't add any value here:
```
{
  camera,
  cameraControls: new OrbitControls(camera, { radius: 1500 }),
  layers: [new VolumeLayer()],
},

```

packages/core/examples/volume_rendering/main.ts
```diff
@@ -1,16 +1,131 @@
...
+  .name("Enable Ray Correction at Bounds")
+  .onChange((show: boolean) => {
+    controls.enableRayCorrection = show;
+    volumeLayer.enableRayCorrection = show;
+  });
+
+const volumeFolder = gui.addFolder("Volume Rendering");
+
+volumeFolder
+  .add(controls, "lod", 0, 2, 1)
```
shlomnissan, line 56: In some instances we use `controls` and in others we reference layer properties directly. I think we should choose one approach so it's consistent unless we're not modifying layer properties at all (.e.g, `showTimePointOverlay`).

packages/core/src/core/chunk_store_view.ts
```diff
@@ -90,6 +90,50 @@ export class ChunkStoreView {
...
+      const isChannelMatch =
+        sliceCoords.c === undefined || sliceCoords.c === chunk.chunkIndex.c;
+      if (!isChannelMatch) continue;
+
+      const priority = this.policy_.priorityMap["visibleCurrent"];
+      this.chunkViewStates_.set(chunk, {
+        visible: true,
+        prefetch: false,
+        priority,
+        orderKey: 0, // All chunks have same priority for volume rendering
```
shlomnissan: I don’t think `orderKey` affects priority, only ordering. If this isn’t relevant for volume rendering it’s still worth a comment but it should be precise about that distinction.

packages/core/src/core/chunk_store_view.ts
```diff
@@ -145,6 +189,10 @@ export class ChunkStoreView {
     return this.currentLOD_;
   }
 
+  public get lowestResLOD(): number {
+    return this.store_.getLowestResLOD();
+  }
```
shlomnissan: @aganders3 I still feel like the relationship between the store and store view abstractions is leaky.

A `ChunkStore` sounds like a passive data structure but it is also acting as a scheduler, a cache, and a policy enforcer. It aggregates view demand, mutates chunk visibility and priority, transitions chunks between unloaded, queued, and loaded, and even decides when data should be disposed. On top of that, it reaches back into views and tells them when to forget chunks. This blurs the direction of ownership. Views are supposed to describe what they need, but the store ends up partially managing how views remember that need. Once that line is crossed, it becomes hard to reason about who owns which piece of state.

The need for this accessor signals that responsibilities between "who owns data" and "who consumes it" are bleeding across.

No-op for this PR.

packages/core/src/core/layer.ts
```diff
@@ -100,6 +105,10 @@ export abstract class Layer {
     return false;
   }
 
+  public hasOverlappingRenderables(): boolean {
```
shlomnissan: I don’t think overlapping renderables is a valid criterion for toggling depth testing. I’d rather have the renderer explicitly ask whether it’s dealing with a volume layer and handle depth testing accordingly. A even cleaner solution would be to add a `depthTest` flag to renderables themselves. This is fundamentally a renderable concern not a layer concern.

shlomnissan (commented):
I still need to review the volume layer and shader code but I wanted to share some early feedback.

packages/core/src/renderers/webgl_renderer.ts
```diff
@@ -125,13 +125,18 @@ export class WebGLRenderer extends Renderer {
...
     this.state_.setBlendingMode(layer.transparent ? layer.blendMode : "none");
-    const shouldUseStencil =
-      layer.objects.length > 0 && layer.hasMultipleLODs();
+    const shouldUseStencil = layer.hasMultipleLODs();
     this.state_.setStencilTest(shouldUseStencil);
     if (shouldUseStencil) {
       this.gl_.clear(this.gl_.STENCIL_BUFFER_BIT);
     }
+
+    const shouldUseDepth = !layer.hasOverlappingRenderables();
```
shlomnissan: I think a `shouldNotUseDepth` condition is better. `shouldUseDepth` hides the real rule and forces a mental inversion. `shouldNotUseDepth` states the condition directly and reads correctly without negation.

packages/core/src/renderers/webgl_renderer.ts
```diff
@@ -187,7 +192,21 @@ export class WebGLRenderer extends Renderer {
     );
     const resolution = [this.canvas.width, this.canvas.height];
 
+    const inverseModelView = mat4.invert(mat4.create(), modelView);
+    const cameraPositionView = vec4.fromValues(0, 0, 0, 1);
+    const cameraPositionModel = vec4.transformMat4(
+      vec4.create(),
+      cameraPositionView,
+      inverseModelView
+    );
```
shlomnissan: I think we should avoid expensive operations unless they are needed them. I assume the inverse is only needed for volume rendering? If that's the case this is probably better:

```ts
case "CameraPositionModel":
  const inverseModelView = mat4.invert(mat4.create(), modelView);
  const cameraPositionView = vec4.fromValues(0, 0, 0, 1);
  const cameraPositionModel = vec4.transformMat4(
    vec4.create(),
    cameraPositionView,
    inverseModelView
  );

  program.setUniform(
    uniformName,
    vec3.fromValues(
      cameraPositionModel[0],
      cameraPositionModel[1],
      cameraPositionModel[2]
    )
  );
  break;
```

packages/core/src/renderers/WebGLState.ts
```diff
@@ -145,6 +149,7 @@ export class WebGLState {
       this.setCullFace(false);
     } else {
       this.setCullFace(true);
+      this.gl_.frontFace(this.gl_.CW);
```
shlomnissan: I’m concerned about the implicit coupling here. `setCullFaceMode` should ideally only control `gl.cullFace`. Inverting the winding order to `CW` as a side effect (and never changing it back) makes the behavior unpredictable in different contexts. If we need `CW` for a specific use case we should make it explicit in the callsite, reverse it when it's no longer needed, and use the same feature caching mechanism.

packages/core/examples/multi_viewport/main.ts
aganders3: There's no reason it needs to be this large, especially if we set the target to the center of the volume.

packages/core/src/core/chunk_store_view.ts
aganders3: I agree in part, though I think this accessor is not actually needed and is more like a workaround.

packages/core/examples/multi_viewport/main.ts
aganders3: I agree we don't want this workaround. Can it just be removed here? The values are always set together so I'm not sure how they would get out of sync.

packages/core/src/core/layer.ts
aganders3: I agree this should ultimately move to the renderable, but I prefer this to having the renderer special-case a layer type.

packages/core/examples/multi_viewport/main.ts
seankmartin: Thanks both! For the moment I removed all of this manual syncing by making the two coordinate objects passed into the volume layer and the image layer use a getter on `t` in [2e5bbb14](https://github.com/chanzuckerberg/idetik/pull/553/commits/2e5bbb14b6c70f6f047de6e49aa8227a5db803ec) -- happy to consider alternatives but at least this it should remove this syncing

packages/core/src/core/layer.ts
seankmartin: Please let me know if https://github.com/chanzuckerberg/idetik/pull/553/commits/e1f9702344363b488a06da08d2d3e910dfdc0833 is what you had in mind, moves it from the layer to the renderable, cheers

packages/core/src/renderers/webgl_renderer.ts
seankmartin: With https://github.com/chanzuckerberg/idetik/commit/e1f9702344363b488a06da08d2d3e910dfdc0833 (if this commit is ok) this shouldn't be needed but agreed and thank you for spotting the awkwardness

packages/core/src/renderers/webgl_renderer.ts
seankmartin: Sounds great, agreed! See https://github.com/chanzuckerberg/idetik/pull/553/commits/039457982a667e5d2830e53f0223b2116f47de14

packages/core/src/renderers/WebGLState.ts
seankmartin: This is because Idetik implements a y-axis flip in the projection matrix. Removing the y-axis flip we can keep the regular CCW triangle ordering, but with the y-flip in the projection matrix it flips the triangle winding order after projection.

I do agree though this is in the wrong place, it should probably happen some levels above this with a comment. I'll make a commit that changes it and please let me know if it makes sense after that, thank you for spotting this

See https://github.com/chanzuckerberg/idetik/pull/553/commits/7477963f5bb349e0fe7908f11df7901bc6039947

packages/core/src/core/chunk_store_view.ts
seankmartin: I've removed this function and changed the volume layer constructor to accept the initial LOD and updated the examples to explicitly set the lowest res LOD in https://github.com/chanzuckerberg/idetik/pull/553/commits/442c02765ad93f40823169ab3b1d3049b8b7345a. This is instead of having lod set to -1 and determining the lowest res available from the chunk store view. It does mean a volume layer could try to get initialized at LOD 0 with a huge amount of chunks though. If you like I can make the parameter non-optional so at least 0 is consciously entered as the LOD in that case

packages/core/src/core/chunk_store_view.ts
seankmartin: https://github.com/chanzuckerberg/idetik/pull/553/commits/d5c025efbbe882f444884ba4eb65f97cf89f5788 hopefully solves

seankmartin:
Outside of the review comments, we changed - https://github.com/chanzuckerberg/idetik/pull/553/commits/e565125349eaa441e9f8b21a7cfbca9f9f07087d. This is because the volume rendering only works with one type of blending, and has to be transparent. The concern was that the layer could be initialised by someone wanting to use, say, additive blending, then be confused why it does not render well. Other layers also seemed to not provide all the parameters that a layer does in the constructor, so hopefully this works

packages/core/examples/volume_rendering/main.ts
seankmartin: Made a few commits which should hopefully clean this up and removed the overlay of time, thanks

seankmartin:
I have marked some threads as resolved when the fix is pretty simple and we're fairly confident that it's not controversial. For some other threads we have made/are making changes that should hopefully resolve them but I haven't marked them as resolved just in case. Please feel free to mark them as resolved if you agree on the change, just so it is clear which points remain to be addressed. Thank you!

packages/core/examples/multi_viewport/main.ts
seankmartin: See https://github.com/chanzuckerberg/idetik/pull/553/commits/2e5bbb14b6c70f6f047de6e49aa8227a5db803ec

packages/core/examples/multi_viewport/main.ts
seankmartin: I updated in https://github.com/chanzuckerberg/idetik/pull/553/commits/6f436e9399430272ccedc9bb01a9280f05789519 so that it looks towards the volume center. But the radius is still pretty large. I'm not sure what units the chunk units are in though. For example we could include

```typescript
  private updateVolumeChunk(volume: VolumeRenderable, chunk: Chunk) {
    const scaleMultiplier = 0.01; // To convert from chunk units to world units if needed
    volume.transform.setScale([
      chunk.shape.x * chunk.scale.x * scaleMultiplier,
      chunk.shape.y * chunk.scale.y * scaleMultiplier,
      chunk.shape.z * chunk.scale.z * scaleMultiplier,
    ]);
    const originOffset = {
      x: (chunk.shape.x * chunk.scale.x) / 2,
      y: (chunk.shape.y * chunk.scale.y) / 2,
      z: (chunk.shape.z * chunk.scale.z) / 2,
    };
    volume.transform.setTranslation([
      (chunk.offset.x + originOffset.x) * scaleMultiplier,
      (chunk.offset.y + originOffset.y) * scaleMultiplier,
      (chunk.offset.z + originOffset.z) * scaleMultiplier,
    ]);
  }
  ```
  in the volume layer if we know that multiplier (if there is one, we should include it per-axis, this is just for demo)

File: packages/core/examples/multi_viewport/main.ts (hunk 14 above)
Comment (seankmartin) [reply packages/core/examples/multi_viewport/main.ts:80]: Done, thank you!

File: packages/core/examples/volume_rendering/main.ts (hunk 20 above)
Comment (seankmartin) [reply packages/core/examples/volume_rendering/main.ts]: Inlined the variable, I think we can discuss the radius/center further in the other comment you linked to, thanks

File: packages/core/src/core/chunk_store_view.ts (hunk 8 above)
Comment (seankmartin) [reply packages/core/src/core/chunk_store_view.ts]: Thanks for clarifying! I just removed the function anyway given other comments

File: packages/core/src/renderers/shaders/volume_frag.glsl (hunk 12 above)
Comment (seankmartin) [reply packages/core/src/renderers/shaders/volume_frag.glsl:21]: See https://github.com/chanzuckerberg/idetik/issues/557

File: packages/core/src/core/chunk_store_view.ts (hunk 23 above)
Comment (andy-sweet) [reply packages/core/src/core/chunk_store_view.ts]: > @aganders3 I still feel like the relationship between the store and store view abstractions is leaky.

> A ChunkStore sounds like a passive data structure but it is also acting as a scheduler, a cache, and a policy enforcer. It aggregates view demand, mutates chunk visibility and priority, transitions chunks between unloaded, queued, and loaded, and even decides when data should be disposed. On top of that, it reaches back into views and tells them when to forget chunks. This blurs the direction of ownership. Views are supposed to describe what they need, but the store ends up partially managing how views remember that need. Once that line is crossed, it becomes hard to reason about who owns which piece of state.

FWIW, I seem to remember the responsibilities of each class being more clear when the view had a reference to the store (i.e. before #545). For example, the `ChunkManager` aggregated and mutated, and the `ChunkStore` was just a cache for a chunk loader. Probably not a discussion to be had buried in a thread in this PR, and not something I think we should immediately change, but definitely worth considering as we move forward with a long term architecture.

File: packages/core/examples/multi_viewport/main.ts (hunk 13 above)
Comment (aganders3) [reply packages/core/examples/multi_viewport/main.ts]: We don't have defined world units, really. The units are effectively whatever is set in the zarr metadata because we scale by the chunk size * scale but the actual units are ignored.

Comment (shlomnissan) [reply packages/core/examples/multi_viewport/main.ts]: I assume we don't have scaling issues in 2D because the ortho camera in this demo is explicitly set to a window in chunk-space units so it's effectively "fit to view". In 3D we're using a default perspective camera with so the orbit radius ends up reflecting whatever scale that dataset implies.

If we want a single viewer that can load arbitrary datasets we probably shouldn't expect users to tune camera parameters per volume. I think two reasonable paths are:
1. Add normalization layer so both 2D/3D land in a consistent scale.
2. Keep data unit as-is but auto-configure cameras from volume bounds.

Thoughts?

File: packages/core/examples/multi_viewport/main.ts [hunk 28]
```diff
@@ -24,29 +25,47 @@ const bottom = 900;
...
+  sliceCoords: volumeCoords,
+  policy: createPlaybackPolicy(),
+  lod: 2,
 });
 
-const sliceCoords2 = { t: 400, z: 300, c: 0 };
-const camera2D2 = new OrthographicCamera(left, right, top, bottom);
-const imageLayer2 = new ChunkedImageLayer({
+const camera2D = new OrthographicCamera(left, right, top, bottom);
+camera2D.zoom(0.65);
```
shlomnissan: I don't mind applying zoom in 2D but this makes the initial scale in viewports further apart.

packages/core/examples/volume_rendering/main.ts
```diff
@@ -1,16 +1,78 @@
-import { Idetik, VolumeLayer, PerspectiveCamera } from "@";
+import { Idetik, VolumeLayer, PerspectiveCamera, OmeZarrImageSource } from "@";
 import { OrbitControls } from "@/objects/cameras/orbit_controls";
+import {
+  createExplorationPolicy,
+  createPlaybackPolicy,
```
shlomnissan: I don't mind the policy toggle but to be clear this example doesn't do any buffering over time so this is just a placeholder right?

packages/core/examples/volume_rendering/main.ts
```diff
@@ -1,16 +1,78 @@
...
+  c: 0,
+};
 
-new Idetik({
+const camera = new PerspectiveCamera();
+const volumeLayer = new VolumeLayer({
+  source,
+  sliceCoords,
+  policy: createExplorationPolicy(),
+  lod: 2,
```
shlomnissan: The policy structure already provides a mechanism for clamping LODs. We should use it instead of passing the LOD directly to the layer.

```ts
const volumeLayer = new VolumeLayer({
  source,
  sliceCoords,
  policy: createExplorationPolicy({lod: {min: 2, max: 2}}),
});
```

packages/core/examples/volume_rendering/main.ts
```diff
@@ -1,16 +1,78 @@
...
+  .name("Opacity scale");
+volumeFolder
+  .add(volumeLayer, "alphaThreshold", 0.8, 1.0, 0.01)
+  .name("Early termination threshold");
+
+const overlaysFolder = gui.addFolder("Debug");
+overlaysFolder.add(volumeLayer, "debugMode").name("Show tile wireframes");
+overlaysFolder
+  .add(volumeLayer, "showEmptyRays")
+  .name("Show rays with length 0");
```
shlomnissan: Showing tile wireframes and degenerate rays are both debugging functions so it feels inconsistent that `debugMode` on the volume layer only controls wireframes. Ideally we would have a `debugConfig` object to manage individual properties but for this PR a simpler approach would be to use two specific flags on the layer: `debugShowWireframes` and `debugShowDegenerateRays`.

packages/core/src/core/chunk_store_view.ts
```diff
@@ -90,6 +90,50 @@ export class ChunkStoreView {
...
+  // TODO: Eventually unify visibility calculations using the camera frustum for both
+  // orthographic (2D slices) and perspective (volume rendering) cameras. This would
+  // replace both updateChunkStates and updateChunkStatesForVolume with a single method
+  // that performs frustum culling based on the camera type.
+  public updateChunkStatesForVolume(
+    sliceCoords: SliceCoordinates,
+    lod: number
+  ): void {
+    const currentTimeIndex = this.store_.getTimeIndex(sliceCoords);
+    const currentTimeChunks = this.store_.getChunksAtTime(currentTimeIndex);
```
shlomnissan, line 99: @aganders3 We're mixing `SliceCoordinates`-based time and explicit time indices throughout this file, sometimes within the same code path. It’s not always clear which representation is the source of truth.

Let's pick a single approach for time in this layer and make helpers accept only that. For example, keep `SliceCoordinates` at the public API boundary, compute `timeIndex` once, and pass to all internal functions.

packages/core/src/core/chunk_store_view.ts
```diff
@@ -90,6 +90,50 @@ export class ChunkStoreView {
...
+
+    if (currentTimeChunks.length === 0) {
+      Logger.warn(
+        "ChunkStoreView",
+        "updateChunkStatesForVolume called with no chunks initialized"
+      );
+      this.chunkViewStates_.clear();
```
shlomnissan, line 106: Why do we clear `chunkViewStates_ ` here? Is it so that `getChunksToRender()` has nothing to iterate on? cc @aganders3

This function is doing two things:
- Detecting an invalid state (no chunks for the resolved time)
- Performing recovery with `chunkViewStates_.clear()`

I think responsibility for detecting an invalid state should live higher up. Clearing here is defensively correct but it feels a bit misplaced.

Also, the error message references `updateChunkStatesForVolume` which could be confusing during debugging since I don’t think that function is called from this path (unless I’m missing something).

packages/core/src/core/chunk_store_view.ts
```diff
@@ -90,6 +90,50 @@ export class ChunkStoreView {
...
+      return;
+    }
+
+    // Reset all existing chunk view states
+    this.chunkViewStates_.forEach(markUnused);
```
shlomnissan: This comment is doing real work to compensate for a non-obvious function name. `markUnused` does not clearly communicate:
- this is a full chunk reset
- that this is part of a two-phase "reset then selectively re-enable"

I think we should rename `markUnused` to `resetChunkViewState`.

packages/core/src/core/chunk_store_view.ts
```diff
@@ -90,6 +90,50 @@ export class ChunkStoreView {
...
+
+    // Mark all chunks at the specified LOD as visible
```
shlomnissan: The loop is doing more than what the comment says:
- Filtering by LOD
- Filtering by channel
- Assigning priority
- Setting view state

At best the comment is incomplete and at worst it makes a reader skim past logic they should actually read. I think we can just delete it. The code logic is easy enough to follow.

packages/core/src/core/chunk_store_view.ts
```diff
@@ -370,7 +414,13 @@ export class ChunkStoreView {
 
   private getZBounds(sliceCoords: SliceCoordinates): [number, number] {
     const zDim = this.store_.dimensions.z;
-    if (zDim === undefined || sliceCoords.z === undefined) return [0, 1];
+    if (zDim === undefined) return [0, 1];
+
+    // If z is undefined, return bounds that encompass all z slices (for volume rendering)
```
shlomnissan, line 415: Right now `sliceCoords.z === undefined` implicitly means "volume rendering". Passing an explicit mode would make that intent clear and avoid silently expanding to full-Z if z is `undefined` for some other reason.

```ts
private getZBounds(sliceCoords, { mode: "volume" | "slice" }));
```

packages/core/src/layers/volume_layer.ts
```diff
@@ -1,24 +1,258 @@
...
+import { Texture3D } from "../objects/textures/texture_3d";
+import { RenderablePool } from "../utilities/renderable_pool";
+import { glMatrix, vec3 } from "gl-matrix";
+import { Camera } from "@/objects/cameras/camera";
+
+export type VolumeLayerProps = {
+  source: ChunkSource;
+  sliceCoords: SliceCoordinates;
+  policy: ImageSourcePolicy;
+  lod?: number;
```
shlomnissan: As mentioned, this property already exists in the image source policy and should not be duplicated.

packages/core/src/layers/volume_layer.ts
```diff
@@ -1,24 +1,258 @@
...
+};
+
+export type OrderingMode = "front-to-back" | "back-to-front";
```
shlomnissan: We only ever sort front-to-back and the type isn’t used outside this file. Remove the enum and always return `diff` in `reorderObjects` to keep the code clear.

packages/core/src/layers/volume_layer.ts
```diff
@@ -1,24 +1,258 @@
...
-    super();
+  private readonly source_: ChunkSource;
+  private readonly sliceCoords_: SliceCoordinates;
+  private readonly visibleChunks_: Map<Chunk, VolumeRenderable> = new Map();
+  private readonly pool_ = new RenderablePool<VolumeRenderable>();
+
+  private sourcePolicy_: ImageSourcePolicy;
+  private chunkStoreView_?: ChunkStoreView;
+  private lod_ = 0;
+  private debugMode_ = false;
```
shlomnissan: As mentioned before, this should be renamed or debugging should be refactored.

packages/core/src/layers/volume_layer.ts
```diff
@@ -1,24 +1,258 @@
...
+
+  private lastLoadedLod_ = -1;
+  private lastLoadedTime_ = -1;
```
shlomnissan: Using `-1` as "not initialized" sentinel is fragile. You should use an explicit state (`undefined` or `null`) so the code encodes intent instead of assumptions (even if these assumptions seems reasonable now).

packages/core/src/layers/volume_layer.ts
```diff
@@ -1,24 +1,258 @@
...
+  public showEmptyRays = false;
+  private color_ = vec3.fromValues(1.0, 1.0, 1.0);
+  public sampleDensity = 128.0; // Samples per unit texture space
+  public maxIntensity = 255.0; // Normalization factor for intensity
+  public opacityScale = 0.1; // Alpha multiplier
+  public alphaThreshold = 0.99; // Early ray termination threshold
```
shlomnissan: If the comment explains what the variable does the name is wrong. Either change the name or delete the comment if you think it's clear.

Suggestions:
- `samplesPerUnit`
- `maxIntensity` this one already stands on its own IMO
- `opacityMultiplier`
- `earlyTerminationAlpha`

packages/core/src/layers/volume_layer.ts
```diff
@@ -1,24 +1,258 @@
...
-    const data = new Int8Array([127, 0, 0, 127, 127, 0, 0, 127]);
-    const texture = new Texture3D(data, 2, 2, 2);
-    texture.unpackAlignment = 1;
-    const renderable = new VolumeRenderable(1, 1, 1, texture);
-    renderable.wireframeEnabled = true;
+  public set lod(value: number) {
+    this.lod_ = value;
+    this.clearObjects();
+    this.updateChunks();
+  }
```
shlomnissan: If we are using the image policy instead this mutator should be deleted.

packages/core/src/layers/volume_layer.ts
```diff
@@ -1,24 +1,258 @@
...
+
+  public get debugMode(): boolean {
+    return this.debugMode_;
+  }
```
shlomnissan: Are we actually using this accessor? If not, delete it. Let's do a quick pass over all getters/setters and keep only the ones that are called. Every accessor is surface area.

packages/core/src/layers/volume_layer.ts
```diff
@@ -1,24 +1,258 @@
...
+  public set debugMode(value: boolean) {
+    this.debugMode_ = value;
+    for (const volume of this.visibleChunks_.values()) {
+      volume.wireframeEnabled = value;
+    }
+  }
+
+  public get sourcePolicy(): Readonly<ImageSourcePolicy> {
+    return this.sourcePolicy_;
+  }
```
shlomnissan: Are we actually using this accessor? If not, delete it. Let's do a quick pass over all getters/setters in this PR and keep only the ones that are called. Every accessor is surface area.

packages/core/src/layers/volume_layer.ts
```diff
@@ -1,24 +1,258 @@
...
+    }
+  }
+
+  public get color(): vec3 {
+    return this.color_;
+  }
+
+  public set color(newColor: vec3) {
+    vec3.copy(this.color_, newColor);
+  }
```
shlomnissan: Are we using color outside this layer right now? If yes, do we actually need to prevent in-place mutation?

packages/core/src/layers/volume_layer.ts
```diff
@@ -1,24 +1,258 @@
...
+
+  private createVolume(chunk: Chunk) {
+    const volume = new VolumeRenderable(Texture3D.createWithChunk(chunk));
+    this.updateVolumeChunk(volume, chunk);
+    return volume;
+  }
+
+  constructor({ source, sliceCoords, policy, lod = 0 }: VolumeLayerProps) {
+    // Volume rendering is always transparent with fixed blend mode
```
shlomnissan: I think the code is clearer than the comment so I would remove it.

packages/core/src/layers/volume_layer.ts
```diff
@@ -1,24 +1,258 @@
...
+  }
 
-    this.addObject(renderable);
-    this.setState("ready");
+  public async onAttached(context: IdetikContext) {
+    this.chunkStoreView_ = await context.chunkManager.addView(
+      this.source_,
+      this.sourcePolicy_
+    );
   }
```
shlomnissan, line 97: @aganders3 Indirectly related to this PR. I think there might be a bug in the chunk manager:

`getOrCreateStore` doesn't seem to manage failure. I think that if `source.open()` rejects, `pendingStores_` keeps a rejected promise forever so future calls keep failing and we never retry. We should clear `pendingStores_` in a finally and log the error.

packages/core/src/layers/volume_layer.ts
```diff
@@ -1,24 +1,258 @@
...
 
-  public update() {
-    // TODO: implement
+  public onDetached(_context: IdetikContext): void {
+    this.releaseAndRemoveChunks(this.visibleChunks_.keys());
+    this.clearObjects();
+    if (!this.chunkStoreView_) return;
```
shlomnissan, line 100: If `onDetached` runs before the view is initialized I assume there’s nothing to release in that case. If so, move the early return to the top to make that contract explicit.

packages/core/src/layers/volume_layer.ts
```diff
@@ -1,24 +1,258 @@
...
+
+    const chunksToRender = this.chunkStoreView_.getChunksToRender(
+      this.sliceCoords_
+    );
+
+    const currentTime = this.sliceCoords_.t ?? -1;
+    const needsUpdate =
+      this.lastLoadedLod_ !== this.lod_ ||
+      this.lastLoadedTime_ !== currentTime ||
+      chunksToRender.length !== this.visibleChunks_.size;
```
shlomnissan: Size equality doesn't imply chunk equality. Given that we always render all chunks for a timepoint/LOD this check looks unnecessary? If we remove it `visibleChunks` should probably be renamed to reflect that it's not a culled set.

packages/core/src/layers/volume_layer.ts
```diff
@@ -1,24 +1,258 @@
...
+
+    if (!needsUpdate) return;
+
+    // Clear and rebuild visible chunks
```
shlomnissan: This comment is misleading. We don’t clear and rebuild visible chunks. We diff the current chunk set against the new one, release chunks that fell out, then clear and rebuild the render list even for chunks that stayed. I'd remove the comment but if we keep it, it should reflect that.

Also, `currentChunkSet` reads like the existing state. `newChunkSet` would be clearer and `visibleChunks` could be renamed to `currentChunks`.

packages/core/src/layers/volume_layer.ts
```diff
@@ -1,24 +1,258 @@
...
+    for (const chunk of chunks) {
+      const volume = this.visibleChunks_.get(chunk);
+      if (volume) {
+        this.pool_.release(poolKeyForChunk(chunk), volume);
+        this.visibleChunks_.delete(chunk);
+      }
+    }
+  }
+
+  public update(_context?: RenderContext) {
```
shlomnissan: - Rename `_context?` to `context`. The leading underscore suggests the parameter is unused.
- @aganders3 I’m not sure when `RenderContext` was added to update but it doesn’t seem like it belongs there. The layer already gets a guaranteed context in `onAttached` so we can store it and avoid optionals and error throwing in update.

packages/core/src/layers/volume_layer.ts
```diff
@@ -1,24 +1,258 @@
...
+    this.chunkStoreView_.updateChunkStatesForVolume(
+      this.sliceCoords_,
+      this.lod_
+    );
+
+    this.updateChunks();
+    if (_context === undefined) {
+      throw new Error(
+        "RenderContext is required for the VolumeLayer update as camera information is used to reorder the chunks."
+      );
```
shlomnissan: As mentioned above this check shouldn't be needed. But if we keep it, move it to the top of `update()`. There's no point updating chunk state and rebuilding objects only to throw right after.

packages/core/src/renderers/WebGLState.ts
```diff
@@ -23,6 +24,16 @@ export class WebGLState {
...
+    this.initializeWindingOrder();
+  }
+
+  private initializeWindingOrder() {
+    // Idetik implements a y-axis flip in the projection matrix in
+    // the webgl renderer. That makes the coordinate system
+    // left-handed, whereas WebGL uses a right-handed system.
+    // To account for this, we need to change the front face
+    // winding order to clockwise.
+    this.gl_.frontFace(this.gl_.CW);
```
shlomnissan: We can inline this into the constructor. Also the comment is misleading. This isn’t about WebGL being right or left handed. WebGL operators in clip space. We flip Y in the projection matrix (see `axisDirection` in `WebGLRenderer`) which mirrors geometry and flips triangle winding. Because of that we need to treat clockwise triangles as front-facing.

I think the original comment in the renderer should change as well:

```ts
// Idetik defines screen-space with +Y pointing downward.
// With the default camera, the basis vectors are:
// (1, 0, 0) → right
// (0, 1, 0) → down
// (0, 0, 1) → out of the screen
//
// To match this convention, we flip Y in the projection matrix.
// This is a mirror transform, which also flips triangle winding.
const axisDirection = mat4.fromScaling(mat4.create(), [1, -1, 1]);
```

packages/core/src/renderers/webgl_shader_program.ts
```diff
@@ -54,6 +54,13 @@ export class WebGLShaderProgram {
           this.gl_.uniform1fv(location, value as Iterable<number>);
         }
         break;
+      case this.gl_.INT:
+        if (typeof value === "number") {
+          this.gl_.uniform1i(location, value);
+        } else {
+          this.gl_.uniform1iv(location, value as Iterable<number>);
```
shlomnissan: I think `uniform1iv` takes an `Int32Array | number[]` not a generic `Iterable<number>`. The cast hides potential runtime issues (e.g. passing a generator). I'd tighten the type here and same for float.

shlomnissan (changes requested):
I've finished reviewing the PR (excluding the ray marching shader which we can merge as is). Most of my comments are nits. I’m blocking for now only because there are enough small things I'd like to tighten up before we merge.

packages/core/src/core/chunk_store_view.ts
aganders3: > For example, keep SliceCoordinates at the public API boundary, compute timeIndex once, and pass to all internal functions.

I think this makes the most sense for the time being. I'd also consider moving time out of `SliceCoordinates` since it is treated differently.

packages/core/src/core/chunk_store_view.ts
aganders3: I'm not entirely sure this is correct, but need to think about it. One problem that arises when trying to maintain a sparse collection of view states here is that you can't just delete them. If they're removed the aggregator does not have a chance to collect the new "no longer necessary" state.

I think it's acknowledged that this whole function is temporary, so I'm okay with it going in and re-evaluating when we get to the above TODO to unify visibility calculations (which I will create an issue for).

> the error message references `updateChunkStatesForVolume` which could be confusing during debugging since I don’t think that function is called from this path

Isn't that this function?

packages/core/src/core/chunk_store_view.ts
aganders3: In general this method is a bit awkward being tied to the Z-dimension. I think it's fine if `SliceCoords` captures either 2D or 3D, but agree it should be explicit. This may also be related to #559 and addressed in that future work.

packages/core/examples/multi_viewport/main.ts
seankmartin: Made this one as an issue at https://github.com/chanzuckerberg/idetik/issues/560, thanks both!

packages/core/examples/multi_viewport/main.ts
seankmartin: We removed this zoom and seems to look fine, thanks

packages/core/examples/volume_rendering/main.ts
seankmartin: I think maybe it is best to remove this because it creates the wrong expectation. The intention was to follow the process shown in some other examples whereby when playback through time was enabled, the policy would switch to a playback policy. 

However, the `updateChunkStates` method in the `chunk_store_view` which would actually use the policy by eventually calling `markTimeChunksForPrefetch` does not support a perspective camera. As such the volume layer does not call `updateChunksStates`.

What I will do is remove the toggle as I think it hopefully makes it clearer that the volume layer is kind of doing its own thing right now and doesn't take advantage of switching between exploration and playback policy for t prefetch.

packages/core/src/core/chunk_store_view.ts
seankmartin: See https://github.com/chanzuckerberg/idetik/issues/558

packages/core/src/core/chunk_store_view.ts
seankmartin: See https://github.com/chanzuckerberg/idetik/issues/561

packages/core/src/layers/volume_layer.ts
seankmartin: Hopefully https://github.com/chanzuckerberg/idetik/pull/553/commits/526980b8c05c7ee32bd6d64885be5ecb3f600973 addresses this, thanks for spotting that

packages/core/src/layers/volume_layer.ts
seankmartin: We use the setter on source policy but not getter, so removed the getter, thanks

packages/core/src/layers/volume_layer.ts
seankmartin: The volume example is using the getter and setter on this one (though has been renamed to debugShowWireframes)

packages/core/src/layers/volume_layer.ts
seankmartin: Made an issue for this one just so it doesn't get lost

packages/core/src/layers/volume_layer.ts
seankmartin: Me moved it to the top and also did the same in the `chunked_image_layer` where this was essentially copied from

packages/core/src/layers/volume_layer.ts
seankmartin: We've changed the name but can't directly remove this check right now, the size check (which agreed is not a sufficient check by itself) is because the chunk store view can load chunks in pieces and if new chunks are loaded which weren't available on the last run of the update loop then they get handled here.

This is not perfect, and will likely want to improve. Very open to how you see that being done best. I would imagine as the volume rendering becomes a little bit less special cased that also might come a bit more generically.

One thing to note is that we've removed the LOD check here since the policy LOD is used instead. If for some reason the LOD changing resulted in the same number of chunks then yes, we run into the issue about chunks not actually being equal.

packages/core/src/layers/volume_layer.ts
seankmartin: Renamed the context, thanks

packages/core/src/layers/volume_layer.ts
seankmartin: Thanks, have moved it up and kept the check for now but will open an issue about keeping the context from `onAttached`

shlomnissan (approved):
Let's merge this :+1:

PR #564, opened 2026-01-08
feat: Multichannel volume rendering example without OIT

seankmartin:
image

Example of multichannel volume rendering by using multiple texture units.


seankmartin:
As a note, for the OIT work there is likely still some useful parts here, namely the channelProps for controlling the contrast limits on each channel, and the setup for the multichannel dataset in the example

seankmartin:
@shlomnissan @aganders3 @andy-sweet @aranega I fixed this to allow rendering chunks as they stream in from different channels. Videos attached. In the blue/green version it takes a while to see anything at LOD 1/0 because the first channel is set to non visible, so all visible example also included.

Two channel:
https://github.com/user-attachments/assets/9f06ec5c-66b7-49c3-8ed1-6ff756a6efa1


Three channel:
https://github.com/user-attachments/assets/db9169e7-691e-4756-bb70-394b2d44b38b

packages/core/src/objects/renderable/volume_renderable.ts
```diff
@@ -32,13 +34,45 @@ export class VolumeRenderable extends RenderableObject {
   }
 
   public override getUniforms(): Record<string, unknown> {
-    const channel = this.channels_[0] ?? validateChannel(this.textures[0], {});
-    const { color, contrastLimits } = channel;
+    const texture = this.textures[0];
+    if (!texture) {
+      throw new Error("No texture set");
+    }
```
aganders3: This means you cannot render a volume if channel 0 is not in the layer's `SliceCoordinates`.

packages/core/src/objects/renderable/volume_renderable.ts
```diff
@@ -32,13 +34,45 @@ export class VolumeRenderable extends RenderableObject {
...
+    const colors = [
+      1, 1, 1,
+      1, 1, 1,
+      1, 1, 1,
+      1, 1, 1,
+    ] // Up to 4 channels, RGB color for each
+    const valueOffset = [0, 0, 0, 0];
+    const valueScale = [1, 1, 1, 1];
+
+    const numChannels = this.textures.length;
```
aganders3: Using only `numChannels` means this will depend on channels loading in order, right? Like we will not start rendering with channel 2 even if it's available until channel 1 is loaded?

Or perhaps worse, might it crash if the textures arrive out of order?

Sorry if this is something you already brought up in the meeting. Instead I think we might need to bind an array of bool indicating available textures?

packages/core/src/layers/volume_layer.ts
```diff
@@ -250,6 +266,7 @@ export function poolKeyForChunk(chunk: Chunk) {
   return [
     `lod${chunk.lod}`,
     `shape${chunk.shape.x}x${chunk.shape.y}x${chunk.shape.z}`,
+    `locationx${chunk.chunkIndex.x}y${chunk.chunkIndex.y}z${chunk.chunkIndex.z}t${chunk.chunkIndex.t}`,
```
aganders3: I think I understand why this change was made, but it feels like they key is now being used for two purposes, and it makes the key too specific for the original usage. I think the idea of the pool is the renderable objects can be reused as long as their textures are compatible.

However with this feature we need to make sure textures are cleared when the renderable is released, since this could cause artifacts mixing channels from previous usage if they're not yet all overwritten.

I haven't looked at the pool code in a while, but this may be worth a closer look.

packages/core/src/objects/renderable/volume_renderable.ts
```diff
@@ -1,29 +1,31 @@
...
 
 export class VolumeRenderable extends RenderableObject {
   public voxelScale: vec3 = vec3.fromValues(1, 1, 1);
-
   private channels_: Required<Channel>[];
 
-  constructor(texture: Texture3D, channels: ChannelProps[] = []) {
+  constructor(
+    texture: Texture3D,
+    textureIndex: number,
```
aganders3: I'm not sure channel should map directly to `textureIndex`. This is kind of related to my other comment. The implementation here feels restrictive not just to four channels (not a big problem IMO) but to the _first_ four channels.

I think this may also interfere if implementing colormaps in the future (in the labels layers we use additional bound textures for colormaps).

aganders3 (commented):
Overall this looks great, and the approach seems sound, but I'm not yet confident it's ready to merge. That said, some of the issues seem beyond the scope here so merging this with known issues and plans to improve could be okay.

I have a few concerns about how this might behave with different datasets or configurations. For example to change the example to render only channel 2 still requires loading all data and using visibility to select the channel to display - this is effectively a regression from main for rendering a single channel.

Some of this could be helped/solved if we update `SliceCoordinates` (for example) to allow selecting specific channels for loading beyond current capabilities (one channel or all channels), because the `ChannelProps` effectively applies to _loaded_ channels. Obviously that's out of scope here, and probably warrants further design discussion.

seankmartin:
> Overall this looks great, and the approach seems sound, but I'm not yet confident it's ready to merge. That said, some of the issues seem beyond the scope here so merging this with known issues and plans to improve could be okay.
> 
> I have a few concerns about how this might behave with different datasets or configurations. For example to change the example to render only channel 2 still requires loading all data and using visibility to select the channel to display - this is effectively a regression from main for rendering a single channel.
> 
> Some of this could be helped/solved if we update `SliceCoordinates` (for example) to allow selecting specific channels for loading beyond current capabilities (one channel or all channels), because the `ChannelProps` effectively applies to _loaded_ channels. Obviously that's out of scope here, and probably warrants further design discussion.

Thanks for the review! Yes agreed, picking subsets of channels, stopping loading of certain channels etc is not ideal here - but just wasn't sure exactly how we want to handle that. Wanted to get some input first on what you do / want to do in Idetik and what assumptions can/can't be made and so on. I also agree that in this PR doing the ideal thing might be out of scope. But the comments you've made have given some ideas so I'll see if we can make some kind of middleground improvement
See https://github.com/chanzuckerberg/idetik/tree/seankmartin-aranega/fix-volume-renderable-channels for that, will merge in when ready, and https://github.com/chanzuckerberg/idetik/commits/seankmartin-aranega/volume-rendering-multichannel-atlas-novisibledownload/ aims to avoid the downloads on non visible channels

packages/core/src/layers/volume_layer.ts
```diff
@@ -40,7 +40,7 @@ export class VolumeLayer extends Layer implements ChannelsEnabled {
   private debugShowWireframes_ = false;
   public debugShowDegenerateRays = false;
   public relativeStepSize = 1.0;
-  public opacityMultiplier = 0.1;
+  public opacityMultiplier = 1.0;
```
shlomnissan, line 42: Can you say more about why this default value changed?

packages/core/src/layers/volume_layer.ts
```diff
@@ -102,6 +102,7 @@ export class VolumeLayer extends Layer implements ChannelsEnabled {
   private createVolume(chunk: Chunk) {
     const volume = new VolumeRenderable(
       Texture3D.createWithChunk(chunk),
+      chunk.chunkIndex.c,
```
shlomnissan: I think this is a bug. If the first channel isn't 0 `VolumeRenderable::getUniforms` breaks.

packages/core/src/layers/volume_layer.ts
```diff
@@ -122,9 +123,18 @@ export class VolumeLayer extends Layer implements ChannelsEnabled {
     const existing = this.currentChunks_.get(chunk);
     if (existing) return existing;
 
+    // Add texture as new channel to existing volume if match
+    for (const [existingChunk, volume] of this.currentChunks_) {
+      if (poolKeyForChunk(existingChunk) === poolKeyForChunk(chunk)) {
```
shlomnissan: Things are getting mixed up here. The pool key is used for reusability. If you want to check grouping, you can do it explicitly:

```
if (existingChunk.chunkIndex.x === chunk.chunkIndex.x &&
    existingChunk.chunkIndex.y === chunk.chunkIndex.y &&
    existingChunk.chunkIndex.z === chunk.chunkIndex.z &&
    existingChunk.chunkIndex.t === chunk.chunkIndex.t)
```

packages/core/src/layers/volume_layer.ts
```diff
@@ -122,9 +123,18 @@ export class VolumeLayer extends Layer implements ChannelsEnabled {
...
+        volume.setTexture(chunk.chunkIndex.c, Texture3D.createWithChunk(chunk));
+        return volume;
+      }
+    }
+
     const pooled = this.pool_.acquire(poolKeyForChunk(chunk));
     if (pooled) {
-      const texture = pooled.textures[0] as Texture3D;
+      const chunkIndex = chunk.chunkIndex.c;
+      const texture = pooled.textures[chunkIndex] as Texture3D;
```
shlomnissan: I think decoupling channel index and texture units is a good idea. 

Can `pooled.textures[chunkIndex]` be undefined if the pooled renderable wasn't previously populated with that channel? If yes, the `as Texture3D` cast hides it.

packages/core/src/layers/volume_layer.ts
```diff
@@ -157,12 +167,12 @@ export class VolumeLayer extends Layer implements ChannelsEnabled {
     const chunksToRender = this.chunkStoreView_.getChunksToRender(
       this.sliceCoords_
     );
+    const numChunksToRender = chunksToRender.length;
```
shlomnissan: I don't think extracting it to a variable doesn't improve readability or performance. Seems like an unnecessary change.

packages/core/src/layers/volume_layer.ts
```diff
@@ -176,11 +186,17 @@ export class VolumeLayer extends Layer implements ChannelsEnabled {
       const volume = this.getVolumeForChunk(chunk);
       volume.wireframeEnabled = this.debugShowWireframes;
       this.currentChunks_.set(chunk, volume);
-      this.addObject(volume);
     }
+    const seenKeys = new Set<string>();
+    this.currentChunks_.forEach((volume, chunk) => {
```
shlomnissan: A simpler approach would be `new Set(this.currentChunks_.values())` to get unique volumes directly, since multiple channel chunks already reference the same object.

packages/core/src/layers/volume_layer.ts
shlomnissan: Yep. Adding `t` defats the purpose of a pool. We should revert this change.

Regarding textures, rather than relying on stale textures from a previous pool entry the renderable should know exactly how many channels it has and the shader should only sample active channels based on channel count.

shlomnissan (changes requested):

packages/core/src/objects/renderable/volume_renderable.ts
```diff
@@ -32,13 +34,45 @@ export class VolumeRenderable extends RenderableObject {
...
 
   public override getUniforms(): Record<string, unknown> {
-    const channel = this.channels_[0] ?? validateChannel(this.textures[0], {});
-    const { color, contrastLimits } = channel;
+    const texture = this.textures[0];
+    if (!texture) {
+      throw new Error("No texture set");
+    }
+    // One index per channel
+    const visible = [1, 1, 1, 1];
```
shlomnissan: The default should probably be `visible = [0, 0, 0, 0]`?

shlomnissan (changes requested):

packages/core/src/layers/volume_layer.ts
seankmartin: This changed to 1.0 as the multiplier as the default value because now we have the `channelProps` we can setup real contrast limits with invlerp on the channels and don't need to rely on the old system which needed some general opacity reduction because we didn't have that setup

packages/core/src/objects/renderable/volume_renderable.ts
seankmartin: Absolutely, this one is changing, thank you

packages/core/src/objects/renderable/volume_renderable.ts
seankmartin: Thank you for spotting that, this has been updated to something more flexible

packages/core/src/objects/renderable/volume_renderable.ts
seankmartin: This is very helpful, thank you for expanding upon what we talked about in the meeting. There is something done for this but I get the feeling it might be a little different to what you had in mind. I'll think on that a bit more

packages/core/src/layers/volume_layer.ts
seankmartin: Thanks both, fixed that -- misunderstood the purpose of the pool here. Thought it was to allow quickly picking up a renderable saved from a different lod/t combination but I get the idea much better now and the concern with stale data. Apologies for the confusion

packages/core/src/layers/volume_layer.ts
seankmartin: I have changed the API but will check if that bug still lingers, thanks for spotting this

packages/core/src/layers/volume_layer.ts
seankmartin: Agreed, this was a leftover from the atlas approach that I forgot to revert when a calculation was happening to determine `numChunksToRender`

packages/core/src/layers/volume_layer.ts
seankmartin: Completely agree thank you, have already done something similar but I'll double check if it matches the suggestion

seankmartin: There is a new function introduced,

```ts
private currentVolumes(): Set<VolumeRenderable> {
    return new Set(this.currentChunks_.values());
  }
```
as part of the volume layer as there are a few places that benefit from this. The only call to it that's strictly necessary is right here so we don't add the object as a renderable multiple times. However, it is also called in


```ts
public setChannelProps(channelProps: ChannelProps[]) {
    this.currentVolumes().forEach((volume) => {
      volume.setChannelProps(channelProps);
    });
    this.channelChangeCallbacks_.forEach((callback) => {
      callback();
    });
  }
```

And in wireframes

```ts
public set debugShowWireframes(value: boolean) {
    if (this.debugShowWireframes_ === value) return;
    for (const volume of this.currentVolumes()) {
      volume.wireframeEnabled = value;
    }
    this.debugShowWireframes_ = value;
  }
```
Those calls are not strictly necessary but I would think are likely a little cleaner than before where we looped over the channel chunks and set the channel props or wireframe property multiple times on the same volume reference

packages/core/src/layers/volume_layer.ts
seankmartin: Perfect, will make this change thank you for the suggestion

packages/core/src/layers/volume_layer.ts
seankmartin: Yes very good spot thank you, that cast should not be there, will fix

I will also think a little more on the pool as I see a different case that could possibly happen from what you bring up here.

Let's say we had just one spatial chunk. If the pool is populated multiple times at that chunk which has 10 channels, then 10 references to the same volume will go into the pool for that chunk. Then later if we disable multiple of the channels, e.g. only keep 2, when we pop the renderables back out of the pool, 8 references get left in the pool.

I see some other comments made about the renderable and the pool that I'm thinking about in the context of the current version at https://github.com/chanzuckerberg/idetik/blob/seankmartin-aranega/fix-volume-renderable-channels/packages/core/src/objects/renderable/volume_renderable.ts

Additionally I think the current `getVolumeForChunk` might be leaving many objects in the pool

packages/core/src/core/color.ts
```diff
@@ -72,6 +72,10 @@ export class Color {
       return new Color(colorLike[0], colorLike[1], colorLike[2], colorLike[3]);
     }
 
+    if (typeof colorLike === "string") {
```
seankmartin, line 75: This allows to directly use a hex string in the volume rendering example which is pretty handy with the GUI but can remove that change here and workaround it if prefer not to add

packages/core/src/layers/volume_layer.ts
```diff
@@ -69,18 +68,14 @@ export class VolumeLayer extends Layer implements ChannelsEnabled {
...
-      chunk.setChannelProps(channelProps);
+    this.currentVolumes().forEach((volume) => {
+      volume.setChannelProps(channelProps);
     });
     this.channelChangeCallbacks_.forEach((callback) => {
       callback();
     });
   }
 
-  public get channelProps(): ChannelProps[] | undefined {
```
seankmartin, line 80: I'll revert removing this function, but I'm not sure if this current setup is ideal.

Since the channelProps don't have anything to watch if they have been changed, only calling in `setChannelProps` I think it could be a bit confusing if you directly call `volumeLayer.channelProps[0].visible = true` and then not have the callbacks or call `setChannelProps` per volume. Perhaps it would be better to change the `ChannelsEnabled` interface to return a `Readonly` type on `channelProps` to indicate that this is for inspection only and not intended for direct mutation of values.

packages/core/src/objects/renderable/volume_renderable.ts
```diff
@@ -1,60 +1,172 @@
...
 
 export class VolumeRenderable extends RenderableObject {
   public voxelScale: vec3 = vec3.fromValues(1, 1, 1);
 
   private channels_: Required<Channel>[];
+  private channelToTextureIndex_: Map<number, number> = new Map();
+  private loadedChannels_: Set<number> = new Set();
 
-  constructor(texture: Texture3D, channels: ChannelProps[] = []) {
+  constructor(channels: ChannelProps[] = []) {
```
seankmartin, line 22: While this is a fairly large change to remove `texture` from the call signature, since we are dealing with chunks essentially streaming into this renderable now this is likely cleaner.

It also allows the renderable to exist standalone without any data in it, which could be useful later for debugging as we can render the wireframe even if there are no chunks (here I've manually disabled adding any chunk data into the renderable). Could be easier for testing adding into an volume renderable

Image

packages/core/src/layers/volume_layer.ts
shlomnissan: > Let's say we had just one spatial chunk. If the pool is populated multiple times at that chunk which has 10 channels, then 10 references to the same volume will go into the pool for that chunk. Then later if we disable multiple of the channels, e.g. only keep 2, when we pop the renderables back out of the pool, 8 references get left in the pool.

I'm not sure I fully understand this comment. When you say "pool is populated multiple times" do you mean multiple chunks that map to the same volume are released back to the pool? If that's the case, I think this code already addresses this case?

```ts
  private releaseAndRemoveChunks(chunks: Iterable<Chunk>) {
    const releasedVolumes = new Set<VolumeRenderable>();
    for (const chunk of chunks) {
      const volume = this.currentChunks_.get(chunk);
      if (volume && !releasedVolumes.has(volume)) {
        volume.clearLoadedChannels();
        this.pool_.release(poolKeyForChunk(chunk), volume);
        releasedVolumes.add(volume);
      }
      this.currentChunks_.delete(chunk);
    }
  }
```

> Additionally I think the current getVolumeForChunk might be leaving many objects in the pool

I think this comment is also addressed by the same code?

If that's not the case, can you explain the problem again?

packages/core/src/layers/volume_layer.ts
shlomnissan: **This is just a suggestion not a blocker**

Right now the layer is doing a lot of work to reconstruct the spatial grouping that the chunk store already implicitly knows about. What if we had something like:

```ts
// Replace currentChunks_ Map<Chunk, VolumeRenderable> with:
private readonly currentVolumes_: Map<string, VolumeRenderable> = new Map();

// Helper to get the spatial key
function spatialKey(chunk: Chunk): string {
  const { x, y, z, t } = chunk.chunkIndex;
  return `${x}:${y}:${z}:${t}`;
}

// Group flat chunk list by spatial position
function groupBySpatialIndex(chunks: Chunk[]): Map<string, Chunk[]> {
  const grouped = new Map<string, Chunk[]>();
  for (const chunk of chunks) {
    const key = spatialKey(chunk);
    let group = grouped.get(key);
    if (!group) {
      group = [];
      grouped.set(key, group);
    }
    group.push(chunk);
  }
  return grouped;
}
```

Then `updateChunks` is simplified to:

```ts
private updateChunks() {
  if (!this.chunkStoreView_) return;

  const chunksToRender = this.chunkStoreView_.getChunksToRender(
    this.sliceCoords_
  );
  const currentTime = this.sliceCoords_.t ?? -1;
  const grouped = groupBySpatialIndex(chunksToRender);

  const needsUpdate =
    this.lastLoadedTime_ !== currentTime ||
    grouped.size !== this.currentVolumes_.size;
  if (!needsUpdate) return;

  // Release volumes no longer needed
  for (const [key, volume] of this.currentVolumes_) {
    if (!grouped.has(key)) {
      volume.clearLoadedChannels();
      this.pool_.release(poolKeyForChunks(volume), volume);
    }
  }

  this.currentVolumes_.clear();
  this.clearObjects();

  for (const [key, chunks] of grouped) {
    const volume = this.getOrCreateVolume(key, chunks);
    volume.wireframeEnabled = this.debugShowWireframes;
    this.currentVolumes_.set(key, volume);
    this.addObject(volume);
  }

  this.lastLoadedTime_ = currentTime;
  if (this.state !== "ready") this.setState("ready");
}
```

and `getVolumeForChunk` is simplified (no 1/2/3 comments):

```ts
private getOrCreateVolume(key: string, chunks: Chunk[]): VolumeRenderable {
  const existing = this.currentVolumes_.get(key);
  if (existing) {
    for (const chunk of chunks) existing.updateVolumeWithChunk(chunk);
    return existing;
  }

  let volume = this.pool_.acquire(poolKeyForChunks(chunks[0]));
  if (volume && this.channelProps_) {
    volume.setChannelProps(this.channelProps_);
  } else {
    volume = new VolumeRenderable(this.channelProps_);
  }

  for (const chunk of chunks) volume.updateVolumeWithChunk(chunk);
  this.updateVolumeTransform(volume, chunks[0]);
  return volume;
}
```

Then:

- `currentVolumes()` deduplicate helper gone.
- The linear scan matching `x/y/z/t` in old step 2 gone.
- The `releasedVolumes` guard in `releaseAndRemoveChunks` gone.
- The `debugShowWireframes` setter and `setChannelProps` can iterate `currentVolumes_.values()` directly.
- The places that previously used `currentVolumes()` just become `this.currentVolumes_.values()`.

packages/core/src/layers/volume_layer.ts
```diff
@@ -99,13 +98,8 @@ export class VolumeLayer extends Layer implements ChannelsEnabled {
...
   }
 
-  private createVolume(chunk: Chunk) {
-    const volume = new VolumeRenderable(
-      Texture3D.createWithChunk(chunk),
-      this.channelProps_
-    );
-    this.updateVolumeChunk(volume, chunk);
-    return volume;
+  private currentVolumes(): Set<VolumeRenderable> {
```
shlomnissan, line 101: See my comment here: https://github.com/chanzuckerberg/idetik/pull/564#discussion_r2907091504 to potentially clean up this file.

packages/core/src/layers/volume_layer.ts
```diff
@@ -119,21 +113,34 @@ export class VolumeLayer extends Layer implements ChannelsEnabled {
...
     }
 
-    return this.createVolume(chunk);
+    // 3. Acquire pooled volume or create new one
+    let volume = this.pool_.acquire(poolKeyForChunk(chunk));
+    if (volume && this.channelProps_) {
+      volume.setChannelProps(this.channelProps_);
+    } else {
+      volume = new VolumeRenderable(this.channelProps_);
+    }
```
shlomnissan, line 139: I find this a little confusing. The `this.channelProps_` check is guarding against calling `setChannelProps` with `undefined`, but the else branch passes `this.channelProps_` (possibly `undefined`) into the constructor anyway. So the "protection" is inconsistent.

Does this work:

```ts
const volume = this.pool_.acquire(poolKeyForChunk(chunk)) ?? new VolumeRenderable();
if (this.channelProps_) {
  volume.setChannelProps(this.channelProps_);
}
```

packages/core/src/objects/renderable/volume_renderable.ts
```diff
@@ -1,60 +1,171 @@
...
```
shlomnissan, line 22: I think removing the `channels` and just defaulting to empty makes more sense?

The channels param is only used to initialize `this.channels_` via `validateChannels(null, channels)`. But with `null` as the texture `validateChannels` can't do meaningful validation. `setChannelProps` gets called anyway either from the layer or after loading chunk data no?

packages/core/src/objects/renderable/volume_renderable.ts
```diff
@@ -1,60 +1,171 @@
...
     this.geometry = new BoxGeometry(1, 1, 1, 1, 1, 1);
-    this.setTexture(0, texture);
-    this.programName = dataTypeToVolumeShader(texture.dataType);
     this.cullFaceMode = "front";
     this.depthTest = false;
-    // TODO handle visibility property of channels
-    this.channels_ = validateChannels(texture, channels);
+    this.channels_ = validateChannels(null, channels);
+    // Requires a name to be set to start, will be replaced once
+    // channel data is loaded and added as a texture to the renderable
```
shlomnissan, line 29: The comment itself is a smell. It's saying "this value is wrong but we'll fix it later."

the constructor is just setting a throwaway default. You could make `programName` start as undefined/empty and have the renderer skip objects with no program set which would be clearer?

packages/core/src/layers/volume_layer.ts
seankmartin: Overall sounds good to me, thank you @shlomnissan! Can look at making this change tomorrow

packages/core/src/layers/volume_layer.ts
seankmartin: Thanks @shlomnissan, yes it should now be fixed. I wrote all this code after realising the problem I was talking here. "do you mean multiple chunks that map to the same volume are released back to the pool" - yes but to try help clarify. What we had before:

In the release step:
1. For each `chunk` get the `volume` for that chunk and release it into the pool
2. Result - The pool contains a reference to the same volume multiple times. In most cases, you will end up with `n_channels * n_spatial_chunks` being in the pool.

In the acquisitation step:
1. For the first `chunk` that corresponds to a unique spatial location, acquire a `volume` from the pool for this chunk key.
2. Result - In the best match scenario, the pool gets left with `(n_channels - 1) * n_spatial_chunks` volume renderable references left in the pool for this LOD.

In addition to the above, even if we fix that bug to clear the pool for each chunk. You could have a different number of chunks in the release and acquisition step because the number of channels to use changed.
The code written here is designed to address both of these cases. Now we have

In the release step:
1. For each `chunk` add the `volume` for that chunk into the pool if that `volume` is unique.
2. Result `n_spatial_chunks` are in the pool.

In the acquisation step:
1. For each `chunk` that corresponds to a unique spatial location, acquire a `volume` from the pool for this chunk key.
2. Result - In the best match scenario, the pool is emptied for this LOD.

packages/core/src/layers/volume_layer.ts
shlomnissan: Def not a blocker. I'm still reviewing this PR but just wanted to share my thoughts on how I would organize it. No need to introduce PR churn, the existing logic in this particular flow looks good to me.

packages/core/src/objects/renderable/volume_renderable.ts
```diff
@@ -1,60 +1,171 @@
...
+    this.loadedChannels_.add(channelIndex);
+  }
+
+  private addChannelTexture(channelIndex: number, chunk: Chunk): void {
+    const texture = Texture3D.createWithChunk(chunk);
+    const textureIndex = this.textures.length;
+
+    this.setTexture(textureIndex, texture);
+    this.channelToTextureIndex_.set(channelIndex, textureIndex);
+    this.programName = dataTypeToVolumeShader(texture.dataType);
```
shlomnissan, line 56: If you have a mismatch data type between channels the program name gets overwritten to the last one added. I think we need an assertion or guard enforcing one data type per volume. Otherwise if mixed types ever showed up it would silently use the wrong shader for some channels.

packages/core/src/objects/renderable/volume_renderable.ts
```diff
@@ -1,60 +1,171 @@
...
+  }
+
+  private updateChannelTexture(textureIndex: number, chunk: Chunk): void {
+    const texture = this.textures[textureIndex];
+
+    if (!(texture instanceof Texture3D)) {
```
shlomnissan, line 62: Is there a code path that would put a non-Texture3D into a tracked texture slot?

packages/core/src/objects/renderable/volume_renderable.ts
```diff
@@ -1,60 +1,171 @@
...
+    }
+
+    texture.updateWithChunk(chunk);
+  }
+
+  public clearLoadedChannels() {
+    this.loadedChannels_ = new Set();
+  }
+
+  public override getUniforms(): Record<string, number[] | number> {
```
shlomnissan, line 75: This is called once per object per frame with no caching. I don't think it's a big deal since volume rendering is GPU-bound, but something to think about as a follow-up optimization (dirty flag + cached uniforms map)

packages/core/src/objects/renderable/volume_renderable.ts
```diff
@@ -1,60 +1,171 @@
...
+      const newTexture = Texture3D.createWithChunk(chunk);
+      this.setTexture(textureIndex, newTexture);
+      return;
+    }
+
+    texture.updateWithChunk(chunk);
+  }
+
+  public clearLoadedChannels() {
+    this.loadedChannels_ = new Set();
```
shlomnissan, line 72: Do you need to clear `channelToTextureIndex_` as well?

shlomnissan (approved):
I feel like this PR has some room for cleanup but non of the things I highlighted in my comments or the things that feel little fragile to me seem like bugs.

This PR currently has over 200 commits which feels crazy to me.

I'm okay to just merge it. Open issues if there are obvious things we can clean up.

seankmartin:
> I feel like this PR has some room for cleanup but non of the things I highlighted in my comments or the things that feel little fragile to me seem like bugs.
> 
> This PR currently has over 200 commits which feels crazy to me.
> 
> I'm okay to just merge it. Open issues if there are obvious things we can clean up.

Sounds good thank you, likely clearer if we do a refactoring pass for these further cleanups. I looked through the commit log briefly and for some reason this seems to have all the commits related to volume rendering (and some other commits from merges and other things) so likely some slightly wonky git history going on there. I think since the merge action is squashing it should be fine though even if the history is a little odd

packages/core/src/layers/volume_layer.ts
seankmartin: That's a good point, I think it would be better to just call `setChannelProps`  even if `this.channelProps_` is undefined. The real intention here was to avoid calling `setChannelProps` on a new volume renderable, and only on one obtained from the pool. In theory if `channelProps_` has become undefined it should actually be passed anyway as empty list to remove the old props that are set on the renderable from the pool

packages/core/src/objects/renderable/volume_renderable.ts
seankmartin: Sounds good to me, I think that is definitely better, thanks

packages/core/src/objects/renderable/volume_renderable.ts
seankmartin: Currently there is not, but because the textures are generic and publicly can be set via `setTexture` it could be that some other kind of texture is stored by a different means at that `textureIndex` and then overwritten here. If preferred we could just throw an error at that point instead but likely good to keep the check here I think since the textures can be set from outside the renderable.

I guess other alternative could be that the volume renderable doesn't allow such public generic `setTexture` calls by overriding it, and handling some logic in the volume renderable to try keep channel textures and regular textures separate. Or something along those lines

packages/core/src/objects/renderable/volume_renderable.ts
seankmartin: Agreed, created as an issue, thank you

packages/core/src/objects/renderable/volume_renderable.ts
seankmartin: Currently we shouldn't but could change the logic a bit to do this instead if preferable.

Right now the textures are kept on the volume renderable to allow reuse of these textures when the renderable is taken out of the pool. The channel map is kept because the textures array could contain mixed types of textures (e.g. a lookup table). So that map is used to track what is a channel texture and separate from any other kind of texture. Because of this, in the current setup if we clear `channelToTextureIndex_` then we'd also need to clear the `textures_` of the volume renderable. If knew something more about the textures, like `Texture3D` is always a channel texture for example, then that would allow reuse without needing this map

We're also looking into avoiding downloading non visible channel data in multi channel volume rendering, which likely ties into this cc @aranega

PR #565, opened 2026-01-09
chore: use latest version of zarrita v0.5

andy-sweet:
### Summary
This updates the version of zarrita to the latest stable v0.5 version instead the pre-release v0.4 version (`next.16`) that was presumably initially used during development, but never bumped to a stable release.

The benefits of bumping this version are as follows.

1) Targets a stable, not pre-release version.
2) Picks up some recent zarrita bug fixes.
3) Drops some zarrita transitive dependencies of @idetik/core (e.g. @zarrita/typedarray is gone because we don't use it)
4) Drops the numcodecs transitive dependency because zarrita now uses the built-in [DecompressionStream](https://developer.mozilla.org/en-US/docs/Web/API/DecompressionStream) for gzip and libz decoders.
5) Cleans up some of the zarrita import statements.
6) All warnings are removed from `npm run build`.

I'm not actually sure how useful (4) is because gzip and libz are not particularly common for zarrs (e.g. zstd and lz4+blosc being more common), so I'm not sure if the numcodecs modules actually get pulled down at runtime in most cases. If they are used, I think the built-in decompressors may be faster (since they don't rely on WASM), but I don't have any evidence for that. The one potential downside of this change is that encoding gzip and libz is not supported by default anymore, but we're not using that currently and it can be enabled by explicitly configuring the numcodecs encoders for zarrita. If we're concerned about this change, we could use the stable zarrita v0.4 instead which still gives us the other benefits.

Because of (6), I upgraded warnings to errors in the rollup config.


andy-sweet:
@aganders3 : I assigned you because I think you configured rollup, which is where I spent most of my time of changes (i.e. being really confused about tslib).

While the PR is not too big, there a few things being mixed up here, which is why I marked this as a draft for now. I think it's worth reviewing as a whole, if only because some of these changes interact in complex ways I don't entirely understand (again tslib). I'm happy to split this up into a few PRs (e.g. tslib dependency, strict engine requirement, bump zarrita to v0.4 or v0.5).

aganders3 (approved):
Thanks! This seems to be working for me. I also tested installed in `idetik-react` to make sure this didn't affect the use of webworkers in downstream applications, and it worked there too.

PR #566, opened 2026-01-09
chore: fix fresh install with LTS node

andy-sweet:
### Summary
This changes fixes a fresh install of idetik where I'm using the latest LTS version of node (v24) and the default npm (v11.6), which is consistent with the engine specification in `package.json` and I think represents a sane default that we should support.

#### Strict engine check

Using an older version of node or npm that is inconsistent with that engine specification makes `npm install` warn, but ultimately succeed. In order to make developers environments more consistent and prevent unnecessary churn of `package-lock.json`, I decided to add an `.npmrc` file to make the engine check fail on that command. I also added an `.nvmrc` file to make it easy for a developer and GitHub actions to use that default version by simply running `nvm use`.

#### tslib dependency

After running `npm install` with those versions, some transitive dependencies are removed from `package-lock.json`. In particular, `tslib` is removed which then causes `npm run build` to fail because `@rollup/plugin-typescript` requires it as a peer dependency. Therefore, I included `tslib` as a dev-dependency since that is consistent with the contents of `package-lock.json` before this change.

I think in general `tslib` should actually be a runtime dependency of `@idetik/core` (maybe as a peer dependency marked as external in the vite config), but it probably doesn't matter right now because our rolled up library doesn't actually use any of the helper functions it provides (e.g. `__extends`). That's likely because we target a recent ECMAScript version (2020) that doesn't need these.


.github/workflows/release.yml
```diff
@@ -26,7 +26,7 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@v4
         with:
-          node-version: 20
+          node-version: 24
```
phoenixAja: should this also come from .nvmrc?

phoenixAja (approved):
install and build work for me

.github/workflows/release.yml
andy-sweet: Yes. Thanks for the catch!

PR #570, opened 2026-01-15
fix: do not temporally prefetch out of channel chunks

### LINKED ISSUE #569: Temporal prefetching over time ignores channel index
When a source contains multiple channels, but a layer only presents one of them (i.e. its instance of `SliceCoordinates` has something like `sliceCoords_.c === 0`), temporal prefetching ignores that channel index (see [markTimeChunksForPrefetch](https://github.com/chanzuckerberg/idetik/blob/ec740a32b4b1f82bbdc0d3f8d27086e8179662d6/packages/core/src/core/chunk_store_view.ts#L357)).

This is different to fetching chunks at the current time, where [updateChunksAtTimeIndex](https://github.com/chanzuckerberg/idetik/blob/ec740a32b4b1f82bbdc0d3f8d27086e8179662d6/packages/core/src/core/chunk_store_view.ts#L312) does explicitly check the channel.

I think this was accidentally removed in #531.

The bug arose with one of the DCA datasets which had 3 channels, but no channel metadata, so was a little tricky to identify.

The simple fix is to just skipping prefetching any chunks not in the selected channel index.

andy-sweet:
### Summary
This ensures that only chunks in the channel specified in `SliceCoordinates` are temporally prefetched. If no channels are specified, then all channels will be prefetched, which is unchanged from `main`.

### Related Issue
Closes #569

### Tests & Checks
I tested the examples.

### DIFF
--- a/packages/core/src/core/chunk_store_view.ts
+++ b/packages/core/src/core/chunk_store_view.ts
@@ -284,6 +284,7 @@ export class ChunkStoreView {
     if (sliceCoords.t !== undefined) {
       this.markTimeChunksForPrefetch(
         currentTimeIndex,
+        sliceCoords,
         viewBounds3D,
         viewBoundsCenter2D
       );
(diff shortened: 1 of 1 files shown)


aganders3 (approved):
Thanks for catching!

Out of curiosity was this failing downstream or just causing extra fetch?

andy-sweet:
> Out of curiosity was this failing downstream or just causing extra fetch?

Just causing the extra fetch(es), though the lack of any explicit failure made that a little hard to track down.

PR #573, opened 2026-01-29
fix: cancel chunk loads when store is removed

### LINKED ISSUE #571: Chunk manager does not cancel chunk loads when store is removed
For context, I encountered this issue when using the DCA viewer using the following steps.

1. Open the viewer with default dataset (public zebrahub volume).
2. Open the network tab.
3. Turn on the temporal prefetch policy.
4. Quickly use some valid s3 dataset.

After the s3 dataset fetches stop for temporal prefetching, I see some fetches for the zebrahub dataset. I'm assuming these were enqueued in step (1) or (3), but were not cancelled on (4) as I would expect.

Note that I believe the corresponding layer created in (1) should have been removed from the layer manager, been detached from the idetik context, and had their view + store removed from the chunk manager.

aganders3:
### Summary
This addresses the bug discovered in the DCA Viewer where switching datasets with loading in-progress does not cancel queued or in-flight chunk requests. The problem is the affected chunks of a disposed view are not processed in the normal aggregation flow via `updateAndCollectChunkChanges`, so the `ChunkManager` is not made aware of the changes. The fix is to move disposal logic into the normal aggregation path.

A possible TODO here would be to raise an error if trying to use a `ChunkStoreView` after disposal.

### Related Issue
Closes #571 

### Tests & Checks
I added automated tests to capture the desired behavior.
I also tested locally with the DCA viewer to confirm changing sources stops loading from a previous source.

### DIFF
--- a/packages/core/test/chunk_store_view.test.ts
+++ b/packages/core/test/chunk_store_view.test.ts
@@ -67,9 +67,8 @@ describe("ChunkStoreView disposal", () => {
 });
 
 function createMockLoader(): ChunkLoader {
-  const dimensions = createSimpleDimensions();
   return {
-    getSourceDimensionMap: () => dimensions,
+    getSourceDimensionMap: createSimpleDimensions,
     loadChunkData: vi.fn(),
     loadRegion: vi.fn(),
   };
--- a/packages/core/test/chunk_store_view.test.ts
+++ b/packages/core/test/chunk_store_view.test.ts
@@ -27,12 +27,12 @@ describe("ChunkStoreView disposal", () => {
 
     view.dispose();
 
-    // next ChunkManager.update() should collect the disposed view's chunks with priority=null
+    // collection should return affected chunks from the disposed view
+    // chunks should have priority=null and visible=false (ready for cancellation)
     collectedChunks = store.updateAndCollectChunkChanges();
 
     expect(collectedChunks.size).toBeGreaterThan(0);
 
-    // chunks should have priority=null and visible=false (ready for cancellation)
     for (const chunk of collectedChunks) {
       expect(chunk.priority).toBe(null);
       expect(chunk.visible).toBe(false);
--- a/packages/core/test/chunk_store_view.test.ts
+++ b/packages/core/test/chunk_store_view.test.ts
@@ -1,0 +1,118 @@
+import { describe, expect, test, vi } from "vitest";
+import { ChunkStore } from "@/core/chunk_store";
+import { ChunkLoader, SourceDimensionMap } from "@/data/chunk";
+import { createNoPrefetchPolicy } from "@/core/image_source_policy";
+import { createTestViewport } from "./helpers";
+
+describe("ChunkStoreView disposal", () => {
+  test("disposed view's chunks ready for cancellation", () => {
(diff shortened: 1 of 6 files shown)


packages/core/test/chunk_store_view.test.ts
```diff
@@ -0,0 +1,118 @@
...
+import { createNoPrefetchPolicy } from "@/core/image_source_policy";
+import { createTestViewport } from "./helpers";
+
+describe("ChunkStoreView disposal", () => {
+  test("disposed view's chunks ready for cancellation", () => {
+    const loader = createMockLoader();
+    const store = new ChunkStore(loader);
+    const policy = createNoPrefetchPolicy();
+    const view = store.createView(policy);
+    const viewport = createTestViewport();
```
andy-sweet, line 13: Optional

Maybe a bit picky, but the parameters of the viewport's camera matter for this test since there must be an overlap between its field of view and that of the source being mocked (because of the visibility checks). Therefore, I would consider parameterizing this function with those parameters so it's more clear here.

packages/core/test/chunk_store_view.test.ts
```diff
@@ -0,0 +1,118 @@
...
+    expect(collectedChunks.size).toBeGreaterThan(0);
+
+    // verify some chunks have priority (are needed)
+    for (const chunk of collectedChunks) {
+      expect(chunk.priority).not.toBe(null);
+    }
+
+    view.dispose();
+
+    // next ChunkManager.update() should collect the disposed view's chunks with priority=null
```
andy-sweet: Optional

I might consider removing this comment since it relies on the coupling between `ChunkStoreView`, `ChunkStore`, and `ChunkManager` whereas this is really just a unit test for `ChunkStoreView`. It is useful in the context of this fix though, so maybe linking to the issue might be a good idea if you keep it (otherwise, this comment might become confusing).

packages/core/test/chunk_store_view.test.ts
```diff
@@ -0,0 +1,118 @@
...
+    for (const chunk of collectedChunks) {
+      expect(chunk.priority).not.toBe(null);
+    }
+  });
+});
+
+function createMockLoader(): ChunkLoader {
+  const dimensions = createSimpleDimensions();
+  return {
+    getSourceDimensionMap: () => dimensions,
```
andy-sweet:
```suggestion
  return {
    getSourceDimensionMap: createSimpleDimensions,
```

andy-sweet (approved):
Works for me. Just some optional comments. The tests are great because they don't rely on any async behavior, which is the problem I ran into when writing tests for ChunkManager to capture the integrated behavior.

packages/core/test/chunk_store_view.test.ts
aganders3: Good point - I removed the reference to the `ChunkManager` but kept a comment there. I think this PR should be enough connection to the original issue.

packages/core/test/chunk_store_view.test.ts
aganders3: This is a good idea, but I'm going to punt on it until we add some tests that need more specific visibility configuration.

Issue #555, opened 2025-12-19
Remove uniforms as being part of the volume layer


seankmartin:
As @shlomnissan @aganders3 have pointed out, having uniforms as part of a layer is not desired. The volume rendering code in https://github.com/chanzuckerberg/idetik/pull/553 uses this pattern to set some uniforms for the rendering.

The suggestion is to instead configure more ergonomic properties on the layer that will be ultimately provided as uniforms from each renderable object

Given that 553 is already pretty sizeable and this would require a bit of moving things around, I'd suggest to make this refactoring in another pass. But if it becomes blocking please let me know. Overall, when we do this we can also revise which uniforms are in use during volume rendering. A related problem is scaling the number of samples based on the relative length of rays, as that will likely also use a uniform. But I'll make that in a separate issue.

The main thing this would do is remove the `getUniforms()` from the layer base object, and move any uniforms on the volume layer into the volume renderable instead.

PR #574, opened 2026-01-30
Add sample density computation based on volume diagonal (longest ray)

aranega:
## Summary

This PR adds the computation of the density of samples for volume rendering based on the length of the ray. This enables the fragment shader to get a consistent number of samples to ensure that a ray through a small chunk gets fewer samples than a ray through a large chunk, proportionally to their world-space sizes.

## Notes

To perform the computation, the dimensions of the volume layer are passed to the `VolumeLayer` constructor. That's not ideal, but the other solution implemented (kept commented in the files to drive discussions) needed the information about the dimension of the full volume in world-space for the current LOD. The dimensions are held by the chunk store and are not directly accessible to the `VolumeLayer`, so basically two options were possible:

1. get the volume world-space dimensions for the current LOD from the chunk store view, which would be weird, the chunk store view is about chunks, not the full volume;
2. compute the world-space dimensions in the `VolumeLayer`, getting the current LOD and the volume dimensions from the chunk store view, which is slightly less weird, but weird anyway as it would expose the raw dimensions from the chunk store view. It's also a bit odd because although technically each LOD could have different bounds, in general the bounds of the whole volume at each LOD would be pretty constant, so having to use the current LOD at all is a bit misleading.

To enable discussions, we decided to pass the volume world-space size in the constructor and kept the solution 2 commented as this solution is slightly less weird as at least it concentrates computation related to the Volume inside a component related to the Volume.

## Related Issue

* This PR adds a new uniform in the `VolumeLayer` which is something that should be removed eventually (#555)

### DIFF
--- a/packages/core/examples/multi_viewport/main.ts
+++ b/packages/core/examples/multi_viewport/main.ts
@@ -50,6 +50,7 @@ const volumeLayer = new VolumeLayer({
   source,
   sliceCoords: volumeCoords,
   policy: createPlaybackPolicy({ lod: { min: 2, max: 2 } }),
+  worldSize: vec3.fromValues(right - left, bottom - top, z.scale * z.shape),
 });
 
 const camera2D = new OrthographicCamera(left, right, top, bottom);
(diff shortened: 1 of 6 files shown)


seankmartin:
Closing in favour of #578

PR #578, opened 2026-02-04
feat: refactor volume rendering to use a relative step size parameter

### LINKED ISSUE #557: Volume rendering - adjust number of samples based on chunk shape and/or scale
For volume rendering the ray marching is performed in model space (where the model is always a unit cube). So the sampling density used to calculate the number of samples when combined with the ray length has no way to be consistent across the volume with different sized chunks. To allow for handling chunks of different sizes we need to provide a scale. See as well https://github.com/chanzuckerberg/idetik/pull/553#discussion_r2631868320

Overall this is related to how the rendering of a single ray in the volume is split into chunks. If we really wanted say 128 samples over the full ray though the whole volume, then we'd need some information above the size of the chunks the ray passes through to accurately the divide the number of samples across the two chunks.

aganders3:
### Summary
This is an alternative to #574. Here I borrow the concept of "relative step size" from vispy. This is basically the step size relative to a single voxel. To me this is quite intuitive: you generally want the step size to be ~one voxel, because sampling the same voxel more than once you don't get new information beyond what you could estimate. I think a default of 1.0 is fine for now. Experimenting, you can also see the step-size artifacts generally disappear below 1.0:

https://github.com/user-attachments/assets/687a0357-0572-4af8-9818-10f7bb7c789a

A benefit here is not needing any volume-level uniforms - only the voxel scale which is already available on the chunk.

### Related Issue
Closes #557 

### Tests & Checks
I tested with manual visual inspection of the zebrafish dataset and adjusted the step size
I also added a debug mode to color voxels by the actual number of steps taken, or the initial calculated ray length (without early termination) to confirm the number of steps "looks" correct given the proxy geometry. Note how the "hot" spot aligns with the chunk diagonal.
Screenshot 2026-02-04 at 11 28 25 AM

https://github.com/user-attachments/assets/a8fa40ad-9391-449c-a3c7-a0f5685a7309

### DIFF
--- a/packages/core/examples/volume_rendering/main.ts
+++ b/packages/core/examples/volume_rendering/main.ts
@@ -64,8 +64,8 @@ gui
 
 const volumeFolder = gui.addFolder("Volume Rendering");
 volumeFolder
-  .add(volumeLayer, "samplesPerUnit", 16, 512, 1)
-  .name("Samples per unit");
+  .add(volumeLayer, "relativeStepSize", 0.25, 3.0, 0.1)
+  .name("Relative step size (voxels)");
 volumeFolder.add(volumeLayer, "maxIntensity", 1, 255, 1).name("Max intensity");
 volumeFolder
   .add(volumeLayer, "opacityMultiplier", 0.01, 1.0, 0.01)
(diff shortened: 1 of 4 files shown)


seankmartin (approved):
Looks great, thanks!

aranega (approved):
Hey @aganders3, sorry for the delay, I reviewed the PR last week and totally forgot to finish the review...
It looks super good! Thanks a lot for the PR! 🙏

Issue #559, opened 2026-01-06
Consider unifying visibility calculations for 2D and 3D visibility testing


aganders3:
Chunk visibility in 2D is using an awkward and inflexible "camera bounds" that makes assumptions about camera orientation. We should be using the frustum to calculate visible chunks. This should generally work for both 2D and 3D, using an intersection of the frustum and any "slicing" (likely from the layer itself). LOD selection may still need separate 2D/3D code paths.

See [#553 (comment)](https://github.com/chanzuckerberg/idetik/pull/553/changes/BASE..afea84f7477e0bf1561fa95c44bd4413764f3237#r2658210714) for some more relevant discussion.

> ```
>   // TODO: Eventually unify visibility calculations using the camera frustum for both
>   // orthographic (2D slices) and perspective (volume rendering) cameras. This would
>   // replace both updateChunkStates and updateChunkStatesForVolume with a single method
>   // that performs frustum culling based on the camera type.
>   public updateChunkStatesForVolume(
>     sliceCoords: SliceCoordinates,
>     lod: number
>   ): void {
> ```

shlomnissan:
I think this issue is a bit unclear.

For true 2D presentation the camera assumptions are intentional. The camera does not rotate around the data and chunk visibility reduces to a straightforward tile/viewport intersection test. That's not awkward or inflexible it's exactly what most 2D tiled viewers do and it's the behavior I would expect.

I think what you mean is rendering 2D slices within a 3D context (orthoslices)? That case is more awkward because chunk visibility becomes a 3D problem: frustum culling combined with a Z-slab or slice constraint. That is a fundamentally different test.

We could argue for a unified visibility path between 2D and 3D but we can also argue against it. LOD selection will almost certainly need separate logic for 2D and 3D anyway (with orthoslices falling into the 3D bucket or maybe not?). If we already need divergent paths there unifying visibility may not actually buy us much.

At this point I'm not convinced either approach is clearly better. Until we see what orthoslices actually look like in practice keeping the paths separate feels like the safer choice. We shouldn't let it block LOD selection for volume rendering.

One more clarification: we already apply frustum culling to geometry. Geometry outside the frustum should not render. This issue seems specifically about chunk loading not rendering and that distinction is sometimes unclear.

We may need clearer terminology for all this.

seankmartin:
[...]

Issue #481, opened 2025-10-06
Enable Data Slicing in Multiple Orientations


phoenixAja:
#### Summary
Add support for slicing datasets in arbitrary orientations (e.g., XY, XZ, YZ planes) within Idetik’s visualization core. This functionality will allow exploration of volumetric data from multiple viewpoints, improving usability and enabling **orthoslices**—a key visualization feature for 3D microscopy and volumetric cell atlas data.

While not critical for DCA (Dynamic Cell Atlas) deliverables, this feature has **significant architectural and performance implications**, and unlocks several downstream capabilities (multi-orientation LOD management, synchronized slice viewers, etc.).

---

#### Background / Context
The Idetik core currently supports 2D slicing in a single orientation (typically XY), relying on the `Region` specification and `ChunkManager` for progressive data loading. However:
- The **ChunkManager** and **ImageLayer** classes are currently optimized for planar XY slicing.
- **Geometry** and **Camera** modules (e.g., `plane_geometry.ts`, `orthographic_camera.ts`) assume axis-aligned image layers.
- Orthogonal slicing (XZ/YZ) will require flexible plane definitions, correct world-to-data transform logic, and efficient memory reuse across slices.

This enhancement extends the architecture to handle arbitrary slicing orientations with minimal duplication or performance regression.

---

#### Goals
- Enable slicing along arbitrary orthogonal planes (XY, XZ, YZ).
- Integrate with `ChunkManager` for streaming chunks along new slice dimensions.
- Maintain compatibility with existing rendering and LOD systems.

PR #579, opened 2026-02-04
[WIP] Additional slice orientation POC

aganders3:
### Summary
`ChunkedImageLayer` currently only supports XY slice orientation (viewing down Z axis). To support multiple orthogonal viewports and arbitrary slice orientations, we needed to generalize the slice coordinate system and chunk management to handle XZ and YZ orientations as well.                                                                                                                                                                                     

To achieve this, this proof-of-concept PR makes the following changes:                                                                                                                                                                                                                              
  1. Generalized `SliceCoordinates` - replaced fixed `{z?, c?, t?}` type with flexible `{orientation: "xy"|"xz"|"yz"|"volume", x?, y?, z?, c?, t?}` sum type and a collection of helper functions (`getSlicePosition`, `getSliceAxis`, `getTextureDimensions`, etc.) to handle orientation-specific logic.
  2. Updated chunk visibility and rendering - refactored `ChunkStoreView` to use orientation-aware slice intersection tests. Chunks now correctly determine visibility for any axis-aligned slice orientation using frustum culling combined with slice plane intersection. A side-effect here is we can use the same code for volume rendering, just skipping the slice plane intersection.
  3. Neighbor-based spatial prefetch - implemented prefetch system that works in chunk index space rather than view space. For each visible chunk, look up neighbors within policy distance using direct array indexing (added `ChunkStore.getChunkByIndex`). This approach naturally extends to future oblique slices since prefetch is defined per-axis rather than in-plane/through-plane.
  4. Texture slicing for all orientations - updated `Texture2DArray` and `ChunkedImageLayer` to extract correct 2D slices from 3D chunks based on orientation.

Obviously this POC is a relatively big PR, though it represents only a small breaking change in how we specify the slice coordinates for certain layers. Consider this PR and the associated document (linked below) as suggestions or a starting point for discussing how to implement these features.

### Related Issue
#481
#559
See [Idetik: Axis-Aligned Slice Orientations](https://docs.google.com/document/d/1pDq8Urzgu8H5cTLrbKvK2bpB8GKRJE0BAW5nqWnwUr8/edit?usp=sharing) for a more detailed description

### Tests & Checks
Added "Slice Orientation" example that lets you select separate slice orientations with the zebrahub dataset.

https://github.com/user-attachments/assets/8ea5cb8e-2b63-408b-9135-19e0451e81b9


shlomnissan:
@aganders3 Looks good. I will review the document on Monday. What's going on with the FPS?

aganders3:
> What's going on with the FPS?

Good question haha. I just did a quick check to make sure nothing weird was going on, and I think the FPS is fine. I was probably just doing something else (zoom) at the time. Please confirm, though 😬.

packages/core/src/data/slice_coordinates.ts
```diff
@@ -0,0 +1,230 @@
...
+    case "xy":
+      return coords.z;
+    case "xz":
+      return coords.y;
+    case "yz":
+      return coords.x;
+    case "volume":
+      return undefined;
+  }
+}
```
aganders3, line 58: The repeated use of `switch (orientation)` in these utilities is kind of OO-disguised-as-functional, but I sort of prefer keeping `SliceCoordinates` as an object and being explicit about this dispatching for this case.

packages/core/src/data/slice_coordinates.ts
```diff
@@ -0,0 +1,230 @@
...
+  orientation: "volume";
+  t?: number;
+  c?: number;
+};
+
+export type SliceCoordinates =
+  | SliceCoordinatesXY
+  | SliceCoordinatesXZ
+  | SliceCoordinatesYZ
+  | SliceCoordinatesVolume;
```
aganders3, line 35: Future work I could see adding `SliceCoordinatesOblique` that would hold a predetermined `SlicePlane`.

packages/core/src/layers/chunked_image_layer.ts
```diff
@@ -262,15 +365,30 @@ export class ChunkedImageLayer extends Layer implements ChannelsEnabled {
   }
 
   private updateImageChunk(image: ImageRenderable, chunk: Chunk) {
+    if (!isAxisAlignedSlice(this.sliceCoords_)) {
+      throw new Error(
+        "ChunkedImageLayer requires axis-aligned slice coordinates"
+      );
+    }
```
aganders3, line 372: I don't like how this is checked at runtime in so many places. There should be a better way to guard this with TypeScript, and not have to check the type of `SliceCoords` repeatedly - especially since the orientation of the slice is effectively immutable once defined.

I have broader questions on keeping `SliceCoordinates` mutable as we do, but I would save that for a separate discussion/PR.

packages/core/examples/slice_orientation/main.ts
```diff
@@ -0,0 +1,191 @@
...
+          camera.setFrame(xMin, xMax, yMin, yMax);
+          camera.transform.setTranslation([xCenter, yCenter, zMax + 10]);
+          camera.transform.setRotation(quat.create());
+
+          currentLayer = new ChunkedImageLayer({
+            source,
+            sliceCoords: { orientation: "xy", z: zCenter, t: 400, c: 0 },
+            policy: createExplorationPolicy(),
+            channelProps: [{ contrastLimits: [0, 200] }],
+          });
```
aganders3, line 118: I think instantiating a new layer when changing orientation could be avoided if we had a way to set a new `sliceCoords` on the layer, but then anything hold a reference to the old `sliceCoords` would also need to be updated with the new reference to be able to mutate it (effectively this does the same thing).

Otherwise I don't think orientation can be mutated in place because you'd need to change orientation and slice axis at the same time. I could be missing something there though.

packages/core/examples/slice_orientation/main.ts
```diff
@@ -0,0 +1,191 @@
...
+          break;
+        }
+
+        case "xz": {
+          // XZ slice: looking from the front (along -Y)
+          camera.setFrame(xMin, xMax, zMin, zMax);
+          camera.transform.setTranslation([xCenter, yMin - 10, zCenter]);
+          const rotationXZ = quat.create();
+          quat.rotateX(rotationXZ, rotationXZ, Math.PI / 2);
+          camera.transform.setRotation(rotationXZ);
```
aganders3, line 138: I think a factory function or a helper orient a camera to face a given slice would be helpful here.

PR #581, opened 2026-02-04
fix: use loading policy LOD min and max to define fallback LOD

### LINKED ISSUE #577: Loading policy LOD limits are not always respected by chunk loading
The LOD limits are respected when determining the [target LOD](https://github.com/chanzuckerberg/idetik/blob/8ca43694e91a6eda5bd1a4d4ba50295dd7c70356/packages/core/src/core/chunk_store_view.ts#L249), but not when [prefetching over time](https://github.com/chanzuckerberg/idetik/blob/8ca43694e91a6eda5bd1a4d4ba50295dd7c70356/packages/core/src/core/chunk_store_view.ts#L366) or when determining the [fallback LOD](https://github.com/chanzuckerberg/idetik/blob/8ca43694e91a6eda5bd1a4d4ba50295dd7c70356/packages/core/src/core/chunk_store_view.ts#L84).

The motivation here is allow applications like the DCA viewer to explicitly limit the LODs (in both directions) when the lowest resolution (current fallback) is too low to be of use.

andy-sweet:
### Summary
This uses the loading policy's LOD min and max to always define the fallback LOD, instead of using the lowest resolution LOD from the store.

This is motivated by the need for some applications to increase the minimum LOD that should be loaded/presented in cases when the lowest resolution LOD in the source is not useful. This is particularly important for the crude temporal prefetching logic, which only fetches the fallback LOD.

### Related Issue
Closes #577

### Tests & Checks
I manually tested the main examples that use chunk streaming, with particular attention on the one modified here that sets the LOD min and max based on the playback state.

### DIFF
--- a/packages/core/examples/ca_wave_dynamics/main.ts
+++ b/packages/core/examples/ca_wave_dynamics/main.ts
@@ -5,7 +5,7 @@ import {
   OmeZarrImageSource,
   OrthographicCamera,
   Color,
-  createExplorationPolicy,
+  createNoPrefetchPolicy,
   createPlaybackPolicy,
 } from "@";
 import { PanZoomControls } from "@/objects/cameras/controls";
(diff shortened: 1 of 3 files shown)


shlomnissan (approved):
LGTM :+1:

andy-sweet:
I'm not quite sure what the plans are for 3D and LOD, so I didn't touch those paths here. It looks like there's a `setLOD(0)` which could maybe be `setLOD(clamp(this.policy_.lod.min, this.policy_.lod.max))` instead, but I'll leave that for a follow-up.

PR #584, opened 2026-02-09
feat: change detection and time prefetch for volume rendering

shlomnissan:
### Summary

This PR refines `ChunkStoreView` to better align image and volume
update paths while avoiding unnecessary per-frame work.

- Add camera change detection for volume rendering using cached
view-projection matrices.
- Skip volume updates when policy, camera, and time are unchanged.
- Add timepoint prefetching for volume rendering at fallback LOD.
- Split image and volume time prefetch logic for clarity.

### Tests & Checks
- All checks have passed
- Manual verification (playback works as expected in the volume rendering examples)

### DIFF
--- a/packages/core/src/core/chunk_store_view.ts
+++ b/packages/core/src/core/chunk_store_view.ts
@@ -1,7 +1,6 @@
 import { Chunk, SliceCoordinates, ChunkViewState } from "../data/chunk";
 import type { ChunkStore } from "./chunk_store";
 import { Viewport } from "./viewport";
-import { Camera } from "../objects/cameras/camera";
 import { OrthographicCamera } from "../objects/cameras/orthographic_camera";
 import { ImageSourcePolicy } from "./image_source_policy";
 import { ReadonlyVec2, vec2, vec3, mat4 } from "gl-matrix";
@@ -179,9 +178,18 @@ export class ChunkStoreView {
     sliceCoords: SliceCoordinates,
     viewport: Viewport
   ): void {
+    // Each call allocates a new mat4 and computes projection * view.
+    // This is intentional for simplicity. If this ever shows up in profiles
+    // avoid multiply entirely by caching and comparing view/projection separately.
+    const viewProjection = mat4.multiply(
+      mat4.create(),
+      viewport.camera.projectionMatrix,
+      viewport.camera.viewMatrix
+    );
+
     const changed =
       this.policyChanged_ ||
-      this.viewProjectionChanged(viewport.camera) ||
+      this.hasViewProjectionChanged(viewProjection) ||
       this.lastTCoord_ !== sliceCoords.t;
 
     if (!changed) return;
@@ -205,14 +213,21 @@ export class ChunkStoreView {
 
     this.chunkViewStates_.forEach(resetChunkViewState);
 
+    const fallbackLOD = this.fallbackLOD();
+
     for (const chunk of currentTimeChunks) {
       if (chunk.lod !== this.currentLOD_) continue;
+      if (!this.isChunkChannelInSlice(chunk, sliceCoords)) continue;
 
-      const isChannelMatch =
-        sliceCoords.c === undefined || sliceCoords.c === chunk.chunkIndex.c;
-      if (!isChannelMatch) continue;
+      const isFallbackLOD = chunk.lod === fallbackLOD;
+      const priority = this.computePriority(
+        isFallbackLOD,
+        true, // isCurrentLOD
+        true, // isVisible
+        false, // isPrefetch
+        true // isChannelInSlice
+      );
 
-      const priority = this.policy_.priorityMap["visibleCurrent"];
       this.chunkViewStates_.set(chunk, {
         visible: true,
         prefetch: false,
@@ -227,6 +242,7 @@ export class ChunkStoreView {
 
     this.policyChanged_ = false;
     this.lastTCoord_ = sliceCoords.t;
+    this.lastViewProjection_ = viewProjection;
   }
 
   public allVisibleFallbackLODLoaded(sliceCoords: SliceCoordinates): boolean {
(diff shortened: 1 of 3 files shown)


packages/core/src/core/chunk_store_view.ts
andy-sweet: Optional: not in this change, but more concise and consistent with other related functions.

```suggestion
if (!this.isChunkChannelInSlice(chunk, sliceCoords)) continue;
```

packages/core/src/core/chunk_store_view.ts
```diff
@@ -382,6 +400,35 @@ export class ChunkStoreView {
...
+    );
+    const fallbackLOD = this.fallbackLOD();
+    const priority = this.policy_.priorityMap["prefetchTime"];
+
+    for (let t = currentTimeIndex + 1; t <= tEnd; ++t) {
+      for (const chunk of this.store_.getChunksAtTime(t)) {
+        if (chunk.lod !== fallbackLOD) continue;
+        if (!this.isChunkChannelInSlice(chunk, sliceCoords)) continue;
+
+        const orderKey = t - currentTimeIndex; // nearer future timepoints first
```
andy-sweet, line 436: In practice (with the browser cache off), I find that the t=400 chunks arrive last which means I have to wait a while before I see anything in the volume demo (which is a regression from main).

I don't see any obvious issues with the code here or anywhere else.

I tried changing the chunk queue size to 1, thinking that would force the closest chunks to the current time index to finish first, but then realized they are scheduled last.

UPDATE: now I see. The definition of `createPlaybackPolicy` means that "prefetchTime" is prioritized ahead of "visibleCurrent". I think the intent of that is to prioritizing prefetching potentially lower resolution chunks over time, compared to high (current LOD) resolution chunks at the current time point. There's some messiness here in terms of the assumption that the temporal prefetch is at the fallback LOD (which may not be true in the future).

The 2D case effectively works around this by using the "fallbackVisible" (instead of "visibleCurrent") in `computePriority`. Maybe the simplest thing for now is to reuse that function in `updateChunksForVolume`. 

In general, this issue could do with some deeper thought, but I think the workaround is simple enough and worth getting into this PR because it enables an improve visualization experience.

This isn't quite a blocking comment, but if we merge this change without some fix, then we should follow up quickly with one.

packages/core/src/core/chunk_store_view.ts
```diff
@@ -455,6 +502,30 @@ export class ChunkStoreView {
...
+    //   - Avoid the allocation by reusing a temporary mat4.
+    //   - Avoid the multiply entirely by caching and comparing view/projection separately.
+    const viewProjection = mat4.multiply(
+      mat4.create(),
+      camera.projectionMatrix,
+      camera.viewMatrix
+    );
+
+    if (this.lastViewProjection_ === null) {
+      this.lastViewProjection_ = mat4.clone(viewProjection);
```
andy-sweet: Nit: the clone here seems overly defensive, since this was created in the previous line and we return on the next line.

```suggestion
      this.lastViewProjection_ = viewProjection;
```

packages/core/src/core/chunk_store_view.ts
```diff
@@ -455,6 +502,30 @@ export class ChunkStoreView {
...
+      return true;
+    }
+
+    if (!mat4.equals(this.lastViewProjection_, viewProjection)) {
+      mat4.copy(this.lastViewProjection_, viewProjection);
```
andy-sweet: Nit: same comment as the clone. I'm not following why we can't directly assign by reference here.
```suggestion
      this.lastViewProjection_ = viewProjection;
```

packages/core/src/core/chunk_store_view.ts
```diff
@@ -173,7 +175,17 @@ export class ChunkStoreView {
...
   }
 
-  public updateChunksForVolume(sliceCoords: SliceCoordinates): void {
+  public updateChunksForVolume(
+    sliceCoords: SliceCoordinates,
+    viewport: Viewport
+  ): void {
+    const changed =
+      this.policyChanged_ ||
+      this.viewProjectionChanged(viewport.camera) ||
```
andy-sweet: Maybe blocking?

If the policy changes and the view projection matrix also changes, we never evaluate this function and we do not update `lastViewProjection_` as intended. In that case, we'll use `viewport.camera` to update the chunk states as intended. But if the actual view projection matrix has not changed on the next call to this function, we won't early return as intended defeating the point of this check.

I think we had a similar issue come up for the 2D path. The current fix we have is only check equivalence here, then always do assignment at the end of the function.

I think we should fix this issue in this change, either with the 2D approach or with something better.

andy-sweet (changes requested):
Generally looks good, but I have one maybe two significant issues with this before I can approve. See the more detailed inline comments.

packages/core/src/core/chunk_store_view.ts
shlomnissan: I followed your suggestion to use `computePriority` for now

packages/core/src/core/chunk_store_view.ts
```diff
@@ -193,14 +213,22 @@ export class ChunkStoreView {
...
+    const fallbackLOD = this.fallbackLOD();
+
     for (const chunk of currentTimeChunks) {
       if (chunk.lod !== this.currentLOD_) continue;
+      if (!this.isChunkChannelInSlice(chunk, sliceCoords)) continue;
 
-      const isChannelMatch =
-        sliceCoords.c === undefined || sliceCoords.c === chunk.chunkIndex.c;
-      if (!isChannelMatch) continue;
+      const isCurrentLOD = chunk.lod === this.currentLOD_;
```
andy-sweet: Nit: based on the condition above, this is always true, so you probably want to treat it in the same ways as `isChannelInSlice` (i.e. pass `true` with a comment).

andy-sweet (approved):
This looks good and works well for me now. Thanks!

PR #585, opened 2026-02-13
chore: generate a GitHub token for release from the Biohub GitHub helper

aganders3:
### Summary
I noticed releases and PR comments are still being by @phoenixAja, which I believe is due to the PAT in the repo secrets.

I [asked in #help-infra](https://chanzuckerbergteam.slack.com/archives/C94RQ5SBV/p1771000992092589) and it sounds like we can generate a key dynamically with this action and the existing org-wide secrets.

Longer-term I think they'd like us to use "release-please" instead, but I don't think it's a high priority.

We should also create a new NPM token, but I'm not sure the best way to do that in a sustainable way. Maybe we add a machine user in 1Password we can share?

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -21,7 +21,6 @@ jobs:
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
-          token: ${{ secrets.RELEASE_TOKEN }}
 
       - name: Setup Node.js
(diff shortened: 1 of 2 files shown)


aganders3:
I'm not sure we have a way to test this other than merging this PR

andy-sweet (approved):
I'm not an expert here, but seems fine to me!

PR #587, opened 2026-02-17
feat: add channel prop support to the volume layer

seankmartin:
### Summary

One thing is common to both https://github.com/chanzuckerberg/idetik/pull/564 and https://github.com/chanzuckerberg/idetik/pull/583 - which is that there needs to be some way to define the properties of the channels for volume rendering. We've extracted that common piece here, as it is more self contained and would be helpful to get feedback on so then #564 and #583 can be adding on top of that.

The `VolumeLayer` has copied some of the `channelProps` functionality from the other image layers with the callbacks on channel change, how to set channels etc. And how the `VolumeRenderable` sets those as a uniform is similar to the `ImageRenderable`. This also helps in moving towards https://github.com/chanzuckerberg/idetik/issues/555 and removing uniforms from the volume layer. In this only the first channel would be supported for passing the uniforms, but it lays the framework for using channel props for multi-channel rendering.

### Tests & Checks

I manually tested the volume rendering examples and checked other examples to see if still working. Visually this shouldn't really change much, outside of the transfer function bounds being a little different for the invlerp mapping.

New:
image

Old:
image

### DIFF
--- a/packages/core/src/layers/volume_layer.ts
+++ b/packages/core/src/layers/volume_layer.ts
@@ -63,10 +63,13 @@ export class VolumeLayer extends Layer implements ChannelsEnabled {
     }
   }
 
-  public setChannelProps(newProps: ChannelProps[]) {
-    this.channelProps_ = newProps;
+  public setChannelProps(channelProps: ChannelProps[]) {
+    this.channelProps_ = channelProps;
     this.currentChunks_.forEach((chunk) => {
-      chunk.setChannelProps(newProps);
+      chunk.setChannelProps(channelProps);
+    });
+    this.channelChangeCallbacks_.forEach((callback) => {
+      callback();
     });
   }
 
--- a/packages/core/src/layers/volume_layer.ts
+++ b/packages/core/src/layers/volume_layer.ts
@@ -123,6 +123,9 @@ export class VolumeLayer extends Layer implements ChannelsEnabled {
       const texture = pooled.textures[0] as Texture3D;
       texture.updateWithChunk(chunk);
       this.updateVolumeChunk(pooled, chunk);
+      if (this.channelProps_) {
+        pooled.setChannelProps(this.channelProps_);
+      }
       return pooled;
     }
 
--- a/packages/core/src/layers/volume_layer.ts
+++ b/packages/core/src/layers/volume_layer.ts
@@ -8,7 +8,7 @@ import { Texture3D } from "../objects/textures/texture_3d";
 import { RenderablePool } from "../utilities/renderable_pool";
 import { glMatrix, vec3 } from "gl-matrix";
 import { Camera } from "../objects/cameras/camera";
-import { ChannelProps, ChannelsEnabled } from "@/objects/textures/channel";
+import { ChannelProps, ChannelsEnabled } from "../objects/textures/channel";
 
 export type VolumeLayerProps = {
   source: ChunkSource;
--- a/packages/core/src/layers/volume_layer.ts
+++ b/packages/core/src/layers/volume_layer.ts
@@ -8,31 +8,34 @@ import { Texture3D } from "../objects/textures/texture_3d";
 import { RenderablePool } from "../utilities/renderable_pool";
 import { glMatrix, vec3 } from "gl-matrix";
 import { Camera } from "../objects/cameras/camera";
+import { ChannelProps, ChannelsEnabled } from "@/objects/textures/channel";
 
 export type VolumeLayerProps = {
   source: ChunkSource;
   sliceCoords: SliceCoordinates;
   policy: ImageSourcePolicy;
+  channelProps?: ChannelProps[];
 };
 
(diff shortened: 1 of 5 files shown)


packages/core/src/layers/volume_layer.ts
```diff
@@ -60,17 +63,51 @@ export class VolumeLayer extends Layer {
     }
   }
 
+  public setChannelProps(newProps: ChannelProps[]) {
```
shlomnissan: Nit: If copied over, might as well use same naming (i.e. `channelProps` instead of `newProps`).

packages/core/src/layers/volume_layer.ts
```diff
@@ -60,17 +63,51 @@ export class VolumeLayer extends Layer {
...
+    this.channelProps_ = newProps;
+    this.currentChunks_.forEach((chunk) => {
+      chunk.setChannelProps(newProps);
+    });
```
shlomnissan, line 77: The chunked image layer invokes callbacks when channel props are set:
```ts
this.channelChangeCallbacks_.forEach((callback) => {
    callback();
});
```
Why did we skip it here?

packages/core/src/layers/volume_layer.ts
```diff
@@ -60,17 +63,51 @@ export class VolumeLayer extends Layer {
     }
   }
 
+  public setChannelProps(newProps: ChannelProps[]) {
+    this.channelProps_ = newProps;
+    this.currentChunks_.forEach((chunk) => {
```
shlomnissan, line 72: When we get a pooled image in the chunked image layer, we set its channel props:

```ts
if (pooled) {
  const texture = pooled.textures[0] as Texture2DArray;
  texture.updateWithChunk(chunk, this.getDataForImage(chunk));
  this.updateImageChunk(pooled, chunk);
  if (this.channelProps_) {
    pooled.setChannelProps(this.channelProps_);
  }
  return pooled;
}
```

Why did we skip it here?

packages/core/src/objects/renderable/volume_renderable.ts
```diff
@@ -4,28 +4,60 @@ import { BoxGeometry } from "../geometry/box_geometry";
...
+  private channels_: Required<Channel>[];
+
+  constructor(texture: Texture3D, channels: ChannelProps[] = []) {
     super();
     this.geometry = new BoxGeometry(1, 1, 1, 1, 1, 1);
     this.setTexture(0, texture);
     this.programName = dataTypeToVolumeShader(texture.dataType);
     this.cullFaceMode = "front";
     this.depthTest = false;
+    // TODO handle visibility property of channels
```
shlomnissan, line 26: Is it a `TODO` because you still want to limit the scope of this PR to one channel? So this PR only covers the plumbing of channel props?

shlomnissan (changes requested):
Looks good! I'm blocking to get more information on:
- https://github.com/chanzuckerberg/idetik/pull/587#discussion_r2834719725
- https://github.com/chanzuckerberg/idetik/pull/587#discussion_r2834733142

packages/core/src/layers/volume_layer.ts
seankmartin: Thank you for spotting this, changing it. This came from the fact that we actually wrote this `setChannelProps` quite a while ago before realising there was an established structure and then essentially copying that code over from the chunked image layer instead. Forgot to update this one in the process, thanks for noticing. The same applies to https://github.com/chanzuckerberg/idetik/pull/587#discussion_r2834719725, so fixing that also

packages/core/src/layers/volume_layer.ts
seankmartin: No reason for skipping, glad you noticed it and apologies for missing in the first place, fixing that problem

packages/core/src/objects/renderable/volume_renderable.ts
seankmartin: Yes exactly, we could plumb it so that visibility is available as a uniform. But I think that might be better not to do right now as OIT and non-OIT would want to handle that visibility differently.

OIT would want to avoid rendering the chunk for that channel altogether if the channel is not visible, so this would be in the renderer. While non-OIT would have access to all the textures in the shader and conditionally sample based on the visibility of the channel, so having a uniform would be useful (it could also compile and use a different shader altogether to avoid the conditional in the shader but that's a bit of overhead to introduce right now)

seankmartin: In general that is also why this PR isn't further updating `getUniforms` right now, as for OIT and without OIT the `getUniforms` would look fairly different

packages/core/src/layers/volume_layer.ts
seankmartin: See https://github.com/chanzuckerberg/idetik/pull/587#discussion_r2838245979

shlomnissan (approved):
LGTM

PR #588, opened 2026-02-20
feat: downsample volume rendering on camera movement (1/2)

shlomnissan:
### Summary

Increases the ray march step size during camera interaction to improve volume rendering
responsiveness. When the camera is moving `RelativeStepSize` is scaled up to `2x`
halving the number of samples per ray. Full quality is restored when the camera stops.

This is the first of two changes:
1. **This PR**: Reduce number of ray march steps per fragment
2. **Next PR**: Render to a lower-resolution framebuffer and blit to screen

Together they reduce both fragment cost and fragment count during interaction.

### Tests & Checks

- Manual verification / downsampled step size when camera is moving.
- All checks have passed.

### DIFF
--- a/packages/core/src/layers/volume_layer.ts
+++ b/packages/core/src/layers/volume_layer.ts
@@ -15,6 +15,8 @@ export type VolumeLayerProps = {
   policy: ImageSourcePolicy;
 };
 
+const INTERACTIVE_STEP_SIZE_SCALE = 2.0;
+
 export class VolumeLayer extends Layer {
   public readonly type = "VolumeLayer";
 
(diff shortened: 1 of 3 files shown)


seankmartin (approved):
looks great, thanks! The only thing I find a bit odd from using it is that I have better FPS while moving the camera than when the camera still which is perhaps a bit counter-intuitive. Maybe some system to avoid re-rendering if no data has changed and the camera hasn't moved could come later. I think this change is good, just wanted to note that when I noticed it

aranega (approved):
Thanks @shlomnissan it looks great!

aganders3:
> Maybe some system to avoid re-rendering if no data has changed and the camera hasn't moved could come later.

I noticed the same when I experimented with something similar. It's not just a quirk, I think it's a real performance issue because there is a 1-frame lag before the camera appears to move. This can feel pretty slow if FPS is low for the high-quality render.

I agree this shouldn't block merging this feature as-is, but it's worth noting. Moving to event-driven rendering could be a pretty big change with significant complication e.g. if the underlying data may change (see napari for an example).

shlomnissan:
> Moving to event-driven rendering could be a pretty big change with significant complication

@aganders3 I think the change itself is limited? Just a flag. The tricky part is making sure every mutation path sets it.

Anyway. Out-of-scope for this PR but I might put a patch together.

PR #589, opened 2026-02-22
fix: Correct volume sorting for volume rendering

seankmartin:
### Summary

Corrects the sorting order to front-to-back in the volume layer to match the expectation in the blending mode set for volume rendering and the compositing in the volume shader. Have also written some tests to ensure this doesn't happen again, I think we had it right before and then broke it in refactoring.

Have also fixed an issue that the chunks were sorted, and then updated, which could cause an annoying flickering if the chunks were being updated pretty regularly. Now the chunks are updated, and then sorted. So this should be more consistent.

Finally, removed a check on if the distances were within a delta to just return 0 instead. I don't expect this is actually needed, but if I'm missing a case where that's needed then happy to revert it.

### Tests & Checks

I wrote automated tests and manually tested. To manually test I bumped the opacity up very high to see if the data closest to the camera appeared to be occluding what was behind it.

### DIFF
--- a/packages/core/src/layers/volume_layer.ts
+++ b/packages/core/src/layers/volume_layer.ts
@@ -7,7 +7,6 @@ import { ImageSourcePolicy } from "../core/image_source_policy";
 import { Texture3D } from "../objects/textures/texture_3d";
 import { RenderablePool } from "../utilities/renderable_pool";
 import { vec3 } from "gl-matrix";
-import { Camera } from "../objects/cameras/camera";
 import { sortFrontToBack } from "../math/sort_by_distance";
 import { ChannelProps, ChannelsEnabled } from "../objects/textures/channel";
 
--- a/packages/core/src/layers/volume_layer.ts
+++ b/packages/core/src/layers/volume_layer.ts
@@ -6,8 +6,9 @@ import { ChunkStoreView, INTERNAL_POLICY_KEY } from "../core/chunk_store_view";
 import { ImageSourcePolicy } from "../core/image_source_policy";
 import { Texture3D } from "../objects/textures/texture_3d";
 import { RenderablePool } from "../utilities/renderable_pool";
-import { glMatrix, vec3 } from "gl-matrix";
+import { vec3 } from "gl-matrix";
 import { Camera } from "../objects/cameras/camera";
+import { sortFrontToBack } from "../math/sort_by_distance";
 
 export type VolumeLayerProps = {
   source: ChunkSource;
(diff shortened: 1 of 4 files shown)


aranega (approved):
Thanks @seankmartin, it looks good to me 🙏

packages/core/src/layers/volume_layer.ts
```diff
@@ -235,30 +234,11 @@ export class VolumeLayer extends Layer implements ChannelsEnabled {
       : 1.0;
 
     this.updateChunks();
+    this.reorderObjects(context.viewport.camera);
```
aganders3: I would just call `sortFrontToBack` here and get rid of `reorderObjects`. I'm not sure why it's public anyway - is it used anywhere else?

packages/core/src/math/sort_by_distance.ts
```diff
@@ -0,0 +1,35 @@
...
+ */
+export function sortFrontToBack(
+  objects: RenderableObject[],
+  camera: Camera
+): RenderableObject[] {
+  const cameraPosition = camera.position;
+  const centerA = vec3.create();
+  const centerB = vec3.create();
+
+  objects.sort((a, b) => {
```
aganders3, line 22: It feels a little odd to me to have a free function sort the objects in-place here, but it might be the best solution for now.

The other option would be to copy and return a new array, but the only mechanism the volume layer has now to update its objects is to clear/add all the objects again in the new order.

aganders3 (approved):
LGTM with a couple optional comments. Thanks for adding the tests here - that looks fine to me.

What are the changes in package-lock.json? Maybe they can be discarded?

seankmartin:
> LGTM with a couple optional comments. Thanks for adding the tests here - that looks fine to me.
> 
> What are the changes in package-lock.json? Maybe they can be discarded?

Thanks! Absolutely they can be discarded, didn't realise I had commit the local changes to the package-lock, thanks for spotting that

packages/core/src/layers/volume_layer.ts
seankmartin: Agreed, not used anywhere else, good idea thank you

seankmartin: The renderer used to do this sorting until it was moved into the layer instead -- and then the method was never changed to private when that refactoring was made

packages/core/src/math/sort_by_distance.ts
seankmartin: I agree that would be cleaner to avoid sorting in place, but for now I think likely best to leave it and then could be made as a follow up PR to change that

aranega (approved):
Looks good to me :) thanks @seankmartin !

PR #591, opened 2026-02-23
perf: early-return to reduce total chunk intersection tests

aganders3:
### Summary
When viewing large multiscale datasets with many chunks (for example stitched OPS data for cellxstate) we see degraded pan/zoom responsiveness. I tracked this down and found this quick optimization to only perform intersection tests for the current and fallback LODs. This can be a significant reduction in total computation on the main thread, improving perceived performance.

I think there are more gains to be had here, probably by using a spatial data structure like an octree to determine visibility.

Another optimization in the layer-per-channel case would be to somehow cache/reuse the visibility results for separate channels.

### Related Issue
N/A

### Tests & Checks
Tested by installing @idetik/core with this change in the DCA viewer and viewing `cellstate/ops0006_20250121/phenotyping_v3.zarr/A/1/0` (`sci-data-staging` S3).

### DIFF
--- a/packages/core/src/core/chunk_store_view.ts
+++ b/packages/core/src/core/chunk_store_view.ts
@@ -347,8 +347,10 @@ export class ChunkStoreView {
       const isFallbackLOD = chunk.lod === fallbackLOD;
       if (!isCurrentLOD && !isFallbackLOD) continue;
 
-      const isInBounds = this.isChunkWithinBounds(chunk, viewBounds3D);
       const isChannelInSlice = this.isChunkChannelInSlice(chunk, sliceCoords);
+      if (!isChannelInSlice) continue;
+
+      const isInBounds = this.isChunkWithinBounds(chunk, viewBounds3D);
 
       const prefetch =
         !isInBounds &&
--- a/packages/core/src/core/chunk_store_view.ts
+++ b/packages/core/src/core/chunk_store_view.ts
@@ -216,13 +216,14 @@ export class ChunkStoreView {
     const fallbackLOD = this.fallbackLOD();
 
     for (const chunk of currentTimeChunks) {
-      if (chunk.lod !== this.currentLOD_) continue;
-      if (!this.isChunkChannelInSlice(chunk, sliceCoords)) continue;
-
+      const isCurrentLOD = chunk.lod === this.currentLOD_;
       const isFallbackLOD = chunk.lod === fallbackLOD;
+      if (!isCurrentLOD && !isFallbackLOD) continue;
+
+      if (!this.isChunkChannelInSlice(chunk, sliceCoords)) continue;
       const priority = this.computePriority(
         isFallbackLOD,
-        true, // isCurrentLOD
+        isCurrentLOD,
         true, // isVisible
         false, // isPrefetch
         true // isChannelInSlice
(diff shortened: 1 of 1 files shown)


packages/core/src/core/chunk_store_view.ts
```diff
@@ -342,11 +343,12 @@ export class ChunkStoreView {
...
     for (const chunk of currentTimeChunks) {
-      const isInBounds = this.isChunkWithinBounds(chunk, viewBounds3D);
-      const isChannelInSlice = this.isChunkChannelInSlice(chunk, sliceCoords);
-
       const isCurrentLOD = chunk.lod === this.currentLOD_;
       const isFallbackLOD = chunk.lod === fallbackLOD;
+      if (!isCurrentLOD && !isFallbackLOD) continue;
+
+      const isInBounds = this.isChunkWithinBounds(chunk, viewBounds3D);
+      const isChannelInSlice = this.isChunkChannelInSlice(chunk, sliceCoords);
```
andy-sweet, line 350: Optional

If this is false, we should probably skip the bounds check above (since that is more costly), but that's a little tricky to do in this PR without more work because of the way `computePriority` handles everything. Probably one for later work/ideas.

andy-sweet: Actually, maybe we can just do the `isChunkChannelInSlice` check first and continue if that's false (as in the volume check) - I think this became simpler after the view/store split.

aganders3: Sorry I'm not following. Which bounds check above do you mean?

packages/core/src/core/chunk_store_view.ts
andy-sweet, line 376: Unrelated to this PR, but I'm now confused why this only sets the chunk view state when it's not null, but the volume case always sets the chunk view state (regardless of whether priority is null).

packages/core/src/core/chunk_store_view.ts
aganders3: Ah I see your second comment now. Yes I think the channel check could come first and also early-return. I was confused by the GitHub UI showing multiple lines for your comment.

packages/core/src/core/chunk_store_view.ts
aganders3: My intention was that `ChunkStoreView.chunkViewStates` be a sparse collection of states for needed (priority != null) chunks, instead of a view state for every chunk. I can take a closer look at the volume case.

packages/core/src/core/chunk_store_view.ts
aganders3: Changed in 7182572 - is this what you meant?

andy-sweet: Yes, that's right. Sorry for the confusion.

packages/core/src/core/chunk_store_view.ts
andy-sweet: Thanks for the clarification. I think the volume path still has some hard assumptions (e.g. all chunks are visible) that mean the priority returned in the loop is always non-null, so the code is roughly equivalent. No need to do anything in this PR.

andy-sweet (approved):
Thanks. Changes look good and all the examples work as expected.

PR #594, opened 2026-02-24
feat: add multi-channel example (grayscale)

seankmartin:
### Summary

Common example to both regular transparency and OIT multichannel volume rendering. To judge that common piece in standalone.
image

### Related Issue
Closes #<issue-number>


aganders3:
Do we need a separate example for this, or should we just modify the existing volume rendering example?

seankmartin:
> Do we need a separate example for this, or should we just modify the existing volume rendering example?

I think it would be fine to just modify the existing example. The multi-viewport example demonstrates using the old data anyway. I'll edit this PR to update the existing example instead, can always roll that back if needed

seankmartin:
Thanks for the note on the example, I think it's cleaner now without a separate one. One other piece we're unsure on is this:

```
volumeLayer.opacityMultiplier = 0.001;
```
If we don't set that the image is basically full white since this PR doesn't fully set up multi-channel rendering yet to respect all the parameters set on the channels. That comes in either the multichannel rendering with/without OIT PRs.

However, even with multichannel rendering setup properly, it can be useful to set this `opacityMultiplier` in the examples. Unsure if this API is ok

packages/core/examples/volume_rendering/main.ts
```diff
@@ -1,44 +1,54 @@
...
-const policy = createPlaybackPolicy({
+const policy = createExplorationPolicy({
   lod: { min: controls.lod, max: controls.lod },
 });
 const volumeLayer = new VolumeLayer({
   source,
   sliceCoords,
   policy,
   channelProps: [
+    {
```
shlomnissan, line 26: To confirm: this is just plumbing right? The channels are not wired up to the volume rendering pipeline?

shlomnissan (approved):
Approved

packages/core/examples/volume_rendering/main.ts
seankmartin: Yes exactly this is just plumbing

seankmartin:
> Approved

thanks for the review!

PR #595, opened 2026-02-24
fix: protect against out bounds time index setting breaking volume image rendering

seankmartin:
### Summary

There was a bug that could be triggered from the volume rendering example or multi-viewport example by setting `t` to the max on the time slider. We fixed that in https://github.com/chanzuckerberg/idetik/pull/597 along with @andy-sweet. 

This PR is not pressing due to the above fix. We originally opened this PR instead of #597. Leaving this open to discuss the possibility to protect against that and other time co-ords being out of bounds.

### Tests & Checks

Manually tested changes.

### DIFF
--- a/packages/core/src/core/chunk_store.ts
+++ b/packages/core/src/core/chunk_store.ts
@@ -89,7 +89,7 @@ export class ChunkStore {
   }
 
   public getChunksAtTime(timeIndex: number): Chunk[] {
-    return this.chunks_[timeIndex];
+    return this.chunks_?.[timeIndex];
   }
 
   public getTimeIndex(sliceCoords: SliceCoordinates): number {
--- a/packages/core/src/core/chunk_store_view.ts
+++ b/packages/core/src/core/chunk_store_view.ts
@@ -74,6 +74,9 @@ export class ChunkStoreView {
   public getChunksToRender(sliceCoords: SliceCoordinates): Chunk[] {
     const currentTimeIndex = this.store_.getTimeIndex(sliceCoords);
     const currentTimeChunks = this.store_.getChunksAtTime(currentTimeIndex);
+    if (currentTimeChunks === undefined) {
+      return [];
+    }
     const currentLODChunks = currentTimeChunks.filter(
       (chunk) =>
         chunk.lod === this.currentLOD_ &&
@@ -129,6 +132,10 @@ export class ChunkStoreView {
 
     const currentTimeIndex = this.store_.getTimeIndex(sliceCoords);
     const currentTimeChunks = this.store_.getChunksAtTime(currentTimeIndex);
+
+    if (currentTimeChunks === undefined) {
+      return;
+    }
 
     if (currentTimeChunks.length === 0) {
       Logger.warn(
@@ -197,6 +204,10 @@ export class ChunkStoreView {
     const currentTimeIndex = this.store_.getTimeIndex(sliceCoords);
     const currentTimeChunks = this.store_.getChunksAtTime(currentTimeIndex);
 
+    if (currentTimeChunks === undefined) {
+      return;
+    }
+
     if (currentTimeChunks.length === 0) {
       Logger.warn(
         "ChunkStoreView",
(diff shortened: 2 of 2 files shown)


packages/core/src/core/chunk_store.ts
```diff
@@ -89,7 +89,7 @@ export class ChunkStore {
   }
 
   public getChunksAtTime(timeIndex: number): Chunk[] {
-    return this.chunks_[timeIndex];
+    return this.chunks_?.[timeIndex];
```
andy-sweet, line 92: I don't think this `?.` adds anything here since `this.chunks_` is readonly and typed as `Chunk[][]` - i.e. `this.chunks_` cannot be undefined.

In js, any use of `[]` with an array can return `undefined`, so maybe the return type should be `Chunk[] | undefined` to be explicit.

andy-sweet (commented):
While this change on its own is fairly benign (and probably mergeable), I don't love the implications this has for the rest of the repo because we may need to add lots of similar defensive checks elsewhere. I'm being a coward by not approving/rejecting, but could do with another reviewer's opinion here.

As I understand js, any use of `[]` on any array can return `undefined`. If we want to check for these in general on lint/compile, I think we should use `"noUncheckedIndexedAccess": true` in `tsconfig.json`. However, that triggers quite a few errors, so may be outside the scope of the intended change here.

For a quick fix for this particular issue that doesn't require discussion, I'd suggest changing the max t-value in the examples to be correct with respect to the source (i.e. 790). In fact, that should be done regardless.

Ideally, I prefer using type definitions like `type NonEmptyArray<T> = [T, ...T[]];` to inform the compiler that certain arrays are non-empty, but that can be awkward to enforce in the implementation.

seankmartin:
> While this change on its own is fairly benign (and probably mergeable), I don't love the implications this has for the rest of the repo because we may need to add lots of similar defensive checks elsewhere. I'm being a coward by not approving/rejecting, but could do with another reviewer's opinion here.
> 
> As I understand js, any use of `[]` on any array can return `undefined`. If we want to check for these in general on lint/compile, I think we should use `"noUncheckedIndexedAccess": true` in `tsconfig.json`. However, that triggers quite a few errors, so may be outside the scope of the intended change here.
> 
> For a quick fix for this particular issue that doesn't require discussion, I'd suggest changing the max t-value in the examples to be correct with respect to the source (i.e. 790). In fact, that should be done regardless.
> 
> Ideally, I prefer using type definitions like `type NonEmptyArray<T> = [T, ...T[]];` to inform the compiler that certain arrays are non-empty, but that can be awkward to enforce in the implementation.

Thanks @andy-sweet! I'll make the simpler change separately to just set 790 as max t val on the volume rendering and multiple viewports example. I'll leave this open for the sake of discussion as well, and thank you for the thoughtful input cc @aranega

packages/core/src/core/chunk_store.ts
aranega: @andy-sweet Yes indeed, it's not necessary, it's a lefover from my side from some quick experiments I did for another subject.

aranega:
> While this change on its own is fairly benign (and probably mergeable), I don't love the implications this has for the rest of the repo because we may need to add lots of similar defensive checks elsewhere. I'm being a coward by not approving/rejecting, but could do with another reviewer's opinion here.

@andy-sweet That was the friction that we felt also with @seankmartin. Perhaps the best would have been to be able to just return an empty list if `undefined` was yield by the array access, but that makes enter the frontend in a kind of fetch loop.

aganders3:
I think I may prefer to throw an error if any value in `SliceCoordinates` does not make sense for a given layer/source.

Otherwise I'm fine with leaving it as-is for now and relying on applications to do the bookkeeping of slice limits for sensible behavior. I think that's a bit easier now that the source dimension map is available from the loader. It's still slightly clunky to need to await a loader from the source first, but I think it's good enough for the time being.

seankmartin:
I will close this for the moment to reduce noise but appreciate the input!

PR #598, opened 2026-02-26
feat: support dynamically adding and removing viewports at runtime

aganders3:
### Summary
Adds support for dynamically adding and removing viewports at runtime. Previously viewports could only be configured at initialization - now you can call `addViewport()` and `removeViewport()` on an Idetik instance while it's running.

This flexibility is helpful for using Idetik with React, where we may need to wait for components to mount before using them as viewport elements. The constructor no longer requires at least one viewport, and the `PixelSizeObserver` now supports observing/unobserving individual elements to handle dynamic viewports properly. Also exports `OrbitControls` and `validateNewViewport` from the core package.

Notably not included here is any cleanup. `removeViewport` takes a reference to the `Viewport` object, but it's currently up to a user to do any layer removal, disposal, etc. if the viewport will be actually discarded, for example.

### Tests & Checks
Manually tested adding/removing viewports in https://github.com/chanzuckerberg/idetik-react/pull/57. Existing validation still works to prevent duplicate IDs or shared elements.

Existing tests and examples all pass with no other changes (backward-compatible).

### DIFF
--- a/packages/core/src/idetik.ts
+++ b/packages/core/src/idetik.ts
@@ -71,6 +71,7 @@ export class Idetik {
    *   viewports: [
    *     {
    *       id: 'main',
+   *       // element omitted - defaults to canvas
    *       camera: camera1,
    *       layers: [layer1]
    *     },
@@ -164,15 +165,16 @@ export class Idetik {
     return viewport;
   }
 
-  public removeViewport(id: string): boolean {
-    const index = this.viewports_.findIndex((v) => v.id === id);
+  public removeViewport(viewport: Viewport): boolean {
+    const index = this.viewports_.indexOf(viewport);
 
     if (index === -1) {
-      Logger.warn("Idetik", `Viewport "${id}" not found, nothing to remove`);
+      Logger.warn(
+        "Idetik",
+        `Viewport "${viewport.id}" not found, nothing to remove`
+      );
       return false;
     }
-
-    const viewport = this.viewports_[index];
 
     if (this.lastAnimationId_ !== undefined) {
       viewport.events.disconnect();
@@ -182,7 +184,7 @@ export class Idetik {
     }
 
     this.viewports_.splice(index, 1);
-    Logger.info("Idetik", `Removed viewport "${id}"`);
+    Logger.info("Idetik", `Removed viewport "${viewport.id}"`);
     return true;
   }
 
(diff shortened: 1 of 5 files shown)


aganders3:
No changes, but now that I have https://github.com/chanzuckerberg/idetik-react/pull/57 open (though draft) I think this is ready for review.

packages/core/src/idetik.ts
```diff
@@ -144,13 +140,61 @@ export class Idetik {
     return this.viewports_;
   }
 
+  public get running(): boolean {
```
andy-sweet, line 143: Optional: this is not used outside of this class, so doesn't need to be public in this change. Maybe it does for some of the related changes, but I don't see it in the linked react PR. I'll let you decide if this should be public - it seems like a reasonable state to expose.

packages/core/src/idetik.ts
```diff
@@ -193,14 +237,15 @@ export class Idetik {
...
-    if (this.lastAnimationId_ === undefined) {
+    if (!this.running) {
       Logger.warn("Idetik", "Idetik runtime not started");
     } else {
       this.sizeObserver_.disconnect();
       for (const viewport of this.viewports_) {
         viewport.events.disconnect();
       }
-      cancelAnimationFrame(this.lastAnimationId_);
+      // safe non-null assertion: this.running is true, so lastAnimationId_ is defined
```
andy-sweet, line 247: Optional: maybe renaming `lastAnimationId` to `runningAnimationId` would make the name more self-descriptive and avoid the need for the use of `running` here and the ! and the explanatory comment.

andy-sweet (approved):
Looks good! All comments are optional.

One other optional thought is to use this functionality in the multiple viewport example by allowing users to switch between 2D/3D/both in some way. I understand that it's not necessary to dynamically add/remove viewports to implement that in the example but it might help to demonstrate this functionality in this repo.

aganders3:
> I understand that it's not necessary to dynamically add/remove viewports to implement that in the example but it might help to demonstrate this functionality in this repo.

I considered adding a dynamic-layout example but wanted to keep this small. I'd probably make a "montage viewer" where you could set the rows/columns of a grid and it would dynamically generate the viewports and show different slices spanning the volume (or timeframes?).

packages/core/src/idetik.ts
aganders3: Good question and I didn't really consider it much.
I think we will take a closer look at public/private and the API surface shortly after open-sourcing, so I'm going to leave it for now.

packages/core/src/idetik.ts
aganders3: Thanks - the rename is a good suggestion but I'm not sure it takes the place of the whole property semantically.

Issue #603, opened 2026-03-10
Allow a method to avoid downloading chunks for non visible channels in multichannel volume rendering


seankmartin:
Currently volume rendering supports multi-channels, and supports using `channelProps` to show/hide individual channels. However, the non-visible channels have chunks fetched for them at the moment. This means that if you have the first few channels in the data hidden, and are rendering the later channels, you need to wait until the fetch for all the hidden first channels completes before you see anything. Example at LOD1 and LOD0 shown here (cache disabled to simulate fresh load where this would be the worst).

This raises some questions though, such as:
1. What does being visible or non-visible control? Should a non-visible chunk still be kept in memory in some cases for performance? Or is visibility and being kept in memory tied together?
2. Is there a difference between a channel that was never made visible and a channel that was once visible but is now non-visible?
3. What should be the relationship between components to manage this and are the existing objects sufficient or do we need a new concept? For example, if we allow slice co-ordinates to influence chunk channel fetches - could the volume layer directly use the channel props to set the slice co-ordinates based on channel visibility or would this create undesired relationships.

Right now we're planning to make a PoC to also help drive this forward and further discussions with @aranega.

https://github.com/user-attachments/assets/fc33e6bf-ef0c-4051-8bcb-2c68629dd957

See also https://github.com/chanzuckerberg/idetik/pull/564#pullrequestreview-3891835046 for context from @aganders3.

aganders3:
Closed by #636

PR #606, opened 2026-03-10
feat: multichannel avoid download of non visible channels

seankmartin:
### Summary

Slice coordinates can include channels as arrays, with individual channels to load - instead of being either a single channel or undefined. This allows avoiding downloading non-visible channels.

### Related Issue
Closes #603

### Tests & Checks

Manual testing


packages/core/src/objects/textures/texture_2d_array.ts
```diff
@@ -62,10 +62,10 @@ export class Texture2DArray extends Texture {
     const height = chunk.shape.y;
     const depth = source.length / (width * height);
     if (
-      this.width != width ||
-      this.height != height ||
-      this.depth_ != depth ||
-      this.dataType != bufferToDataType(source)
+      this.width !== width ||
```
seankmartin, line 65: Did the linter complain about the use of `!=` comparisons and not doing imports using `type` when all that is used is the type?

I think these are good to change! But was wondering if they are just a nice to fix or necessary for the PR. Might be helpful for later reviewers if they know the same since I think the changes here are quite a bit smaller than they look in the diff since a good few changes are to have type imports and more solid comparisons

seankmartin:
I don't seem to be able to change the base branch so will close and reopen against main

PR #607, opened 2026-03-11
refactor: remove uniforms from volume layer

### LINKED ISSUE #555: Remove uniforms as being part of the volume layer

seankmartin:
### Summary
Removes uniforms being at the layer level which was introduced for volume rendering. Instead moves those uniforms to the volume renderable, and reverts the changes made to the webgl renderer to support layer uniforms.

### Related Issue
Closes #555

### Tests & Checks

Manually tested examples

### DIFF
--- a/packages/core/src/core/layer.ts
+++ b/packages/core/src/core/layer.ts
@@ -125,12 +125,4 @@ export abstract class Layer {
   protected clearObjects() {
     this.objects_ = [];
   }
-
-  /**
-   * Get uniforms for shader program. Override in derived classes that need custom uniforms.
-   * @returns Object containing uniform name-value pairs
-   */
-  public getUniforms(): Record<string, unknown> {
-    return {}; // Default implementation returns no uniforms
-  }
 }
--- a/packages/core/src/layers/volume_layer.ts
+++ b/packages/core/src/layers/volume_layer.ts
@@ -211,12 +211,14 @@ export class VolumeLayer extends Layer implements ChannelsEnabled {
     ]);
   }
 
-  private updateVolumeUniforms(volume: VolumeRenderable) {
-    volume.debugShowDegenerateRays = this.debugShowDegenerateRays;
-    volume.relativeStepSize =
-      this.relativeStepSize * this.interactiveStepSizeScale_;
-    volume.opacityMultiplier = this.opacityMultiplier;
-    volume.earlyTerminationAlpha = this.earlyTerminationAlpha;
+  private updateVolumeUniforms() {
+    for (const volume of this.currentVolumes()) {
+      volume.debugShowDegenerateRays = this.debugShowDegenerateRays;
+      volume.relativeStepSize =
+        this.relativeStepSize * this.interactiveStepSizeScale_;
+      volume.opacityMultiplier = this.opacityMultiplier;
+      volume.earlyTerminationAlpha = this.earlyTerminationAlpha;
+    }
   }
 
   private releaseAndRemoveChunks(chunks: Iterable<Chunk>) {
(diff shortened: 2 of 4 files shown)


shlomnissan (changes requested):
Thanks for the PR! I like the intent of cleaning up the uniform plumbing but I'm not sure this is a net improvement.

The main concern is state duplication. Before, the layer was the single source of truth for volume-wide properties and the renderer pulled them via `getUniforms()`. Now those properties live on both the layer and each renderable kept in sync by `updateVolumeUniforms()` copying them every frame. That's more state to manage not less.

I also think `updateVolumeUniforms` is a bit of a misnomer. It's syncing layer properties to renderables not updating shader uniforms.

I don't think there's anything wrong with layers owning state that maps to uniforms. The thing I'd change is the `getUniforms(): Record<string, unknown>` interface. It's loose and generic. I'd prefer the renderer just do an explicit check (if layer is a volume layer) and set each uniform manually. That keeps the layer as the single source of truth avoids the untyped bag of values API and doesn't introduce any state duplication.

I know that's what we do with renderables, but the generic interface would make sense to me if we had many layer types all providing uniforms. With just one I think it's premature abstraction that trades type safety for flexibility we don't need.

seankmartin:
> Thanks for the PR! I like the intent of cleaning up the uniform plumbing but I'm not sure this is a net improvement.
> 
> The main concern is state duplication. Before, the layer was the single source of truth for volume-wide properties and the renderer pulled them via `getUniforms()`. Now those properties live on both the layer and each renderable kept in sync by `updateVolumeUniforms()` copying them every frame. That's more state to manage not less.
> 
> I also think `updateVolumeUniforms` is a bit of a misnomer. It's syncing layer properties to renderables not updating shader uniforms.
> 
> I don't think there's anything wrong with layers owning state that maps to uniforms. The thing I'd change is the `getUniforms(): Record<string, unknown>` interface. It's loose and generic. I'd prefer the renderer just do an explicit check (if layer is a volume layer) and set each uniform manually. That keeps the layer as the single source of truth avoids the untyped bag of values API and doesn't introduce any state duplication.
> 
> I know that's what we do with renderables, but the generic interface would make sense to me if we had many layer types all providing uniforms. With just one I think it's premature abstraction that trades type safety for flexibility we don't need.



> Thanks for the PR! I like the intent of cleaning up the uniform plumbing but I'm not sure this is a net improvement.
> 
> The main concern is state duplication. Before, the layer was the single source of truth for volume-wide properties and the renderer pulled them via `getUniforms()`. Now those properties live on both the layer and each renderable kept in sync by `updateVolumeUniforms()` copying them every frame. That's more state to manage not less.
> 
> I also think `updateVolumeUniforms` is a bit of a misnomer. It's syncing layer properties to renderables not updating shader uniforms.
> 
> I don't think there's anything wrong with layers owning state that maps to uniforms. The thing I'd change is the `getUniforms(): Record<string, unknown>` interface. It's loose and generic. I'd prefer the renderer just do an explicit check (if layer is a volume layer) and set each uniform manually. That keeps the layer as the single source of truth avoids the untyped bag of values API and doesn't introduce any state duplication.
> 
> I know that's what we do with renderables, but the generic interface would make sense to me if we had many layer types all providing uniforms. With just one I think it's premature abstraction that trades type safety for flexibility we don't need.

thank you for the thoughtful review and discussion in the meeting! Closing based on that to allow layers to have uniforms

PR #615, opened 2026-03-23
feat: relax constraints on T and C chunk sizes

### LINKED ISSUE #600: Support OME-Zarr sources with multiple channels in chunks
Currently, we assume and [validate that the chunk size of an OME-Zarr source is 1](https://github.com/chanzuckerberg/idetik/blob/df0e4f86fe6416fcc6111d5269ca442437fd3014/packages/core/src/core/chunk_store.ts#L271). 

This means that we don't have to slice any chunk data when preparing texture data.

The goal of this issue is to support OME-Zarr source's with multiple channels in chunks. For an example, see the following example in one of our internal [s3 buckets](s3://czi-dynamic-cell-atlas-staging/allencell-static-hipsc/1000860/7be5fdb9aec8aa956a839ffae3e2d9fc28fe715117a8239bb6417fc13d87c5c0_raw.ome.zarr/0/zarr.json) where each chunk contains 3 channels.

The motivation of this issue is to support all DCA OME-Zarr images, which may absorb channels into a single chunk to create bigger chunks. See the [DCA imaging standard](https://friendly-adventure-7j9rgl2.pages.github.io/v0.1/array-standard.html#chunk-size) to see the reasoning (i.e. channel chunks length should be 1, not must be 1).

aganders3:
### Summary
This changes the OME Zarr image loader to support loading channels and timepoints from datasets with chunking in T and/or C.

This is the simplest fix I could think of to unblock visualization of affected datasets, but it has known inefficiency. This will result in duplicate fetches and redundant chunk decompression. The detrimental effect on UX is mitigated by the browser cache and WebWorker decompression off the main thread.

This can be considered a temporary workaround, and a longer-term solution should be investigated. Some options may include:
* Simple in-flight request deduplication
* Opportunistic chunk filling (combined with less aggressive chunk disposal)
* Full source chunk cache

### Related Issue
Closes #600

### Tests & Checks
I tested locally by installing this in `idetik-dca-viewer` and accessing `NIST/exp0.zarr`, which has chunk size > 1 in T.

Screenshot 2026-03-23 at 11 18 13 AM

### DIFF
--- a/packages/core/src/data/ome_zarr/image_loader.ts
+++ b/packages/core/src/data/ome_zarr/image_loader.ts
@@ -104,7 +104,7 @@ export class OmeZarrImageLoader {
 
     validateTightlyPackedChunk(receivedChunk);
 
-    chunk.data = this.sliceSourceChunk(
+    chunk.data = this.sliceReceivedChunk(
       chunk,
       receivedChunk.data,
       receivedChunk.stride
@@ -121,10 +121,10 @@ export class OmeZarrImageLoader {
 
   // trim any padding (XYZ padding for edge chunks)
   // and extract the channel/timepoint
-  private sliceSourceChunk(
+  private sliceReceivedChunk(
     chunk: Chunk,
-    sourceData: ChunkData,
-    sourceStride: number[]
+    receivedChunkData: ChunkData,
+    receivedChunkStride: number[]
   ): ChunkData {
     const cLod = this.dimensions_.c?.lods[chunk.lod];
     const tLod = this.dimensions_.t?.lods[chunk.lod];
@@ -134,20 +134,19 @@ export class OmeZarrImageLoader {
 
     // internal chunks are compact 3D XYZ, with size 1 in C and T
     const compactSize = chunk.shape.x * chunk.shape.y * chunk.shape.z;
-    const compactData = new (sourceData.constructor as ChunkDataConstructor)(
-      compactSize
-    );
+    const compactData =
+      new (receivedChunkData.constructor as ChunkDataConstructor)(compactSize);
 
     const cStride = this.dimensions_.c
-      ? sourceStride[this.dimensions_.c.index]
+      ? receivedChunkStride[this.dimensions_.c.index]
       : 0;
     const tStride = this.dimensions_.t
-      ? sourceStride[this.dimensions_.t.index]
+      ? receivedChunkStride[this.dimensions_.t.index]
       : 0;
     const zStride = this.dimensions_.z
-      ? sourceStride[this.dimensions_.z.index]
+      ? receivedChunkStride[this.dimensions_.z.index]
       : 0;
-    const yStride = sourceStride[this.dimensions_.y.index];
+    const yStride = receivedChunkStride[this.dimensions_.y.index];
 
     // note: this assumes tczyx ordering
     const baseOffset = tOffsetInSource * tStride + cOffsetInSource * cStride;
@@ -157,7 +156,10 @@ export class OmeZarrImageLoader {
       for (let y = 0; y < chunk.shape.y; y++) {
         const srcStart = zStart + y * yStride;
         const srcEnd = srcStart + chunk.shape.x;
-        compactData.set(sourceData.subarray(srcStart, srcEnd), destOffset);
+        compactData.set(
+          receivedChunkData.subarray(srcStart, srcEnd),
+          destOffset
+        );
         destOffset += chunk.shape.x;
       }
     }
--- a/packages/core/src/data/ome_zarr/image_loader.ts
+++ b/packages/core/src/data/ome_zarr/image_loader.ts
@@ -71,15 +71,27 @@ export class OmeZarrImageLoader {
     if (this.dimensions_.z) {
       chunkCoords[this.dimensions_.z.index] = chunk.chunkIndex.z;
     }
+
+    // internal (ChunkStore) chunks have size 1 in C and T
+    // so divide by the actual chunkSize to get the chunkCoord here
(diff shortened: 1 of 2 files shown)


packages/core/src/data/ome_zarr/image_loader.ts
```diff
@@ -136,29 +139,51 @@ export class OmeZarrImageLoader {
...
+    const tStride = this.dimensions_.t
+      ? receivedChunkStride[this.dimensions_.t.index]
+      : 0;
     const zStride = this.dimensions_.z
       ? receivedChunkStride[this.dimensions_.z.index]
       : 0;
     const yStride = receivedChunkStride[this.dimensions_.y.index];
+
+    // note: this assumes tczyx ordering
+    const baseOffset = tOffsetInSource * tStride + cOffsetInSource * cStride;
```
andy-sweet: Nit: consider `srcOffset` for consistency with `srcStart`.

packages/core/src/data/ome_zarr/image_loader.ts
```diff
@@ -112,20 +124,11 @@ export class OmeZarrImageLoader {
...
-      receivedShape.z > chunk.shape.z;
-
-    if (receivedChunkHasPadding) {
-      chunk.data = this.trimChunkPadding(
-        chunk,
-        receivedChunk.data,
-        receivedChunk.stride
-      );
-    } else {
-      chunk.data = receivedChunk.data;
```
andy-sweet, line 127: Optional

I wonder if we can easily reintroduce a similar optimization skip when we know that `sliceReceivedChunk` will just copy all the chunk data without striding. I think this case will need some of the checks above, plus checks against `cOffsetInSource` and `tOffsetInSource`. The check should likely be in `sliceReceivedChunk` and if true should early return the chunk data itself.

Marking as optional because I'm not quite sure how complex the code will be and if it's worth the effort. However, I'd lean towards doing this if possible because doing a strided memory copy when unnecessary feels wrong in a supposedly performance conscious library.

andy-sweet (approved):
This looks good to me and works for me too.

Approving because this closes #600, enabling new usage.

It's probably worth creating a follow-up issue to address the performance overhead of fetching/decompressing certain chunks multiple times.

I think it's also be worth addressing the new performance overhead this introduces in cases where the slicing/trimming is not needed, but I'll leave that decision to you.

packages/core/src/data/ome_zarr/image_loader.ts
aganders3: By the time we're there, I think we can just check the size/offset. What do you think?

See [b538bd7](https://github.com/chanzuckerberg/idetik/pull/615/commits/b538bd7439966683bc98c51a5f6c65db589ab4c6)

aganders3: I guess this runs the risk of having the dimensions/shape be wrong, but still come out to the same product...

aganders3: Let me know what you think of [3bc0e48](https://github.com/chanzuckerberg/idetik/pull/615/commits/3bc0e489fe74b31be4c9c9da4c65c32eaf837b2f)

aganders3:
Thanks for the review!

> It's probably worth creating a follow-up issue to address the performance overhead of fetching/decompressing certain chunks multiple times.

I agree - I'll open an issue if/when this is merged.

packages/core/src/data/ome_zarr/image_loader.ts
```diff
@@ -136,29 +135,76 @@ export class OmeZarrImageLoader {
...
+  // trim any padding (XYZ padding for edge chunks)
+  // and extract the channel/timepoint
+  private sliceReceivedChunk(
     chunk: Chunk,
-    receivedChunkData: ChunkData,
-    receivedChunkStride: number[]
+    receivedChunk: zarr.Chunk<zarr.DataType>
   ): ChunkData {
+    const receivedChunkData = receivedChunk.data as ChunkData;
+    const receivedChunkStride = receivedChunk.stride;
```
andy-sweet: Nit: passing these in avoids the need for the static cast since `receivedChunk.data` has already passed a dynamic cast, so I think it's better to stick with original function signature and call even though the new one is shorter.

packages/core/src/data/ome_zarr/image_loader.ts
```diff
@@ -136,29 +135,76 @@ export class OmeZarrImageLoader {
...
+
+    const receivedShape = {
```
andy-sweet: I know premature optimization is always a concern, but I think it's silly to make an object that was just made before the call to this function.

Could we pack more of the checks/casts that occur just before this function is called into the function itself?

Something like the following

https://github.com/chanzuckerberg/idetik/compare/aganders3/flexible-c-and-t-chunking...andy-sweet/move-checks?expand=1

andy-sweet (approved):
Still looks good to me, though there's a minor optimization I suggested with a commit that you could merge here.

packages/core/src/data/ome_zarr/image_loader.ts
andy-sweet: Or you could pass it in.

packages/core/src/data/ome_zarr/image_loader.ts
andy-sweet: Or follow the suggestion in my other comment to move some of the checks into this method.

packages/core/src/data/ome_zarr/image_loader.ts
aganders3: Thanks - this is more readable, I'll merge it in.

PR #617, opened 2026-03-24
feat: multichannel avoid downloading chunks for channels that have never been visible

### LINKED ISSUE #603: Allow a method to avoid downloading chunks for non visible channels in multichannel volume rendering

seankmartin:
### Summary

This introduces two main changes:
1. Slice coordinates can include channels as arrays, with individual channels to load - instead of being either a single channel or undefined. This allows avoiding downloading chunks for channels that are not in the slice co-ordinates.
2. A way for chunked image layers and volume layer to use their channel props to set the slice co-ordinates channel index to be an array of all channels that have ever been visible. As such, for layers, the main aim is to avoid downloading data for channels that are of no interest because they were never looked at. This is mostly to avoid issues around needing to dispose data / mixing between different LODs / quickly toggling on/off visibility etc.

https://github.com/user-attachments/assets/5225d968-196f-422c-83de-0644b97ab1c6

### Related Issue
Closes #603

### Tests & Checks

Manual testing


aganders3:
I sent this as a DM to Sean — posting here for posterity.

I'm a little ambivalent about the API, but can I see how you got here. I think partly there is some awkwardness because of the different 2D and 3D approaches of channel-per-layer vs. channels-in-layer. This makes the responsibilities of the channel props and slice coords a bit muddled, in my opinion.

In my mind, the slice coords should determine which channels are fetched. Then the channel props would determine which are visible. Channel props would map (in order) to the channels indicated in the slice coords, so there's no need to add effectively "dummy" or hidden values for channels not fetched. I think this creates a timing issue with the volume layer, where you need to update both slicecoords and channel props simultaneously...

I need to think about this a bit more. Two quick ideas I haven't thought through:
* channel properties could be a map instead of a list
* channels could actually be removed from `SliceCoordinates`

seankmartin:
> I sent this as a DM to Sean — posting here for posterity.
> 
> I'm a little ambivalent about the API, but can I see how you got here. I think partly there is some awkwardness because of the different 2D and 3D approaches of channel-per-layer vs. channels-in-layer. This makes the responsibilities of the channel props and slice coords a bit muddled, in my opinion.
> 
> In my mind, the slice coords should determine which channels are fetched. Then the channel props would determine which are visible. Channel props would map (in order) to the channels indicated in the slice coords, so there's no need to add effectively "dummy" or hidden values for channels not fetched. I think this creates a timing issue with the volume layer, where you need to update both slicecoords and channel props simultaneously...
> 
> I need to think about this a bit more. Two quick ideas I haven't thought through:
> 
> * channel properties could be a map instead of a list
> * channels could actually be removed from `SliceCoordinates`

Also bringing over some ideas from DMs, thanks Ashley for bringing this up and the discussion.

First I think is how this current API in this PR was landed at. We had originally thought of doing the following:
1. Slice co-ordinates could contain channel indices to avoid loading all the channels.
2. For any channel that is in the slice co-ordinates, a layer (say the volume layer, since I think the main example where one layer uses multiple channels at a time) could set visibility of that channel, and visibility would only affect drawing.

In this way, the visibility of a channel and its load status were more independent. However, we were concerned this could be a bit awkward for a user to work with. They would have to know ahead of time which channels they would ever need, and couldn't use visibility to later decide whether to load a channel or not. Overall I agree that despite the aim here, I also don't feel super sold on the API it results in.

The main running idea to change this at the moment is that `channelProperties` become a map. Channel props as a map enables either:
1. Removing channels from the `sliceCoords` and using the presence of a channel in the channel props map determine loading state (visibility would then just control rendered/not rendered, and be more independent from loading).
2. Keep channels in the `sliceCoords` but consider channels that are not in the channel props to be non visible. So sliceCoords still continue to be the determination of what is loaded.

The primary drawback with channels being an array is that you need to have dummy information for channels you don't care about and also that it makes it a bit awkward to update them (such as how Ashley points out that both the sliceCoords and channelProperties need to update at the same time right now).

I personally prefer the idea where channels are not part of the sliceCoords. Channels feel cleaner being a separate concept from a co-ordinate for a slice from a volume. But that might just be from a bias to tend to think of a channel as a special dimension that is separated from spatial dimensions or a time dimension. So I definitely get it if it is preferred to keep them in the sliceCoords as maybe then it starts to feel like the `channelProperties` are doing too much.

If I'm not mistaken, channels being excluded from the sliceCoords also allows a fairly clean way to later mark a channel that it needs to be both loaded and visible. That channel can get added into the channel props map.

For @aganders3 please let me know if you feel I've inaccurately brought across anything from our conversation here, and thank you again

shlomnissan:
Replaced with https://github.com/chanzuckerberg/idetik/pull/636

PR #622, opened 2026-03-27
feat: Downsampling during camera movement 

aranega:
### Summary              

This PR adds XY downsampling for transparent layers (e.g. volumes) during camera movement (when the camera is actively rotating/panning). Transparent layers render into a smaller framebuffer (e.g. 4x fewer pixels), which is then composited back to the main canvas. When the camera stops, rendering returns to full resolution.     

https://github.com/user-attachments/assets/f07f1a65-15f8-4b95-888a-a0711a53d3f2

                                                                            
### Implementation                                                                                                                                  
 
`DownsampledFramebuffer` manages a frame buffer with a single RGBA color attachment (lazy resize with texture recreation), while `DownsamplingCompositePass` renders a fullscreen triangle that samples the downsampled texture. In the renderer `beginDownsampling` and `endDownsampling` wrap the `renderTransparentLayers` call. The framebuffer is owned by the renderer, created lazily on first use, and resized as needed. `downsamplingFactor` is a per-viewport config option (defaults to 1 = no downsampling). The `cameraControls.isMoving` signal controls when downsampling activates.                                                                                                                                        
                                                           
  ### Tests & Checks                                                                                                                                          
  
Manually tested with the volume_rendering and multi_viewport examples (both configured with downsamplingFactor: 4)

### DIFF
--- a/packages/core/src/renderers/webgl_renderer.ts
+++ b/packages/core/src/renderers/webgl_renderer.ts
@@ -140,6 +140,10 @@ export class WebGLRenderer extends Renderer {
       }
     }
 
+    for (const layer of transparent) {
+      layer.update(renderContext);
+    }
+
     if (viewportIsVisible) {
       const isDownsampling = this.beginDownsampling(
         viewport,
@@ -149,7 +153,6 @@ export class WebGLRenderer extends Renderer {
 
       this.state_.setDepthMask(false);
       for (const layer of transparent) {
-        layer.update(renderContext);
         if (layer.state === "ready") {
           this.renderLayer(layer, viewport.camera, frustum);
         }
--- a/packages/core/src/renderers/webgl_renderer.ts
+++ b/packages/core/src/renderers/webgl_renderer.ts
@@ -60,6 +60,7 @@ class DownsamplingCompositePass {
   }
 
   private end() {
+    this.gl_.bindVertexArray(null);
     this.state_.setDepthMask(true);
   }
 
@@ -226,6 +227,7 @@ export class WebGLRenderer extends Renderer {
     this.state_.setViewport(viewportBox);
     this.applyScissor(scissorBox);
     this.downsamplingPass_.draw(this.downsamplingFramebuffer_!, this.programs_);
+    this.bindings_.invalidateActiveGeometry();
   }
 
   // Returns the scissor box if clipping is needed,
--- a/packages/core/src/renderers/webgl_renderer.ts
+++ b/packages/core/src/renderers/webgl_renderer.ts
@@ -1,21 +1,19 @@
-import { Renderer } from "../core/renderer";
-import { WebGLShaderProgram } from "./webgl_shader_program";
-import { WebGLShaderPrograms } from "./webgl_shader_programs";
-import { Logger } from "../utilities/logger";
-
-import { WebGLBuffers } from "./webgl_buffers";
-import { WebGLTextures } from "./webgl_textures";
-
-import { Layer } from "../core/layer";
-import { WebGLState } from "./WebGLState";
-import { RenderableObject } from "../core/renderable_object";
-import { Geometry, Primitive } from "../core/geometry";
-import { Box2 } from "../math/box2";
-import { Viewport } from "../core/viewport";
-import { Camera } from "../objects/cameras/camera";
-
 import { mat4, vec2, vec3, vec4 } from "gl-matrix";
-import { Frustum } from "../math/frustum";
+import type { Geometry, Primitive } from "../core/geometry";
+import type { Layer } from "../core/layer";
+import type { RenderableObject } from "../core/renderable_object";
+import { Renderer } from "../core/renderer";
+import type { Viewport } from "../core/viewport";
+import { Box2 } from "../math/box2";
+import type { Frustum } from "../math/frustum";
+import type { Camera } from "../objects/cameras/camera";
+import { Logger } from "../utilities/logger";
+import { WebGLState } from "./WebGLState";
+import { WebGLBuffers } from "./webgl_buffers";
+import { DownsampledFramebuffer } from "./webgl_framebuffers";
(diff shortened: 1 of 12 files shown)


packages/core/src/renderers/webgl_renderer.ts
```diff
@@ -116,6 +166,96 @@ export class WebGLRenderer extends Renderer {
...
+    this.state_.setViewport(viewportBox);
+    this.applyScissor(scissorBox);
+    this.downsamplingPass_.draw(this.downsamplingFramebuffer_!, this.programs_);
+    this.bindings_.invalidateActiveGeometry();
+  }
+
+  // Returns the scissor box if clipping is needed,
+  // - undefined if not
+  // - null if viewport is outside canvas
+  private computeScissorBox(viewportBox: Box2): Box2 | undefined | null {
```
seankmartin, line 225: Would probably be nicer if this was instead a jsdoc comment describing the different return types

packages/core/src/renderers/webgl_renderer.ts
```diff
@@ -100,14 +140,24 @@ export class WebGLRenderer extends Renderer {
       }
     }
 
-    this.state_.setDepthMask(false);
-    for (const layer of transparent) {
-      layer.update(renderContext);
-      if (layer.state === "ready" && viewportIsVisible) {
-        this.renderLayer(layer, viewport.camera, frustum);
+    if (viewportIsVisible) {
```
seankmartin, line 147: There's one note here in a change in behaviour between opaque and transparent layers. I'm not really sure which one is desired, but I imagine them being different isn't ideal.

For opaque layers, if the viewport is non visible, we still call `layer.update(renderContext)` for every opaque layer. However, with the change in this PR we don't call `layer.update(renderContext)` for any transparent layer that is part of a non visible viewport.

Is there a reason to update layers in non-visible viewports as opaque layers do? Or in other words, which would be correct:
1. Layers in non visible viewports continue to `update` with the `renderContext` as they did before this PR
2. Layers in non visible viewports do not `update` with the `renderContext`, which would be new behaviour

packages/core/src/renderers/webgl_renderer.ts
```diff
@@ -116,6 +166,96 @@ export class WebGLRenderer extends Renderer {
     return this.textures_.textureInfo;
   }
 
+  private renderTransparentLayers(
```
seankmartin: This is really minor, but I feel it would be better to move this back to being inline or to change the opaque rendering to also be in a function like `renderTransparentLayers`. Right now it looks like there's a big difference between the two, but really there is not. This function is pretty simple and only used in one place so I'd lean towards having it inline

packages/core/src/renderers/webgl_renderer.ts
```diff
@@ -116,6 +166,96 @@ export class WebGLRenderer extends Renderer {
...
+  // Returns the scissor box if clipping is needed,
+  // - undefined if not
+  // - null if viewport is outside canvas
+  private computeScissorBox(viewportBox: Box2): Box2 | undefined | null {
+    const rendererBox = new Box2(
+      vec2.fromValues(0, 0),
+      vec2.fromValues(this.width, this.height)
+    );
+    if (Box2.equals(viewportBox.floor(), rendererBox.floor())) {
+      return undefined; // full canvas, no scissor needed
```
seankmartin: don't think we need this comment, the top level func comment would suffice (same for L245 and L247)

seankmartin (commented):
Overall looks good thanks! Wanted to note one difference in behaviour for the Idetik team to weigh in on. The others are very small comments

packages/core/src/renderers/webgl_renderer.ts
aganders3: Adding JSDoc comments for the public API is on our roadmap, but we've mostly been taking the "all or nothing" approach and not using it for now. I think the only JSDoc comment we have is on the runtime itself.

packages/core/src/renderers/webgl_renderer.ts
aganders3: Thanks - I ran into this recently when adding a second viewport to one of our viewers, but haven't had a chance to move that forward or bring it up.

I'm not sure what the correct behavior should be. The result of `update` on a layer in a non-visible viewport may depend on _why_ it's not visible, because it will try to calculate visible chunks, etc. Right now I think `viewportIsVisible` means the viewport has non-zero size and overlaps the canvas.

So I think we have a couple of options:
1. call `update()` on layers in non-visible viewports, use CSS (`visibility: hidden` or whatever, I can't remember) to hide viewports. Viewports/layers can maintain "visible" chunks without actually rendering, so they're ready to render if/when they become visible again.
2. don't call `update()` on non-visible viewports, they will maintain the renderable objects they had when hidden (not sure if there is a race here)

What do you think?

seankmartin: Thanks Ashley, makes a lot of sense. I modified the multi-viewport example to allow toggling between showing the volume and slice viewport and making the shown one fullscreen, or showing both in half and half to get a feel for this. I separated t for each viewport, as otherwise at LOD 2 on volume rendering both fetch very similar data. At higher resolution LODs for volume rendering this would be a bit different

Here is the version that calls update on non-visible viewports:
https://github.com/user-attachments/assets/39b5cdb5-c68a-45f7-a10d-6a4a704d18f9

And the version that does not call update on non-visible viewports:
https://github.com/user-attachments/assets/958eabd2-d10a-4d1e-a2dd-b393c561d3db

Both feel pretty reasonable (I ran the non-update first so cache would be a bit better on the always update version), but it can be nice in the version that does call update how quickly you can see the result. If someone was playing back through time while changing visible viewports this could feel pretty good.

Overall I think it would be reasonable not to call update, but I think for the sake of scoping in this PR I'd lean that we change this PR to call `update` on transparent layers always. And can open up a follow up PR/issue about viewport visibility/layer updates.

PR #623, opened 2026-03-27
chore: add LICENSE file for stats.js

aganders3:
Adding the license for stats.js


andy-sweet (approved):
Good enough to approve.

A couple of possible future improvements.

1) Remove this stats.js code as part of the library and make it a utility for any applications that live in this repo.

2) If we plan to add more vendored code in the future, a sub-directory per library is probably cleaner.

PR #627, opened 2026-03-30
feat: publicly expose chunk queue stats on runtime

aganders3:
### Summary
This exposes the chunk queue stats via the runtime. The goal is to allow applications to understand loading states to be able to communicate loading/progress.

This also exposes the `Overlay` type for external implementors.

### Related Issue
N/A

### Tests & Checks
Tested in a downstream application using the chunk queue stats to create a loading indicator

### DIFF
--- a/packages/core/src/core/chunk_manager.ts
+++ b/packages/core/src/core/chunk_manager.ts
@@ -1,5 +1,10 @@
 import { ChunkSource } from "../data/chunk";
 import { ChunkQueue } from "../data/chunk_queue";
+
+export type QueueStats = {
+  pending: number;
+  running: number;
+};
 import { ChunkStore } from "./chunk_store";
 import { ChunkStoreView } from "./chunk_store_view";
 import { ImageSourcePolicy } from "./image_source_policy";
@@ -8,6 +13,13 @@ export class ChunkManager {
   private readonly stores_ = new Map<ChunkSource, ChunkStore>();
   private readonly pendingStores_ = new Map<ChunkSource, Promise<ChunkStore>>();
   private readonly queue_ = new ChunkQueue();
+
+  public get queueStats(): QueueStats {
+    return {
+      pending: this.queue_.pendingCount,
+      running: this.queue_.runningCount,
+    };
+  }
 
   public async addView(
     source: ChunkSource,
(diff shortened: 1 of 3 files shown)


packages/core/src/index.ts
```diff
@@ -1,4 +1,5 @@
 export { Idetik } from "./idetik";
+export type { Overlay } from "./idetik";
```
andy-sweet, line 2: Do you actually need to export this type for functional reasons? Or is this just to help clarity for whoever imports this (e.g. allowing them to explicit an object with this type)?

packages/core/src/core/chunk_manager.ts
```diff
@@ -9,6 +14,13 @@ export class ChunkManager {
   private readonly pendingStores_ = new Map<ChunkSource, Promise<ChunkStore>>();
   private readonly queue_ = new ChunkQueue();
 
+  public get queueStats(): QueueStats {
```
andy-sweet, line 17: Optional: returning a new object for these stats feels a bit like overkill, when they could just be fairly simple and descriptive properties like `numPending` and `numRunning`. But if you plan to add more stats in the future, then introducing an object is sensible.

andy-sweet (approved):
No blocking comments.

I had imagined that `Overlay` would become `PreRenderCallback` and `PostRenderCallback` as a more generic way to providing some way into the idetik animation loop. But I think it's fine to expose the type now and clean up later.

packages/core/src/index.ts
aganders3: It's not strictly necessary, but I think it makes the typing downstream a little clearer. I'm not very confident in how TS handles type inference or duck typing, but I think it will be nice to have this in any API documentation as well.

packages/core/src/core/chunk_manager.ts
aganders3: Thanks - in general I agree and I'm not sure exposing properties like this on the runtime is really the correct place anyway. I suppose for now this could even just be `queueDepth` and return the sum.

I think the chunk manager in general has access to more useful information about chunks, but maybe that suggests calling this `chunkStats` instead.

I'm not too worried about breaking this later, so I'm going to merge for now and revisit when we consider the public API for a 1.0 release.

PR #628, opened 2026-03-31
feat: volume frustum chunk culling and performance improvements

seankmartin:
### Summary

The primary goal of this PR was to add frustum culling at the chunk level for volume rendering. I had originally also aimed to harmonize the slice and volume paths in the chunk store view a little, similar to https://github.com/chanzuckerberg/idetik/pull/579 (thanks for the pointer to this Ashley!). But with the performance issues mentioned below, that kind of blew this PR up in size. So scaled back to just modifying the culling of chunks for volumes and aiming to improve performance.

This change brought up some pretty major performance issues when directly implemented. Most of these weren't introduced by the frustum culling, but the frustum culling either highlighted the issue or the frustum culling extra operations were compounding with some performance problems already there. I'll note which ones were pre-existing because they could be pulled out of this PR as separate improvements if needed.

From profiling these seemed to be (in rough order):
1. Volume layer memory management inefficiencies (pre-existing issue). The volume layer was previously clearing all volumes whenever there were different chunks coming from the chunk store view. This didn't have a huge impact before, but with chunks updating rapidly this caused a very large bottleneck with new object creation and texture updates. Fixing this was likely the biggest improvement.
2. Recomputing the frustum and camera position (new issue). The first version of this PR obtained the frustum and position directly from the camera. But the first operation creates a new frustum each time, and the latter operation clones the camera position vector. These operations happened for each chunk, and sometimes also per timepoint. This inefficiency was removed at the chunk store level, not modifying the camera.
3. Recomputing chunk bounds and centers regularly in the chunk store view (pre-existing issue, compounded with frustum culling). The chunk store view was regularly recomputing the chunk bounds and center. Since the chunk is created once, we can cache these bounds. Right now, they are cached in the chunk store view, but we could alternatively lazily compute and then cache it on the chunk itself. As a note, the 2D code also uses this.
4. Volume sorting was recomputing the proxy geometry centers on each comparison (pre-existing issue, compounded because with frustum culling we can now use higher res LODs). The centers are now all computed before the sort, not during the sort.
5. Frustum intersection checks repeated across time (new issue). For time prefetching, we were doing many frustum intersection for chunks that shared the same spatial position and bounds. We introduced a cache for these results. Because we don't want to enforce that across t everything is chunked the same (unsure if a requirement or not for Idetik), the bounds are also cached and checked for equality on top of a spatial index + lod key. If we know that everything is chunked the same across t that check could be removed, but the performance impact is quite low of this check. Definitely the map access + bounds check is lower cost than the full frustum intersection test.

### Related Issue
Closes #<issue-number>

### Tests & Checks

Manually tested changes. Introduced the old example for this, but that can be removed, just something that anyone can use to test the feel of the perf changes. Verified chunk culling through logs, breakpoints, and network requests.

Right now the image mask example fails to load the data but that's the same on main branch.

### DIFF
--- a/packages/core/src/core/chunk_store_view.ts
+++ b/packages/core/src/core/chunk_store_view.ts
@@ -6,6 +6,7 @@ import { ImageSourcePolicy } from "./image_source_policy";
 import { ReadonlyVec2, vec2, vec3, mat4 } from "gl-matrix";
 import { Box2 } from "../math/box2";
 import { Box3 } from "../math/box3";
+import type { Frustum } from "../math/frustum";
 import { Logger } from "../utilities/logger";
 import { clamp } from "../utilities/clamp";
 
@@ -30,6 +31,10 @@ export class ChunkStoreView {
   private sourceMaxSquareDistance2D_: number;
   private maxSquareDistance3D_: number | null = null;
   private readonly chunkViewStates_: Map<Chunk, ChunkViewState> = new Map();
+  private readonly chunkBoundsCache_ = new Map<
+    Chunk,
+    { bounds: Box3; center: vec3 }
+  >();
 
   private isDisposed_ = false;
 
@@ -215,7 +220,8 @@ export class ChunkStoreView {
     this.chunkViewStates_.forEach(resetChunkViewState);
 
     const fallbackLOD = this.fallbackLOD();
-
+    const cameraFrustum = viewport.camera.frustum;
+    const cameraPosition = viewport.camera.position;
     this.maxSquareDistance3D_ = null;
     for (const chunk of currentTimeChunks) {
       const isCurrentLOD = chunk.lod === this.currentLOD_;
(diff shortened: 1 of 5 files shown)


packages/core/src/core/chunk_store_view.ts
```diff
@@ -28,7 +29,19 @@ export class ChunkStoreView {
...
   private sourceMaxSquareDistance2D_: number;
+  private maxSquareDistance3D_: number | null = null;
   private readonly chunkViewStates_: Map<Chunk, ChunkViewState> = new Map();
+  private readonly chunkBoundsCache_ = new Map<
+    Chunk,
+    { bounds: Box3; center: vec3 }
+  >();
+  // This frustum cache can be simplified if we can assume that
+  // chunks at the same xyz indices at a given LOD across time
+  // all share the same bounds (can remove the bounds here and bounds check)
```
aganders3, line 40: This seems like a safe assumption to me...

packages/core/src/math/sort_by_distance.ts
```diff
@@ -16,17 +16,20 @@ export function sortFrontToBack(
...
+  }
 
-    const cam2aDistance = vec3.squaredDistance(cameraPosition, centerA);
-    const cam2bDistance = vec3.squaredDistance(cameraPosition, centerB);
+  objects.sort((a, b) => {
+    const cam2aDistance = vec3.squaredDistance(cameraPosition, centers.get(a)!);
+    const cam2bDistance = vec3.squaredDistance(cameraPosition, centers.get(b)!);
 
     return cam2aDistance - cam2bDistance;
   });
```
aganders3, line 35: I think this can be slightly more efficient by pre-computing the squared distance as well.
```suggestion
  const squaredDistances = new Map<RenderableObject, vec3>();
  for (const obj of objects) {
    const bbox = obj.boundingBox;
    const center = vec3.create();
    vec3.add(center, bbox.max, bbox.min);
    vec3.scale(center, center, 0.5);
    squaredDistances.set(obj, vec3.squaredDistance(cameraPosition, center));
  }

  objects.sort((a, b) => squaredDistances.get(a)! - squaredDistances.get(b)!);
```

PR #630, opened 2026-03-31
fix: replace UUID generator with incremental IDs

aganders3:
### Summary
This will remove the CodeQL warnings security@ requested we resolve before making the repo public.

This switches to use simple incrementing counters with optional prefixes for ID generation.

This could be even simpler if we don't use the same function for assigning the fallback ID for viewports. That could probably be moved/removed but it's pretty straightforward. We were only using the UUID generator there for convenience anyway.

### Related Issue
Closes https://github.com/chanzuckerberg/idetik/security/code-scanning/3

### Tests & Checks
Tests all still pass


aganders3:
Note `Node.uuid` is public so changing the name here is technically a breaking change, but I don't think it's used anywhere.

shlomnissan (approved):

PR #636, opened 2026-04-03
feat: support multi-channel selection in slice coordinates

shlomnissan:
### Summary

The public API for channels is probably best configured as follows:

- Each layer accepts a channel props array that covers configs for all channels.
- `SliceCoordinates::c` should always be an array.
- `SliceCoordinates` decides what gets downloaded, decoupled from channel visibility.

This gives us a unified and explicit API.

This PR updates the API so each layer includes channel props for all channels, makes `SliceCoordinates::c` an array, and handles loading individual channels for volume rendering.

This PR replaces #617. It's smaller but more files changed because it required updating examples.  **Note this PR breaks the public API**.

### Tests & Checks

All checks have passed.

### DIFF
--- a/packages/core/examples/image_labels_overlay_with_value_picking/main.ts
+++ b/packages/core/examples/image_labels_overlay_with_value_picking/main.ts
@@ -41,7 +41,7 @@ const yStopPoint = yLod.size * yLod.scale;
 
 const sliceCoords = {
   t: tStartPoint,
-  c: phaseChannelIndex,
+  c: [phaseChannelIndex],
   z: zMidPoint,
 };
 
@@ -75,7 +75,11 @@ const imageLayer = new ChunkedImageLayer({
   sliceCoords,
   policy: createExplorationPolicy(),
   transparent: true,
-  channelProps: [{ contrastLimits: phaseContrastLimits }],
+  channelProps: [
+    { visible: true, contrastLimits: phaseContrastLimits },
+    { visible: false },
+    { visible: false },
+  ],
 });
 
 // Function to create label layer with current mode
--- a/packages/core/examples/image_labels_overlay_with_value_picking/main.ts
+++ b/packages/core/examples/image_labels_overlay_with_value_picking/main.ts
@@ -26,10 +26,6 @@ const lod = 0;
 const loader = await imageSource.open();
 const dimensions = loader.getSourceDimensionMap();
 
-// Phase contrast limits were chosen qualitatively.
-const phaseChannelIndex = 0;
-const phaseContrastLimits: [number, number] = [20, 200];
-
 const tStartPoint = 0;
 
 const zLod = dimensions.z!.lods[0];
@@ -41,7 +37,7 @@ const yStopPoint = yLod.size * yLod.scale;
 
 const sliceCoords = {
   t: tStartPoint,
-  c: [phaseChannelIndex],
+  c: [0],
   z: zMidPoint,
 };
 
@@ -76,7 +72,7 @@ const imageLayer = new ChunkedImageLayer({
   policy: createExplorationPolicy(),
   transparent: true,
   channelProps: [
-    { visible: true, contrastLimits: phaseContrastLimits },
+    { visible: true, contrastLimits: [20, 200] },
     { visible: false },
     { visible: false },
   ],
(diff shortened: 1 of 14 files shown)


shlomnissan:
@aganders3 I think we can improve the API if we make the `visible` field required but this change is out of scope for this PR. Would like to hear your thoughts

packages/core/examples/image_labels_overlay_with_value_picking/main.ts
```diff
@@ -75,7 +75,11 @@ const imageLayer = new ChunkedImageLayer({
   sliceCoords,
   policy: createExplorationPolicy(),
   transparent: true,
-  channelProps: [{ contrastLimits: phaseContrastLimits }],
+  channelProps: [
+    { visible: true, contrastLimits: phaseContrastLimits },
+    { visible: false },
+    { visible: false },
+  ],
```
aganders3, line 78: Optional: construct the channel props dynamically since `phaseChannelIndex` is already defined.


```
const channelCount = dimensions.c!.lods[0].size;
const channelProps = Array.from({ length: channelCount }, (_, i) =>
  i === phaseChannelIndex
    ? { visible: true, contrastLimits: phaseContrastLimits }
    : { visible: false }
);
```

packages/core/examples/volume_rendering/main.ts
```diff
@@ -11,7 +11,7 @@ const source = OmeZarrImageSource.fromHttp({ url });
 const sliceCoords = {
   t: 0,
   z: undefined,
-  c: undefined, // Show all channels
+  c: [], // Show all channels
```
aganders3: I'm a little unsure about using `[]` to specify "all channels". This would initially suggest "no channels" to me. I'm not sure that will be a common need, but I could see it being the default state in an app that lets you select which channels to download.

I think the type still supports `undefined` with these changes, what do you think of using that for "all channels"?

aganders3 (approved):
I like the API here. It feels like a clear separation of responsibilities to have the slice coordinates control the fetch, and channel properties control visibility. I was resistant to requiring `channelProperties` required for all source channels, but I think it's the right choice and simple enough if `visible` is the only required field. 

One question comes to mind: does channel order in the slice coordinates affect anything? For example does it influence blending? For now with `ChunkedImageLayer` blending is affected by the layer order (one channel per layer), but that API will eventually change.

> I think we can improve the API if we make the visible field required

I agree - hopefully with this we can also make `channelProps_` non-nullable on the layers and get rid of `initialChannelProps_`.

shlomnissan:
> One question comes to mind: does channel order in the slice coordinates affect anything? For example does it influence blending? For now with ChunkedImageLayer blending is affected by the layer order (one channel per layer), but that API will eventually change.

@aganders3 Good question! I think it depends how we implement blending if we had a single layer that supports multiple channels. If channels within the layer use additive blending then it's order-independent by default, right?

packages/core/examples/image_labels_overlay_with_value_picking/main.ts
shlomnissan: I inlined `phaseChannelIndex` and `phaseContrastLimits` instead. I prefer an explicit approach. Constructing dynamically at this scale feels like unnecessary complexity to me.

packages/core/examples/volume_rendering/main.ts
shlomnissan: Yep that makes sense. In this case we can either say `[]` nothing to render or `[]` error. Based on your reasoning I think the former is the right approach? I modified the volume and chunked image layers.

aganders3:
> If channels within the layer use additive blending then it's order-independent by default, right?

Yes, and I think that's probably the most common use case for channels.

I think sometimes alpha compositing is used when the background is a more dense image, like bright field. But that's also a _mixed_ blending mode that gets more complicated (additive fluorescent channels overlaid with alpha compositing on the bright field image). Maybe the answer there is to use a separate layer for the background image. I don't think that's a terrible compromise/API.

I suppose that could mean restricting blending channels in a single layer to commutative modes? In that case `channel` could be a set, but the array is also fine and simple.

I'm just musing, this is not blocking of course.

shlomnissan:
> I suppose that could mean restricting blending channels in a single layer to commutative modes? In that case channel could be a set, but the array is also fine and simple.

@aganders3 or even more explicit: always assume additive for a single layer with multiple channels. Any other configuration requires multiple layer, blending order depends on layer order. That's the nice thing of supporting both.

PR #638, opened 2026-04-08
feat: add `getVisibleDataRange` to `ChunkedImageLayer` and `VolumeLayer`

aganders3:
### Summary
This PR adds per-channel min/max statistics derived from currently visible chunks. This is intended to support basic auto-contrast adjustment in downstream apps.

Per-chunk min/max is computed once when chunk data loads and cached on the chunk itself. `getVisibleDataRange()` iterates the already-tracked visible chunks and aggregates by channel index, returns {} when no chunks have loaded yet; callers are expected to handle this. Downstream apps already access dtype range via the loader as a fallback (though it may be nice to make this more ergonomic, it requires an async call to get the loader).

To keep the core implementation light, this does not add any kind of callback or notification when it may have changed (e.g. new chunks loaded, changing camera, timepoint, slice, etc.). This means downstream applications may need to poll or have user interaction (e.g. "auto" button next to contrast sliders).

### Related Issue
Closes N/A

### Tests & Checks
Tested with logging and by installing locally in `idetik-dca-viewer`, where I implemented some basic auto-contrast features.

### DIFF
--- a/packages/core/src/data/chunk.ts
+++ b/packages/core/src/data/chunk.ts
@@ -34,8 +34,14 @@ export type ChunkViewState = {
   orderKey: number | null;
 };
 
+export type ChunkDataStats = {
+  min: number;
+  max: number;
+};
+
 export type Chunk = {
   data?: ChunkData;
+  dataStats?: ChunkDataStats;
   state: "unloaded" | "queued" | "loading" | "loaded";
   lod: number;
   shape: {
@@ -115,6 +121,37 @@ export type ChunkLoader = {
 
   loadChunkData(chunk: Chunk, signal: AbortSignal): Promise<void>;
 };
+
+export function computeChunkDataStats(data: ChunkData): ChunkDataStats {
+  let min = Infinity;
+  let max = -Infinity;
+  for (let i = 0; i < data.length; i++) {
+    const v = data[i];
+    if (v < min) min = v;
+    if (v > max) max = v;
+  }
+  return { min, max };
+}
+
+export type ChannelDataStats = Record<number, ChunkDataStats>;
+
+export function computeChannelDataStats(
+  chunks: Iterable<Chunk>
+): ChannelDataStats {
+  const result: ChannelDataStats = {};
+  for (const chunk of chunks) {
+    if (!chunk.dataStats) continue;
+    const c = chunk.chunkIndex.c;
+    const existing = result[c];
+    if (!existing) {
+      result[c] = { min: chunk.dataStats.min, max: chunk.dataStats.max };
+    } else {
+      existing.min = Math.min(existing.min, chunk.dataStats.min);
+      existing.max = Math.max(existing.max, chunk.dataStats.max);
+    }
+  }
+  return result;
+}
 
 export function coordToIndex(lod: SourceDimensionLod, coord: number): number {
   return Math.round((coord - lod.translation) / lod.scale);
--- a/packages/core/src/data/chunk.ts
+++ b/packages/core/src/data/chunk.ts
@@ -34,14 +34,14 @@ export type ChunkViewState = {
   orderKey: number | null;
 };
 
-export type ChunkDataStats = {
+export type ChunkDataRange = {
   min: number;
   max: number;
 };
 
 export type Chunk = {
   data?: ChunkData;
-  dataStats?: ChunkDataStats;
+  dataRange?: ChunkDataRange;
   state: "unloaded" | "queued" | "loading" | "loaded";
   lod: number;
   shape: {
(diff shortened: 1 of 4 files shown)


aganders3:
I'll reiterate that this only allow very basic auto-contrast and contrast bounds setting, but I think it's better than nothing. If we want to improve we will likely need to calculate histograms. That would be a great feature but lower-priority.

Here's an example of what kind of thing we can achieve with just this visible min/max. On the left is our current "guess" and on the right is with simple auto-contrast based on min/max. There is a small chance this makes things appear worse IMO (see some of our vizarr links for examples) but I think it's an improvement in most cases, especially with a conservative algorithm.
Screenshot 2026-04-08 at 6 31 14 PM


In the end putting this info in the OMERO metadata is still going to provide the best experience, but some auto-contrast is a nice-to-have fallback and helpful in certain processing workflows.

packages/core/src/data/ome_zarr/image_loader.ts
```diff
@@ -105,6 +106,7 @@ export class OmeZarrImageLoader {
       );
     }
     chunk.rowAlignmentBytes = rowAlignment;
+    chunk.dataStats = computeChunkDataStats(chunk.data);
```
shlomnissan: I think using "range" is more descriptive unless we plan to add additional stats or maybe I'm missing something?

```ts
chunk.dataRange = computeChunkDataRange(chunk.data)
```

packages/core/src/data/chunk.ts
```diff
@@ -116,6 +122,37 @@ export type ChunkLoader = {
   loadChunkData(chunk: Chunk, signal: AbortSignal): Promise<void>;
 };
 
+export function computeChunkDataStats(data: ChunkData): ChunkDataStats {
```
shlomnissan: When measuring with the zebrafish image, average pass is 10ms but sometimes it spikes on my end.

If there's a way to do this in a webworker without added complexity (library of some sort) it might be better but I'm not sure. Thoughts?

packages/core/src/data/chunk.ts
```diff
@@ -116,6 +122,37 @@ export type ChunkLoader = {
...
+  let min = Infinity;
+  let max = -Infinity;
+  for (let i = 0; i < data.length; i++) {
+    const v = data[i];
+    if (v < min) min = v;
+    if (v > max) max = v;
+  }
```
shlomnissan, line 132:
```suggestion
  for (const v of data) {
    min = Math.min(min, v);
    max = Math.max(max, v);
  }
```

packages/core/src/data/chunk.ts
```diff
@@ -116,6 +122,37 @@ export type ChunkLoader = {
...
+  return { min, max };
+}
+
+export type ChannelDataStats = Record<number, ChunkDataStats>;
```
shlomnissan: I don't think you need to export/import this. Return value for indirection functions in layers will be deduced.

shlomnissan:
> returns {} when no chunks have loaded yet; callers are expected to handle this

Are you saying callers already have the data they need to handle it? I'm trying to understand what would be the recommended pattern for calling this function as a consumer.

shlomnissan (approved):
LGTM

packages/core/src/data/ome_zarr/image_loader.ts
aganders3: My initial idea was that this would include a per-chunk histogram, but it was more complicated than its worth for a quick implementation. I agree `dataRange` is a better name and will change it.

packages/core/src/data/chunk.ts
aganders3: Thanks for profiling it. I think it will be easy to move into the webworker call just above it where we fetch/decompress the data already. It's probably worth doing that since we're now sometimes slicing channels/time out of there. I'll open another PR for that if it's as straightforward as I think.

packages/core/src/data/chunk.ts
aganders3: I tested this and the index loop is significantly faster for some reason, so I'm going to leave it as-is.

aganders3:
> > returns {} when no chunks have loaded yet; callers are expected to handle this
>
> Are you saying callers already have the data they need to handle it? I'm trying to understand what would be the recommended pattern for calling this function as a consumer.

I think there are two cases we might call this as a consumer.

The simple case is to call it on-demand, e.g. the user clicks an "auto contrast" button on a channel control.

The more complex case is to call it automatically, e.g. when the image first loads. We don't currently have affordances for when a specific layer has either started or finished downloading or displaying what it needs. Currently even the layer doesn't know some of this information — it gets a certain list of chunks to render, but it does not know if that set is complete (so it could even get an empty list). Without additional events or callbacks here, a downstream application might just have to poll a bit on first load or something.

PR #639, opened 2026-04-08
chore: only remove stale volumes instead of clearing all

shlomnissan:
### Summary

This PR fixes a performance issue in `VolumeLayer` where `updateChunks` was
clearing all volumes from `currentVolumes_` on every update, forcing full teardown
and recreation of every volume even when most volumes hadn't changed.

Original fix by @seankmartin https://github.com/chanzuckerberg/idetik/pull/628

### Tests & Checks

- All checks have passed

### DIFF
--- a/packages/core/src/layers/volume_layer.ts
+++ b/packages/core/src/layers/volume_layer.ts
@@ -145,7 +145,9 @@ export class VolumeLayer extends Layer implements ChannelsEnabled {
 
   public onDetached(_context: IdetikContext): void {
     if (!this.chunkStoreView_) return;
-    this.releaseAndRemoveVolumes(this.currentVolumes_.values());
+    for (const volume of this.currentVolumes_.values()) {
+      this.releaseAndRemoveVolume(volume);
+    }
     this.clearObjects();
     this.chunkStoreView_.dispose();
     this.chunkStoreView_ = undefined;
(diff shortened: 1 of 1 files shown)


packages/core/src/layers/volume_layer.ts
```diff
@@ -206,12 +208,10 @@ export class VolumeLayer extends Layer implements ChannelsEnabled {
...
-  private releaseAndRemoveVolumes(volumes: Iterable<VolumeRenderable>) {
-    for (const volume of volumes) {
-      volume.clearLoadedChannels();
-      this.pool_.release(this.volumeToPoolKey_.get(volume)!, volume);
-      this.volumeToPoolKey_.delete(volume);
-    }
+  private releaseAndRemoveVolume(volume: VolumeRenderable) {
+    volume.clearLoadedChannels();
+    this.pool_.release(this.volumeToPoolKey_.get(volume)!, volume);
+    this.volumeToPoolKey_.delete(volume);
```
aganders3, line 214: Could/should this also just delete from `this.currentVolumes_`?

aganders3 (approved):

packages/core/src/layers/volume_layer.ts
shlomnissan: It could but I think we will need to pass the key as a parameter or do a reverse lookup. Maybe keeping the delete at the call site is cleaner

aganders3: I see I totally missed the separate key. You're right this is cleaner.

PR #646, opened 2026-04-10
feat: remove deprecated layers and rename `ChunkedImageLayer` -> `ImageLayer`

aganders3:
### Summary
This is the first step in the process of deprecating layers and cleaning up examples.

#### Summary of changes
- Removed `ProjectedLineLayer`
- Removed `TracksLayer`
- Removed `ImageLayer` and `ImageSeriesLayer`
- Renamed `ChunkedImageLayer` -> `ImageLayer`
- Deleted examples for removed layers: image2d_from_omezarr5d_u16, image_mask_overlay, image_series_from_omezarr5d_u8, image2d_from_omezarr5d_u16layer_blending, projected_lines, tracks
- Remaining examples all use the new (chunked) `ImageLayer`
- Scale bar overlay added to the chunk_streaming example

### Tests & Checks
Tests pass locally
Examples all work

### Next steps
1. Label layers will be replaced with a single chunked label layer implementation. This will also allow removing the ImageSeriesLoader and related series-loading infrastructure that currently backs the series layer pattern.
2. Many of the examples can be consolidated into the chunk_streaming example. I think we should rename this example and add a source selector to still cover a variety of datasets.
3. The old `ImageLayer` implementation supported multi-channel images, but currently we use one-channel-per layer and layer blending to display multi-channel images with the new `ImageLayer` implementation. Updating `ImageLayer` to support multi-channel images directly is also on the roadmap, but not top priority.


shlomnissan (approved):
LGTM

PR #648, opened 2026-04-10
feat: port `LabelImageLayer` and `LabelImageSeriesLayer` to new chunky `LabelLayer`

aganders3:
### Summary
This adds `LabelLayer`, which combines both `LabelImageLayer` and `LabelImageSeriesLayer` to a single layer using the same chunk-streaming infrastructure as `ImageLayer`. Much of the code in `src/layers/label_layer.ts` is either copied or from or very similar to the existing `ImageLayer` (formerly `ChunkedImageLayer`).

This also allows removing the now-unused `ImageSeriesLoader` and a bunch more now-unused loading code, including the "Region" type. I think there may be room for a type like this (for example for describing a global "crop") in the future, but I don't think we need to maintain it until we need it.

After consolidating the layers, I also combined the two label examples into one.

### Related Issue
Closes N/A

### Tests & Checks
Tests pass locally
Remaining examples all work as expected


shlomnissan (approved):
LGTM :+1:

PR #649, opened 2026-04-10
refactor: remove texture info and overlay

shlomnissan:
### Summary

This PR removes the texture info accessor and the overlay that used it.

- The overlay was used for debugging purposes so it's safe to remove it.
- The texture info is bad because it exposes renderer information via public API.
- It's also added to the WebGL renderer instead of the `Renderer` base class.

I'm not against having some stats object that clients can consume but it needs
to be more structured than just texture buffer information.


aganders3 (approved):
LGTM but I am +1 on defining some runtime stats for consumers

PR #652, opened 2026-04-11
feat: experimental WebGPU renderer

shlomnissan:
### Summary

Experimental WebGPU renderer. Currently only supports the image layer.
This PR replaces the renderer in first 3 examples to use the WebGPU
renderer instead of WebGL.


aganders3 (approved):
Overall this looks great to me! I've got more to learn but I'm happy to get it in as-is and work from there.

Issue #626, opened 2026-03-30
Make non-visible viewport layer updates configurable


seankmartin:
I'm not sure what the correct behavior should be. The result of `update` on a layer in a non-visible viewport may depend on _why_ it's not visible, because it will try to calculate visible chunks, etc. Right now I think `viewportIsVisible` means the viewport has non-zero size and overlaps the canvas.

So I think we have a couple of options:
1. call `update()` on layers in non-visible viewports, use CSS (`visibility: hidden` or whatever, I can't remember) to hide viewports. Viewports/layers can maintain "visible" chunks without actually rendering, so they're ready to render if/when they become visible again.
2. don't call `update()` on non-visible viewports, they will maintain the renderable objects they had when hidden (not sure if there is a race here)

_Originally posted by @aganders3 in https://github.com/chanzuckerberg/idetik/pull/622#discussion_r3001207116_

seankmartin:
I modified the multi-viewport example to allow toggling between showing the volume and slice viewport and making the shown one fullscreen, or showing both in half and half to get a feel for this. I separated t for each viewport, as otherwise at LOD 2 on volume rendering both fetch very similar data. At higher resolution LODs for volume rendering this would be a bit different

Here is the version that calls update on non-visible viewports:
https://github.com/user-attachments/assets/39b5cdb5-c68a-45f7-a10d-6a4a704d18f9

And the version that does not call update on non-visible viewports:
https://github.com/user-attachments/assets/958eabd2-d10a-4d1e-a2dd-b393c561d3db

Both feel pretty reasonable (I ran the non-update first so cache would be a bit better on the always update version), but it can be nice in the version that does call update how quickly you can see the result. If someone was playing back through time while changing visible viewports this could feel pretty good.

It might be best to consider to allow the update to be configurable on the viewport. Because I'm not sure we can accurately infer why the viewport is non-visible and what the desired behaviour for that viewport should be

aganders3:
Closed by #653

PR #653, opened 2026-04-13
refactor: cleanup WebGL render function

shlomnissan:
### Summary

This PR cleans up the WebGL render function:
1. Removed `viewportIsVisible` hidden and out-of-bounds cases bail immediately.
2. Update loop separated from render loop (obvious that layer updates always happen regardless of visibility).
3. Renamed `rendererBox` to `surfaceBox` matches domain terminology.

Best way to review is reading the entire function instead of the diff.

All checks have passed.

### DIFF
--- a/src/renderers/webgl_renderer.ts
+++ b/src/renderers/webgl_renderer.ts
@@ -61,53 +61,55 @@ export class WebGLRenderer extends Renderer {
   }
 
   public render(viewport: Viewport) {
-    let viewportIsVisible =
-      getComputedStyle(viewport.element).visibility !== "hidden";
+    this.renderedObjects_ = 0;
+    this.renderedObjectsPerFrame_ = 0;
+
+    const { opaque, transparent } = viewport.layerManager.partitionLayers();
+    for (const layer of [...opaque, ...transparent]) {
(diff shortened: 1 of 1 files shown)


src/renderers/webgl_renderer.ts
```diff
@@ -132,6 +131,7 @@ export class WebGLRenderer extends Renderer {
       this.gl_.clear(this.gl_.STENCIL_BUFFER_BIT);
     }
 
+    const frustum = camera.frustum;
```
aganders3: There's a potential performance regression here, as the frustum is not cached and will be recalculated for each layer in a viewport.

I don't think this is a big deal, given we're not currently stacking tons of layers into viewports. We can solve that later if it becomes an issue.

aganders3 (approved):
Looks good - nice cleanup

I think this closes #626 and I agree here that we should call `layer.update()` regardless of visibility.

src/renderers/webgl_renderer.ts
shlomnissan: Ah that's a good call, I didn't look at internals. I'll update it before merging.

PR #655, opened 2026-04-16
refactor: remove unused texture array from image renderables

shlomnissan:
### Summary

This PR removes the texture array path from the image renderable and deletes related
types and shader program.

The initial implementation of the image layer used texture arrays for multiple channels.
Since we currently use one layer per channel, a texture array isn't serving a real purpose.
This adds unnecessary complexity and overhead.

@aganders3 I was hoping to delete `Texture2DArray` altogether or at least make it an internal
type but noticed it was exported and used by the point layer and the point example program.
Keeping it is fine but I'm a little worried about clients instantiating it directly. The cleanest
fix here is to make points own its atlas internally rather than accepting a texture from the
outside, but that's out of scope for this PR.

### Tests & Checks

- All checks have passed


aganders3 (approved):
LGTM

PR #657, opened 2026-04-30
Volume rendering in compute shaders

shlomnissan:
### Summary

WIP


harishkrishnamurthy:
Found this while testing #658 (frame timing overlay) against this branch:

```
Uncaught Error: WebGPU volume rendering only supports unsigned data types (got floatVolume)
    at shaderNameForObject (webgpu_renderer.ts:511:13)
    at WebGPURenderer.renderObject (webgpu_renderer.ts:269:21)
    at webgpu_renderer.ts:230:14
    at Array.forEach (<anonymous>)
    at WebGPURenderer.renderLayer (webgpu_renderer.ts:228:19)
    at WebGPURenderer.render (webgpu_renderer.ts:176:14)
    at Idetik.animate (idetik.ts:256:22)
    at idetik.ts:284:12
```

The organelle_box dataset in the volume_rendering example uses float data, which hits this error on the WebGPU path. Just flagging in case it's not already on your radar.

PR #660, opened 2026-05-04
docs: add basic vitepress documentation site

aganders3:
### Summary

This adds a vitepress docs site and deploys it to GitHub Pages via a GH Actions workflow. This replaces the current examples-only deployment. The content is meant to be temporary filler, with this PR focused on the setup.

Examples are removed from the public site, but kept for local development if you run `npm run examples`.

Use `npm run docs:dev` to run the vitepress dev server that rebuilds docs on change.

Here's what this branch looks like. The theme is the default, but I changed it to use the Biohub purple.
Screenshot 2026-06-26 at 9 17 13 PM
Screenshot 2026-06-26 at 9 17 28 PM

I think the only thing we may need to change if/when moving to idetik.org is to update the "base path" for the build.

See #699, which adds TypeDoc API docs.


aganders3:
@shlomnissan please take another look when you have a chance - I removed the co-hosted links to the dev examples.

.github/workflows/deploy-docs.yml
```diff
@@ -0,0 +1,41 @@
+name: Deploy docs
+
+on:
+  push:
+    branches: [main]
+  workflow_dispatch:
```
shlomnissan: - is this for adding the "run workflow" button?
- when would we trigger this workflow manually?
- if it is intentional, maybe add a comment?

shlomnissan (approved):
LGTM :+1:

.github/workflows/deploy-docs.yml
aganders3: Yeah this is so you can run the workflow manually, but probably not needed. Good catch, I'll remove it!

PR #661, opened 2026-05-08
perf: restore increased worker count, and remove `getVisibleDataRange`

aganders3:
### Summary
In idetik-remote, chunks were being cancelled before they could finish loading, producing a mostly-black volume with occasional flashes. Two regressions were stacking: PR #647 cut the worker count from up to 8 down to 3, and PR #638 added a full min/max scan (`computeChunkDataRange`) to the chunk-load critical path. Each chunk's worker time roughly doubled at the same time as the worker pool was cut. This is generally fine for local rendering where the network is the bottleneck, but it's very bad for server-side rendering. Oops!

This PR effectively reverts #647 and #638. We lose `getVisibleDataRange` which was intended for use in an auto-contrast feature for [idetik-dca-viewer](https://github.com/chanzuckerberg/idetik-dca-viewer), but that has not been merged anyway.

### Related Issue
N/A

### Tests & Checks
manually tested this branch with idetik-remote to compare against the baseline performance (v0.14.0)

### DIFF
--- a/src/data/chunk.ts
+++ b/src/data/chunk.ts
@@ -32,14 +32,8 @@ export type ChunkViewState = {
   orderKey: number | null;
 };
 
-export type ChunkDataRange = {
-  min: number;
-  max: number;
-};
-
 export type Chunk = {
   data?: ChunkData;
-  dataRange?: ChunkDataRange;
   state: "unloaded" | "queued" | "loading" | "loaded";
   lod: number;
   shape: {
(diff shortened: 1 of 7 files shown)


aganders3:
I was able to achieve decent quality at LOD1 with 20FPS playback on idetik-remote with this branch. Chunks dropped occasionally, but prefetching more (32) helped and I think performance in general is comparable to 0.14.0.

https://github.com/user-attachments/assets/765a5c7c-b6f9-4ea6-95eb-a1cbb211af7d

shlomnissan (approved):
LGTM 👍

PR #662, opened 2026-05-11
docs: consolidate 2D image examples and allow custom URL input

aganders3:
### Summary

Consolidates the 2D image examples into a single `image_2d` with a dataset selector instead of one example per hardcoded URL.

The core of the `chunk_streaming` example is preserved, and the `ca_wave_dynamics` is included as a preset data source. The example now includes four OME-Zarr datasets covering both Zarr v2 and v3 (OME-Zarr 0.4 and 0.5): zebrahub, GFP Ca wave dynamics, marmoset neurons, and a fused exaSPIM reconstruction (very large image).

This also adds a custom URL field to allow visualization of other public datasets. A new info overlay (bottom-right) shows the dataset label, OME-Zarr / Zarr version, shape, LOD count, and voxel scale.

The exaSPIM preset is intentionally a large dataset (~7.9M chunks at LOD 0) with known performance issues on `main`. This example now serves as a baseline performance test.

Screenshot 2026-05-11 at 11 52 19 AM

Screenshot 2026-05-11 at 11 53 12 AM

### Related Issue
I will create two new issues that are well-exhibited by the exaSPIM example:
1. performance issues when panning/zooming
2. Chunk half-voxel offset issue, causing LODs to not line up properly (actually, this is already captured in https://github.com/chanzuckerberg/idetik/issues/369)

Both the large and small exaSPIM images also exhibit the issues described in #429.

### Tests & Checks
- `npm run compile` and `npm test -- --run` pass.
- Manually verified each preset loads and renders correctly
- exaSPIM preset loads, but shows expected slow pan/zoom
- Confirmed custom URL field works for another public dataset (https://aind-open-data.s3.amazonaws.com/exaSPIM_844958_2026-04-13_12-25-27/SPIM.ome.zarr/tile_000001_ch_561.zarr)

PR #664, opened 2026-05-11
fix: detect voxel-cell convention from OME-Zarr LOD translation pattern

### LINKED ISSUE #369: Allow image layers rendering offset to be configured
In PR #357 [I noticed](https://github.com/chanzuckerberg/idetik/pull/357#issuecomment-3156705458) that a newly added example defines OME-Zarr LOD transforms that seem to assume that the first voxel is effectively centered at (0, 0, 0) in the virtual coordinate system of the pixel data.

This is well explained in the [simple ITK docs](https://simpleitk.readthedocs.io/en/master/fundamentalConcepts.html#images).

@aganders3 [suggested a fix](https://github.com/chanzuckerberg/idetik/pull/357#issuecomment-3160291922) that seemed to confirm this.

The goal of this issue is to provide some way to configure image layers so that they can respect the standard image-processing mindset (where pixels are centered at (0, 0)) vs. a standard computer graphics mindset (where pixels are centered at (0.5, 0.5)).

In my experience, most scientific images tend to follow the image-processing mindset.

It is also worth noting that sometimes the lack of configuration has no important effects, especially when the imaging data is not otherwise transformed.

aganders3:
### Summary

Multiscale pyramids may use one of two conventions for the origin of each LOD: the _corner_ or the _center_ of the first voxel. To correct for this offset, a zarr will include a half-voxel-per-LOD translation pattern ("voxel-center" convention `trans_k = trans_{k-1} + 0.5 * (scale_k - scale_{k-1})`) or with constant translations across LODs ("voxel-corner" convention). Both are valid; the choice is set by the dataset author, but not declared anywhere explicitly.

Without normalization, downstream code (chunk visibility, rendering placement) gets the wrong extent for one or the other convention. We see both conventions in our sample data.

This fix detects the convention per-axis in `inferSourceDimensionMap` by checking whether each LOD's translation delta matches the half-voxel prediction. For center-convention dimensions, we shift translations by `-0.5 * scale` per-axis per-LOD. After this, `chunk.offset = trans + chunkIdx * chunkSize * scale` is always the voxel-cell corner, and the rest of our code can use `[offset, offset + shape * scale]` without regard for the source data convention.

See here the before/after of the "marmoset neurons" dataset. Note the offset in the corner, resolved by this PR (all corners look good after this change). The effect can be severe in 3D datasets with many levels (e.g. the exaSPIM dataset added in #662).

Screenshot 2026-05-11 at 4 02 40 PM
Screenshot 2026-05-11 at 4 02 30 PM

This effect can be especially jarring in 3D data where the "fallback" chunks show a very different slice compared to the current LOD.

### Related Issue
Closes #369

### Tests & Checks
Tested four datasets: zebrahub, ca wave dynamics, marmoset neurons, and exaSPIM

### DIFF
--- a/src/data/ome_zarr/image_loader.ts
+++ b/src/data/ome_zarr/image_loader.ts
@@ -159,10 +159,7 @@ function inferSourceDimensionMap(
     // pattern, shift translations by -0.5 * scale per LOD so downstream code
     // uses corner extents uniformly. Skip non-spatial axes (c, t) where the
     // convention doesn't apply and ChunkStore enforces translation == 0.
-    if (
-      image.axes[index].type === "space" &&
-      isVoxelCenterConvention(lods)
-    ) {
+    if (image.axes[index].type === "space" && isVoxelCenterConvention(lods)) {
       for (const lod of lods) lod.translation -= 0.5 * lod.scale;
     }
     return {
--- a/src/data/ome_zarr/image_loader.ts
+++ b/src/data/ome_zarr/image_loader.ts
@@ -150,6 +150,21 @@ function inferSourceDimensionMap(
         translation: translation[index],
       });
     }
+    // Normalize translations on spatial axes so chunk.offset = trans + chunkIdx
+    // * chunkSize * scale uniformly refers to the voxel-cell corner. Two
+    // conventions appear in the wild: "voxel-center" pyramids encode the
+    // half-voxel shift across LODs (translation_k - translation_{k-1} === 0.5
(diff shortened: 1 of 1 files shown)


aganders3:
Here is a demonstration where the effect when zooming is quite jarring:

### on `main`

https://github.com/user-attachments/assets/bcd988b2-e347-4b18-9a7c-c4df09e05d85

### with this PR
https://github.com/user-attachments/assets/f8e4845e-97a5-42eb-b136-daf2687e06cb

shlomnissan (approved):
LGTM :+1:

Issue #602, opened 2026-03-10
Cache volume renderable uniforms to avoid per object per frame calculations


seankmartin:
This is called once per object per frame with no caching. I don't think it's a big deal since volume rendering is GPU-bound, but something to think about as a follow-up optimization (dirty flag + cached uniforms map)

_Originally posted by @shlomnissan in https://github.com/chanzuckerberg/idetik/pull/564#discussion_r2907585838_

Issue #166, opened 2025-02-11
Profile one-off calculations in TS vs GLSL


aganders3:
I think the following should be more performant, though needs a bit more work on the typescript side.

```suggestion
    float texel = texture(texture0, TexCoords).r;
    float value = (texel + offset) / scale;
```

_Originally posted by @andy-sweet in https://github.com/chanzuckerberg/imaging-active-learning/pull/152#discussion_r1932625564_

PR #665, opened 2026-05-13
perf: faster chunk visibility testing

### LINKED ISSUE #663: Poor pan/zoom performance in `ImageLayer` for large images
Pan/zoom on an `ImageLayer` is slow for large datasets, e.g. the fused exaSPIM added to the examples in #662.

I believe the bottleneck is `ChunkStoreView.updateChunksFor{Image,Volume}` iterating every chunk at the current time index to classify visible/prefetch/fallback and establish loading priority. The downstream filters in (`ChunkStoreView.getChunksToRender`, `ChunkStoreView.allVisibleFallbackLODLoaded`) do the same: they `filter()` the full per-time-index `Chunk[]`. This is effectively 3x iteration on the full set of chunks with every `ImageLayer.update` (I believe same for `VolumeLayer`, possibly worse as it also holds multiple channels).

The iteration cost is `O(chunks-per-timepoint)` per frame, and it becomes a problem with larger
datasets like the fused exaSPIM data. The dataset is 1×1×21041×40660×66934 voxels (~7.9M chunks instantiated).

To see this, load the **exaSPIM Fused** preset in the new `image_2d` example (`#dataset=exaspim`).

To attribute the cost, open DevTools → Performance, record a few seconds of pan/zoom, and look at the call tree. The dominant self-time entries are inside `ChunkStoreView.updateChunksForImage` and inside the `.filter()` calls in `ChunkStoreView.getChunksToRender` and `ChunkStoreView.allVisibleFallbackLODLoaded`. In total this is 300ms+ per frame (on the main thread) on my laptop for the fused exaSPIM dataset.

aganders3:
### Summary
This restructures the `chunks` array to `Chunk[lod][t][c][z][y][x]` and uses direct indexing to reduce the total number of intersection tests used to determine chunk visibility and priority.

On `main`, `updateChunksForImage` iterates every chunk at the current time index testing each against the view bounds, channel filter, and LOD predicate. This is O(chunks_per_timepoint) per-frame work. With the new layout, `chunkIndexRange(bounds, lod)` converts the world-space view bounds directly to a chunk-index range, and `iterateChunksInBox` walks only the chunks the view actually overlaps. Per-frame work now scales with the visible region instead of the dataset.

The fallback LOD runs through the same bounded path against the prefetch AABB, so in-view fallback chunks act as a backdrop while current-LOD chunks load, and the prefetch padding stays warm for short pans. Datasets whose coarsest LOD is still pretty large (e.g. OPS) no longer pay an O(all_fallback_lod_chunks) cost per frame.

With this refactor, `getChunksToRender` and `allVisibleFallbackLODLoaded` iterate `chunkViewStates_` (the per-view output map, already reflecting the latest update) instead of re-scanning all chunks on every render call. They also no longer take `sliceCoords`, they just operate on the state established by the prior `updateChunksFor{Image,Volume}` call.

The volume path (`updateChunksForVolume`) is intentionally left unchanged, just re-expressed through the new iterators. Extending bound-visibility to the volume layer is a natural follow-up.

### Related Issue
Closes #663

### Tests & Checks
I instrumented the code with `performance.measure` and saw the per-frame visibility checking cost drop from 300ms+ to <1ms. I did not include these measurement spans in this PR because they add noise, but I think it would be helpful to instrument the code thoroughly in a follow-up to allow us to track regressions, etc. Possibly related to #602 #166 and other perf-related issues.

The difference is easily felt with the exaSPIM data in the main `image_2d` example. You can also see the FPS in the stats.js overlay go from 3-5fps to 100+fps. Here are before/after comparisons of zooming and panning.

#### on `main`

https://github.com/user-attachments/assets/7ca425a4-76c8-4843-9271-e093759cfa9d

#### with this PR

https://github.com/user-attachments/assets/58d803b8-3f0f-490f-9829-c9c4fd3c05b2

### DIFF
--- a/examples/image_2d/main.ts
+++ b/examples/image_2d/main.ts
@@ -322,12 +322,26 @@ function formatDatasetInfo(
     .replace(/</g, "&lt;")
     .replace(/>/g, "&gt;");
 
+  let totalChunks = 0;
+  for (let lod = 0; lod < dims.numLods; ++lod) {
+    const xLod = dims.x.lods[lod];
+    const yLod = dims.y.lods[lod];
+    const zLod = dims.z?.lods[lod];
+    const chunksX = Math.ceil(xLod.size / xLod.chunkSize);
+    const chunksY = Math.ceil(yLod.size / yLod.chunkSize);
+    const chunksZ = zLod ? Math.ceil(zLod.size / zLod.chunkSize) : 1;
+    const chunksT = tL?.size ?? 1;
+    const chunksC = cL?.size ?? 1;
+    totalChunks += chunksT * chunksC * chunksZ * chunksY * chunksX;
+  }
+
   return (
     `<div class="label">${escapedLabel}</div>` +
     `<div class="meta">` +
     [
       `shape:  ${shape}`,
       `LODs:   ${dims.numLods}`,
+      `chunks: ${totalChunks.toLocaleString()}`,
       `voxel:  ${scale}${unit ? " " + unit : ""}`,
     ].join("\n") +
     `</div>`
--- a/src/core/chunk_store.ts
+++ b/src/core/chunk_store.ts
@@ -11,7 +11,8 @@ import { ChunkStoreView } from "./chunk_store_view";
 import { ImageSourcePolicy } from "./image_source_policy";
 
 export class ChunkStore {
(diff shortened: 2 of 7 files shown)


shlomnissan (approved):
LGTM

PR #666, opened 2026-05-14
feat: updates to projected lines (WebGL)

aganders3:
### Summary
This simplifies the existing `ProjectedLine` renderable and fixes a few longstanding issues with it. It now supports properly mitered corners, closed paths, and width defined in screen pixels. All of this leas to more uniform appearance and consistent, predictable behavior. This also fixes issues with how lines appeared in viewports, fixing the appearance of the axes helper in a 4-up view, for example.

While here, I removed support for "tapering" along the path. This was previously used to convey a sense of temporal movement for cell tracking, but I was never happy with it. We can bring that back in better shape if/when required.

Note this only touches the WebGL rendering of these lines. I plan to also port this to the WebGPU renderer to learn more about WebGPU and the current implementation of that renderer.

### Related Issue
Closes N/A

#### on `main`
Screenshot 2026-05-14 at 1 00 54 PM

#### with this PR

Screenshot 2026-05-14 at 11 23 02 AM

### DIFF
--- a/src/core/geometry.ts
+++ b/src/core/geometry.ts
@@ -22,9 +22,9 @@ export const GeometryAttributeIndex: Record<GeometryAttributeType, number> = {
   next_position: 3,
   previous_position: 4,
   direction: 5,
-  color: 7,
-  size: 8,
-  marker: 9,
+  color: 6,
+  size: 7,
+  marker: 8,
 };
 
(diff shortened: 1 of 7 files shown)


shlomnissan (approved):
Nice refactor! Looking forward to WebGPU implementation!!

PR #670, opened 2026-05-22
feat: chunk frustum culling for volume rendering path

aganders3:
### Summary

This adds a view intersection test to the volume path. Though we still have separate paths for slice and volume rendering, this makes them more similar in structure.

For *volume*, the iteration bound is now `cameraFrustumWorldAabb`; precise intersection testing uses `Frustum.intersectsWithBox3`. Previously this path walked every chunk at its LOD via `iterateAllChunksAtLod` — that helper is now unused and thus removed. Volume rendering does not include spatial prefetching, because it's both more complicated and more costly than in the 2D case.

Volume time-prefetch uses the same path and is now bounded (spatially) the same way as current-time.

This PR also generalizes the slice-axis helpers for future work enabling slicing in other orientations: `sliceChunkIndex(axis, lod, position)` replaces `getZBounds`'s world-space slab math with an integer chunk lookup, and `sliceSlabBounds(axis, sliceIdx)` gives the world extent of one chunk on a given axis.

Likewise this removes `lastViewBounds2D_` and uses `lastViewProjection_`/`hasViewProjectionChanged` for both paths.

### Priority assignment summary

Spatial regions used in the code:
- **`cameraFrustum`** — the camera's actual truncated-pyramid frustum.
- **`cameraFrustumWorldAabb`** — the world-axis-aligned bounding box of the 8 frustum corners. Used as the spatial-iteration bound to limit number of intersection tests. This may be a pretty loose bound for a perspective camera, especially with a large `far` value, so I'm not sure how much performance improvement we will see in practice.

Image (slice) path uses `viewBounds3D` and `prefetchAabb`. The volume path has no equivalent of `prefetchAabb` yet. A future volume spatial-prefetch might naturally be an inflated frustum, but is harder to describe in "number of chunks" that the loading policy currently captures.

#### Priority for volume current-time chunks

| Chunk position | Iterated? | Resulting priority |
|---|---|---|
| Inside `cameraFrustum`, current LOD | Yes | `visibleCurrent` (or `fallbackVisible` if `currentLOD === fallbackLOD`) |
| Inside `cameraFrustum`, fallback LOD (distinct from current) | Yes | `fallbackVisible` |
| Inside `cameraFrustumWorldAabb` but outside `cameraFrustum` | Yes | **evicted** *(skipped by precise frustum filter)* |
| Outside `cameraFrustumWorldAabb` | No | **evicted** |

#### Priority for volume temporal prefetch (fallback LOD only)

| Chunk position | Iterated? | Resulting priority |
|---|---|---|
| Inside `cameraFrustum` | Yes | `prefetchTime` |
| Inside `cameraFrustumWorldAabb` but outside `cameraFrustum` | Yes | **evicted** *(skipped by precise frustum filter)* |
| Outside `cameraFrustumWorldAabb` | No | **evicted** |

"Evicted" means: the chunk's entry in `chunkViewStates_` is left at the reset state `{ priority: null }`. Aggregation across views then sets `chunk.priority = null` and the chunk's data is freed.

### Related issue
I think this may replace #628, but we should re-evaluate as that PR points out a number of performance issues.

### Tests & Checks
Tested manually with the multi-viewport example that displays a volume and 2D slice simultaneously. With these changes you can see logs indicating the chunk manager loading or discarding chunks as you pan the images in/out of the frame. On `main` chunks stay loaded even with cameras panned such that nothing is visible.

### DIFF
--- a/src/core/chunk_store.ts
+++ b/src/core/chunk_store.ts
@@ -1,5 +1,6 @@
 import { Chunk, SourceDimensionMap, ChunkLoader } from "../data/chunk";
 import { almostEqual } from "../utilities/almost_equal";
+import { clamp } from "../utilities/clamp";
 import { Logger } from "../utilities/logger";
 import { ChunkStoreView } from "./chunk_store_view";
 import { ImageSourcePolicy } from "./image_source_policy";
@@ -103,6 +104,23 @@ export class ChunkStore {
 
   public hasChunksAtTime(timeIndex: number): boolean {
     return this.chunks_[0]?.[timeIndex] !== undefined;
+  }
+
+  // For axis-aligned slicing: which chunk on `axis` at this LOD contains the
+  // given world-space `position`. Returns the clamped chunk index (always
+  // valid). Skips slab-arithmetic dance — caller gets the integer directly.
+  public sliceChunkIndex(
+    axis: "x" | "y" | "z",
+    lod: number,
+    position: number
+  ): number {
+    const dim = this.dimensions_[axis];
+    if (dim === undefined) return 0;
+    const ld = dim.lods[lod];
+    const voxelIdx = Math.floor((position - ld.translation) / ld.scale);
+    const chunkIdx = Math.floor(voxelIdx / ld.chunkSize);
+    const chunkCount = Math.ceil(ld.size / ld.chunkSize);
+    return clamp(chunkIdx, 0, chunkCount - 1);
   }
 
   public get lodCount() {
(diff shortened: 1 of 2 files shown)


shlomnissan:
@aganders3 is this PR still relevant? if yes, should we resolve merge conflicts?

aganders3:
It's stale and low priority at the moment. I'll close and reconsider once some of the other work on chunk residency, budgets, etc. are merged.

PR #671, opened 2026-05-27
chore: remove unused `PromiseScheduler` and `PromiseQueue`

aganders3:
## Summary
- Remove `PromiseScheduler` and `AbortError` (`src/data/promise_scheduler.ts`) along with its test file
- Remove `PromiseQueue` from `src/data/ome_zarr/image_loader.ts`

These were only used by the old image layer that did its own chunk loading. All image chunk loading now goes through `ChunkManager`, so these classes are dead code.

## Test plan
- [x] `npm run compile` passes
- [x] `npm run lint` passes
- [x] `npm test` passes (119 tests)


shlomnissan (approved):
Nice!

PR #675, opened 2026-06-05
refactor: fold layer manager into the viewport

shlomnissan:
### Summary

this pr removes the layer manager and folds layer ownership into the viewport.
the viewport now exposes layers and `partitionLayers` is inlined. it also drops
the unused `addLayersChangeCallback` system and the redundant
`removeByIndex` method along the way. 

### Tests & Checks

- all checks have passed.


aganders3 (approved):
Nice

PR #678, opened 2026-06-05
refactor: use fixed marker atlas in `Points` renderable, remove public `Texture2DArray` export

aganders3:
### Summary
Points no longer takes a runtime `Texture2DArray` marker atlas. Instead it exposes a fixed set of marker shapes via a `Marker = "circle" | "square" | "triangle"` type, mapped to internal indices and backed by a procedurally generated atlas shared across all instances. This lets us drop `Texture2DArray` from the public API.

Point color now accepts `ColorLike`, matching the rest of the public API.

### Related Issue
Closes N/A

### Tests & Checks
Tested the points example and it looks the same
All other tests pass

### DIFF
--- a/examples/points/main.ts
+++ b/examples/points/main.ts
@@ -6,7 +6,6 @@ import {
   Points,
   OrthographicCamera,
   OmeZarrImageSource,
-  Texture2DArray,
   LayerState,
   ImageLayer,
   Color,
@@ -60,8 +59,7 @@ class Particles extends Layer {
   public readonly type = "Particles";
   private readonly points_: vec3[] = [];
   private readonly color_: Color;
-  private readonly markerIndex_: number = 0;
-  private static markerAtlas_: Texture2DArray;
+  private readonly marker_: Marker;
   private needsUpdate_: boolean = true;
   private depth_: number = 0;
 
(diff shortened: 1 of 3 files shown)


shlomnissan (approved):
thanks!

overall i think this change is good because it removes the texture array but i'm not sure about the point renderable and this example in general.

a few non-blocking questions i have:

1. should we add a points layer to core?
2. should the point renderable generate sprites? users can provide images?
3. what's the primary difference between points/labels? is it just the data source?
4. since it's not actively used, should we get rid of the point renderable in first release?
5. why is `point_picking.ts` in the layers directory?

aganders3:
> 1. should we add a points layer to core?

I don't have a strong opinion about this. It would be pretty easy to make a basic "Points" layer type part of core. I think adding a chunked multiscale version would be better, but I'd want to explore what formats people would expect to use. Maybe we can figure out what they're using on the atlas project?

> 2. should the point renderable generate sprites? users can provide images?

Sure, I think adding sprites would be a reasonable enhancement. For the API we'd need to consider how to handle the `color` attribute in the case of an RGB sprite.

> 3. what's the primary difference between points/labels? is it just the data source?

Points are single (point) locations in a continuous space, where labels are rasterized pixel labels. Some use cases may overlap, but they're typically distinct annotation types. I also think points could be used to display embeddings.

> 4. since it's not actively used, should we get rid of the point renderable in first release?

I understand this desire, but I'd rather include it. I think our only actively used layers by this definition are the image and volume layers. We have shipped apps using the Points renderable, and I think Utz is probably using it in the copick web prototype. I'm still pretty new to JS - is there a JS convention for including "private"/unstable API?

> 5. why is point_picking.ts in the layers directory?

I think we can remove this - I hadn't noticed it.

PR #679, opened 2026-06-08
refactor: expose memory stats accessor on idetik

shlomnissan:
### Summary

add a read-only `memoryStats` accessor on idetik for baseline memory
instrumentation ahead of experimenting with shared residency.

the accessor reports:
- cpu-resident chunk bytes + chunk count
- gpu texture bytes + texture count
- js heap usage (chromium `performance.memory`, optional)

### Tests & Checks

- all checks have passed.


srivarra (approved):
tysm!

PR #681, opened 2026-06-10
feat: add async `OmeZarrImageSource` API that pre-opens the loader

### LINKED ISSUE #562: Possible bug in the chunk manager
@aganders3 Indirectly related to this PR. I think there might be a bug in the chunk manager:

`getOrCreateStore` doesn't seem to manage failure. I think that if `source.open()` rejects, `pendingStores_` keeps a rejected promise forever so future calls keep failing and we never retry. We should clear `pendingStores_` in a finally and log the error.

_Originally posted by @shlomnissan in https://github.com/chanzuckerberg/idetik/pull/553#discussion_r2662477210_

aganders3:
### Summary
Opening a source and getting basic information about it is an unnecessarily clunky process: `OmeZarrImageSource.fromHttp({ url })` (sync), then `source.open()` (async), then get dimensions/channel count out of the returned loader (`loader.getSourceDimensionMap()`).

This PR changes that to make `OmeZarrImageSource.fromHttp` and `OmeZarrImageSource.fromFilesystem` async factories that also open the source. This also adds `getDimensions` and `getChannelCount` on the source, exposing information from the now-opened loader. In the future we can refactor further to possibly consolidate the source/loader, but I think this API should be more ergonomic and doesn't need changes with that potential refactor.

I also added the try/except/finally block to fix #562 since it's somewhat related. If a source failed to open, `ChunkManager` cached the rejected promise forever, so it never retried. A follow-up PR will simplify this code further, but it's worth this quick correctness fix here.

### Related Issue
Closes #562

### Tests & Checks
Tested existing examples, and they all work as expected

### DIFF
--- a/src/data/chunk_manager.ts
+++ b/src/data/chunk_manager.ts
@@ -9,6 +9,7 @@ export type QueueStats = {
 import { ChunkStore } from "./chunk_store";
 import { ChunkStoreView } from "./chunk_store_view";
 import { ImageSourcePolicy } from "../core/image_source_policy";
+import { Logger } from "../utilities/logger";
 
 export class ChunkManager {
   private readonly stores_ = new Map<ChunkSource, ChunkStore>();
@@ -54,11 +55,19 @@ export class ChunkManager {
     const newPending = initializeStore();
     this.pendingStores_.set(source, newPending);
 
-    const store = await newPending;
-    this.stores_.set(source, store);
-    this.pendingStores_.delete(source);
-
-    return store;
+    try {
+      const store = await newPending;
+      this.stores_.set(source, store);
+      return store;
+    } catch (error) {
+      Logger.error(
+        "ChunkManager",
+        `Failed to open chunk source: ${String(error)}`
+      );
+      throw error;
+    } finally {
+      this.pendingStores_.delete(source);
+    }
   }
 
   public update() {
--- a/src/data/ome_zarr/image_source.ts
+++ b/src/data/ome_zarr/image_source.ts
@@ -111,7 +111,7 @@ export class OmeZarrImageSource {
   }
 
   /**
-   * Creates an OmeZarrImageSource from a local filesystem directory.
+   * Creates and opens an OmeZarrImageSource from a local filesystem directory.
    *
    * @param directory return value of `window.showDirectoryPicker()` which gives the browser
    *    permission to access a directory (only works in Chrome/Edge)
--- a/src/data/ome_zarr/image_source.ts
+++ b/src/data/ome_zarr/image_source.ts
@@ -11,6 +11,7 @@ import {
   parseOmeZarrImage,
   Version as OmeZarrVersion,
 } from "./metadata_loaders";
+import { SourceDimensionMap } from "../chunk";
 
 type OmeZarrImageSourceProps = {
   location: Location<Readable>;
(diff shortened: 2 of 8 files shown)


src/data/ome_zarr/image_source.ts
```diff
@@ -33,12 +34,16 @@ export class OmeZarrImageSource {
...
 
+  private loader_?: OmeZarrImageLoader;
+
   private constructor(props: OmeZarrImageSourceProps) {
     this.location = props.location;
     this.version = props.version;
   }
 
   public async open(): Promise<OmeZarrImageLoader> {
+    if (this.loader_) return this.loader_;
```
shlomnissan, line 45: we memoize the loader itself not the promise so two concurrent opens would each run the full open sequence. it's not a big deal but i think memoizing the promise is safer.

src/data/ome_zarr/image_source.ts
```diff
@@ -33,12 +34,16 @@ export class OmeZarrImageSource {
...
   readonly version?: OmeZarrVersion;
 
+  private loader_?: OmeZarrImageLoader;
+
   private constructor(props: OmeZarrImageSourceProps) {
     this.location = props.location;
     this.version = props.version;
   }
 
   public async open(): Promise<OmeZarrImageLoader> {
```
shlomnissan, line 44: what's the point of `open()` after this change? i know it's part of the chunk source contract but i think image source is the only implementer? can we just remove it?

src/data/ome_zarr/image_source.ts
```diff
@@ -69,25 +74,40 @@ export class OmeZarrImageSource {
...
+    return this.loader_.getSourceDimensionMap();
+  }
+
+  public getChannelCount(): number {
+    return this.getDimensions().c?.lods[0].size ?? 1;
   }
 
   /**
-   * Creates an OmeZarrImageSource from an HTTP(S) URL.
+   * Creates and opens an OmeZarrImageSource from an HTTP(S) URL.
```
shlomnissan, line 96: i think we need to update the comment for `fromFileSystem` as well.

src/data/chunk_manager.ts
```diff
@@ -54,11 +55,19 @@ export class ChunkManager {
     const newPending = initializeStore();
     this.pendingStores_.set(source, newPending);
 
-    const store = await newPending;
-    this.stores_.set(source, store);
-    this.pendingStores_.delete(source);
-
-    return store;
+    try {
```
shlomnissan, line 58: not blocking: retry is the right fix but i think this rejection propagates to `layer.onAttached` which `Viewport.addLayer` fire-and-forgets so each retry now surfaces as a fresh unhandled rejection + log line, where the old cached rejection failed once. worth giving attach failures a real error path in a follow-up.

shlomnissan (approved):
looks good. it would be nice to get rid of `open()` if we can but not blocking.

aganders3:
Good eye - `open()` can go away, but is still used internally (though it'll generally just hit the cached path. It just keeps this PR a little smaller to leave it at the moment.

src/data/ome_zarr/image_source.ts
aganders3: Good point - I think this code will change in a follow-up cleanup, and this is simpler for now.

PR #682, opened 2026-06-12
feat: move z-slicing to the gpu

shlomnissan:
### Summary

move 2D image and label rendering to gpu-side z-slicing. instead of slicing
the chunk on the cpu into a 2D texture and re-uploading on every z change, we
now upload the whole chunk once as a 3D texture and select the slice in the
shader via a normalized z coordinate. scrubbing z becomes a uniform update
with no texture traffic.

chunk data stays cpu-resident, so picking is unchanged. also fixes the shader
define injection, which previously produced a malformed line and only
worked by accident.

### Tests & Checks

- all tests have passed


aganders3:
What do you think about encoding both slice and axis as a `mat3` instead of a hard-coded Z? The shaders here bake in Z-slicing by building `vec3(TexCoords, ZTexCoord)`. If we replace it we can use `sliceTransform * vec3(uv, 1)`. That keeps the shaders orientation-agnostic so it's easier to add XZ, YZ, and eventually oblique slices without shader changes.

We still don't support additional slice orientation, but I'd rather be moving in the direction of generalizing that rather than further codifying Z as the slice axis.

Also we should consider how to de-duplicate VRAM usage across multiple layers. Our dedup today is only at the CPU/`ChunkStore` level. This isn't a new problem (two volume layers take double VRAM) but this change will significantly increase VRAM usage for image layers. I know this is part of a series of changes you have planned, so I'm happy to address that later.

shlomnissan:
@aganders3 

> What do you think about encoding both slice and axis as a mat3 instead of a hard-coded Z?

ah i didn't even think about it. my instinct is that it probably doesn't matter at this point. it's a one-line change in the shader which doesn't unlock different slice orientation. i think it makes more sense to make this change in a cleanup pr that removes baked slice orientation assumptions across the app.

> I know this is part of a series of changes you have planned

yeah so this will come as part of "transfer data to gpu as soon as it's available." the way i implemented the prototype is by moving texture ownership to the chunk and have the chunk manager upload the data. then the renderable takes a reference to an existing texture that's already on the gpu. this approach will resolve deduplication.

src/layers/image_layer.ts
```diff
@@ -117,7 +117,14 @@ export class ImageLayer extends Layer implements ChannelsEnabled {
...
-    this.resliceIfZChanged();
+
+    const zPointWorld = this.sliceCoords_.z;
+    if (zPointWorld !== undefined && this.zPrevPointWorld_ !== zPointWorld) {
+      for (const [chunk, image] of this.visibleChunks_) {
+        image.zTexCoord = this.zTexCoordForChunk(chunk);
+      }
+      this.zPrevPointWorld_ = zPointWorld;
+    }
   }
```
aganders3: Can we just skip this change detection check and remove `zPrevPointWorld_` since we're not really doing any CPU work anymore when z changes? We still need the loop but it should be quite cheap now as it's just setting the value for the uniform.

src/renderers/webgpu/webgpu_renderer.ts
```diff
@@ -412,6 +412,7 @@ class WebGPURenderer extends Renderer {
       valueOffset: uniforms.ValueOffset,
       valueScale: uniforms.ValueScale,
       opacity: this.currentLayerOpacity_ * objectOpacity,
+      zTexCoord: uniforms.ZTexCoord,
```
aganders3, line 415: Unrelated, but is there a convention of capitalizing uniforms in WebGL? Should we just make these consistent? (not asking for it in this PR)

aganders3 (approved):
Cool to see slicing on the GPU! This is less code than I was expecting.

src/renderers/webgpu/webgpu_renderer.ts
shlomnissan: it's not a convention, i don't know that there is a webgl standard convention. in most of my shader code i typically use `u_camelCase` but then other uniforms this codebase are capitalized, and some are lower case, so yeah we're not consistent. i'll open another pr to clean this up later.

src/layers/image_layer.ts
shlomnissan: ahh good catch!

PR #683, opened 2026-06-16
refactor!: make the layer/chunk data path synchronous

### LINKED ISSUE #399: `LayerManager.add` hides async behavior

aganders3:
### Summary
Adding a layer to a `Viewport` (previously `LayerManager`) was hiding async behavior: `addLayer` wasn't async but called the async `onAttached`. In several layers `onAttached` awaited `ChunkManager.addView`, which in turn awaited `source.open()`. That hidden async caused a race — two layers sharing a source could both miss the not-yet-resolved store and each create their own, defeating the whole point of keying stores by source.

This makes the data path synchronous end-to-end:
- Since #681, `OmeZarrImageSource` opens its loader in the `fromHttp`/`fromFileSystem` factories. This PR follows up, removing the lazy `open()` pathway so the loader is now non-null. Opening a source now _always_ happens up front in a factory so failures surface at construction instead of at attach.
- `ChunkManager.addView` is now synchronous: look up or create the store, return a fresh view. The `pendingStores_` promise-cache and its `getOrCreateStore` dance existed only to guard the race the async-ness created, so they're deleted.
- Layer lifecycle hooks `onAttach`/`onDetach` are now synchronous, which is fine for how we use them, especially given the comment in `viewport.ts` that they may go away. I think this is a step in that direction if we want to take it.

BREAKING CHANGE: `ChunkSource.open()` is removed; use `getLoader()` (sources open eagerly via `fromHttp`/`fromFileSystem`).

### Related Issue
Follow-up for #681 
Closes #399 

### Tests & Checks
Existing automated tests pass, manually tested examples

### DIFF
--- a/src/core/layer.ts
+++ b/src/core/layer.ts
@@ -61,10 +61,8 @@ export abstract class Layer {
 
   public onEvent(_: EventContext): void {}
 
-  // called when a layer is added to a viewport
   public onAttached(_context: IdetikContext): void {}
 
-  // called when a layer is removed from a viewport
   public onDetached(_context: IdetikContext): void {}
 
   public get objects() {
--- a/src/core/layer.ts
+++ b/src/core/layer.ts
@@ -61,11 +61,11 @@ export abstract class Layer {
 
   public onEvent(_: EventContext): void {}
 
-  // Synchronous attach/detach. The source must be opened before a layer is attached
-  // (see ChunkManager.viewFor) so creating the layer's view needs no async step.
-  public attach(_context: IdetikContext): void {}
+  // called when a layer is added to a viewport
+  public onAttached(_context: IdetikContext): void {}
 
-  public detach(): void {}
+  // called when a layer is removed from a viewport
+  public onDetached(_context: IdetikContext): void {}
 
   public get objects() {
     return this.objects_;
--- a/src/core/layer.ts
+++ b/src/core/layer.ts
@@ -61,13 +61,11 @@ export abstract class Layer {
 
   public onEvent(_: EventContext): void {}
 
-  // TODO: Consider making this an abstract method once chunk manager
-  // integration is finalized. Most layers will likely need access to the chunk
-  // manager, but for now, we allow optional overrides to avoid requiring
-  // placeholder implementations.
-  public async onAttached(_context: IdetikContext) {}
+  // Synchronous attach/detach. The source must be opened before a layer is attached
+  // (see ChunkManager.viewFor) so creating the layer's view needs no async step.
+  public attach(_context: IdetikContext): void {}
 
-  public onDetached(_context: IdetikContext): void {}
+  public detach(): void {}
 
   public get objects() {
     return this.objects_;
--- a/src/data/chunk_manager.ts
+++ b/src/data/chunk_manager.ts
@@ -25,11 +25,8 @@ export class ChunkManager {
     return chunkMemoryStats();
   }
 
-  /**
-   * Creates a view for a source. The source is opened by its factory
-   * (`OmeZarrImageSource.fromHttp(...)` / `fromFileSystem(...)`), so this is synchronous.
-   */
-  public viewFor(
+  /** Adds a view onto `source`'s store, creating the store on first use. */
+  public addView(
     source: ChunkSource,
     policy: ImageSourcePolicy
   ): ChunkStoreView {
@@ -38,7 +35,7 @@ export class ChunkManager {
       store = new ChunkStore(source.getLoader());
       this.stores_.set(source, store);
     }
-    return store.createView(policy);
+    return store.addView(policy);
   }
 
   public update() {
--- a/src/data/chunk_manager.ts
+++ b/src/data/chunk_manager.ts
@@ -25,7 +25,6 @@ export class ChunkManager {
     return chunkMemoryStats();
   }
 
-  /** Adds a view onto `source`'s store, creating the store on first use. */
   public addView(
     source: ChunkSource,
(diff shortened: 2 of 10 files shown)


src/core/layer.ts
```diff
@@ -61,12 +61,10 @@ export abstract class Layer {
 
   public onEvent(_: EventContext): void {}
 
-  // TODO: Consider making this an abstract method once chunk manager
-  // integration is finalized. Most layers will likely need access to the chunk
-  // manager, but for now, we allow optional overrides to avoid requiring
-  // placeholder implementations.
-  public async onAttached(_context: IdetikContext) {}
+  // called when a layer is added to a viewport
```
shlomnissan: if this isn't obvious from the context we should change the method's name to make it explicit instead of adding comments. i would just drop the comments.

src/data/ome_zarr/image_source.ts
```diff
@@ -34,18 +35,25 @@ export class OmeZarrImageSource {
...
   readonly version?: OmeZarrVersion;
 
-  private loader_?: OmeZarrImageLoader;
+  private readonly loader_: OmeZarrImageLoader;
 
+  /**
+   * Private: construct via a static factory (`fromHttp`/`fromFileSystem`). Factories open
+   * the loader up front and pass it in, so a source's loader is non-null.
+   * errors opening a source surface early, and `getLoader`/`getDimensions` always succeed
+   */
```
shlomnissan: this feels more like a change summary than a useful inline comment.

src/data/ome_zarr/image_source.ts
```diff
@@ -101,13 +106,9 @@ export class OmeZarrImageSource {
...
     props: HttpOmeZarrImageSourceProps
   ): Promise<OmeZarrImageSource> {
-    const store = new FetchStore(props.url);
-    const source = new OmeZarrImageSource({
-      location: new Location(store),
-      version: props.version,
-    });
-    await source.open();
-    return source;
+    const location = new Location(new FetchStore(props.url));
```
shlomnissan, line 104: do we currently use the `fromFileSystem` factory? if not, i suggest we drop it and rename this function.

src/data/chunk_manager.ts
```diff
@@ -27,47 +25,17 @@ export class ChunkManager {
     return chunkMemoryStats();
   }
 
-  public async addView(
+  /** Adds a view onto `source`'s store, creating the store on first use. */
```
shlomnissan: are these comments generated by claude?

src/layers/image_layer.ts
```diff
@@ -73,13 +73,13 @@ export class ImageLayer extends Layer implements ChannelsEnabled {
     this.onPickValue_ = onPickValue;
   }
 
-  public async onAttached(context: IdetikContext) {
+  public onAttached(context: IdetikContext): void {
```
shlomnissan: - do we need to specify `void` return value? this was added to all three layers.
- this check doesn't belong in each subclass. it's an attachment-lifecycle invariant and that lifecycle is owned by the base layer class not the type, i would move it there:

```ts
// src/core/layer.ts

export abstract class Layer {
  private attached_ = false;

  public onAttached(context: IdetikContext): void {
    if (this.attached_) {
      throw new Error(
        `${this.type} cannot be attached to multiple viewports simultaneously.`
      );
    }
    this.attached_ = true;
    this.attach(context);
  }

  public onDetached(context: IdetikContext): void {
    if (!this.attached_) return;
    this.detach(context);
    this.attached_ = false;
  }

  protected attach(_context: IdetikContext): void {}
  protected detach(_context: IdetikContext): void {}
}
```

sub-classes call `attach` and `detach` with the existing logic of `onAttached` and `onDetached` minus the guard.

shlomnissan (approved):
nice change 👍 

i would like to see the layer base class refactor before it goes in and maybe removing the auto-generated comments. otherwise, looks good!

src/data/chunk_manager.ts
aganders3: Yeah, Claude loves comments because I think it helps the model work correctly. I reviewed and deleted a bunch but I like your suggestions here to remove even more.

src/layers/image_layer.ts
aganders3: I guess `void` is the default return type? I'm still not great with the gradual typing stuff. Removed.

aganders3:
Thanks!

I think we can remove the web filesystem stuff. It was used in the SABER app, but it's easy enough to add back when we need it. I started removing here but pulling the thread led to a lot of changes so I'll open new PR for that.

I also agree wish codifying the 1:1 layer:viewport guard at the base class level, but I'm going to do that in a separate PR as well.

aganders3:
Oh oops I just saw this part of your comment:
> I would like to see the layer base class refactor before it goes in

I added #687 so I'll make a quick follow-up PR for that shortly. Sorry!

PR #686, opened 2026-06-17
feat: read texture pixel for gpu picking

shlomnissan:
### Summary

move image and label picking to gpu-side readback. instead of reading the
value under the cursor from cpu-resident chunk data, we now read the single
texel back from the gpu texture.

this removes the last reader of `chunk.data` so nothing outside the upload path
touches it anymore.

### Tests & Checks

- all checks have passed.
- visual verification.

https://github.com/user-attachments/assets/1f888d1d-ba42-40b3-be36-7604ae1c4f61


aganders3 (approved):
I still have some questions, but mostly due to just my lack of experience with framebuffers, readpixels, and WebGPU in general.

I'm happy to approve to unblock you and because I think GPU picking is in general a good idea. Any issues I'm likely to miss anyway, and surely can be fixed or reverted if needed.

PR #689, opened 2026-06-18
ci: keep breaking changes in the 0.x range (breaking -> minor)

aganders3:
### Summary
I accidentally released v1.0.0 with #683 - sorry!

This will prevent accidentally bumping the major version again for the time being.

Merging this should also trigger a release with the minor version bumped appropriately.


shlomnissan (approved):
:+1:

PR #690, opened 2026-06-18
feat: add base-class guard for multiple layer attachment

### LINKED ISSUE #687: Add guard for 1:1 layer:viewport in layer base class
- this check doesn't belong in each subclass. it's an attachment-lifecycle invariant and that lifecycle is owned by the base layer class not the type, i would move it there:

```ts
// src/core/layer.ts

export abstract class Layer {
  private attached_ = false;

  public onAttached(context: IdetikContext): void {
    if (this.attached_) {
      throw new Error(
        `${this.type} cannot be attached to multiple viewports simultaneously.`
      );
    }
    this.attached_ = true;
    this.attach(context);
  }

  public onDetached(context: IdetikContext): void {
    if (!this.attached_) return;
    this.detach(context);
    this.attached_ = false;
  }

  protected attach(_context: IdetikContext): void {}
  protected detach(_context: IdetikContext): void {}
}
```

sub-classes call `attach` and `detach` with the existing logic of `onAttached` and `onDetached` minus the guard.

_Originally posted by @shlomnissan in https://github.com/chanzuckerberg/idetik/pull/683#discussion_r3437275176_

aganders3:
### Summary
The "a layer can't be attached to two viewports at once" check was duplicated across `ImageLayer`, `LabelLayer`, and `VolumeLayer` and relied on the presence of `layer.chunkStoreView_` as a proxy for attachment. It's an attachment-lifecycle invariant, so this moves it into the `Layer` base class. This removes any distinction between layers that _can_ be attached to multiple viewports, and layers that _can't_.

Two notes:
1. `Viewport.addLayer` now calls `onAttached` _before_ registering the layer, so a layer that fails to attach isn't left in `viewport.layers_`.
2. `attached_` is set only after `attach()` succeeds, so a failed attach leaves the layer unattached.

An alternative; we could instead have layers keep track of _which_ viewport they're attached to (if any) instead of having "attached_" and then needing to also pass in the viewport on `update`. Moving from one viewport to another would then be like a reparenting operation.

Note that there can be a resource leak if a layer ends up partially initialized but throws during `onAttached`. For example the layers that use the chunk store now can throw on channel validation _after_ getting a `ChunkStoreView`. For now I think this will just be up to the layer to clean up (that is: make `attach()` atomic). This is pre-existing behavior, so I can address in a follow-up.

### Related Issue
Closes #687
Follow-up from https://github.com/chanzuckerberg/idetik/pull/683#discussion_r3437275176.

### Tests & Checks
All examples still work, no changes needed
Existing tests pass
A few new tests added to check the behavior

### DIFF
--- a/src/core/layer.ts
+++ b/src/core/layer.ts
@@ -33,6 +33,7 @@ export abstract class Layer {
 
   private objects_: RenderableObject[] = [];
   private state_: LayerState = "initialized";
+  private attached_ = false;
   private readonly callbacks_: StateChangeCallback[] = [];
 
   private opacity_: number;
@@ -61,9 +62,25 @@ export abstract class Layer {
 
   public onEvent(_: EventContext): void {}
 
-  public onAttached(_context: IdetikContext): void {}
+  public onAttached(context: IdetikContext): void {
+    if (this.attached_) {
+      throw new Error(
+        `${this.type} cannot be attached to multiple viewports simultaneously.`
+      );
+    }
+    this.attach(context);
+    this.attached_ = true;
+  }
 
-  public onDetached(_context: IdetikContext): void {}
+  public onDetached(context: IdetikContext): void {
+    if (!this.attached_) return;
+    this.detach(context);
+    this.attached_ = false;
+  }
+
+  protected attach(_context: IdetikContext): void {}
+
+  protected detach(_context: IdetikContext): void {}
 
   public get objects() {
     return this.objects_;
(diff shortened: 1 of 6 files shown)


shlomnissan:
> we could instead have layers keep track of which viewport they're attached

i'm not sure it's a good idea because it inverts ownership. can layers do things to their viewport? and now the layer isn't a passive renderable unit anymore. also transient context can't go stale. if we add a field we need to maintain it.

shlomnissan (approved):
LGTM! I responded to your PR summary note :+1:

PR #692, opened 2026-06-19
refactor: add texture upload/dispose api to renderer

shlomnissan:
### Summary

adds abstract `uploadTexture` and `disposeTexture` to the renderer base and
implements them in both backends. this is going to be used for uploading textures
directly to the gpu without having to do it lazily through the existing render path

### Tests & Checks

- all checks have passed.


aganders3 (approved):
Looks good to me.

Do we even need to override these on the renderers? I guess it makes sense rather than creating an interface for `WebGLTextures` and `WebGPUTexturePool`.

shlomnissan:
> I guess it makes sense rather than creating an interface for WebGLTextures and WebGPUTexturePool

yeah, i thought this approach was simpler

PR #693, opened 2026-06-19
refactor: decouple chunk store from chunk loader

shlomnissan:
### Summary

the chunk store now takes a source dimension map instead of a chunk loader,
and load orchestration moves entirely to the chunk manager via the source
loader getter. drops the redundant load-data pass-through on the store and
swaps the chunk manager store registry from a map to a small array lookup.
the chunk store view test now passes dimensions instead of a mock loader.

### Tests & Checks

- all checks have passed.

### DIFF
--- a/src/data/chunk.ts
+++ b/src/data/chunk.ts
@@ -99,7 +99,7 @@ export type SliceCoordinates = {
 };
 
 export type ChunkSource = {
-  getLoader(): ChunkLoader;
+  get loader(): ChunkLoader;
 };
 
 export type ChunkLoader = {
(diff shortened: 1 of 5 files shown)


aganders3 (approved):
LGTM

Out of curiosity what motivates the switch from `Map` to `{ source: ChunkSource; store: ChunkStore }[]`? Is it for performance?

shlomnissan:
> Out of curiosity what motivates the switch

yeah a map for a handful of entries isn't worth the overhead. in this case it's probably negligible but avoiding maps in runtime loops is typically a good idea.

PR #694, opened 2026-06-22
refactor: decouple chunk store from viewport

shlomnissan:
### Summary

the chunk store only ever needed a handful of view values to pick chunks, but
it took the entire viewport which coupled the data layer to the rendering layer.
this PR passes the necessary values instead of the viewport. each value now
comes from a function on its natural owner.

while here i inlined the one-field `RenderContext` wrapper.

### Tests & Checks

- all checks have passed.


aganders3 (approved):
Nice!

PR #698, opened 2026-06-23
feat: eagerly upload chunk textures to the GPU

shlomnissan:
### Summary

this pr changes when chunk textures are sent to the gpu. instead of uploading a
chunk lazily the first time it is drawn, each chunk's texture is now uploaded as
soon as its data finishes loading and that gpu texture is owned centrally for
the chunk's whole lifetime rather than by individual layers. once a chunk is on
the gpu its cpu copy is released to save memory, and its texture is freed when
the chunk is no longer needed. the image, label, and volume layers now share
these textures instead of each building and tracking their own.

there's a minor temporary regression: because a chunk's texture is freed the
moment it is no longer needed scrubbing through time can briefly show gaps
where the previous frame used to stay on screen while the next one loaded.
that smoothness returns once eviction is deferred under a budget, keeping
recently used textures resident. this does not affect volume rendering.

this change touches quite a few files but most of the changes are minor.

### Tests & Checks

- all checks have passed

### DIFF
--- a/src/core/renderable_object.ts
+++ b/src/core/renderable_object.ts
@@ -20,11 +20,17 @@ export abstract class RenderableObject extends Node {
   private cullFaceMode_: CullingMode = "none";
 
   public setTexture(index: number, texture: Texture) {
-    const oldTexture = this.textures_[index];
-    if (oldTexture !== undefined) {
-      this.staleTextures_.push(oldTexture);
-    }
     this.textures_[index] = texture;
+  }
+
+  protected clearTextures() {
+    this.textures_.length = 0;
+  }
+
+  protected markStaleTexture(texture: Texture | undefined) {
+    if (texture !== undefined) {
+      this.staleTextures_.push(texture);
+    }
   }
 
   public popStaleTextures() {
--- a/src/layers/image_layer.ts
+++ b/src/layers/image_layer.ts
@@ -11,13 +11,13 @@ import {
   validateChannelPropsCount,
 } from "../core/channel";
 import { ImageRenderable } from "../objects/renderable/image_renderable";
-import { Texture3D } from "../objects/textures/texture_3d";
 import { Color } from "../math/color";
 import { EventContext } from "../core/event_dispatcher";
 import { vec2, vec3 } from "gl-matrix";
 import { handlePointPickingEvent, PointPickingResult } from "./point_picking";
 import { clamp } from "../utilities/clamp";
 import { RenderablePool } from "../utilities/renderable_pool";
+import { Texture } from "../objects/textures/texture";
 
 export type ImageLayerProps = LayerOptions & {
   source: ChunkSource;
@@ -128,8 +128,13 @@ export class ImageLayer extends Layer implements ChannelsEnabled {
     if (!this.chunkStoreView_) return;
     if (this.state !== "ready") this.setState("ready");
 
+    const visibleChunksResident = Array.from(this.visibleChunks_.keys()).every(
+      (chunk) => chunk.texture !== undefined
+    );
+
     if (
       this.visibleChunks_.size > 0 &&
+      visibleChunksResident &&
       !this.chunkStoreView_.allVisibleFallbackLODLoaded() &&
       !this.isPresentationStale()
     ) {
@@ -147,8 +152,7 @@ export class ImageLayer extends Layer implements ChannelsEnabled {
 
     this.clearObjects();
     for (const chunk of orderedByLOD) {
-      if (chunk.state !== "loaded") continue;
-      const image = this.getImageForChunk(chunk);
+      const image = this.getImageForChunk(chunk, chunk.texture!);
       this.visibleChunks_.set(chunk, image);
       this.addObject(image);
     }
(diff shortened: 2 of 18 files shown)


src/core/renderable_object.ts
```diff
@@ -20,13 +20,19 @@ export abstract class RenderableObject extends Node {
...
-    const oldTexture = this.textures_[index];
-    if (oldTexture !== undefined) {
-      this.staleTextures_.push(oldTexture);
-    }
     this.textures_[index] = texture;
   }
 
+  protected clearTextures() {
+    this.textures_.length = 0;
+  }
```
aganders3, line 28: Should this mark all textures stale before clearing? Is the caller responsible for that?

src/layers/image_layer.ts
```diff
@@ -147,8 +152,7 @@ export class ImageLayer extends Layer implements ChannelsEnabled {
 
     this.clearObjects();
     for (const chunk of orderedByLOD) {
-      if (chunk.state !== "loaded") continue;
-      const image = this.getImageForChunk(chunk);
+      const image = this.getImageForChunk(chunk, chunk.texture!);
```
aganders3, line 155: Is this `chunk.texture!` safe? This is iterating over a new list, not verified by the `visibleChunksResident` guard above.

Why did you change the signature to take `chunk.texture` separately, instead of accessing it from the chunk inside the method?

src/objects/renderable/label_image_renderable.ts
```diff
@@ -73,6 +73,8 @@ export class LabelImageRenderable extends RenderableObject {
   }
 
   public setColorMap(colorMap: LabelColorMap) {
+    this.markStaleTexture(this.textures[1]);
+    this.markStaleTexture(this.textures[2]);
```
aganders3, line 77: Why did you change this from `setTexture` automatically marking old textures stale? I'm not opposed to the change, just curious about the motivation.

src/objects/renderable/volume_renderable.ts
```diff
@@ -32,49 +32,50 @@ export class VolumeRenderable extends RenderableObject {
...
-    }
-
-    texture.updateWithChunk(chunk);
+  private updateChannelTexture(textureIndex: number, texture: Texture): void {
+    this.setTexture(textureIndex, texture);
   }
 
   public clearLoadedChannels() {
     this.loadedChannels_ = new Set();
   }
```
aganders3, line 71: nit: consider inlining this into `reset`

src/objects/renderable/volume_renderable.ts
```diff
@@ -32,49 +32,50 @@ export class VolumeRenderable extends RenderableObject {
...
 
   public updateVolumeWithChunk(chunk: Chunk): void {
+    if (!chunk.texture) return;
```
aganders3, line 35: Is it possible for this to silently early-return and never get called again, leaving the renderable with stale data?

Should this throw instead?

src/renderers/webgl_textures.ts
```diff
@@ -135,6 +135,7 @@ export class WebGLTextures {
     this.gl_.bindTexture(textureType, textureId);
     this.configureTextureParameters(texture, textureType);
     this.uploadTextureData(texture, info, textureType);
+    this.currentTexture_ = null;
```
aganders3, line 138: not blocking: I this this merged in #692, but the responsibilities of `bindTexture` and `uploadTexture` seem a bit mixed here. It could just be my confusion, or a naming thing. What do you think?

aganders3 (approved):
I left some questions, but nothing blocking and I think this looks good.

> there's a minor temporary regression: because a chunk's texture is freed the
moment it is no longer needed scrubbing through time can briefly show gaps
where the previous frame used to stay on screen while the next one loaded.
that smoothness returns once eviction is deferred under a budget, keeping
recently used textures resident. this does not affect volume rendering.

Does this not affect volume rendering, or is it that volume rendering behavior is unchanged? I think the volume layer never got the same delay feature.

shlomnissan:
> Does this not affect volume rendering, or is it that volume rendering behavior is unchanged? I think the volume layer never got the same delay feature

volume rendering behavior is unchanged. the volume layer discards the previous volumes as soon as it needs an update. with memory budget and caching this doesn't have to be the case.

src/core/renderable_object.ts
shlomnissan: - the stale texture path is used for renderable owned textures (e.g., label color colormap)
- chunk textures are owned by the chunk manager now which is responsible for releasing them
- this function is used specifically for clearing out texture references, not releasing underlying data
- i'm going to rename this function to `clearTextureReferences` so it's clearer

as discussed on zoom, there are better ways of creating a separation such that ownership of renderable-specific textures move up the hierarchy (e.g., the label renderable owns the label colormap). it's something to think about but i decided not to deal with it in this PR to limit its scope.

src/layers/image_layer.ts
shlomnissan: yes it is safe because `getChunksToRender()` skips chunks that have no texture assigned to them

src/objects/renderable/label_image_renderable.ts
shlomnissan: automatically marking old textures as stale makes assumptions about ownership. the chunk manager is now responsible for directly allocating and deallocating chunk textures. these textures should never be disposed lazily by the stale texture flow.

src/objects/renderable/volume_renderable.ts
shlomnissan: - this is not called for chunks that don't have a textures
- this precondition is guaranteed by `getChunksToRender()`
- the guard is there to make typescript happy. otherwise dead code
- throw seems like the wrong approach. if it ever happened it will self correct in the next frame

src/renderers/webgl_textures.ts
shlomnissan: can you say more about responsibilities being mixed?

- `bindTexture` is responsible for two things:
  - if the texture exists, bind it so it can be used by the renderer
  - if it doesn't exists, upload the data and then bind (currently used for renderable-owned textures)
- `uploadTexture` is responsible for taking `texture.data` and uploading it to the GPU

maybe the naming is a bit confusing because there's `WebGLTextures.bindTexture` and there's `gl.bindTexture` (the latter is used internally by `uploadTexture`)? or is it something else?

PR #699, opened 2026-06-27
docs: add TypeDoc API reference skeleton

aganders3:
### Summary
This uses TypeDoc to create an API reference. Most items remain undocumented, but the public API is given "groups" using JSDoc like `/** @group Layer Configuration */`. I don't know much about TypeDoc so config was all set up using Claude.

Currently these groups are:
* Runtime
* Data Loading
* Layers
* Layer Configuration
* Renderable Objects
* Cameras & Controls

I'm happy to iterate on these, but here's what it looks like:
Screenshot 2026-06-26 at 9 08 05 PM

### Tests & Checks
no changes to the code, examples still work, etc.

note: this builds on #660 so the target branch is that to help see the diff

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -20,9 +20,10 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
         with:
-          node-version: 20
+          node-version-file: '.nvmrc'
           cache: npm
       - run: npm ci
+      - run: npm run docs:api
       - run: npx vitepress build --base /idetik/
       - uses: actions/configure-pages@v4
       - uses: actions/upload-pages-artifact@v3
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,7 @@ dist-ssr
 
 # Docs
 .vitepress/cache
+docs/api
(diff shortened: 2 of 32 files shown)


shlomnissan:
@aganders3 are you still working on this PR? the vitepress PR is approved :)

aganders3:
Oh sorry, I think this is ready for review I just had it in draft so I could read it through once more. Again this is about structure, not intended to be final content.

There are a couple unrelated changes here, but they're fairly small. I updated the README and switching the "Getting Started" guide to a volume layer

.github/workflows/deploy-docs.yml
```diff
@@ -20,9 +20,10 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
         with:
-          node-version: 20
```
aganders3, line 23: I think this caused the previous build to fail :(

https://github.com/chanzuckerberg/idetik/actions/runs/28454996849/job/84327226438

shlomnissan (approved):
i skimmed through it and it looks good. i will probably review the generated docsite once it's merged and share some high level thoughts / make changes.

PR #700, opened 2026-06-30
feat: gate chunk uploads with memory budget

shlomnissan:
### Summary

this PR gates chunk uploads based on a memory budget. the runtime exposes
a memory budget param with a default value of 2GB. setting to`Infinity` means
no limits (possibly undesirable with deferred eviction). the actual gating is done
by the chunk manager prior to queue handoff.

this change is scoped to gating only. no changes to existing eviction logic.
deferred eviction will be implemented in the next PR.

### Tests & Checks

- all checks have passed
- manual verification (lower memory budget results in partial loading)

https://github.com/user-attachments/assets/c6b726b1-8d9e-4e3e-84fd-59bd8435e3c9

### DIFF
--- a/src/data/chunk_manager.ts
+++ b/src/data/chunk_manager.ts
@@ -20,13 +20,20 @@ export class ChunkManager {
 
   private readonly uploadTexture_?: (texture: Texture) => void;
   private readonly disposeTexture_?: (texture: Texture) => void;
+  private readonly memoryLimitBytes_: number;
 
   constructor(
     uploadTexture?: (texture: Texture) => void,
-    disposeTexture?: (texture: Texture) => void
+    disposeTexture?: (texture: Texture) => void,
+    memoryLimitBytes: number = Infinity
   ) {
     this.uploadTexture_ = uploadTexture;
     this.disposeTexture_ = disposeTexture;
+    this.memoryLimitBytes_ = memoryLimitBytes;
+  }
+
+  public get memoryLimitBytes(): number {
+    return this.memoryLimitBytes_;
   }
 
   public get queueStats(): QueueStats {
--- a/src/data/chunk_manager.ts
+++ b/src/data/chunk_manager.ts
@@ -20,15 +20,18 @@ export class ChunkManager {
 
   private readonly uploadTexture_?: (texture: Texture) => void;
   private readonly disposeTexture_?: (texture: Texture) => void;
+  private readonly getGpuResidentBytes_: () => number;
   private readonly memoryLimitBytes_: number;
 
   constructor(
     uploadTexture?: (texture: Texture) => void,
     disposeTexture?: (texture: Texture) => void,
+    getGpuResidentBytes: () => number = () => 0,
     memoryLimitBytes: number = Infinity
   ) {
     this.uploadTexture_ = uploadTexture;
     this.disposeTexture_ = disposeTexture;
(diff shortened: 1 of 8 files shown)


src/data/ome_zarr/image_loader.ts
```diff
@@ -101,6 +107,14 @@ export class OmeZarrImageLoader {
...
 }
 
+function bytesPerElementForDtype(dtype: zarr.DataType): number {
+  const bits = Number(/\d+/.exec(String(dtype))?.[0]);
+  if (!Number.isFinite(bits) || bits <= 0) {
+    throw new Error(`Cannot determine byte size for zarr dtype "${dtype}"`);
+  }
+  return bits / 8;
+}
+
```
aganders3, line 130: Looking at the [core data types](https://zarr-specs.readthedocs.io/en/latest/v3/data-types/index.html#core-data-types) this should pretty much work, except probably for `bool`. I'd consider just making a record of dtype -> bytes per element.

src/data/chunk_manager.ts
```diff
@@ -80,6 +91,31 @@ export class ChunkManager {
...
+  ) {
+    if (candidates.length === 0) return;
+
+    candidates.sort((a, b) => comparePriority(a.chunk, b.chunk));
+
+    let committedBytes = this.getGpuResidentBytes_();
+
+    for (const { source, chunk } of candidates) {
+      const bytes = this.chunkBytes(source, chunk);
+      if (committedBytes + bytes > this.memoryLimitBytes_) continue;
```
aganders3, line 105: What do you think of logging this case?

Just to confirm my understanding: this means a chunk will skip fetch but it could be (likely is) something like a prefetch chunk, not a currently visible chunk?

src/objects/textures/texture.ts
```diff
@@ -79,6 +79,26 @@ export function textureBytesPerChannel(texture: Texture) {
...
+  let total = 0;
+
+  for (let i = 0; i < levels; ++i) {
+    total += bytesPerTexel * depth * width * height;
+    width = Math.max(1, width >> 1);
+    height = Math.max(1, height >> 1);
+  }
+
+  return total;
+}
```
aganders3, line 103: I don't think we're ever using mipmaps anyway, but for a 3D texture should the depth should also be shifted for each level?

aganders3 (approved):
LGTM

src/data/ome_zarr/image_loader.ts
shlomnissan: good idea!

src/data/chunk_manager.ts
shlomnissan: this is a temporary state: (in-flight chunk data + chunk data in GPU) exceeds memory budget so no more downloads. it's the gate.

this is where the deferred eviction branch attempts to dispose "evictable" chunks in the next PR:

https://github.com/chanzuckerberg/idetik/pull/702/changes#diff-c7880ad704ac5f2d27c866d5dd37e5f66737e685a4c6f0cfb3a0934f783b5ec0R126

src/objects/textures/texture.ts
shlomnissan: good call! this function was already doing future work so might as well close the gap. i applied it specifically to 3D textures, mipmaps don't halve the depth of 2D texture arrays.

PR #702, opened 2026-07-02
feat: add deferred eviction by priority

shlomnissan:
### Summary

this pr adds deferred eviction by priority. chunks are evicted only under
memory pressure based on `memoryLimitMB`.

the stale frame regression introduced by #698 is resolved but still not
applied to volumes. i will add a fix for it in a subsequent pr.

@srivarra i'm going to post another pr after this one for circular
prefetching so it's probably better for range prefetching to come after that.

### Tests & Checks

- all checks have passed
- manual verification (video shows chunks of previous frames are not evicted by default)

https://github.com/user-attachments/assets/92e5de01-0fba-44f0-8a60-9a52025528ba

### DIFF
--- a/src/data/chunk.ts
+++ b/src/data/chunk.ts
@@ -38,6 +38,7 @@ export type ChunkViewState = {
 export type Chunk = {
   data?: ChunkData;
   texture?: Texture;
+  releasedAt?: DOMHighResTimeStamp;
   state: "unloaded" | "queued" | "loading" | "loaded";
   lod: number;
   shape: {
--- a/src/data/chunk_manager.ts
+++ b/src/data/chunk_manager.ts
@@ -4,7 +4,7 @@ import { chunkMemoryStats, clearChunkData } from "./chunk_memory";
 import { ChunkStore } from "./chunk_store";
 import { ChunkStoreView } from "./chunk_store_view";
 import { ImageSourcePolicy } from "../core/image_source_policy";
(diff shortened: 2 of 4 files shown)


srivarra (approved):
Played around with it and was able to see chunks get evicted with a strict budget.

I noticed that changing `memoryLimitMB` at runtime currently means tearing down and recreating the Idetik instance which drops the whole GPU chunk cache. Is there a way to avoid this? (doesn't need to be changed in this pr).

shlomnissan:
> I noticed that changing memoryLimitMB at runtime currently means tearing down and recreating the Idetik instance which drops the whole GPU chunk cache. Is there a way to avoid this? (doesn't need to be changed in this pr).

@srivarra that's a great callout. the parameter object has different mutation models:
- `viewports` are expected to be modified by `addViewport`/`removeViewport`
- `overlays` on the other hand are aliased
- the rest of the values are not runtime changeable

construction parameter objects should typically be read-only (otherwise leads to ambiguity and couples to caller memory). `overlays` feels like an oversight.

i'll post a quick follow-up after this PR is merged:
- pass `overlays` by value and add `addOverlay`/`removeOverlay`
- add `setMemoryLimitMB` mutator.

@aganders3 @srivarra 
- do we want mutators for `maxConcurrentRequests` and `maxGpuUploadsPerUpdate`?
- ashley remind me why we don't get rid of overlays altogether?

PR #705, opened 2026-07-07
feat: allow updating memory limit and overlays at runtime

shlomnissan:
### Summary

this PR make idetik runtime params a construction-time snapshot with runtime
setters. it adds runtime mutators for memory limits and overlays.

https://github.com/chanzuckerberg/idetik/pull/702#issuecomment-4906222994

i decided to skip mutators for `maxConcurrentRequests` and `maxGpuUploadsPerUpdate`. trivial to implement but i'm not sure we need to adjust these values at runtime.

### Tests & Checks

- all checks have passed

### DIFF
--- a/src/data/chunk_manager.ts
+++ b/src/data/chunk_manager.ts
@@ -22,7 +22,7 @@ export class ChunkManager {
   private readonly uploadTexture_?: (texture: Texture) => void;
   private readonly disposeTexture_?: (texture: Texture) => void;
   private readonly getGpuResidentBytes_: () => number;
-  private readonly memoryLimitBytes_: number;
+  private memoryLimitBytes_: number;
   private readonly maxGpuUploadsPerUpdate_: number;
 
   private readonly resident_ = new Set<Chunk>();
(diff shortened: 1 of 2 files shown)


aganders3 (approved):
Looks good.

I think there may be a path where the updated memory budget doesn't really take effect until the visible chunks change, because `enqueueWithinBudget` returns early if `candidates.length === 0`. I'm not certain though, and it may even be desirable behavior.

srivarra (approved):
Awesome ty, I don't think my use cases will be needing mutators for `maxConcurrentRequests` and `maxGpuUploadsPerUpdate` afik.

shlomnissan:
> and it may even be desirable behavior

@aganders3 i don't know about desirable but maybe not a big deal. the only lazy path is:
- all queued chunks are loaded
- user reduces memory budget
- loaded chunks no longer fit the new budget
- eviction won't start until visible chunks change

note that increasing memory budget has an immediate impact because there are candidates waiting.

i think that's okay?

aganders3:
Yes, I agree and it's not a blocker at all for me.

What I mean by desirable is that I think it's good any currently visible and loaded chunks will not be evicted. Additional behavior at the edges of this (e.g. evicting prefetched chunks when the budget is lowered) might be nice, but doesn't need to be a priority.

One more thought for posterity (not for this PR): we should perhaps warn the user in the case that loading all visible chunks exceeds the budget.

PR #706, opened 2026-07-10
feat: add bidirectional temporal prefetch

### LINKED ISSUE #701: Prefetching chunks backwards in T
Currently prefetching works by looking forwards in T, it would be nice to be able to also prefetch to the previous timepoint so users can scrub back and forth.

Here's a recording of scrubbing forward and backwards in time, with a config of `prefetch: { x: 0, y: 0, z: 0, t: 2 }`

https://github.com/user-attachments/assets/548d8ad1-ddb2-44f8-934d-efd3f2613f33

I'm happy to contribute a pr for this as well.

srivarra:
### Summary

You can now prefetch back in `t` which allows for a slightly smoother experience with scrubbing back and forth.

`prefetch.t` now accepts a tuple of numbers `[backward, forward]`. You can still pass in a scalar value for `t` and that'll default to just forward. I originally thought about converting a scalar that gets passed into `t` into a tuple of `[0, forward]` but saw that would break things in downstream packages.  

Made sure it worked with circular prefetching.

### Related Issue
Closes #701

### Tests & Checks

Tested by allowing you to adjust forward and backward t prefetch range in the image2d example. Not sure if the changes to the example should be kept.

https://github.com/user-attachments/assets/28486dfa-1d27-40b9-b096-20c7d4e376a0

### DIFF
--- a/examples/image_2d/main.ts
+++ b/examples/image_2d/main.ts
@@ -106,6 +106,26 @@ const xLod = dimensions.x.lods[0];
 const yLod = dimensions.y.lods[0];
 const zLod = dimensions.z?.lods[0];
 const tLod = dimensions.t?.lods[0];
+const maxTemporalPrefetch = Math.min(Math.max((tLod?.size ?? 1) - 1, 0), 25);
+const temporalPrefetchState = {
+  backward: 0,
+  forward: Math.min(20, maxTemporalPrefetch),
+};
+let playbackRateHz = 0;
+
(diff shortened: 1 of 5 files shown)


shlomnissan:
how do you assess the overall value of prefetching past frames now that we don't immediately evict past frames? Is there still a clear use case for it?

srivarra:
@shlomnissan after chatting, yeah looks like the deferred eviction solves my issue!

PR #707, opened 2026-07-14
refactor: generalize image slicing to plane/slice axes

shlomnissan:
### Summary

this PR routes hard-coded XY operations through a `SliceAxes` mapping.
in-plane u/v + a slice normal w. this is fixed on XY for now. no behavior
change. this is the groundwork for selectable slice orientation.

### Tests & Checks

- all checks have passed.

### DIFF
--- a/src/layers/image_layer.ts
+++ b/src/layers/image_layer.ts
@@ -2,7 +2,7 @@ import { Layer, LayerOptions } from "../core/layer";
 import { Viewport } from "../core/viewport";
 import { OrthographicCamera } from "../objects/cameras/orthographic_camera";
 import type { IdetikContext } from "../idetik";
-import { Chunk, ChunkSource, SliceCoordinates } from "../data/chunk";
+import { Chunk, ChunkSource, SliceAxes, SliceCoordinates } from "../data/chunk";
 import { ChunkStoreView, INTERNAL_POLICY_KEY } from "../data/chunk_store_view";
 import { ImageSourcePolicy } from "../core/image_source_policy";
 import {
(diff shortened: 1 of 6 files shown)


srivarra (approved):

src/layers/image_layer.ts
```diff
@@ -280,8 +285,9 @@ export class ImageLayer extends Layer implements ChannelsEnabled {
     } else {
       image.wireframeEnabled = false;
     }
-    image.transform.setScale([chunk.scale.x, chunk.scale.y, 1]);
-    image.transform.setTranslation([chunk.offset.x, chunk.offset.y, 0]);
+    const { u, v } = this.axes_;
+    image.transform.setScale([chunk.scale[u], chunk.scale[v], 1]);
+    image.transform.setTranslation([chunk.offset[u], chunk.offset[v], 0]);
```
aganders3, line 290: We can talk about and/or change this later, but I still feel like this should place and orient the slice in world space correctly relative to the volume. That would allow the same renderables to be shown e.g. as cut planes in the 3D view.

aganders3 (approved):
Nice work - this is a good step towards alternate orientations.

Any thoughts on generalizing further for oblique slices?

One note not in the diff: the shaders still use `u_zTexCoord` but maybe that should also be renamed?

shlomnissan:
@aganders3 

> Any thoughts on generalizing further for oblique slices?

i think it's possible but it introduces complexity that we probably don't need right now? spatial slice coordinates becomes a clip plane, the math is probably straight-forward, plane intersection vs chunk AABB. the geometry part is annoying: polygons that sits on the clip planes with 3D texture coordinates? at this point i would consider virtual textures so we have a single plane geometry that's aligned with the clip plane.

i don't know, feels like a whole lot of unnecessary work right now.

> the shaders still use u_zTexCoord

yeah i have TODOs in this PR that flag this.

src/layers/image_layer.ts
shlomnissan: i agree. the approach i had in mind is permutation for 2D slice views but i'm happy to change it.

i don't think it impacts this PR other than `sliceTexCoord` path but i can remove it in the next PR.

any ideas on how we should deal with the camera if we place the slice in world space relative to the volume? i think we make a few assumptions around camera looks down world -Z.

PR #708, opened 2026-07-15
refactor: slice images by world-space fragment position

shlomnissan:
### Summary

this PR adds image slicing based on world-space fragment position. planes get placed
at their true world position and fragment shaders derive the 3D texture coordinate from
the fragment's world position. it's another step forward toward selectable slice
orientations.

this approach is based on @aganders3 suggestion:

https://github.com/chanzuckerberg/idetik/pull/707#discussion_r3584072542

### Tests & Checks

- all checks have passed
- visual verification

https://github.com/user-attachments/assets/a993896b-2b86-4315-aceb-daa1c1039d35

### DIFF
--- a/src/data/chunk.ts
+++ b/src/data/chunk.ts
@@ -3,6 +3,7 @@ import {
   TextureUnpackRowAlignment,
 } from "../objects/textures/texture";
 import { Logger } from "../utilities/logger";
+import { mat4 } from "gl-matrix";
 
 const chunkDataTypes = [
   Int8Array,
@@ -142,4 +143,16 @@ export function coordToChunkIndex(
 ): number {
   const index = coordToIndex(lod, coord);
   return Math.floor(index / lod.chunkSize);
+}
+
+export function worldToTexCoordForChunk(chunk: Chunk, axes: SliceAxes): mat4 {
+  const m = mat4.create();
+  for (const axis of ["x", "y", "z"] as const) {
+    const scale = 1 / (chunk.scale[axis] * chunk.shape[axis]);
+    const centerShift = axis === axes.w ? 0.5 * chunk.scale[axis] : 0;
+    const component = AxisComponent[axis];
+    m[component * 4 + component] = scale;
+    m[12 + component] = (centerShift - chunk.offset[axis]) * scale;
+  }
+  return m;
 }
--- a/src/data/chunk.ts
+++ b/src/data/chunk.ts
@@ -3,7 +3,7 @@ import {
   TextureUnpackRowAlignment,
 } from "../objects/textures/texture";
 import { Logger } from "../utilities/logger";
-import { mat4 } from "gl-matrix";
+import { mat4, vec3 } from "gl-matrix";
 
 const chunkDataTypes = [
   Int8Array,
@@ -146,13 +146,14 @@ export function coordToChunkIndex(
 }
 
 export function worldToTexCoordForChunk(chunk: Chunk, axes: SliceAxes): mat4 {
-  const m = mat4.create();
+  const scale = vec3.create();
+  const translation = vec3.create();
   for (const axis of ["x", "y", "z"] as const) {
-    const scale = 1 / (chunk.scale[axis] * chunk.shape[axis]);
     const centerShift = axis === axes.w ? 0.5 * chunk.scale[axis] : 0;
     const component = AxisComponent[axis];
-    m[component * 4 + component] = scale;
-    m[12 + component] = (centerShift - chunk.offset[axis]) * scale;
+    scale[component] = 1 / (chunk.scale[axis] * chunk.shape[axis]);
+    translation[component] = centerShift - chunk.offset[axis];
   }
-  return m;
+  const m = mat4.fromScaling(mat4.create(), scale);
+  return mat4.translate(m, m, translation);
 }
--- a/src/objects/cameras/orthographic_camera.ts
+++ b/src/objects/cameras/orthographic_camera.ts
@@ -5,6 +5,8 @@ import { Box2 } from "../../math/box2";
 const DEFAULT_ASPECT_RATIO = 1.77; // 16:9
 const DEFAULT_WIDTH = 128;
 const DEFAULT_HEIGHT = 128 / DEFAULT_ASPECT_RATIO;
+const DEFAULT_NEAR = -1e6;
+const DEFAULT_FAR = 1e6;
 
 /**
  * A camera using an orthographic (parallel) projection.
@@ -33,16 +35,16 @@ export class OrthographicCamera extends Camera {
    * @param right - Right edge of the view frame, in world units.
    * @param top - Top edge of the view frame, in world units.
    * @param bottom - Bottom edge of the view frame, in world units.
-   * @param near - Near clipping plane distance. Defaults to `0.0`.
-   * @param far - Far clipping plane distance. Defaults to `100.0`.
+   * @param near - Near clipping plane distance. Defaults to `-1e6`.
+   * @param far - Far clipping plane distance. Defaults to `1e6`.
    */
   constructor(
     left: number,
     right: number,
     top: number,
     bottom: number,
-    near = 0.0,
-    far = 100.0
+    near = DEFAULT_NEAR,
+    far = DEFAULT_FAR
   ) {
     super();
     this.near_ = near;
--- a/src/objects/renderable/label_image_renderable.ts
+++ b/src/objects/renderable/label_image_renderable.ts
@@ -1,5 +1,6 @@
 import { RenderableObject } from "../../core/renderable_object";
 import { PlaneGeometry } from "../../objects/geometry/plane_geometry";
+import { mat4 } from "gl-matrix";
 import { Texture, TextureDataType } from "../../objects/textures/texture";
 import { Color } from "../../math/color";
 import { Texture2D } from "../textures/texture_2d";
@@ -48,12 +49,11 @@ export class LabelImageRenderable extends RenderableObject {
   private outlineSelected_: boolean;
   private selectedValue_: number | null;
 
-  // The layer overwrites this per chunk with a texel-centered slice coordinate.
-  // 0.5 (the texture's center) is a safe placeholder until it does.
-  public sliceTexCoord = 0.5;
+  public worldToTexCoord: mat4 = mat4.create();
 
   constructor(props: LabelImageRenderableProps) {
     super();
+    this.depthTest = false;
     this.geometry = new PlaneGeometry(props.width, props.height, 1, 1);
     this.setTexture(0, validateImageData(props.imageData));
     const colorCycleTexture = this.makeColorCycleTexture(props.colorMap.cycle);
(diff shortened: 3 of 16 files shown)


src/renderers/shaders/scalar_image_frag.glsl
```diff
@@ -17,12 +17,13 @@ uniform vec3 u_color;
...
+uniform mat4 u_worldToTexCoord;
 
-in vec2 v_texCoords;
+in vec3 v_positionWorld;
 
 void main() {
-    float texel = float(texture(u_imageSampler, vec3(v_texCoords, u_zTexCoord)).r);
+    vec3 texCoords = (u_worldToTexCoord * vec4(v_positionWorld, 1.0)).xyz;
+    float texel = float(texture(u_imageSampler, texCoords).r);
     float value = (texel + u_valueOffset) * u_valueScale;
```
srivarra: spent some time [learning about webgl2](https://webgl2fundamentals.org/webgl/lessons/webgl-how-it-works.html), and I'm wondering if `texCoords` could be computed in vertices and then itnerpolated here? This should hopefully keep the same result while avoiding the matrix transformation per fragment. Any thoughts?

shlomnissan: ha! that's a miss on my part. yes, moving the multiply to the vertex shader is exactly the right thing. no reason to compute this in fragment shader.

great callout!

shlomnissan: https://github.com/chanzuckerberg/idetik/pull/708/changes/1175940392611e8b786ef92b46a96181ac07e49f

srivarra (approved):
Very exciting!

src/data/chunk.ts
```diff
@@ -143,3 +144,15 @@ export function coordToChunkIndex(
...
+export function worldToTexCoordForChunk(chunk: Chunk, axes: SliceAxes): mat4 {
+  const m = mat4.create();
+  for (const axis of ["x", "y", "z"] as const) {
+    const scale = 1 / (chunk.scale[axis] * chunk.shape[axis]);
+    const centerShift = axis === axes.w ? 0.5 * chunk.scale[axis] : 0;
+    const component = AxisComponent[axis];
+    m[component * 4 + component] = scale;
+    m[12 + component] = (centerShift - chunk.offset[axis]) * scale;
+  }
+  return m;
```
aganders3: nit: might be more readable with `mat4.fromTranslation` and `scale`, but this is fine. It just took me a minute to grok this.

src/objects/cameras/orthographic_camera.ts
```diff
@@ -5,6 +5,8 @@ import { Box2 } from "../../math/box2";
 const DEFAULT_ASPECT_RATIO = 1.77; // 16:9
 const DEFAULT_WIDTH = 128;
 const DEFAULT_HEIGHT = 128 / DEFAULT_ASPECT_RATIO;
+const DEFAULT_NEAR = -1e6;
+const DEFAULT_FAR = 1e6;
```
aganders3, line 9: I'm okay with this, but it can be a bit confusing to have a negative near-plane. Effectively this allows the camera to see "behind" itself.

I'm also concerned about float precision and depth buffer quantization with these large values, but they're just defaults so not a blocker.

src/objects/renderable/label_image_renderable.ts
```diff
@@ -39,12 +40,11 @@ export class LabelImageRenderable extends RenderableObject {
...
   private selectedValue_: number | null;
 
-  // The layer overwrites this per chunk with a texel-centered slice coordinate.
-  // 0.5 (the texture's center) is a safe placeholder until it does.
-  public sliceTexCoord = 0.5;
+  public worldToTexCoord: mat4 = mat4.create();
 
   constructor(props: LabelImageRenderableProps) {
     super();
+    this.depthTest = false;
```
aganders3: We already use `this.gl_.depthFunc(this.gl_.LEQUAL);`, isn't that good enough? Will setting `depthTest = false` for labels allow labels to draw over an image even if they're physically behind in world space?

src/renderers/shaders/label_image_frag.glsl
```diff
@@ -24,24 +25,26 @@ vec4 unpackRgba(uint packed) {
...
+    vec3 stepY = modelToTexCoord[1].xyz;
 
     // Check 8-connected neighbors
     for (int dx = -1; dx <= 1; dx++) {
         for (int dy = -1; dy <= 1; dy++) {
             if (dx == 0 && dy == 0) continue; // Skip center pixel
 
-            vec2 neighborCoords = v_texCoords + vec2(float(dx), float(dy)) * texelSize;
+            vec3 neighborCoords =
+                texCoords + float(dx) * stepX + float(dy) * stepY;
```
aganders3, line 40: Obviously out-of-scope for this PR but this is a reminder to myself to look into using frag derivatives here.

aganders3 (approved):
Nice! This looks great.

src/data/chunk.ts
shlomnissan: yeah much more readable. i don't have a strong grip over gl-matrix idioms.

src/objects/cameras/orthographic_camera.ts
shlomnissan: - it's perfectly reasonable to have a negative near plane. ortho projection has no eye point it's just a world-space box mapped to the canonical view volume. as i understand it ome-zarr doesn't guarantee a positive range so we need to extend the range in both directions. note that currently `setFrame` puts the camera in z = 0 looking down -z so positive z values in ome are actually behind the camera.
- ortho is linear in z ([no perspective 1/z messiness](https://www.shlom.dev/articles/reverse-z-perfect-hack/)). this default range gives us uniform ~0.12 world units which is probably fine? if we do have multiple planes with distinct sources sitting at 0.05 units apart we're in z fighting territory but we can just adjust the default values.

src/objects/renderable/label_image_renderable.ts
shlomnissan: the label's layer source in idetik example has no depth value so it's always pinned to z = 0. i disabled depth tests by default which made sense to me when i wrote the code but i didn't think this through: if we have 3D orthoslices with labels layer overlay things will get messed up.

i suppose we can make this configurable? what do you think? this is a general problem with the world slicing approach: even a slight variation in the transform offset of the sources can cause issues.

src/objects/cameras/orthographic_camera.ts
aganders3: > it's perfectly reasonable to have a negative near plane

I agree, but it breaks the "camera" analogy. For example in three.js the default near plane is a small positive value, and they don't let you set negative values (they do let you set 0). The alternative is to require translating the camera.

> this default range gives us uniform ~0.12 world units

I think this is also generally fine but I don't have a good grasp of what units we will see. For example if a zarr dataset for some reason specifies "m" for units and a very small scale value to get into microns.

In any case I'm happy to leave it as-is and override for specific data/application if necessary. Targeting DCA-spec data also probably helps us here.

src/objects/renderable/label_image_renderable.ts
aganders3: This is a good question - I didn't realize the example was a 2D + t label over a slice of a 3D + t image. This does feel like something that perhaps needs to be configurable, but I'm not sure the right way to do it.

I could imagine setting the transform of the label layer using the slice coord of the image layer. This would only work in the "matching" orientation.

In the case of 3D orthoslices I would assume a matching 3D labels layer, otherwise I'm not sure what I'd expect to display.

shlomnissan: how does neuroglancer deals with this?

src/renderers/shaders/label_image_frag.glsl
shlomnissan: this is pretty straightforward (assuming we're keeping the loop). i can just add the fix to the next commit.

src/renderers/shaders/label_image_frag.glsl
```diff
@@ -58,14 +55,17 @@ bool isEdgePixel(DATA_TYPE centerValue) {
 }
 
 void main() {
-    DATA_TYPE texel = texture(u_imageSampler, vec3(v_texCoords, u_zTexCoord)).r;
+    DATA_TYPE texel = texture(u_imageSampler, v_texCoords).r;
+
+    vec3 stepX = dFdx(v_texCoords);
```
shlomnissan, line 60: @aganders3 derivatives can't be computed inside non-uniform branches so computed here and passed to the function.

src/objects/renderable/label_image_renderable.ts
aganders3: [Here's a link to this data in Neuroglancer](https://neuroglancer-demo.appspot.com/#!%7B%22dimensions%22:%7B%22T%22:%5B1%2C%22s%22%5D%2C%22Z%22:%5B1.7399999999999997e-7%2C%22m%22%5D%2C%22Y%22:%5B1.494e-7%2C%22m%22%5D%2C%22X%22:%5B1.494e-7%2C%22m%22%5D%7D%2C%22displayDimensions%22:%5B%22X%22%2C%22Y%22%2C%22Z%22%5D%2C%22position%22:%5B62%2C0%2C378%2C540%5D%2C%22crossSectionScale%22:1%2C%22projectionOrientation%22:%5B-0.0874595120549202%2C-0.17883099615573883%2C-0.10617391765117645%2C0.9742162823677063%5D%2C%22projectionScale%22:1590.201050667412%2C%22layers%22:%5B%7B%22type%22:%22image%22%2C%22source%22:%7B%22url%22:%22https://public.czbiohub.org/organelle_box/datasets/A549/2024_11_07_A549_SEC61_DENV_tracking.zarr/B/3/000000/%7Czarr2:%22%2C%22transform%22:%7B%22matrix%22:%5B%5B1%2C0%2C0%2C0%2C0%2C0%5D%2C%5B0%2C1%2C0%2C0%2C0%2C0%5D%2C%5B0%2C0%2C25%2C0%2C0%2C0%5D%2C%5B0%2C0%2C0%2C1%2C0%2C0%5D%2C%5B0%2C0%2C0%2C0%2C1%2C0%5D%5D%2C%22outputDimensions%22:%7B%22T%22:%5B1%2C%22s%22%5D%2C%22C%27%22:%5B1%2C%22%22%5D%2C%22Z%22:%5B1.7399999999999997e-7%2C%22m%22%5D%2C%22Y%22:%5B1.494e-7%2C%22m%22%5D%2C%22X%22:%5B1.494e-7%2C%22m%22%5D%7D%7D%7D%2C%22localDimensions%22:%7B%22C%27%22:%5B1%2C%22%22%5D%7D%2C%22localPosition%22:%5B0%5D%2C%22tab%22:%22source%22%2C%22shaderControls%22:%7B%22normalized%22:%7B%22range%22:%5B0%2C141%5D%2C%22window%22:%5B0%2C177%5D%7D%7D%2C%22name%22:%22000000%22%7D%2C%7B%22type%22:%22image%22%2C%22source%22:%22https://public.czbiohub.org/organelle_box/datasets/A549/2024_11_07_A549_SEC61_DENV_cropped.zarr/B/3/000000/%7Czarr2:%22%2C%22localDimensions%22:%7B%22C%27%22:%5B1%2C%22%22%5D%7D%2C%22localPosition%22:%5B1%5D%2C%22tab%22:%22source%22%2C%22name%22:%220000001%22%7D%5D%2C%22selectedLayer%22:%7B%22visible%22:true%2C%22layer%22:%22000000%22%7D%2C%22layout%22:%224panel-alt%22%7D) for posterity.

PR #709, opened 2026-07-20
refactor: support signed data types for labels

shlomnissan:
### Summary

this PR adds support for signed data types in label image renderables. this
change solved the issue i was having with rendering segmentation data
in copick. it turns out that some of the data was signed even though the
values themselves are exclusively positive.

if there's a correctness concern around this change i can transform the data
on the copick server as an alternative.

### Tests & Checks

- all checks have passed.
- verified copick segmentation layer is working locally via npm link.

Screenshot 2026-07-20 at 3 00 09 PM


srivarra (approved):
Looks good. Do the webgpu shaders support labels?

aganders3 (approved):

shlomnissan:
> Looks good. Do the webgpu shaders support labels?

@srivarra not yet.. i don't feel comfortable adding anything new to it because the architecture of the webpu renderer is bad. i want to allocate some time to figure out the right thing for it but didn't get a chance yet.

PR #716, opened 2026-07-24
feat: multichannel image layer

aganders3:
### Summary
We currently use a stencil buffer to prevent over-drawing multiple LODs of the same channel/label. This is evaluated per-layer, so the current state is that this restricts the Image layer to a single channel on `main`. This creates unnecessary complication and asymmetry with the Volume layer API (which supports multiple channels in a single layer). 

This PR removes the layer-level `hasMultipleLODs` in favor of a renderable-level `coverageGroup`. The renderer uses the `coverageGroup` as a reference for the stencil, preventing draws if the current group is equal to the current stencil buffer. Thus, layers are expected to partition their objects to be coverage-contiguous. In the case of multi-LOD blending this means sorting to be fine -> coarse LOD and grouping by channel, assigning the channel to the `coverageGroup`.

Since we only use this feature to encode channel, we could just name it `channel`, but that abstraction is broken in the Label and Volume layers where labels don't typically have a "channel" and the Volume layer relies on blending channels in its shader, as well as blending spatial chunks.

This PR also updates the Image layer to support multiple channels. Behavior is very similar to the Volume layer: `SliceCoords` determines which channels are _loaded_ and channel properties determines _if and how_ they are drawn.

### Tests & Checks
Modified HCS and labels examples to each use a multi-channel image layer

### DIFF
--- a/src/core/layer.ts
+++ b/src/core/layer.ts
@@ -44,7 +44,7 @@ export interface LayerProps {
 export abstract class Layer {
   public abstract readonly type: string;
 
-  private objects_: RenderableObject[] = [];
+  private readonly renderGroups_ = new Map<number | null, RenderableObject[]>();
   private state_: LayerState = "initialized";
   private attached_ = false;
   private readonly callbacks_: StateChangeCallback[] = [];
@@ -95,8 +95,11 @@ export abstract class Layer {
 
   protected detach(_context: IdetikContext): void {}
 
-  public get objects() {
-    return this.objects_;
+  public get renderGroups(): ReadonlyMap<
+    number | null,
+    readonly RenderableObject[]
+  > {
+    return this.renderGroups_;
   }
 
   public get state() {
@@ -121,12 +124,20 @@ export abstract class Layer {
     this.callbacks_.forEach((callback) => callback(newState, prevState));
   }
 
-  protected addObject(object: RenderableObject) {
-    this.objects_.push(object);
+  protected addObject(
+    object: RenderableObject,
+    coverageGroup: number | null = null
+  ) {
+    const members = this.renderGroups_.get(coverageGroup);
+    if (members) {
+      members.push(object);
+    } else {
+      this.renderGroups_.set(coverageGroup, [object]);
+    }
   }
 
   protected clearObjects() {
-    this.objects_ = [];
+    this.renderGroups_.clear();
   }
 
   /**
--- a/src/core/layer.ts
+++ b/src/core/layer.ts
@@ -115,10 +115,6 @@ export abstract class Layer {
     this.callbacks_.splice(index, 1);
   }
 
-  public hasMultipleLODs(): boolean {
-    return false;
-  }
-
   protected setState(newState: LayerState) {
     const prevState = this.state_;
     this.state_ = newState;
@@ -127,13 +123,6 @@ export abstract class Layer {
 
   protected addObject(object: RenderableObject) {
     this.objects_.push(object);
-  }
-
-  protected removeObject(object: RenderableObject) {
-    const index = this.objects_.indexOf(object);
-    if (index !== -1) {
-      this.objects_.splice(index, 1);
-    }
   }
 
   protected clearObjects() {
--- a/src/core/layer.ts
+++ b/src/core/layer.ts
@@ -44,7 +44,10 @@ export interface LayerProps {
 export abstract class Layer {
   public abstract readonly type: string;
 
-  private readonly renderGroups_ = new Map<number | null, RenderableObject[]>();
+  private readonly coverageGroups_ = new Map<
+    number | null,
+    RenderableObject[]
+  >();
   private state_: LayerState = "initialized";
   private attached_ = false;
   private readonly callbacks_: StateChangeCallback[] = [];
@@ -95,11 +98,11 @@ export abstract class Layer {
 
   protected detach(_context: IdetikContext): void {}
 
-  public get renderGroups(): ReadonlyMap<
+  public get coverageGroups(): ReadonlyMap<
     number | null,
     readonly RenderableObject[]
   > {
-    return this.renderGroups_;
+    return this.coverageGroups_;
   }
 
   public get state() {
@@ -128,16 +131,16 @@ export abstract class Layer {
     object: RenderableObject,
     coverageGroup: number | null = null
   ) {
-    const members = this.renderGroups_.get(coverageGroup);
+    const members = this.coverageGroups_.get(coverageGroup);
     if (members) {
       members.push(object);
     } else {
-      this.renderGroups_.set(coverageGroup, [object]);
+      this.coverageGroups_.set(coverageGroup, [object]);
     }
   }
 
   protected clearObjects() {
-    this.renderGroups_.clear();
+    this.coverageGroups_.clear();
   }
 
   /**
--- a/src/core/renderable_object.ts
+++ b/src/core/renderable_object.ts
@@ -11,6 +11,12 @@ export abstract class RenderableObject extends Node {
   public wireframeEnabled = false;
   public wireframeColor = Color.WHITE;
   public depthTest = true;
+  // Objects sharing a non-null coverageId are overlapping representations of
+  // the same region (e.g. LODs of one channel):
+  //   * the renderer stencils them so each pixel is only drawn once per coverage group
+  //   * `null` means the object is drawn normally with no coverage dedup
+  //   * Objects with the same id must be contiguous in a layer's object list
+  public coverageId: number | null = null;
   private readonly textures_: Texture[] = [];
   private staleTextures_: Texture[] = [];
   private readonly transform_ = new TrsTransform();
--- a/src/core/renderable_object.ts
+++ b/src/core/renderable_object.ts
@@ -8,17 +8,9 @@ import { Color } from "../math/color";
 import { CullingMode } from "../renderers/webgl_state";
 
 export abstract class RenderableObject extends Node {
-  public visible = true;
   public wireframeEnabled = false;
   public wireframeColor = Color.WHITE;
   public depthTest = true;
-  // objects sharing a non-null coverageGroup are overlapping representations of the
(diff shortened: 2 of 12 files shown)


shlomnissan (commented):
why do you think "groups" is cleaner / more flexible? chunks are already grouped by channel. this change introduces another grouping abstraction on top of one that already exists.

perf for groups matches main but that's also its ceiling. the stencil reference approach can do better than main: if the ref is a frame-global counter (bumped once per layer-channel group) instead of restarting per layer we only clear the stencil once per frame and in webgpu that removes the reason for pass-per-layer entirely, so the whole frame renders in a single pass (the comment i had `webgpu_renderer` already points at this).

i think the fix can be small: `REPLACE/NOTEQUAL` with `channel + 1` as the ref. the data path is the real work but i think it's the same as what you have here? i didn't scope it so i might be missing something. either way, if we want to be forward leaning here i would pick the stencil reference approach.

aganders3:
Ah yeah you're right. I got stuck thinking we needed to use the stencil as a bitmask but I hadn't considered that grouping/sorting by channel encodes that already. I'm going to refactor this to use the `channel + 1` method you suggest.

aganders3:
Sorry, thinking about it some more...I'm trying to remember a week back after vacation what was on my mind.

The stencil reference improves the rendering efficiency but we still need to pass _something_ to the renderer to indicate the current channel being drawn in order to bump that reference, and I think the concept of a "group" or "coverageId" makes more sense there. We could encode this as "channel", but it's specific to multi-channel image data and "channel" doesn't really apply to e.g. a labels layer, but that's just a naming thing. Maybe more importantly the volume layer supports channels but does not need/use the stencil buffer in the same way.

`RenderGroup` does two things here: decouples LOD ordering (within a group) from channel partitioning (across groups), and enforces that channel partitioning to be correct by construction instead of convention. (It also provides a logical place to later enforce LOD ordering if we want).

Maybe that's not worth it if we're still implicitly depending on LOD order anyway. `getChunksToRender` is currently _not_ actually channel-contiguous, it's just `[...currentLODChunks, ...lowResChunks]`, so the layer still needs reorder them. I'm open to that - I will update the PR to see what it looks like.

src/core/layer.ts
```diff
@@ -129,13 +125,6 @@ export abstract class Layer {
     this.objects_.push(object);
   }
 
-  protected removeObject(object: RenderableObject) {
```
shlomnissan, line 132: did you remove it because it's unused or is there another reason that's tied to this change?

src/core/renderable_object.ts
```diff
@@ -8,9 +8,17 @@ import { Color } from "../math/color";
...
   public wireframeEnabled = false;
   public wireframeColor = Color.WHITE;
   public depthTest = true;
+  // objects sharing a non-null coverageGroup are overlapping representations of the
+  // same region (e.g. LODs of one channel), and the renderer draws each pixel only
+  // once per group; `null` draws normally with no coverage dedup. Groups are
+  // layer-local non-negative integers, and objects in the same group must be
+  // contiguous in the layer's object list. A renderer warns when a layer breaks
+  // either rule, or exceeds however many groups it can track.
+  public coverageGroup: number | null = null;
```
shlomnissan: renderable objects should describe what to draw. how a given frame presents them (visibility, coverage grouping) is computed by the layer, so the layer should own them. these are public mutable fields the layer overwrites on rebuild. properties like `wireframeEnabled`/`depthTest` represent user intent, visibility and coverage group are layer outputs.

maybe instead of `layer.objects` you have an accessor returning `{coverage, members: RenderableObject[]}[]` with hidden objects omitted. contiguity becomes true by construction and both fields come off `RenderableObject`.

src/core/renderable_object.ts
```diff
@@ -8,9 +8,17 @@ import { Color } from "../math/color";
 import { CullingMode } from "../renderers/webgl_state";
 
 export abstract class RenderableObject extends Node {
+  public visible = true;
   public wireframeEnabled = false;
   public wireframeColor = Color.WHITE;
   public depthTest = true;
+  // objects sharing a non-null coverageGroup are overlapping representations of the
```
shlomnissan: lets try to avoid auto-generate comments.

shlomnissan (changes requested):
blocking to consider the following:

https://github.com/chanzuckerberg/idetik/pull/716/changes#r3760139863

src/layers/image_layer.ts
```diff
@@ -97,7 +97,7 @@ export class ImageLayer extends Layer implements ChannelsEnabled {
     onPickValue,
     ...layerOptions
   }: ImageLayerProps) {
-    super(layerOptions);
+    super({ blendMode: "additive", ...layerOptions });
```
shlomnissan, line 101: i think this is a sensible default but it does change what opacity does on an image layer: before (blend "none") opacity only landed in the alpha channel with RGB unscaled. now RGB is scaled by opacity and the alpha channel gets opacity^2.

src/layers/image_layer.ts
```diff
@@ -227,18 +217,27 @@ export class ImageLayer extends Layer implements ChannelsEnabled {
     this.releaseAndRemoveChunks(nonVisibleChunks);
 
     this.clearObjects();
+
+    // `getChunksToRender` yields finest-first but not channel-contiguous, so
```
shlomnissan: the code makes it clear. i don't think this comment adds anything.

src/renderers/webgl_renderer.ts
```diff
@@ -137,43 +137,86 @@ export class WebGLRenderer extends Renderer {
...
     this.gl_.stencilMask(0xff);
-    this.gl_.stencilFunc(this.gl_.EQUAL, clearValue, 0xff);
-    this.gl_.stencilOp(this.gl_.KEEP, this.gl_.KEEP, this.gl_.INCR);
+    this.gl_.stencilOp(this.gl_.KEEP, this.gl_.KEEP, this.gl_.REPLACE);
+  }
+
+  // check that the layer has set `coverageGroup` properly:
+  // * groups must fit in the stencil buffer [0, 255)
+  // * objects in the same group must be contiguous in the object list
+  private hasValidCoverage(layer: Layer): boolean {
```
shlomnissan: i think with this suggested refactor: https://github.com/chanzuckerberg/idetik/pull/716/changes#r3760139863

this entire function can go away.

src/renderers/webgl_renderer.ts
```diff
@@ -137,43 +137,86 @@ export class WebGLRenderer extends Renderer {
...
+    object: RenderableObject,
+    camera: Camera
+  ) {
     object.popStaleTextures().forEach((texture) => {
       this.textures_.dispose(texture);
     });
 
     if (!object.programName) return;
     this.state_.setCullFaceMode(object.cullFaceMode);
+    // Depth *testing* is per-object; depth *writing* follows the per-pass policy
```
shlomnissan, line 187: this comment makes sense to me in the context of the change.

src/renderers/webgl_renderer.ts
```diff
@@ -137,43 +137,86 @@ export class WebGLRenderer extends Renderer {
...
-    const shouldUseStencil = layer.hasMultipleLODs();
-    this.state_.setStencilTest(shouldUseStencil);
-    if (shouldUseStencil) {
+
+    if (this.hasValidCoverage(layer)) {
       this.gl_.clear(this.gl_.STENCIL_BUFFER_BIT);
     }
 
-    layer.objects.forEach((object, i) => {
+    let coverageGroup: number | null | undefined = undefined;
```
shlomnissan: this kinda setup is why i dislike typescript. single variable with three states that mean different things.

you could add `let first = true` and it will be much easier to read.

src/renderers/webgl_renderer.ts
```diff
@@ -137,43 +137,86 @@ export class WebGLRenderer extends Renderer {
...
   }
 
   private renderLayer(layer: Layer, camera: Camera, frustum: Frustum) {
-    if (layer.objects.length === 0) return;
     this.state_.setBlendingMode(layer.blendMode);
-    const shouldUseStencil = layer.hasMultipleLODs();
-    this.state_.setStencilTest(shouldUseStencil);
-    if (shouldUseStencil) {
+
+    if (this.hasValidCoverage(layer)) {
```
shlomnissan: i think this can also go away with the grouped-list refactor.

src/core/layer.ts
aganders3: Mostly because it's unused, but possibly did not get added back after a refactor. Especially with objects in groups, I'm inclined to keep this removed until we need it for something.

src/layers/image_layer.ts
aganders3: This is a good call. We now have `blendMode` used for blending both within and between layers, which is for sure confusing. The opacity^2 is a bug though - I need to think about it.

src/core/renderable_object.ts
aganders3: I agree it's better at the layer level. This is similar to the initial version of this PR that had a `RenderGroup` container layers could use to organize renderable objects.

I'm not quite sure I get how `{coverage, members: RenderableObject[]}[]` enforces contiguity by construction. I think a `Map<number | null, RenderableObject[]>` would is more intuitive. What do you think?

src/layers/image_layer.ts
aganders3: I understand the problem here, but I think it's best sequenced after we finish this work if you're okay with that.

My first instinct is to change all the shaders to output premultiplied colors, and adjust the blend functions accordingly. I'll open an issue.

src/renderers/webgl_renderer.ts
shlomnissan: i meant to say "it only makes sense to me in the context of this change. maybe we should remove it"

shlomnissan (approved):
looks great!

src/renderers/webgl_renderer.ts
aganders3: Oops I removed it in the wrong branch, so this will land in #730

PR #718, opened 2026-07-28
feat: add slice orientation to image and label layers

shlomnissan:
### Summary

this PR completes the selectable slice orientation work. image and label
layers now accept an orientation prop (`XY` | `XZ` | `YZ`, default `XY`).
the slice axes are threaded to the chunk store view which drives culling,
LOD selection, and prefetch in plane axes, and the slice plane transform
is rotated to match. 

one stopgap worth noting: dynamic LOD selection requires an orthographic camera (LOD is derived from the world-space view rect, which only ortho cameras provide). both image and label layers used to throw on a perspective camera. to support orthoslices in a 3D view i removed that and instead treat the whole slice plane as visible with LOD chosen as if the plane spanned the viewport. so in 3D the full slice loads and zooming won't increase LOD. making view-dependent LOD work under perspective is doable but out of scope here.

### Tests & Checks

- all checks have passed
- visual verification (modified the multiple viewport example)

Screenshot 2026-07-28 at 10 56 20 AM

### DIFF
--- a/src/data/chunk_store_view.ts
+++ b/src/data/chunk_store_view.ts
@@ -21,7 +21,7 @@ export class ChunkStoreView {
   private policy_: ImageSourcePolicy;
   private policyChanged_ = false;
   private currentLOD_: number = 0;
-  private readonly axes_: SliceAxes = { u: "x", v: "y", w: "z" };
+  private readonly axes_: SliceAxes;
   private readonly scale0_: number;
   private lastViewBounds2D_: Box2 | null = null;
   private lastViewProjection_: mat4 | null = null;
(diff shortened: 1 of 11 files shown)


src/data/chunk_store_view.ts
```diff
@@ -606,30 +626,25 @@ export class ChunkStoreView {
   }
 
   private getPaddedBounds(bounds: Box3): Box3 {
```
srivarra, line 628: [question(non-blocking)](https://pullpo.io/cc?l=question&d=non-blocking) 

Why do you have to pad here in the first place?

srivarra (approved):
LGTM, just a side question about how idetik is designed.

src/data/chunk_store_view.ts
shlomnissan: we pad the visible bounds before using them for spatial prefetching. this extends the region beyond the viewport to cover adjacent currently-off-screen chunks in 2D. since users can pan prefetching this padded area lets us load nearby content preemptively and eliminate pop-in

srivarra: Ohhh that makes a lot of sense.

PR #720, opened 2026-07-30
feat: switch slice orientation at runtime

shlomnissan:
### Summary

this PR adds runtime orientation switching to the orthographic camera and
the image and label layers at runtime. this lets downstream users change
orientation without having to manage multiple sets of camera/layer per
slice orientation.

### Tests & Checks

- all checks have passed.
- shape verified by copick integration (in progress).

### DIFF
--- a/src/layers/image_layer.ts
+++ b/src/layers/image_layer.ts
@@ -62,8 +62,9 @@ export class ImageLayer extends Layer implements ChannelsEnabled {
 
   private readonly source_: ChunkSource;
   private readonly sliceCoords_: SliceCoordinates;
-  private readonly axes_: SliceAxes;
-  private readonly planeRotation_: quat;
+  private axes_: SliceAxes;
+  private planeRotation_: quat;
+  private orientation_: SliceOrientation;
   private readonly onPickValue_?: (info: PointPickingResult) => void;
   private readonly visibleChunks_: Map<Chunk, ImageRenderable> = new Map();
   private readonly pool_ = new RenderablePool<ImageRenderable>();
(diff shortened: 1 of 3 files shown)


nclack (approved):

src/layers/image_layer.ts
```diff
@@ -162,6 +167,39 @@ export class ImageLayer extends Layer implements ChannelsEnabled {
...
+      return;
+    }
+
+    this.releaseAndRemoveChunks(this.visibleChunks_.keys());
+    this.clearObjects();
+    this.chunkStoreView_?.dispose();
+    this.lastPresentationTimeStamp_ = undefined;
+    this.lastPresentationTimeCoord_ = undefined;
+    this.orientation_ = orientation;
+    this.axes_ = sliceAxesFor(orientation);
```
srivarra, line 189: Should `sliceCoords_` be updated here as well? Switching the orientation moves `axes_.w`.

srivarra (changes requested):
Looks good, just one thing that may or may not be a problem.

src/layers/image_layer.ts
shlomnissan: slice coordinates are managed by the application and are often shared across layers (see new multiple viewport example) so it shouldn't change here. 

there's an edge case where if the app only did XY and it switches to XZ the `w` coordinate might be undefined, but this is a configuration issue that `setOrientation` makes easier to run into.

i think i will leave it as is for now, but it might be worth considering logging a warning like "sliceCoords has no XXX coordinate for orientation XXX chunks will fall back to their own offsets."

srivarra (approved):

PR #725, opened 2026-08-05
docs: add a custom header with version badge to the API reference

shlomnissan:
### Summary

this PR adds a dynamic version badge to the API reference index page. it also
adds some context at the top of the page.

### Tests & Checks

- all checks have passed
- visual verification

Screenshot 2026-08-05 at 1 22 54 PM

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -18,6 +18,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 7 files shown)


srivarra (approved):
Do you build docs for every version? Suppose Idetik get's updated to `v0.40.0`, can I also view the docs for `0.35.1`?

aganders3:
They're built for every version, but I think only the most recent is actually published. Maybe there's a vitepress plugin or something that would give us a version picker?

PR #730, opened 2026-08-12
feat: add a depth pass to enable occlusion on multi-channel image layer (blending)

aganders3:
### Summary
In #716 we moved blending for multi-channel images from separate layers into a single layer. However, this means the layer's `blendMode` is now overloaded to control blending of objects _within a layer_ as well as blending _between layers_. We also dropped the explicit `Layer.transparent` attribute in #668, instead deriving it from the `blendMode`.

The main issue this causes is that image layers now appear transparent in some circumstances, because they never write depth.

This PR adds a `Layer.occludes` attribute to disambiguate this. Initially this looks like we're resurrecting the `Layer.transparent` attribute but it's subtly different, hence the different name. `transparent` gated blending and was entirely derivable from `blendMode`. `occludes` gates depth writing, which isn't derivable — an additive multi-channel image blends internally but should still read as solid to other layers.

This also updates the multi-viewport example to use a multi-channel dataset.

Note: in the 3D view a volume layer would now be erased wherever a slice covers it. Previously the example drew the volume first, so it composited into an empty framebuffer; now non-occluding layers (the volume in this example) always draw after occluders (images, in this example) regardless of their order in layers. This is not depth occlusion (the volume layer opts out of depth testing) but a result of the premultiplied blending "under" the already-drawn slices. I plan to fix this in a follow-up PR by: 1) changing the volume layer blending to compositing "over" with back-to-front chunk order, and 2) writing this depth pass into a _texture_ that can later be sampled in the volume shader to terminate rays at the scene depth.

### Tests & Checks
See the multi-viewport example - without this PR the sections in the 3D view would show through each other instead of appearing as three opaque planes.

#### `main`

Screenshot 2026-08-20 at 11 32 07 AM

#### with this PR

Screenshot 2026-08-20 at 11 32 41 AM

### DIFF
--- a/examples/multi_viewport/main.ts
+++ b/examples/multi_viewport/main.ts
@@ -19,29 +19,27 @@ import GUI from "lil-gui";
 const url =
   "https://public.czbiohub.org/royerlab/zebrahub/imaging/multi-view/ZMNS001.ome.zarr/";
 
-// shared source between viewports/layers
 const source = await OmeZarrImageSource.fromHttp({ url });
 
-const lod = 0;
+const baseLod = 0;
 const dims = source.getDimensions();
 const axisRange = (axis: "x" | "y" | "z"): readonly [number, number] => {
-  const d = dims[axis]!.lods[lod];
+  const d = dims[axis]!.lods[baseLod];
   return [d.translation, d.translation + d.size * d.scale] as const;
 };
 const [xMin, xMax] = axisRange("x");
 const [yMin, yMax] = axisRange("y");
 const [zMin, zMax] = axisRange("z");
-const step = (axis: "x" | "y" | "z") => dims[axis]!.lods[lod].scale;
+const step = (axis: "x" | "y" | "z") => dims[axis]!.lods[baseLod].scale;
 
-const timeCount = dims.t?.lods[lod].size ?? 1;
+const timeCount = dims.t?.lods[baseLod].size ?? 1;
 const center = vec3.fromValues(
   (xMin + xMax) / 2,
   (yMin + yMax) / 2,
   (zMin + zMax) / 2
 );
 
-// shared timepoint across all viewports
-const sharedTime = { t: Math.floor((timeCount - 1) / 2) };
+const currentTime = { t: Math.floor((timeCount - 1) / 2) };
 
 const channelProps: ChannelProps[] = [
   { color: Color.CYAN, contrastLimits: [0, 1200] },
@@ -50,7 +48,7 @@ const channelProps: ChannelProps[] = [
 
 const sliceCoords = {
   get t() {
-    return sharedTime.t;
+    return currentTime.t;
   },
   x: center[0],
   y: center[1],
@@ -101,7 +99,6 @@ const camera3D = new PerspectiveCamera({ near: 1.0 });
 
 new Idetik({
   canvas: document.querySelector<HTMLCanvasElement>("#canvas")!,
-  memoryLimitMB: 4096,
   viewports: [
     createSliceViewport("slice-xy", "XY", xRange, yRange),
     {
@@ -130,7 +127,7 @@ const gui = new GUI({ width: 300 });
 
 addDimensionSlider({
   gui,
-  sliceCoords: sharedTime,
+  sliceCoords: currentTime,
   dimensionName: "t",
   minValue: 0,
   maxValue: timeCount - 1,
--- a/examples/multi_viewport/main.ts
+++ b/examples/multi_viewport/main.ts
@@ -97,10 +97,11 @@ const xRange = [xMin, xMax] as const;
 const yRange = [yMin, yMax] as const;
 const zRange = [zMin, zMax] as const;
 
-const camera3D = new PerspectiveCamera();
+const camera3D = new PerspectiveCamera({ near: 1.0 });
 
 new Idetik({
   canvas: document.querySelector<HTMLCanvasElement>("#canvas")!,
+  memoryLimitMB: 4096,
   viewports: [
     createSliceViewport("slice-xy", "XY", xRange, yRange),
     {
--- a/examples/multi_viewport/main.ts
+++ b/examples/multi_viewport/main.ts
@@ -1,4 +1,6 @@
 import {
+  ChannelProps,
+  Color,
   Idetik,
   ImageLayer,
   OmeZarrImageSource,
@@ -16,52 +18,45 @@ import { vec3 } from "gl-matrix";
 import GUI from "lil-gui";
 
 const url =
-  "https://public.czbiohub.org/royerlab/zebrahub/imaging/single-objective/ZSNS001.ome.zarr/";
-const left = 150;
-const right = 950;
-const top = 100;
-const bottom = 900;
+  "https://public.czbiohub.org/royerlab/zebrahub/imaging/multi-view/ZMNS001.ome.zarr/";
 
-// values copied from source
-const z = { translate: 0.0, scale: 1.24, shape: 448 };
-const zMin = z.translate;
-const zMax = z.translate + z.scale * z.shape - z.scale;
+// shared source between viewports/layers
+const source = await OmeZarrImageSource.fromHttp({ url });
 
-const volumeCenter = vec3.fromValues(
-  (right + left) / 2,
-  (top + bottom) / 2,
+const lod = 0;
+const dims = source.getDimensions();
+const axisRange = (axis: "x" | "y" | "z"): readonly [number, number] => {
+  const d = dims[axis]!.lods[lod];
+  return [d.translation, d.translation + d.size * d.scale] as const;
+};
+const [xMin, xMax] = axisRange("x");
+const [yMin, yMax] = axisRange("y");
+const [zMin, zMax] = axisRange("z");
+const step = (axis: "x" | "y" | "z") => dims[axis]!.lods[lod].scale;
+
+const timeCount = dims.t?.lods[lod].size ?? 1;
+const center = vec3.fromValues(
+  (xMin + xMax) / 2,
+  (yMin + yMax) / 2,
   (zMin + zMax) / 2
 );
 
-// shared source between viewports
-const source = await OmeZarrImageSource.fromHttp({ url });
+// shared timepoint across all viewports
+const sharedTime = { t: Math.floor((timeCount - 1) / 2) };
 
-// Shared timepoint across all viewports
-const sharedTime = { t: 400 };
-
-// Volume layer - no z coordinate to render entire volume
-const volumeCoords = {
-  get t() {
-    return sharedTime.t;
-  },
-  c: [0],
-};
-const camera3D = new PerspectiveCamera();
-const volumeLayer = new VolumeLayer({
-  source,
-  sliceCoords: volumeCoords,
-  policy: createPlaybackPolicy({ lod: { min: 2, max: 2 } }),
-  channelProps: [{ contrastLimits: [0, 200] }],
-});
+const channelProps: ChannelProps[] = [
+  { color: Color.CYAN, contrastLimits: [0, 1200] },
+  { color: Color.MAGENTA, contrastLimits: [0, 400] },
+];
 
 const sliceCoords = {
   get t() {
     return sharedTime.t;
   },
-  x: (left + right) / 2,
-  y: (top + bottom) / 2,
-  z: 300,
-  c: [0],
+  x: center[0],
+  y: center[1],
+  z: center[2],
+  c: [0, 1],
 };
 
 function createSliceLayer(orientation: SliceOrientation) {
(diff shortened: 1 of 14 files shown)


src/core/layer.ts
```diff
@@ -54,10 +55,16 @@ export abstract class Layer {
...
-  constructor({ opacity = 1.0, blendMode = "none" }: LayerProps = {}) {
+  constructor({
+    opacity = 1.0,
+    blendMode = "none",
+    occludes,
+  }: LayerProps = {}) {
     this.opacity_ = clamp(opacity, 0.0, 1.0);
     this.blendMode = blendMode;
+    this.occludes = occludes ?? blendMode === "none";
   }
```
srivarra, line 71: `occludes` defaults from `blendMode` only during construction. Changing `blendMode` later leaves the layer's depth behavior unchanged. Should it be able to change?

srivarra (approved):
Looks good, just one question about how we want `occludes` and `blendMode` to interact.

src/core/layer.ts
aganders3: Good point, thanks. I don't love the API here in any case, but what do you think of changing to a getter/setter?

```
  public get occludes() {
    return this.occludes_ ?? this.blendMode === "none";
  }
  public set occludes(value: boolean) {
    this.occludes_ = value;
  }
```

So if you never set occludes it tracks `blendMode` (unblended occludes, blended doesn't); if you set it, your value wins from then on.

examples/multi_viewport/main.ts
```diff
@@ -16,60 +17,53 @@ import { vec3 } from "gl-matrix";
...
-
-const volumeCenter = vec3.fromValues(
-  (right + left) / 2,
-  (top + bottom) / 2,
-  (zMin + zMax) / 2
-);
+  "https://public.czbiohub.org/royerlab/zebrahub/imaging/multi-view/ZMNS001.ome.zarr/";
 
-// shared source between viewports
+// shared source between viewports/layers
```
shlomnissan: nit: given that this file reuses `source` and `sharedTime` i think both comments stating that it's shared are unnecessary. i would also rename `sharedTime` to `currentTime`. if we really want to make it explicit, keep `sharedTime` and rename `source` to `sharedSource`.

i know it's not new to this PR so take it or leave it.

examples/multi_viewport/main.ts
```diff
@@ -16,60 +17,53 @@ import { vec3 } from "gl-matrix";
...
-// Shared timepoint across all viewports
-const sharedTime = { t: 400 };
-
-// Volume layer - no z coordinate to render entire volume
-const volumeCoords = {
-  get t() {
-    return sharedTime.t;
-  },
-  c: [0],
+const lod = 0;
```
shlomnissan: nit: i think renaming to `baseLod` makes this constant clearer.

examples/multi_viewport/main.ts
```diff
@@ -16,60 +17,53 @@ import { vec3 } from "gl-matrix";
...
+const dims = source.getDimensions();
+const axisRange = (axis: "x" | "y" | "z"): readonly [number, number] => {
```
shlomnissan, line 26: this makes me wonder if we should introduced a more structured return type after we open a source that already includes this type of metadata along with the computed extent.

examples/multi_viewport/main.ts
```diff
@@ -99,26 +93,28 @@ function createSliceViewport(
...
-const yRange = [top, bottom] as const;
+const xRange = [xMin, xMax] as const;
+const yRange = [yMin, yMax] as const;
 const zRange = [zMin, zMax] as const;
 
+const camera3D = new PerspectiveCamera({ near: 1.0 });
+
 new Idetik({
   canvas: document.querySelector<HTMLCanvasElement>("#canvas")!,
+  memoryLimitMB: 4096,
```
shlomnissan: is the default value not enough? if not, should we consider updating `DEFAULT_MEMORY_LIMIT_MB`?

src/core/layer.ts
shlomnissan: i'd leave it as is and add a comment. With a getter/setter you get an internal field with three states (`true` / `false` / `undefined = derive` from `blendMode`) and once you assign occludes it's pinned. un-pinning would mean assigning `undefined` through a API weirdness.

The sequence that would need live re-derivation: pinning occludes then mutating `blendMode` at runtime and expecting the layer to change passes seems unlikely to me anyway.

src/renderers/webgl_renderer.ts
```diff
@@ -121,27 +115,70 @@ export class WebGLRenderer extends Renderer {
 
     const frustum = viewport.camera.frustum;
 
-    this.state_.setDepthMask(true);
-    for (const layer of opaque) {
-      if (layer.state === "ready") {
-        this.renderLayer(layer, viewport.camera, frustum);
-      }
+    const occludingLayers: Layer[] = [];
+    const transparentLayers: Layer[] = [];
```
shlomnissan: does `transparentLayers` brings back the conflation this PR is trying to kill: `occludes` is not derived from blending? maybe `nonOccludingLayers` is a better name?

src/renderers/webgl_renderer.ts
```diff
@@ -121,27 +115,70 @@ export class WebGLRenderer extends Renderer {
...
+    for (const layer of viewport.layers) {
+      if (layer.state !== "ready") continue;
+      (layer.occludes ? occludingLayers : transparentLayers).push(layer);
     }
 
+    // polygon offset nudges objects slightly *away* to prevent z-fighting when
+    // rendering color in the next passes with depth LEQUAL
+    this.state_.setPolygonOffset({ factor: 1, units: 1 });
```
shlomnissan: do you need to push away and then pass towards (L137)? i think that in principle nudging objects away should be enough? if that's the case, polygon offset should be encapsulated within `renderDepthOnly`.

src/renderers/webgl_renderer.ts
```diff
@@ -121,27 +115,70 @@ export class WebGLRenderer extends Renderer {
...
+    this.renderDepthOnly(occludingLayers, viewport.camera, frustum);
+    this.state_.setPolygonOffset(null);
+
     this.state_.setDepthMask(false);
-    for (const layer of transparent) {
-      if (layer.state === "ready") {
-        this.renderLayer(layer, viewport.camera, frustum);
+    this.state_.setDepthFunc(this.gl_.LEQUAL);
```
shlomnissan: i assume this needs to be set to `LEQUAL` because `renderDepthOnly` uses `LESS`? so it's set in one function and reset in another. either move both into the `renderDepthOnly` or introduce a more granular state reset (`clear` resets the color/depth buffers but we should probably have two functions, `clear` and `resetState`. render passes typically end with a clear state before the next one begin.

src/renderers/webgl_renderer.ts
```diff
@@ -121,27 +115,70 @@ export class WebGLRenderer extends Renderer {
...
+    this.state_.setPolygonOffset({ factor: -1, units: -1 });
+    this.renderLayers(transparentLayers, viewport.camera, frustum);
+    this.state_.setPolygonOffset(null);
+
+    this.renderedObjects_ = this.renderedObjectsPerFrame_;
+  }
+
+  // TODO: this runs the full shader (incl. texture samples) just to
+  // discard it. Replace with depth-only variants that share each
+  // program's vertex shader but use a cheap fragment shader
```
shlomnissan: do we need a `TODO` for this? i think this fits well in this PR.

src/renderers/webgl_renderer.ts
```diff
@@ -121,27 +115,70 @@ export class WebGLRenderer extends Renderer {
...
+  private renderDepthOnly(layers: Layer[], camera: Camera, frustum: Frustum) {
+    this.state_.setColorMask(false);
+    this.state_.setDepthMask(true);
+    this.state_.setDepthTesting(true);
+    this.state_.setDepthFunc(this.gl_.LESS);
+    this.state_.setStencilTest(false);
```
shlomnissan: that's what i mean by introducing a reset. a pass in opengl typically starts by configuring the state and ends by restoring the state to a reasonable default.

src/renderers/webgl_renderer.ts
```diff
@@ -121,27 +115,70 @@ export class WebGLRenderer extends Renderer {
...
 
-    this.renderedObjects_ = this.renderedObjectsPerFrame_;
+    this.state_.setColorMask(true);
+  }
+
+  private renderLayers(layers: Layer[], camera: Camera, frustum: Frustum) {
+    for (const layer of layers) {
+      this.renderLayer(layer, camera, frustum);
+    }
   }
```
shlomnissan: at this point you might as well inline this function.

src/renderers/webgl_state.ts
```diff
@@ -180,6 +209,26 @@ export class WebGLState {
...
   }
 
+  public setStencilMask(mask: number) {
+    if (this.stencilWriteMask_ !== mask) {
+      this.gl_.stencilMask(mask);
+      this.stencilWriteMask_ = mask;
+    }
+  }
+
+  public setStencilOp(fail: GLenum, zfail: GLenum, zpass: GLenum) {
```
shlomnissan: We always use the same ops right? if that's the case we should probably hardcode these values in this object's constructor or something along these lines.

shlomnissan (changes requested):
only thing that's blocking for me is adding the depth-only fragment shader. i feel it should be part of this PR but happy to unblock if there's a reason you decided to hold off. also, a couple of potential cleanups should probably be addressed before merging but none of them is blocking.

examples/multi_viewport/main.ts
aganders3: My fault - this was left in from some debugging. I think 2048 is enough for now.

src/renderers/webgl_renderer.ts
aganders3: Thanks for the push - it ended up being a smaller change than I anticipated to add the "empty" frag shaders.

src/renderers/webgl_renderer.ts
aganders3: I think you're right, I removed the second polygon offset.

There is still an issue with depth, though. With the default near/far in `PerspectiveCamera` this offset is not enough to prevent artifacts. This is why I set the near plane to be 1.0 in this example. This maybe goes with your comment above regarding a more structured return from `source.getDimensions();` that could provide better data-aware camera parameters.

aganders3:
Thanks! Great suggestions. I think I applied everything that makes sense to include here.

examples/multi_viewport/main.ts
aganders3: Good idea - I opened https://github.com/chanzuckerberg/idetik/issues/745 to capture this.

src/core/layer.ts
aganders3: I added a basic comment here. We might want it to be a doc comment later when docs are improved.

src/renderers/shaders/points_depth_frag.glsl
```diff
@@ -0,0 +1,16 @@
...
+
+precision mediump float;
+
+uniform mediump sampler2DArray u_markerAtlas;
+
+flat in uint v_marker;
+
+// Depth-only counterpart to points_frag.glsl. Point sprites are alpha tested
+// against the marker atlas, so the test has to be repeated here or the whole
+// square sprite would occlude.
```
shlomnissan: i think these GLSL comments are superfluous. the function is implied in the name

src/renderers/shaders/depth_frag.glsl
```diff
@@ -0,0 +1,5 @@
+#version 300 es
+
+// Depth-only pass: the rasterizer writes depth from the vertex stage, so there
+// is nothing for the fragment stage to do.
```
shlomnissan: i think these GLSL comments are superfluous. the function is implied in the name

src/renderers/shaders/index.ts
```diff
@@ -28,8 +30,11 @@ type ShaderCode = {
   vertexDefines?: ReadonlyMap<string, string>;
   fragment: string;
   fragmentDefines?: ReadonlyMap<string, string>;
+  depthOnlyFragment?: string;
```
shlomnissan: i don't think treating depth-only programs separately makes a lot of sense. a lot of this structure feels unnecessary to me unless i'm missing something. why not create depth specific programs (add new entries to `type Shader`? it's already a variant namespace.

src/renderers/webgl_state.ts
```diff
@@ -24,13 +26,17 @@ export class WebGLState {
   private stencilFunc_: GLenum | null = null;
   private stencilRef_: number | null = null;
   private stencilFuncMask_: number | null = null;
+  private polygonOffsetFactor_: number | null = null;
+  private polygonOffsetUnits_: number | null = null;
```
shlomnissan: we can probably hardcode the values and expose `setPolygonOffset(enabled)` since it's only used once, but i don't feel strongly about it

shlomnissan (approved):

shlomnissan:
approved :+1: with a couple of comments. i would like to see the extra depth-specific machinery simplified before this goes in.

PR #733, opened 2026-08-12
refactor: standardize constructors on a single props object

shlomnissan:
### Summary

this PR aligns all public constructors on the options-object style:

- `OrthographicCamera`: replaces the positional (left, right, top, bottom)
  arguments with a props object, and setFrame() now takes the same frame
  fields, removing the inconsistent argument order between the two
- `Idetik`: folds the trailing positional renderer argument into the props
- `ImageRenderable`: replaces positional arguments with a props object and
  renames channels to channelProps to match the rest of the API
- `VolumeRenderable`: accepts optional channelProps at construction
- `PerspectiveCamera`: renames the constructor parameter from options to props

it also makes policy optional on the image, label, and volume layers,
defaulting to the exploration policy, and updates all call sites in tests
and examples.

### Tests & Checks

- all checks have passed.

### DIFF
--- a/src/idetik.ts
+++ b/src/idetik.ts
@@ -27,7 +27,6 @@ type IdetikProps = {
   memoryLimitMB?: number;
   maxConcurrentRequests?: number;
   maxGpuUploadsPerUpdate?: number;
-  renderer?: Renderer;
 };
 
 export type IdetikContext = {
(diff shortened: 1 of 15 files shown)


src/idetik.ts
```diff
@@ -83,10 +84,11 @@ export class Idetik {
    *   The `id` property is optional but useful for referencing specific viewports later.
    * @param params.overlays - Optional array of overlay objects that update each frame (e.g., for HUD elements)
    * @param params.showStats - Optional flag to display performance statistics
+   * @param params.renderer - Optional renderer instance; defaults to a WebGL renderer bound to the canvas
```
srivarra: [question](https://pullpo.io/cc?l=question) 
~Isn't there only a WebGL renderer, or is a renderer instance something else?~

Ah nvm, you could have a different canvas or something else that houses the renderer's instance.

srivarra (approved):
Looks good, just got 1 question.

src/idetik.ts
shlomnissan: this was just left over from the experimental webgpu renderer. it should have been removed along with that renderer. i updated the PR so it's no longer there.

PR #737, opened 2026-08-14
feat: composite volume with images in 3D

aganders3:
In #730 there is an added depth-only pass to help disambiguate blending within a layer (e.g. channels) and between layers (occlusion). This PR also restores the volume to the multi-viewport example.

Here's what it looks like on `main`:
Screenshot 2026-08-20 at 11 34 03 AM

In this PR we conditionally render that depth-only pass into a depth _texture_, which can be subsequently sampled. This is useful in the volume layer where we can terminate rays at the scene depth, allowing us to compose the volume layer with slices or other opaque objects.

Neuroglancer does something similar, but because it renders everything to offscreen framebuffers depth is just another attachment. They then bind the opaque depth to the volume shader as `uDepthSampler`. We render to the default framebuffer, which WebGL does not let you sample depth from. We might be able to optimize similarly with an offscreen pipeline, but I think that's out of scope here.

That's half the battle, but our "premultiplied" blend mode used in the volume layer also assumes front-to-back draw order and attempts to blend "under" with ONE_MINUS_DST_ALPHA. This is why the whole volume is occluded in the example above. Instead this PR reverses the draw order and changes to composite premultiplied "over" (updating the name to `premultipliedOver` as well). The cost here is that we no longer have the opportunity to do certain cross-chunk volume optimizations, such as early ray termination (we can still do early-termination *within* a chunk). We're not making use of this at the moment anyway.

Here's the result with this PR:

Screenshot 2026-08-20 at 11 34 37 AM


src/renderers/webgl_renderer.ts
```diff
@@ -128,10 +135,16 @@ export class WebGLRenderer extends Renderer {
     this.renderDepthOnly(occludingLayers, viewport.camera, frustum);
     this.state_.setPolygonOffset(null);
 
-    this.state_.setDepthMask(false);
-    this.state_.setDepthFunc(this.gl_.LEQUAL);
     this.renderLayers(occludingLayers, viewport.camera, frustum);
 
+    this.sceneDepthBound_ = false;
+    if (transparentLayers.some((l) => l.readsSceneDepth)) {
+      this.renderSceneDepthToTexture(occludingLayers, viewport.camera, frustum);
```
srivarra: `VolumeLayer` always reads the depth of the scene, so volume viewports still allocate and clear an empty depth target every frame. Could we add `occludingLayers.length > 0` into the conditional here as well? It's a very small perf bump in the `multi_viewport` example cause we don't need to clear an empty depth target each frame. But it may be more useful at larger datasets.

srivarra (dismissed):
Looks good, just one small optimization that could be nice.

src/renderers/webgl_renderer.ts
```diff
@@ -170,11 +183,80 @@ export class WebGLRenderer extends Renderer {
...
+      gl.TEXTURE_2D,
+      0,
+      gl.DEPTH_COMPONENT24,
+      w,
+      h,
+      0,
+      gl.DEPTH_COMPONENT,
+      gl.UNSIGNED_INT,
+      null
+    );
```
srivarra: Do you think it's important to have this texture be accounted for in `gpuTextureCount()` and `gpuTextureBytes()`? I did some tests locally and it doesn't seem to get that large (at 3840 x 2160 it takes up like 31.6 MiB).

examples/multi_viewport/main.ts
```diff
@@ -93,6 +94,21 @@ function createSliceViewport(
   };
 }
 
+// Volume layer - no z coordinate to render entire volume
```
shlomnissan: nit: this comment seems to be out of line and not very useful

src/core/layer.ts
```diff
@@ -24,6 +24,7 @@ export interface LayerProps {
   opacity?: number;
   blendMode?: BlendMode;
   occludes?: boolean;
+  readsSceneDepth?: boolean;
```
shlomnissan: this is kinda weird. it's a user-facing value that's only used internally by the volume layer which hard-codes its value. it's not a preference, it's how the layer's shader works. i think adding a protected member is probably better:

```ts
protected readsSceneDepth_: boolean = false;
```

with an accessor for the renderer. and then override it in the volume layer:

```ts
protected override readsSceneDepth_ = true;
```

src/core/layer.ts
```diff
@@ -56,15 +57,24 @@ export abstract class Layer {
...
     this.blendMode = blendMode;
-    this.occludes = occludes ?? blendMode === "none";
+    this.occludes = occludes ?? (blendMode === "none" && !readsSceneDepth);
+    this.readsSceneDepth = readsSceneDepth;
+
+    if (this.occludes && this.readsSceneDepth) {
+      throw new Error(
+        `${this.constructor.name} cannot both occlude and read scene depth.`
+      );
+    }
```
shlomnissan: if you follow the suggestion above, this probably moves to layer's `onAttached` function. this becomes more of an internal invariant.

src/math/sort_by_distance.ts
```diff
@@ -33,3 +33,24 @@ export function sortFrontToBack(
...
+ * Uses the camera position compared to the center of each object's
+ * bounding box to determine distance.
+ *
+ * Use with "over" (src-over-dst) compositing, so nearer objects are
+ * drawn on top by draw order.
+ * @param objects - Array of renderable objects to sort.
+ * @param camera - The camera to calculate distances from
+ * @returns Input objects sorted-in-place from farthest to closest
+ */
+export function sortBackToFront(
```
shlomnissan: what's the point of exporting these functions in a separate module and implementing one in terms of the other? this feels like unnecessary generalization to me. i don't think idetik should provide these kind of helpers. i would implement `sortBackToFront` in the volume layer as a private function and delete this module entirely.

src/renderers/shaders/volume_frag.glsl
```diff
@@ -97,6 +105,19 @@ void main() {
...
+            vec2 uvVp = (gl_FragCoord.xy - u_viewportOrigin) / u_resolution;
+            vec3 ndc = vec3(uvVp * 2.0 - 1.0, sceneWindow * 2.0 - 1.0);
+            vec4 mp = u_mvpInverse * vec4(ndc, 1.0);
+            vec3 occModel = mp.xyz / mp.w;
+            float tScene = dot(occModel - u_cameraPositionModel, RayDirModel);
+            if (tScene < tEnter) discard;
+            tExit = min(tExit, tScene);
+        }
+    }
+
```
shlomnissan, line 120: this is kinda bad imo. six uniforms, four exist to translate between coordinate spaces. also inline variable names are unhelpful (`uvVp`, `mp`,  `occModel`, etc). a couple of suggestions:

- you probably don't need `u_sceneDepthResolution`. the FBO is same size as the canvas, use `texelFetch` instead.
- you also don't need `u_viewportOrigin` and `u_resolution`. the vertex shader already computes clip position so if you pass it as a varying, the fragment shader can do something like `vec2 ndcXY = v_clipPosition.xy / v_clipPosition.w;` which is the standard screen-space-UV from clip space approach.
- you can hoist whatever's left to a function:

```glsl
float distanceAlongRayAtWindowDepth(float windowDepth, vec3 rayOrigin, vec3 rayDir) {
    vec3 ndc = vec3(v_clipPosition.xy / v_clipPosition.w, windowDepth * 2.0 - 1.0);
    vec4 positionModel = u_mvpInverse * vec4(ndc, 1.0);
    return dot(positionModel.xyz / positionModel.w - rayOrigin, rayDir);
}
```

src/renderers/webgl_renderer.ts
```diff
@@ -37,6 +37,12 @@ export class WebGLRenderer extends Renderer {
   private renderedObjectsPerFrame_ = 0;
   private stencilRef_ = 0;
   private currentViewportSize_: [number, number] = [0, 0];
+  private currentViewportOrigin_: [number, number] = [0, 0];
```
shlomnissan: If you follow the previous suggestion this probably goes away.

src/renderers/webgl_renderer.ts
```diff
@@ -37,6 +37,12 @@ export class WebGLRenderer extends Renderer {
...
+
+  private sceneDepthFbo_: WebGLFramebuffer | null = null;
```
shlomnissan: i find passes are much easier to reason about when they're encapsulated in their own object with resources managed there instead of in the main renderer. i would add a `WebGLDepthPass` that owns the FBO and depth texture:

- `bind()` lazily creates/resizes the target, then binds the FBO.
- `unbind()` restores the default framebuffer.
- `bindTexture(unit)` binds the depth texture so the volume pass can sample it.
- `destroy()` deletes the FBO and texture; right now the FBO is never freed.

The pass should own the render target not the render loop.

src/renderers/webgl_renderer.ts
shlomnissan: good catch!

src/renderers/webgl_textures.ts
```diff
@@ -40,12 +44,17 @@ export class WebGLTextures {
     return this.textureCount_;
   }
 
+  public get reservedUnit() {
```
shlomnissan: reserve for what and who? if we were to implement virtual textures we may also need reserved units.

i used to have reserved units in vglx and moved away from it because most of the bindings have the same lifetime but i guess it makes sense here.

i would consider having something like `reservePersistentUnit()` that hands out units from the back `maxUnits - 1` and have the owner keep track of it (i.e. `GLDepthPass` or whatever)

shlomnissan (changes requested):
enough to refactor here that i feel comfortable requesting changes: simplifying the shader code and encapsulating some of this work is worth it.

src/math/sort_by_distance.ts
aganders3: Agree we shouldn't make these public - I just implemented here as its where the previous version was. We only need one right now though and it's only used in one place, so happy to move it.

src/renderers/webgl_renderer.ts
aganders3: Agree - thanks!
I don't think it applies to the `multi_viewport` example though, does it? We do have occluding layers in the volume viewport there.

src/renderers/webgl_renderer.ts
aganders3: Yes, I think this is a worthwhile change.

src/renderers/shaders/volume_frag.glsl
aganders3: Much nicer!

src/renderers/webgl_textures.ts
aganders3: I think this mechanism makes more sense with the `WebGLDepthPass` encapsulation.

examples/multi_viewport/main.ts
```diff
@@ -91,6 +92,20 @@ function createSliceViewport(
...
+  sliceCoords: {
+    get t() {
+      return currentTime.t;
+    },
+    c: [0, 1],
+  },
+  policy: createPlaybackPolicy({ lod: { min: volumeLod, max: volumeLod } }),
+  channelProps,
+});
+volumeLayer.opacityMultiplier = 0.01;
```
shlomnissan, line 107: this value feels random. is there a way to derive it so opacity multiplier 1.0 means a ray crossing the whole volume at full intensity saturates about once? maybe we can open an issue.

src/core/layer.ts
```diff
@@ -58,6 +58,8 @@ export abstract class Layer {
   public blendMode: BlendMode;
   public occludes: boolean;
 
+  protected readsSceneDepth_ = false;
```
shlomnissan: this flag is not about what the shader does, it's more of a contract with the renderer. so maybe renaming to `requiresSceneDepth_` is better.

src/layers/volume_layer.ts
```diff
@@ -31,6 +30,9 @@ const INTERACTIVE_STEP_SIZE_SCALE = 2.0;
 export class VolumeLayer extends Layer implements ChannelsEnabled {
   public readonly type = "VolumeLayer";
 
+  // rays are terminated at the nearest opaque surface
```
shlomnissan: this comment describe an effect this line doesn't even produce. i'm generally not a big fan of inline comment but rephrasing can make it a little more useful:

```suggestion
  // the ray marcher uses scene depth to stop rays at occluding geometry
```

src/layers/volume_layer.ts
```diff
@@ -279,3 +281,18 @@ export function poolKeyForChunk(chunk: Chunk) {
     `align${chunk.rowAlignmentBytes}`,
   ].join(":");
 }
+
+function sortBackToFront(objects: VolumeRenderable[], camera: Camera) {
```
shlomnissan, line 284: :+1:

src/renderers/shaders/volume_frag.glsl
shlomnissan: MUCH NICER!

src/renderers/shaders/volume_frag.glsl
```diff
@@ -46,6 +47,18 @@ uniform vec4 u_valueScale;
 uniform vec4 u_channelOpacity;
 uniform vec3 u_color[4];
 
+// Optionally terminate rays at some predetermined scene depth
+uniform bool u_hasSceneDepth;
+uniform sampler2D u_sceneDepth;
+uniform mat4 u_mvpInverse;
+
+// How far along the ray the occluding surface is (precomputed scene depth)
```
shlomnissan: bad comment. the first part restates the name and "precomputed scene depth" documents the caller not the function. i would just delete it.

src/renderers/webgl_depth_pass.ts
```diff
@@ -0,0 +1,99 @@
+import { WebGLState } from "./webgl_state";
+import { WebGLTextures } from "./webgl_textures";
+
+export class WebGLDepthPass {
+  public readonly textureUnit: number;
+
+  private readonly gl_: WebGL2RenderingContext;
+  private readonly state_: WebGLState;
+  private readonly textures_: WebGLTextures;
```
shlomnissan: i don't think this is the right shape.

to support this depth pass you added a second texture model to `WebGLTextures` (`createDepthTarget`, `bindTarget`, `disposeTarget`, and the targets map). none of it has to do with what `WebGLTextures` is for: a texture-object cache with an upload pipeline.

the changes to `disposeAll` is a clear tell because it has to enumerate things it doesn't own.

meanwhile `WebGLDepthPass` already mostly owns the resource. it owns the raw texture handle, the framebuffer, the sizing logic, etc. it only delegates creation and deletion so it reads as split to me.

imo a better shape is letting the pass create/bind/delete its own depth texture. so basically:
- `WebGLTextures` drops the public methods and the map, back to what it was
- unit allocate stays centralized in `reservePersistentUnit()`. the renderer owns the unit and takes care of binding it `bindTexture()` is deleted.
- `WebGLDepthPass` handles depth texture creation and active/bind.

i know it means that we're not accounting for the chunk memory budget but i can make an argument that it's better because this is not evictable memory? either way a handful of MBs is probably not a big deal.

src/renderers/webgl_depth_pass.ts
```diff
@@ -0,0 +1,99 @@
...
+  public renderToTexture(width: number, height: number, draw: () => void) {
+    this.bindFBO(width, height);
+    this.state_.setDepthMask(true);
+    this.state_.setDepthFunc(this.gl_.LESS);
+    this.gl_.clear(this.gl_.DEPTH_BUFFER_BIT);
+    draw();
+    this.unbindFBO();
+  }
+
+  private bindFBO(width: number, height: number) {
```
shlomnissan: i feel like `bindFBO` and `unbindFBO` can be inlined. they don't and probably won't have any other callers and `gl_.bindFramebuffer` is already clear about what's going on.

src/renderers/webgl_depth_pass.ts
```diff
@@ -0,0 +1,99 @@
...
+    if (this.texture_) this.textures_.disposeTarget(this.texture_);
+    if (this.framebuffer_) this.gl_.deleteFramebuffer(this.framebuffer_);
+    this.texture_ = null;
+    this.framebuffer_ = null;
+    this.width_ = 0;
+    this.height_ = 0;
+  }
+
+  private ensureTarget(width: number, height: number) {
+    if (this.texture_ && this.width_ === width && this.height_ === height) {
```
shlomnissan: i would split this to improve clarity:

- move the check to the call site (maybe even wrap it in a function `hasValidTarget`)
- rename the function to `createRenderTarget` or something along those lines
- depth texture creation should probably live here as well

something like this:

```ts
if (!this.hasValidTarget(width, height)) {
  this.createRenderTarget(width, height);
}

this.gl_.bindFramebuffer(this.gl_.FRAMEBUFFER, this.framebuffer_);
```

src/renderers/webgl_renderer.ts
```diff
@@ -121,9 +126,23 @@ export class WebGLRenderer extends Renderer {
...
 
+    const needsSceneDepth =
+      occludingLayers.length > 0 &&
+      nonOccludingLayers.some((l) => l.readsSceneDepth);
+
+    if (needsSceneDepth) {
+      this.depthPass_.renderToTexture(this.width, this.height, drawOccluders);
+      this.depthPass_.bindTexture();
+      this.resetState();
+    }
```
shlomnissan: i'm not sure why `renderPrePass` and `renderToTexture` are grouped.

`renderPrePass` uses none of the depth pass class state. these functions are grouped by theme instead of by data but they're doing unrelated jobs.

i also think state ownership is split across now. i think the renderer should be responsible for managing the state in general (no need to pass the state object to the depth pass):

```ts

// add two functions to the renderer

private renderDepthPrePass(layers: Layer[], camera: Camera, frustum: Frustum) {
  this.state_.setColorMask(false);
  this.state_.setDepthFunc(this.gl_.LESS);
  this.state_.setPolygonOffset(true);

  this.drawDepthOnly(layers, camera, frustum);

  this.resetState();
}

private renderDepthTarget(layers: Layer[], camera: Camera, frustum: Frustum) {
  this.depthPass_.bindTarget(this.width, this.height);
  this.gl_.clear(this.gl_.DEPTH_BUFFER_BIT);
  this.state_.setDepthMask(true);
  this.state_.setDepthFunc(this.gl_.LESS);
  
  this.drawDepthOnly(layers, camera, frustum);

  this.depthPass_.unbindTarget();
  this.resetState();
}
```

and call them from here.

shlomnissan (changes requested):

shlomnissan:
my blocking comments:

https://github.com/chanzuckerberg/idetik/pull/737#discussion_r3896939956
https://github.com/chanzuckerberg/idetik/pull/737#discussion_r3897649799

all other comments are just nits.

aganders3:
Thanks! I wasn't expecting another review here yet, but I was probably headed in some wrong directions so the earlier feedback is helpful. I'll try to incorporate these suggestions and re-request review when done.

examples/multi_viewport/main.ts
aganders3: 100% agree, I don't understand this parameter at all really. There should be a data-independent way of setting a reasonable default, or at least based on the size/scale of the data. I'll take a look.

src/renderers/webgl_renderer.ts
aganders3: @shlomnissan's comment below changed my mind on this: this memory cannot be evicted, which is the main use for the memory tracking at the moment. I think it's small enough to ignore for now, but we can revisit if it makes the memory budget unintuitive.

aganders3:
Okay I think this is ready for another review. Thanks for all the attention on this so far.

I changed the name of `WebGLDepthPass` to `WebGLDepthTarget`, which I think is more accurate regarding the responsibilities. "Pass" reads more like a verb to me that would including drawing, etc. I'm also open to changing back if this is an established convention or if you disagree.

src/renderers/webgl_renderer.ts
```diff
@@ -121,8 +129,16 @@ export class WebGLRenderer extends Renderer {
...
+
+    this.currentViewportHasSceneDepth_ =
+      occludingLayers.length > 0 &&
+      nonOccludingLayers.some((l) => l.requiresSceneDepth);
+
+    this.renderDepthPrePass(occludingLayers, viewport.camera, frustum);
+
+    if (this.currentViewportHasSceneDepth_) {
+      this.renderDepthTarget(occludingLayers, viewport.camera, frustum);
+    }
```
shlomnissan, line 141: `renderDepthPrePass` and `renderDepthTarget` reads as two flavors of the same thing. maybe something like this is clearer:

```ts
this.primeDepthBuffer(occludingLayers, viewport.camera, frustum);

if (needsSceneDepth) {
  this.renderDepthTexture(occludingLayers, viewport.camera, frustum);
}
```

shlomnissan (approved):
looks good :+1:

one non-blocking concern!

the rendering code is accumulating logic that isn't self-describing. i already re-derive the stencil code from scratch every couple of months and this feature adds two more things future-me will find confusing:

- why the depth pre-pass always runs. the answer is an invariant: the color pass never writes depth, so occlusion is established up front even when a viewport is simple enough not to need it. but why does the color pass never writes depth? nothing in the code says that.
- why the depth target exists at all. it's not redundant with the pre-pass, it's there because the volume shader needs depth as data to clamp rays mid-fragment which the hardware depth test can't do. again invisible unless you already know it and easy to forget.

none of this is a problem with the implementation. the approach is right. it's that the why lives in our heads and in the PR discussion instead of the repo.

maybe we need a document explaining the visibility model in detail (color pass never writes depth, pre-pass feeds the fixed-function test, depth target feeds shaders that read depth, stencil resolves coplanar overlap, etc). maybe we can slo think harder about naming that hints at purpose (something that ties the depth target to scene depth for shader sampling rather than depth generically.

either way, this is good to go, but i think documenting this will help us in the future and better to do it now before we forget. maybe we can fold it into a guide on the docsite?

aganders3:
Thanks! We chatted in our sync but to summarize: I agree the structure is confusing and too much is in our collective memory. My concern is more that the structure makes sense only for exactly our current feature set, but may be inflexible and need to be continually re-thought when new features are added that are not compatible.

I think starting by documenting the renderer in detail is probably the right approach.

PR #738, opened 2026-08-17
refactor: replace the LabelColorMap class with plain color map props

shlomnissan:
### Summary

this PR removes the `LabelColorMap` class from the public API. color maps are
now plain `LabelColorMapProps` objects validated internally, following the
same pattern as channel props. `LabelLayer` still validates eagerly at
construction, and its `colorMap` getter returns the validated shape.

### Tests & Checks

- all checks have passed

### DIFF
--- a/src/objects/renderable/label_image_renderable.ts
+++ b/src/objects/renderable/label_image_renderable.ts
@@ -37,14 +37,14 @@ export type LabelColorMapProps = {
   cycle?: ReadonlyArray<ColorLike>;
 };
 
-export type ValidatedLabelColorMap = {
-  lookupTable: ReadonlyMap<number, Color>;
-  cycle: ReadonlyArray<Color>;
+export type LabelColorMap = {
+  readonly lookupTable: ReadonlyMap<number, Color>;
+  readonly cycle: ReadonlyArray<Color>;
 };
 
 export function validateColorMap(
   props: LabelColorMapProps = {}
-): ValidatedLabelColorMap {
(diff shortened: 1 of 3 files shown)


src/objects/renderable/label_image_renderable.ts
```diff
@@ -34,24 +34,28 @@ function validateCycle(cycle?: ReadonlyArray<ColorLike>): ReadonlyArray<Color> {
...
+  cycle?: ReadonlyArray<ColorLike>;
 };
 
-export class LabelColorMap {
-  public readonly lookupTable: ReadonlyMap<number, Color>;
-  public readonly cycle: ReadonlyArray<Color>;
+export type ValidatedLabelColorMap = {
+  lookupTable: ReadonlyMap<number, Color>;
+  cycle: ReadonlyArray<Color>;
+};
```
srivarra: This allows us to replace the colormap lookup table directly, but doing so doesn’t update the GPU textures. Should assigning a new map automatically call `setColorMap`()?

srivarra (changes requested):
Do we want the ability to change colormaps on the fly?

src/objects/renderable/label_image_renderable.ts
shlomnissan: great point, but making assignment trigger a GPU update would require accessors or a wrapper object. it's probably better to just make the fields themselves `readonly`. i'll push an update.

srivarra (approved):

PR #747, opened 2026-08-31
feat(core)!: accept constructed viewport instances

### LINKED ISSUE #740: Viewer State, Serialization and Deserialization
# Idetik viewer state and shareable links

## Problem or use case

Idetik would benefit from a portable state format so applications incorporating the viewer can create shareable links. Because Idetik is a library to build other tools, its state must remain separate from application layout and workflow state.

The application built with Idetik owns its application state and delegates viewer state to Idetik, and a shared link should contain both.

## Proposed improvement

Idetik should capture and restore the parts of the viewer it owns:

- Dataset source descriptors
- Coordinate scopes and bindings
- Viewports
- Navigation, including camera projection, pose, and control state
- Layers and layer ordering
- Channel selection and appearance
- Loading policies, including prefetch and level-of-detail settings
- Image and volume rendering settings

The application owns everything that's not the Idetik viewer, this would include: 
- UI and pane layout
- playback scheduling
- workflow state
- and creation of the shareable link. 

It can refer to Idetik layers, viewports, and coordinate scopes by a stable ID instead of keeping another copy of their values.

**State model**

The state itself is plain JSON, and `Zod` checks it when ingesting. Datasets and coordinate scopes appear once in the JSON, and layers refer to them by ID.

```ts
type IdetikState = {
  datasets: DatasetSourceState[];
  coordinateScopes: CoordinateScopeState[];
  viewports: ViewportState[];
};
```

A dataset source descriptor holds only what Idetik needs to reopen the data:
- a stable ID
- source type
- location
- format

Idetik reloads image metadata from the source, and stable IDs help keep references intact across a round trip.

**Coordinate scopes**

The coordination model comes from Mark Keller's [`use-coordination`](https://github.com/keller-mark/use-coordination) and the [paper](https://ieeexplore.ieee.org/document/10771108). I used it while building Vitessce dashboards a few years ago, and it worked well. We only (maybe) need the idea that values live in scopes, and views refer to those scopes.

Multiple views may observe one coordinate. For example, XY, XZ, YZ, and volume views may share one time value while retaining different spatial positions.

Copying `t = 42` into every layer preserves the current frame but loses the fact that those values must continue changing together (such as in the [`multi_viewport` example](https://github.com/chanzuckerberg/idetik/tree/main/examples/multi_viewport)). A coordinate scope stores the value once, and layers reference it by stable ID:

```json
{
  "coordinateScopes": [{ "id": "global-time", "axis": "t", "value": 42 }],
  "viewports": [
    {
      "id": "xy",
      "layers": [
        { "id": "xy-image", "coordinateScopes": { "t": "global-time" } }
      ]
    },
    {
      "id": "volume",
      "layers": [
        {
          "id": "volume-image",
          "coordinateScopes": { "t": "global-time" }
        }
      ]
    }
  ]
}
```

When Idetik restores the state, both layers receive the same `CoordinateScope` instance. Views can share `t` while keeping `x`, `y`, and `z` independent. We can use the same pattern for other viewer values as well. I'm not sure if we want to implement a subset of `use-coordination` or add it as a dependency.

**Interface**

```ts
const state = idetik.toState();

idetik.applyState(compatibleState);

const restored = await Idetik.fromState(state, {
  canvas,
  resolveViewportElement,
});
```

- `toState()` captures the live viewer as JSON-serializable plain data.
- `applyState()` updates an existing compatible viewer without tearing it down. Changes to the dataset or some other ones, like adding another viewport, still require a rebuild.
- `fromState()` builds a new viewer from a state document. It resolves datasets, rebuilds coordinate scopes, navigation, and layers, then returns a stopped Idetik runtime.

Shareable links only really need `toState()` and `fromState()`.

But `applyState()` lets us apply compatible changes to a running viewer (like moving the camera around, or zooming in / out). I found this useful when trying seeing how this would work, but not sure if it's necessary.

**Link serialization**

The application using Idetik combines both states before it creates the link:

```ts
type SharedView<ApplicationState> = {
  idetik: IdetikState;
  application: ApplicationState;
};
```

Each application defines its own `ApplicationState`. It can contain UI state and references to Idetik IDs without duplicating viewer values.

From there, the application can turn the artifact into JSON, compress it (optional), and URL-encode it. Opening the link reverses that process and passes the Idetik part to `fromState()`. Keeping link handling in the application means the same Idetik state can also go into a file or server-backed snapshot so the viewer can automatically open up to an interesting angle, time point or view of the OME-Zarr.

Stuff like DOM elements, caches, loaded chunks, renderer objects should stay outside of the state.

## Misc

I took a look at Zustand and it can definitely work here. Its [`persist` middleware](https://zustand.docs.pmnd.rs/reference/integrations/persisting-store-data#createjsonstorage) can serialize a store to JSON and hydrate it from custom storage. While Zustand can serialize for us, it might require rebuilding Idetik around its store. We'd still need Zod as well though.

<details>
<summary>Zustand example</summary>

```ts
import { createStore } from "zustand/vanilla";
import { persist, createJSONStorage } from "zustand/middleware";

const store = createStore<IdetikStore>()(
  persist(
    (set) => ({
      // state and actions
    }),
    {
      name: "idetik-viewer",
      storage: createJSONStorage(() => localStorage),

      // Persist the document, not actions or runtime objects.
      partialize: (store) => ({
        viewer: store.viewer,
      }),
    }
  )
);
```

</details>

<details>
<summary>Further reading</summary>

- [Zustand persistence](https://zustand.docs.pmnd.rs/reference/integrations/persisting-store-data#createjsonstorage)
- [Use-Coordination paper](https://doi.org/10.1109/VIS55277.2024.00041) and [reference implementation](https://github.com/keller-mark/use-coordination)
- [Vitessce view config JSON](https://vitessce.io/docs/view-config-json/)
- [Neuroglancer StateBuilder](https://www.caveconnecto.me/nglui/usage/statebuilder/)

</details>

srivarra:
Make viewport construction imperative and add camera state

### Summary

Applications now construct `Viewport` instances before passing them to `Idetik`. This removes the `ViewportProps` conversion path and lets applications own viewport construction.

Temporary integration: this branch also carries [#757](https://github.com/chanzuckerberg/idetik/pull/757)'s resident multiscale rendering changes so [idetik-dca-viewer#220](https://github.com/chanzuckerberg/idetik-dca-viewer/pull/220) can pin one core revision containing both changes.

This is a breaking change. The `Idetik` constructor and `Idetik.addViewport()` now accept `Viewport` instances, and `Idetik.getViewport()` has been removed.

### Related Issue

Related to #740

### Tests & Checks

- Ran `npm run compile`
- Ran `npx vitest run`, tests passed.


shlomnissan:
would `feat` tag automatically release `v1.0.0`? i don't think we want that

shlomnissan:
i think this should be split into two PRs:
- imperative viewport construction
- camera state serialization

srivarra:
@shlomnissan 

> would feat tag automatically release v1.0.0? i don't think we want that

I don't think so as #689 seems to keep them within the minor range.

> i think this should be split into two PRs.

Makes sense to me.

examples/image2d_from_omezarr4d_hcs/main.ts
```diff
@@ -48,19 +49,16 @@ const camera = new OrthographicCamera({
...
-  canvas: document.querySelector<HTMLCanvasElement>("canvas")!,
-  viewports: [
-    {
-      camera,
-      cameraControls: new PanZoomControls(camera),
-    },
-  ],
+const canvas = document.querySelector<HTMLCanvasElement>("canvas")!;
+const viewport = new Viewport({
+  element: canvas,
```
shlomnissan: - passing the canvas as feels like boilerplate but it's the right approach. i feel like the name can be more self-describing (e.g., `domElement` or `container`) but i don't know if it actually adds any value.
- i did notice that `getBufferReact` in viewport cast to `HTMLCanvasElement` unconditionally. it's a pre-existing issue but might be worth fixing in this PR.

examples/image2d_from_omezarr4d_hcs/main.ts
```diff
@@ -48,19 +49,16 @@ const camera = new OrthographicCamera({
...
+  camera,
+  cameraControls: new PanZoomControls(camera),
 });
+const idetik = new Idetik({ canvas, viewports: [viewport] });
 
 idetik.start();
 
-const viewport = idetik.viewports[0];
```
shlomnissan, line 63: this line validates the motivation behind the change. i hoped we could also remove `get viewports()` but it's used by the overlays. i think we should get rid of those as well but that's a whole different thing.

examples/image_2d/main.ts
```diff
@@ -158,24 +159,31 @@ const orbitTarget = vec3.fromValues(
 type Projection = "orthographic" | "perspective";
 
 const layers = [imageLayer];
+const canvas = document.querySelector<HTMLCanvasElement>("#canvas")!;
 
-function makeViewportProps(projection: Projection) {
+function makeViewport(projection: Projection): Viewport {
```
shlomnissan: @aganders3 non-blocking since this PR only adapts existing code. #742 added this toggle to explore chunk visibility/LOD behavior under different projections but i'm not sure examples are the right place for exploration tooling. we still need to decide the scope of the examples. my feeling is they should demonstrate capabilities, and experiments like 2D images with 3D controls belong in a sandbox or a dedicated example. fine for now since it's the only example that demonstrate add/remove viewports at runtime.

one note on `makeViewport` itself: since all its inputs are module-scope constants the indirection isn't buying much. @srivarra i think we could construct the two viewport instances at module scope and make the toggle a pure `removeViewport`/`addViewport` swap. that also demonstrates re-attaching a previously removed viewport. fine to fold into this PR or defer not a big deal either way.

examples/volume_rendering/main.ts
```diff
@@ -51,18 +52,19 @@ const volumeLayer = new VolumeLayer({
   channelProps,
 });
 
+const canvas = document.querySelector<HTMLCanvasElement>("#canvas")!;
+const viewport = new Viewport({
```
shlomnissan: if the viewport is not used anywhere in the file i think we should prefer inlining when we construct the runtime, but i don't feel too strongly about it.

```ts
const idetik = new Idetik({
  canvas,
  viewports: [
    new Viewport({
      element: canvas,
      camera,
      cameraControls: new OrbitControls(camera, {
        radius: 100,
        target: vec3.fromValues(40, 40, 10), // Volume center,
      }),
      layers: [volumeLayer],
    }),
  ],
  showStats: true,
});
```

src/core/viewport.ts
```diff
@@ -86,14 +74,43 @@ export class Viewport {
...
+    if (this.context_) {
+      for (const layer of this.layers_) {
+        layer.onDetached(this.context_);
+      }
+    }
+    this.layers_ = [];
+  }
+
+  public attachToIdetik(context: IdetikContext): void {
+    if (this.context_) {
```
shlomnissan: the fact that the viewport is managing the context at all doesn't make a lot of sense to me. 

what if we invert this so the viewport never needs the context? layers would own their attachment state directly (we can replace `attached_` boolean with `context_: IdetikContext | null` where null means unattached.

the runtime's main loop then reconciles lazily: each frame as it iterates viewports and their layers any layer with a null context gets `onAttached` called with the runtime's context. detach stays eager but no longer needs plumbing: since the layer holds its own context `removeLayer` just calls `layer.onDetached()` and `removeViewport` sweeps the viewport's layers the same way.

then `attachToIdetik`/`detachFromIdetik` and the viewport's `context_` disappear. `addLayer` becomes a pure push, `viewport.ts` no longer imports anything from `idetik.ts` and the attach-rollback try/catch blocks (here and in the Idetik constructor) go away.

one concern is the "layers can't be shared in multiple viewports" invariant but that's something that viewport validation can handle. 

i think this should be integrated as part of this change or as an immediate follow-up.

src/core/viewport.ts
```diff
@@ -203,26 +220,8 @@ export function validateNewViewport(
   }
 }
 
-function validateViewportProps(viewportProps: ResolvedViewportProps[]): void {
-  for (let i = 0; i < viewportProps.length; i++) {
-    validateNewViewport(viewportProps[i], viewportProps.slice(0, i));
+export function validateViewports(viewports: Viewport[]): void {
```
shlomnissan: `validateNewViewport` and `validateViewports` enforce cross-viewport invariants that only the runtime can see so they should be private (module-scoped?) functions in `idetik.ts`. `validateViewports` is a one-caller wrapper and can be inlined (or the constructor could just call `this.addViewport(v)` per viewport which already does validate + attach and would remove the duplicated rollback block too). 

`validateNewViewport` rename to `validateViewport` once it's private. tests can move as well.

shlomnissan (changes requested):
overall looks good :+1:

the only hard blocker i have is moving `validateNewViewport` and `validateViewports`  to the runtime module, but i do want to get your thoughts on removing the context from the viewport object. do you think it should land in this PR or an immediate follow-up? i think the change will be relatively small but it's up to you.

https://github.com/chanzuckerberg/idetik/pull/747#discussion_r3906488330
https://github.com/chanzuckerberg/idetik/pull/747#discussion_r3906525735

src/core/viewport.ts
srivarra: These points make sense to me, I'll move the validation into `idetik.ts`.

src/core/viewport.ts
srivarra: I think this makes sense to me, as the `Viewport` should own its layers without storing a reference to Idetik's runtime state. And to restate, each `Layer` owns it's attachment context, and Idetik can attach pending layers before rendering.

I'll try to integrate this change here.

examples/image2d_from_omezarr4d_hcs/main.ts
srivarra: I like the name `domElement` so I'll pick that one. Might as well go through and fix the viewport cast too. Was your suggestion to rename `canvas` or the `element` prop?

srivarra:
@shlomnissan 

I think this is ready for a re-review.

I removed `validateNewViewport` and `validateViewports` from `viewport.ts` and added `validateViewport` in `idetik.ts`. There's a new private function `renderViewport` which checks layer attachment each frame.

I also removed the context from the `Viewport` so it doesn't import from `idetik.ts` and `addLayer` just pushes a `Layer`.

Lemme know what you think.

examples/image_2d/main.ts
```diff
@@ -158,34 +159,39 @@ const orbitTarget = vec3.fromValues(
...
+  right: xMax,
+  top: yMin,
+  bottom: yMax,
+});
+const perspectiveCamera = new PerspectiveCamera({
+  fov: FOV_DEGREES,
+  near: orbitRadius / 100,
+  far: orbitRadius * 10,
+});
+const viewports: Record<Projection, Viewport> = {
```
shlomnissan: nit: is it necessary to specify the type in this case?

examples/image_2d/main.ts
```diff
@@ -329,10 +335,9 @@ debugFolder
   .add(projectionState, "projection", ["orthographic", "perspective"])
   .name("Projection")
   .onChange((projection: Projection) => {
-    const current = idetik.viewports[0];
-    current.removeAllLayers();
-    idetik.removeViewport(current);
-    idetik.addViewport(makeViewportProps(projection));
+    idetik.removeViewport(viewport);
```
shlomnissan, line 338: there's a bug when switching viewports.

if you load the example, resize the window, and switch the camera, the state of the camera's aspect ratio is stale (see screenshot). i think you can fix it by calling `viewport.updateSize()` in `Idetik.addViewport`.

Image

examples/multi_viewport/main.ts
```diff
@@ -101,9 +102,9 @@ new Idetik({
   canvas: document.querySelector<HTMLCanvasElement>("#canvas")!,
   viewports: [
     createSliceViewport("slice-xy", "XY", xRange, yRange),
-    {
+    new Viewport({
```
shlomnissan, line 122: nit: viewports are bound to dom elements so they are not order dependent. changing this
```ts
viewports: [
  createSliceViewport("slice-xy", "XY", xRange, yRange),
  new Viewport({id: "3d", ...}),
  createSliceViewport("slice-xz", "XZ", xRange, zRange),
  createSliceViewport("slice-yz", "YZ", yRange, zRange),
],
```
to something like this
```ts
viewports: [
  createSliceViewport("slice-xy", "XY", xRange, yRange),
  createSliceViewport("slice-xz", "XZ", xRange, zRange),
  createSliceViewport("slice-yz", "YZ", yRange, zRange),
  new Viewport({id: "3d", ...}),
],
```

makes this block more readable.

examples/volume_rendering/main.ts
```diff
@@ -51,17 +52,19 @@ const volumeLayer = new VolumeLayer({
...
 });
 
+const canvas = document.querySelector<HTMLCanvasElement>("#canvas")!;
 const idetik = new Idetik({
-  canvas: document.querySelector<HTMLCanvasElement>("#canvas")!,
+  canvas,
   viewports: [
-    {
+    new Viewport({
+      domElement: canvas,
```
shlomnissan, line 60: the boilerplate is nagging me. making `domElement` optional and having deferred initialization for the empty case is more streamlined. on the other hand setting the dom element explicitly feels more structured. what do you think?

src/core/layer.ts
```diff
@@ -50,7 +50,8 @@ export abstract class Layer {
     RenderableObject[]
   >();
   private state_: LayerState = "initialized";
-  private attached_ = false;
+  private attachment_: { context: IdetikContext; viewport: Viewport } | null =
```
shlomnissan: this gives a layer a reference back to its owner which inverts the ownership direction (viewports own layers). as far as I can tell it exists only so `Viewport.removeLayer` and `removeAllLayers` can call `onDetached` without knowing whether this viewport is the one that's actually attached. the stored `context` is there for the same reason: those callers have none to pass.

i think the idetik runtime can own this instead:

- `Layer`: just an `attached` boolean. `onAttached`/`onDetached` both take the context, nothing stored.
- `Idetik`: owns the `(layer, viewport)` pairs it attached. uses them for the ownership check in `renderViewport`, detaches them in `removeViewport`, and each frame detaches any whose viewport is gone or no longer lists the layer.
- `Viewport.removeLayer`/`removeAllLayers` just splice.

the one tradeoff is `Viewport.removeLayer` would detach on the next frame instead of synchronously so the "eagerly detaches" assertions in `viewport.test.ts` and `idetik.test.ts` would change. a one-frame delay releasing chunk subscriptions seems fine?

shlomnissan (changes requested):
i haven't reviewed the entire PR yet. holding off for your feedback on: 

https://github.com/chanzuckerberg/idetik/pull/747#discussion_r3962265572

it's borderline blocking but i want to get your thoughts first.

examples/image_2d/main.ts
srivarra: nope, I can remove it.

examples/volume_rendering/main.ts
srivarra: I think making domElement optional makes sense. If omitted, we could use the runtime's canvas each time the viewport is added, updating the input target and camera size, and explicitly supplied elements would stay fixed.

We already construct layers independently and finish their setup on attachment, so this would follow a similar pattern.

I'll try it out and see if I run into any problems.

---


I looked into this a bit more. Making it optional would mean deferring the viewport's element-dependent setup until registration, including input handling and sizing. I think that's more than what I want to do in this pr, so I'll take a longer look in a followup pr.

src/core/layer.ts
srivarra: I spent some time playing around with this implementation, and I couldn't see a noticeable visual difference from deferring detachment to the next frame. It still feels instant enough to me.

I think it makes sense for the runtime to track which viewport a layer is attached to, while the layer handles its own chunk fetching.

The main caveat I found is moving a layer between separate Idetik runtimes. Within one runtime, the move can happen in one render pass because removals are processed before new attachments. Across runtimes, runtime B can't attach the layer until runtime A releases it.

I tested this with A stopped. Removing the layer from A's viewport left its attachment intact, so B threw an ownership error and stopped scheduling frames, even though `running` still reported `true`. Resuming A or removing A's viewport released the attachment, explicitly stopping and starting B then completed the transfer.

I'm not sure if multiple runtimes is a recommended way to use Idetik tbh.

So I don't think the frame delay is a visual concern during normal rendering, but maybe the stopped-runtime behavior and recovery steps are worth making explicit later on.

shlomnissan:
@srivarra this PR includes some of ashley's resident chunks code. can you clean it up so i can review it as a single unit?

PR #749, opened 2026-08-31
feat(core): expose orbit controls interaction state

srivarra:
### Summary

Applications need the camera state to be serializable for persistence and shareable links. Perspective and orthographic cameras now serialize their projection settings and transforms and can be restored with `fromJSON()`. Mostly follows from what `three.js` does.

`OrbitControls.toJSON()` serializes its radius, angles, and target. Applications restore the camera first, then construct `OrbitControls` with the saved orbit parameters.

### Related Issue

Related to #740

### Tests & Checks

- Added JSON round-trip tests for perspective and orthographic cameras.
- Added an orbit-control restoration test.
- Ran `npm run compile`
- Ran `npx vitest run`


src/objects/cameras/camera.ts
```diff
@@ -2,6 +2,34 @@ import { Node } from "../../core/node";
...
+    }
+  | {
+      type: "OrthographicCamera";
+      width: number;
+      height: number;
+      near: number;
+      far: number;
+      orientation: SliceOrientation;
+      transform: CameraTransformJSON;
+    };
```
shlomnissan: i think there's a clear split here: interaction state (the transform and orbit parameters) and configuration (the projection parameters). i don't think we need to capture configuration because the application derives it from the source anyway.

interestingly `camera.transform` is already public (the test itself uses it directly) so applications can already read and restore that part.

the only real gap is orbit controls state, which is private. but maybe we don't need `toJSON`/`OrbitControlsJSON` for that either. could this entire change reduce to getters on orbit controls for radius/yaw/pitch/target?

read-and-restore then round-trips through the existing constructor, apps own their own encoding, and we don't commit to a stable wire format.

the exception would be if library-owned session persistence is actually on the roadmap, but we already kinda agreed that's probably the wrong direction.

what do you think?

shlomnissan (changes requested):
i'm really sorry about all the back-and-forth on this.

src/objects/cameras/camera.ts
srivarra: > could this entire change reduce to getters on orbit controls for radius/yaw/pitch/target?

yess, that simplifies this by so much, I like that, no need to go about developing a versioned json schema.

shlomnissan (approved):
LGTM :+1:

PR #754, opened 2026-09-09
docs: document loading policies on a synthetic topic page

shlomnissan:
### Summary

this PR adds a loading policies page to the docsite. loading policies are free
factory functions so i had to create "topic pages" that can group related functions
into custom pages.

### Tests & Checks

- all checks have passed
- visual verification

Screenshot 2026-09-09 at 1 17 40 PM

### DIFF
--- a/src/core/image_source_policy.ts
+++ b/src/core/image_source_policy.ts
@@ -1,13 +1,24 @@
+/**
+ * A category of chunk request ordered by a loading policy.
+ */
+export type PriorityCategory =
+  | "fallbackVisible"
+  | "prefetchTime"
+  | "visibleCurrent"
+  | "fallbackBackground"
+  | "prefetchSpace";
+
 const ALL_CATEGORIES = [
   "fallbackVisible",
   "prefetchTime",
   "visibleCurrent",
   "fallbackBackground",
   "prefetchSpace",
-] as const;
+] as const satisfies readonly PriorityCategory[];
 
-export type PriorityCategory = (typeof ALL_CATEGORIES)[number];
-
+/**
+ * @hidden
+ */
 export type ImageSourcePolicyProps = {
   profile?: string;
   prefetch: {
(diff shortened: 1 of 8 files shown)


src/core/image_source_policy.ts
```diff
@@ -24,65 +35,51 @@ export type ImageSourcePolicyProps = {
...
   profile: string;
+  /** How far to prefetch beyond the visible region per axis. */
   prefetch: {
+    /** Chunks to prefetch along x. */
     x: number;
+    /** Chunks to prefetch along y. */
     y: number;
+    /** Chunks to prefetch along z. */
     z: number;
+    /** Timepoints to prefetch ahead. */
```
aganders3: I don't see these doc comments in the output.

aganders3 (approved):
I admit to not taking the time to quite understand all the code changes here for adding topic pages, but it looks reasonable and the output looks good (tested locally).

src/core/image_source_policy.ts
```diff
@@ -1,13 +1,24 @@
...
   "fallbackBackground",
   "prefetchSpace",
-] as const;
-
-export type PriorityCategory = (typeof ALL_CATEGORIES)[number];
+] as const satisfies readonly PriorityCategory[];
 
+/**
+ * @hidden
+ */
```
srivarra, line 21: [question(non-blocking)](https://pullpo.io/cc?l=question&d=non-blocking) 
does this need to stay hidden? it's already exported in `index.ts`?

src/core/image_source_policy.ts
```diff
@@ -118,7 +123,14 @@ export function createPlaybackPolicy(
   return createImageSourcePolicy(mergeProps(base, overrides));
 }
 
-/** @hidden */
+/**
+ * Creates a loading policy that loads only visible chunks.
+ *
+ * No spatial or temporal prefetching happens, which minimizes memory
+ * use and network traffic for static scenes.
+ *
```
srivarra: Maybe "Creates a loading policy with spatial and temporal prefetching disabled. Layers still load their selected and fallback resolution levels." would be more accurate. That avoids implying volume loading is limited to what's on screen. Pretty minor, up to you.

src/core/image_source_policy.ts
```diff
@@ -136,6 +148,53 @@ export function createNoPrefetchPolicy(
...
+export function createImageSourcePolicy(
+  config: ImageSourcePolicyProps
+): ImageSourcePolicy {
+  validatePolicyProps(config);
+
+  const prefetch = {
+    x: config.prefetch.x,
+    y: config.prefetch.y,
+    z: config.prefetch.z ?? 0,
+    t: config.prefetch.t ?? 0,
```
srivarra, line 155: `NaN` and `Infinity` pass the prefetch check. Is that meaningful in anyway?

src/core/image_source_policy.ts
```diff
@@ -136,6 +148,53 @@ export function createNoPrefetchPolicy(
...
+        return acc;
+      },
+      {} as Record<PriorityCategory, number>
+    )
+  );
+
+  const lod = {
+    min: config.lod?.min ?? 0,
+    max: config.lod?.max ?? Number.MAX_SAFE_INTEGER,
+    bias: config.lod?.bias ?? 0.5,
```
srivarra, line 172: [nitpick(non-blocking)](https://pullpo.io/cc?l=nitpick&d=non-blocking) 
We can pass in negative values in for min/max, I don't think negative numbers have meaning for lods afaik, we should put a little check in another pr.

src/core/image_source_policy.ts
```diff
@@ -136,6 +148,53 @@ export function createNoPrefetchPolicy(
...
+
+  const resolved: ImageSourcePolicy = {
+    profile: config.profile ?? "custom",
+    prefetch,
+    priorityOrder: Object.freeze([...config.priorityOrder]),
+    priorityMap,
+    lod,
+  };
+
+  return Object.freeze(resolved);
```
srivarra, line 183: [question](https://pullpo.io/cc?l=question) 
`Object.freeze()` prevents replacing top-level properties:


```ts
const policy = createExplorationPolicy();

policy.profile = "different"; // Blocked.
policy.prefetch = { x: 0, y: 0, z: 0, t: 0 }; // Blocked.
```

But `prefetch` and `lod` remain mutable:

```ts
policy.prefetch.x = 99; // Allowed.
policy.lod.bias = 3;   // Allowed.
```

Is the intent for the entire policy to be immutable, with changes made by creating a new policy and swapping it in? If so, should we freeze these nested objects too?

srivarra (approved):
The docs look good overall, I think my questions arn't exactly related to the documentation, and more on the how we want these objects to behave overall.

src/core/image_source_policy.ts
shlomnissan: yeah, i wanted to create a nested breakdown but it's too messy. the page still outlines `ImageSourcePolicy` without comments. i think it's probably good enough for now?

either way, i deleted the inline comments as they are not picked up anyway.

src/core/image_source_policy.ts
shlomnissan: similar reason: https://github.com/chanzuckerberg/idetik/pull/754#discussion_r4007938808

types are currently displayed as 2-column tables (name/description). nested types don't render well. one way to resolve this is making `prefetch` and `lod` values their own types and documenting those, but that's a bigger change, and i think `ImageSourcePolicy` already captures what the properties are.

i know it's not great, but better than nothing?

srivarra: gotcha, works for me.

src/core/image_source_policy.ts
shlomnissan: negative LODs are not a problem. the chunk store view clamps them into the dataset range. but there are other issues that your previous message flagged: non integer bounds and NaN in LOD fields.

src/core/image_source_policy.ts
shlomnissan: i think that `NaN` breaks the viewer and `Infinity` will probably prefetch the entire plane along the axis. so both need fixing.

src/core/image_source_policy.ts
shlomnissan: intent is to make it all immutable. change detection is by identity.

shlomnissan:
@srivarra you mentioned three issues that needs fixing:

- handle prefetch axis set to `NaN` and `Infinity`
- handle any LOD fields set to `NaN`
- freeze nested objects in `createImageSourcePolicy`

they are out of scope for this PR but i will submit a quick follow up patch

PR #756, opened 2026-09-16
feat: prefetch time chunks at the current LOD

aganders3:
### Summary
This changes temporal prefetching behavior to prefetch timepoints at the current LOD, instead of the fallback LOD. Temporal prefetch priority is also changed, as `main` right now prioritizes prefetch chunks over current-t current-LOD chunks, which no longer makes sense with this change.

The change to the volume behavior is included but irrelevant for now, given we always clamp the policy for a volume layer to a single LOD.

### Tests & Checks
I tested locally with our examples and some untracked debugging scripts.


shlomnissan (approved):
LGTM :+1:

PR #757, opened 2026-09-16
feat: draw all resident chunks in view; ordered by distance from current LOD

aganders3:
### Summary

This change is motivated by work on livescreen data and is meant to work in concert with #756, as discussed in Slack. We want to:
* prefetch at the current LOD (#756)
* always draw the best data we've already loaded

This means prefetched data is not necessarily "wasted" if you zoom in or out across an LOD boundary during playback.

It's long been a pet peeve of mine that zooming into the image it gets worse before it gets better. This PR is also a step towards fixing that!

#### How it works
`getChunksToRender` currently returns only the current LOD plus the coarse fallback. This PR changes it to return any resident (has a texture) chunk that _would be_ marked visible at the current slice — same timepoint, a channel of interest, spanning the slice coordinate, and intersecting the view rect. They are sorted by distance from the current LOD (with ties going to the coarser chunk), so this Just Works™ with the existing stencil-based rendering.

This is all opportunistic loading. These chunks are not marked visible, so they remain available for eviction. If current demand (e.g. from prefetching) is causing memory pressure, these chunks may be quickly evicted to make room. That might require some tweaking.

#### Why residency moves from `ChunkManager` to `ChunkStore`
The view returns renderable chunks by iterating resident chunks _for its own source_. This is much more efficient than iterating all chunks looking for those in view and finding the resident ones. `ChunkManager` kept one global `resident_` set spanning every source, but a view can't filter them by source, so a chunk from a different source could get drawn by the wrong layer.

### Tests & Checks

- all checks and tests pass
- tested locally with existing examples and some untracked visual tests


src/data/chunk_store_view.ts
```diff
@@ -23,9 +23,11 @@ export class ChunkStoreView {
   private currentLOD_: number = 0;
   private readonly axes_: SliceAxes;
   private readonly scale0_: number;
+  private mode_: "image" | "volume" | null = null;
```
aganders3: This "mode" is implicit before this change, but I think this makes what is happening a bit more readable.

Regardless this feels like motivation to either unify 2D and 3D approaches, or separate this into two classes.

aganders3, line 32: This "mode" is implicit before this change, but I think this makes what is happening a bit more readable. This gets set by successful calls to either `updateChunksForImage` or `updateChunksForVolume`.

Regardless this feels like motivation to either unify 2D and 3D approaches, or separate this into two classes.

src/data/chunk_store_view.ts
```diff
@@ -23,9 +23,11 @@ export class ChunkStoreView {
...
   private lastViewBounds2D_: Box2 | null = null;
   private lastViewProjection_: mat4 | null = null;
-  private lastSliceBounds_?: [number, number];
+
+  private lastSliceCoordW_?: number;
```
aganders3, line 36: Chunks at different LODs may have a different slice span, so change detection is moved to the point coord.

PR #758, opened 2026-09-16
feat: add scroll zoom option to orbit and pan-zoom controls

shlomnissan:
### Summary

this PR adds an optional `scrollZoom` property to both camera controls.

the motivation comes from the inline viewers in the docsite. when a user
scrolls through the page and the cursor happens to be over a viewer, the
wheel event zooms the viewer instead of scrolling the page.

### Tests & Checks

- all checks have passed.


src/index.ts
```diff
@@ -58,6 +58,7 @@ export type {
   OrthographicCameraProps,
 } from "./objects/cameras/orthographic_camera";
 export { PanZoomControls } from "./objects/cameras/controls";
+export type { PanZoomControlsProps } from "./objects/cameras/controls";
```
aganders3, line 61: I'm not familiar enough with TS, do we also need to export `ScrollZoomMode` or can it just be inferred? Is the literal enumeration included in the docs?

aganders3 (approved):
LGTM

I'm kind of surprised this requires a library change (aside from the modifier key thing), but that's just my inexperience with browser events.

srivarra (approved):
Tried it out with a little example, works great!

src/index.ts
shlomnissan: inferred is fine

shlomnissan:
> I'm kind of surprised this requires a library change

@aganders3 look at mapbox live examples and try to scroll through it. that's what i want to implement

PR #759, opened 2026-09-17
docs: how idetik works architecture guide

shlomnissan:
### Summary

this PR adds a new guide "how idetik works" that describes the architecture
at a high level with a couple of diagrams.

this summary is based on my own understanding. i tried to keep it readable. let me know if i missed something or if i got something wrong. best way to review this is to pull the branch and read through. 

### Tests & Checks

- all checks have passed

https://github.com/user-attachments/assets/83887581-0ba8-4ace-8a8a-684e2a83108e


docs/guide/how-idetik-works.md
```diff
@@ -0,0 +1,130 @@
...
+
+Viewports and layers can be added and removed while the runtime is running. A layer belongs to one viewport at a time and adding it to a second throws.
+
+Synchronizing views follows from the plain-object model. Two layers constructed with the same slice coordinate object show the same time point and move together when it changes. Layers that share a channel appearance array recolor together. Layers on the same source share every chunk the source loads, whichever viewport they draw in.
+
+## Layers and Renderables
+
+A layer decides what to draw and how events should be handled. It reports its lifecycle through a [`LayerState`](/api/classes/Layer.html#layerstate) that begins at `initialized` and reaches `ready` once the layer has something to show. The renderer skips layers that are not ready and applications can subscribe to state changes to defer work until the first data has arrived.
```
srivarra: I'm not sure this is accurate totally. Built-in streaming layers switch to ready on their first update after attachment, before any chunks have necessarily arrived. The renderer can therefore draw an empty layer. I think, ready means eligible to draw, not that the layer has rendered chunks to show.

```suggestion
A layer decides what to draw and how to handle input. Its [`LayerState`](/api/classes/Layer.html#layerstate) controls whether the renderer draws it. The renderer calls `update()` every frame, but skips drawing until the state is `ready`. Built-in streaming layers switch to `ready` on their first update after attachment, whether or not any chunks have arrived. Applications can subscribe to state changes, but `ready` is not a signal that image data is available.
```

docs/guide/how-idetik-works.md
```diff
@@ -0,0 +1,130 @@
...
+});
+```
+
+The element also routes input. Each viewport listens for pointer and wheel events on its own element so the browser decides which viewport a pointer is over. A layer may stop propagation to keep an event from the controls, for instance, to drag a handle without panning the view. Controls see the event last.
+
+Cameras are plain objects and controls are an interface. [`PanZoomControls`](/api/classes/PanZoomControls.html) and [`OrbitControls`](/api/classes/OrbitControls.html) write to the camera's transform directly, and the next frame draws the result. Any object that implements [`CameraControls`](/api/interfaces/CameraControls.html) can drive a camera.
+
+Viewports and layers can be added and removed while the runtime is running. A layer belongs to one viewport at a time and adding it to a second throws.
+
+Synchronizing views follows from the plain-object model. Two layers constructed with the same slice coordinate object show the same time point and move together when it changes. Layers that share a channel appearance array recolor together. Layers on the same source share every chunk the source loads, whichever viewport they draw in.
```
srivarra: Sharing the `channelProps` array doesn't automatically keep the layers synchronized. Existing renderables copy their channel settings, so the application must call `setChannelProps()` on each layer after a change. We should mention those calls rather than state that sharing the array is enough.


```suggestion
Synchronizing views follows from the plain-object model. Two layers constructed with the same slice coordinate object show the same time point and move together when it changes. Channel appearance is scoped to each layer, so the application must call `setChannelProps()` on each layer to recolor several views together. Layers on the same source share every chunk the source loads, whichever viewport they draw in.
```

docs/guide/how-idetik-works.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Streaming Chunks
+
+A source describes an image and knows how to load pieces of it. [`OmeZarrImageSource`](/api/classes/OmeZarrImageSource.html) reads the metadata of a multiscale OME-Zarr group when it is created and exposes, for each axis and each resolution level, the size, the chunk size, and the scale and translation in world units. Nothing else is fetched until a layer asks.
+
+In each frame, a streaming layer asks for the chunks under its view. It picks the resolution level whose elements are closest to one screen pixel at the current zoom, then requests the chunks at that level that intersect the view, padded by the policy's prefetch distance and ranked by category and by distance from the center of the view.
+
+Alongside that level it always requests one coarse level as a backdrop. The backdrop ranks first, so the view is covered quickly at low resolution and refined as finer chunks land. When the slice changes, the layer keeps showing the old one until the new backdrop has fully arrived, so scrubbing through time shows whole frames rather than half-loaded ones.
```
srivarra: suggestion
I think this should describe the timeout so readers do not treat whole-frame presentation as a guarantee.

```suggestion
Alongside that level it always requests one coarse level as a backdrop. The backdrop ranks first, so the view is covered quickly at low resolution and refined as finer chunks land. When the slice changes, image and label layers keep showing the previous slice until the new backdrop has fully loaded or one second passes. After that timeout, they display whichever chunks are available.
```

docs/public/diagrams/architecture_diagram_0.png
srivarra: nitpick
I think the layer should be slightly modified to:

```
Layer  ×N
each instance belongs to one viewport
```

up to you though.

srivarra (approved):
I think it looks good, nothing here is blocking, and it's all just nitpick suggestions.

docs/public/diagrams/architecture_diagram_0.png
aganders3: Agree - or perhaps something like "each instance may be attached to at most one viewport".

docs/guide/how-idetik-works.md
aganders3: I think this is more accurate, but it sounds like perhaps an inconsistency in the API? Should channel properties instead be mutated references like `SliceCoordinates`?

docs/guide/how-idetik-works.md
```diff
@@ -0,0 +1,130 @@
+# How Idetik Works
+
+This guide describes Idetik's architecture at a high level. It is for anyone who wants to extend the library or reason about how it loads data and uses memory. It is not a reference: the [API Reference](/api/) has the exact signatures.
+
+## Objects and Ownership
+
+Every Idetik application is a small tree of objects. At the root is the [`Idetik`](/api/classes/Idetik.html) runtime. It owns the canvas, the renderer, the chunk manager that streams data for it, and any number of viewports. A [`Viewport`](/api/classes/Viewport.html) owns a camera, optional camera controls, an element that defines its area, and an ordered stack of layers. A [`Layer`](/api/classes/Layer.html) owns the renderable objects it draws.
```
aganders3:
```suggestion
Every Idetik application is a small tree of objects. At the root is the [`Idetik`](/api/classes/Idetik.html) runtime. It owns the canvas, the renderer, the chunk manager that streams data for it, and any number of viewports. A [`Viewport`](/api/classes/Viewport.html) owns a camera, optional camera controls, an element that defines its area, and an ordered stack of layers. A [`Layer`](/api/classes/Layer.html) manages a collection of [renderable objects](/api/classes/RenderableObject.html) to be drawn by the renderer.
```

docs/guide/how-idetik-works.md
```diff
@@ -0,0 +1,130 @@
...
+
+Two objects are never constructed by the application. The renderer and the chunk manager are created by the runtime and reached only through it. Everything else is built by the application and handed in. That is what makes the library composable: a camera works with any controls, a source feeds any layer, and a layer draws in any viewport.
+
+The tree is not a scene graph. Each renderable object carries one transform expressed directly in world coordinates and nothing inherits a transform from a parent. Nesting here means lifetime and update order, not spatial containment.
+
+Idetik has no user interface and holds no application state. The application creates its objects and passes them to Idetik by reference. It changes them and Idetik reads them on the next frame. There is no store to dispatch to and no event to emit. This one principle explains much of what follows.
+
+## One Coordinate System
+
+Everything in Idetik is in world units. An OME-Zarr image declares a scale and a translation for each axis at each resolution level, and Idetik uses them to place every chunk in physical space, typically micrometers.
```
aganders3: Consider adding a warning/callout here that Idetik only accounts for _scale_ but not the actual _units_ at this time. This is not a problem unless source units disagree (e.g. an image specified in Å with an overlay specified in nm).

docs/guide/how-idetik-works.md
```diff
@@ -0,0 +1,130 @@
...
+
+Once every viewport has drawn, the chunk manager runs. It gathers what all the layers asked for, admits requests against the memory budget, hands them to the loading queue, uploads a few finished chunks to the GPU, and evicts if it must. The overlays run last.
+
+![One frame in Idetik, and the path of a chunk through three frames](/diagrams/architecture_diagram_1.png)
+
+_One frame, then the path of a single chunk. Layers request while drawing, the manager acts after every viewport has drawn, and a chunk on screen in frame N+2 was requested in frame N._
+
+The ordering has a consequence the diagram makes visible. Layers declare their needs while drawing, and the manager acts on the aggregate afterwards. A chunk requested in one frame is fetched and decoded off the main thread, arrives during a later frame, is uploaded at the end of that frame, and is drawn by the layer in the frame after that.
```
aganders3:
```suggestion
The ordering has a consequence the diagram makes visible. Layers declare their needs while drawing, and the manager acts on the aggregate afterwards. A chunk requested in one frame is fetched and decoded off the main thread, arrives during a later frame, is uploaded at the end of that frame, and is ready to be drawn in the frame after that.
```

docs/guide/how-idetik-works.md
```diff
@@ -0,0 +1,130 @@
...
+
+The element also routes input. Each viewport listens for pointer and wheel events on its own element so the browser decides which viewport a pointer is over. A layer may stop propagation to keep an event from the controls, for instance, to drag a handle without panning the view. Controls see the event last.
+
+Cameras are plain objects and controls are an interface. [`PanZoomControls`](/api/classes/PanZoomControls.html) and [`OrbitControls`](/api/classes/OrbitControls.html) write to the camera's transform directly, and the next frame draws the result. Any object that implements [`CameraControls`](/api/interfaces/CameraControls.html) can drive a camera.
+
+Viewports and layers can be added and removed while the runtime is running. A layer belongs to one viewport at a time and adding it to a second throws.
+
+Synchronizing views follows from the plain-object model. Two layers constructed with the same slice coordinate object show the same time point and move together when it changes. Layers that share a channel appearance array recolor together. Layers on the same source share every chunk the source loads, whichever viewport they draw in.
+
+## Layers and Renderables
```
aganders3: Maybe add a note here about subclassing `Layer` to implement a new layer type. This could also be a separate guide at some point.

docs/guide/how-idetik-works.md
```diff
@@ -0,0 +1,130 @@
...
+
+Requests are served most urgent first, a bounded number at a time, and a request is cancelled when its chunk drops out of every view. Fetching and decoding happen on a pool of workers which read the compressed chunk and transfer the result back without a copy.
+
+The manager uploads chunks to the GPU as 3D textures as soon as they load, a few per frame so a burst of arrivals never stalls, and frees the CPU copy at once.
+
+The prefetch distances, the ordering of priority categories, and the range of levels are the three knobs of a loading policy. [Loading Policies](/api/namespaces/LoadingPolicies.html) describes presets and how to build them.
+
+## The Memory Budget
+
+Every chunk resident on the GPU counts against one budget, set by [`memoryLimitMB`](/api/classes/Idetik.html#property-memorylimitmb) and shared by all viewports of a runtime. The default is two gigabytes. The manager admits a request only if the chunk will fit, evicting less important resident chunks to make room if it must. A chunk visible in any view is never evicted.
```
aganders3: Should we add a note here about setting this limit? Might want to set it close to your GPU memory size, but it can also exceed it. Do we know if that is platform dependent?

aganders3 (approved):
A couple small comments, but this looks great!

docs/guide/how-idetik-works.md
shlomnissan: it does feel like API inconsistency which is why i assumed it worked this way, but thinking through it, maybe it's for the best? maybe `sliceCoordinates` should be more like channel properties? i'm not sure.

what's the more likely scenario: multiple layers sharing channel properties and slice coordinates, or each layer having it's own (sharing `t` is kind of a special case really)?

docs/guide/how-idetik-works.md
shlomnissan: this suggestion might be more accurate but i'm not sure it clarifies behavior. i need to revisit the timeout implementation. it doesn't feel like the right behavior to me.

docs/guide/how-idetik-works.md
shlomnissan: i will add a note but keep "exceeding" out of it for now. i haven't tested this on multiple platforms.

